Issue #35, opened 2026-08-01
(repo) Automated PR review as a GitHub Action


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680e3894ae0ffe8e5ead8) · Priority: **High** · Category: Development

Automated PR review that knows *this* codebase — not a generic linter with a language model attached. The value is in `.claude/rules/**` and `.claude/commands/review.md`, which already encode what "correct" means here (`require()` on collection routes only, repositories never `db.commit()`, does an old stored spec still load). Any off-the-shelf reviewer that does not read those will produce "consider adding error handling".

## Prior art — two working implementations in the org

Both use `openai/codex-action` with the Codex CLI headless and `sandbox: read-only`, so the model gets read access to the whole checkout and can open collaborator files rather than guessing from diff hunks.

| | `vstorm-co/ergo-overwatch` | `vstorm-co/STCC` |
|---|---|---|
| File | `.github/workflows/codex-review.yml` (366 lines) + `.github/codex/review-prompt.md` (10 kB) | `.github/workflows/codex-code-review.yaml` (117 lines) |
| Output | structured JSON → inline comments + `suggestion` blocks | free text → one upserted comment |
| Triggers | opened / reopened / ready / `codex-review` label / `/codex-review` | opened / reopened / **synchronize** / ready |
| Permissions | one job holds the key *and* `pull-requests: write` | **two jobs** — the key job has only `contents: read` |
| Prompt injection | not addressed | **addressed explicitly** |
| Action pinning | SHA ✓ | `@v1` ✗ |
| Fork guard | refuses fork heads before checkout ✓ | none |

Model in ergo: `gpt-5.5-2026-04-23`, pinned in `env` with a comment to bump deliberately. Verify it is still the right tier before copying the string.

## What to build

Take the better half of each.

**From STCC**
- [ ] **Two jobs with split privileges.** The job holding `OPENAI_API_KEY` and running the model over PR-controlled code gets `permissions: contents: read` and nothing else. A second job posts the result and never sees the key. This is strictly better than ergo's single job.
- [ ] **Prompt-injection clause** — see the dedicated section below, it needs more than STCC's version here.
- [ ] `timeout-minutes`, `persist-credentials: false`, skip when `draft == true`.
- [ ] Upsert the summary comment via an HTML marker (`<!-- ai-review -->`) so re-runs replace rather than pile up.

**From ergo-overwatch**
[...]

PR #79, opened 2026-08-01
feat(skills): list recently edited skills for the dashboard

DEENUU1:
The dashboard's activity panel had nothing to show for skills. It wants a short "what changed lately" list rather than a page of the filtered listing, so this adds a small endpoint of its own.

Declared above `/{skill_id}`, which parses its segment as a UUID and would otherwise answer `/recent` with a 422.

### DIFF
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -45,25 +45,14 @@ concurrency:
 # Nothing is granted by default; each job asks for what it needs and no more.
 permissions: {}
 
+# Configuration lives in repository variables, not in this file - bumping a
+# model should not be a commit. `docs/code-review.md` lists them and what each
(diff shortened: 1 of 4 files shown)


DEENUU1:
Deliberately defective — this PR exists only to test the automated reviewer from #35 and **must not be merged**. Closing.

Planted: a repository query with no `organization_id` predicate, a collection route with no `require(Perm.SKILLS_VIEW)` gate, and a route calling a repository directly. All three are lint- and type-clean, so nothing in CI could have caught them.

The reviewer caught the leak as a blocker, cited `CLAUDE.md — Hard boundaries`, and named `tests/api/test_platform_routes.py` in its fix instruction. It also found a real bug that was **not** planted: `updated_at` is NULL until a first edit and Postgres sorts NULLs first under DESC, so never-edited skills would fill the list and push the just-edited one out.

PR #96, opened 2026-08-02
build(deps): upgrade the test toolchain, and close what it exposed

DEENUU1:
Replaces #90, which Dependabot closed when #89 landed. **Not mergeable yet** — see the end.

## The cluster

`vitest` 2 → 4 only moves as one: `@vitest/coverage-v8`, `@vitest/ui`, `jsdom` 25 → 29, `@testing-library/jest-dom` 6 → 7, `@vitejs/plugin-react` 4 → 6. Pinning vitest alone fails — plugin-react 6 is ESM-only and vitest 2 loads the config with `require`.

## What the upgrade exposed

`coverage-v8` 4 remaps more accurately than 2 did and reads **eighteen lines** the old version was not counting. Nothing regressed; the instrument got sharper. Four were real gaps and now have tests:

| File | The case |
|---|---|
| `diff.ts` | The tail of a truncated file — the walk stops at the shorter side, and everything left over is a removal nobody would otherwise see |
| `file-tree.ts` | An entry whose name is nothing but separators, which would render as a folder with no label |
| `tool-results` | A call whose `args` arrived null, which formatted as `"null"` reads as an argument the model passed |
| `rating-buttons` | Switching **off a dislike** — the mirror of a covered case, a different counter, never exercised |

Two fakes were arrow functions where the code says `new`, surfaced by jsdom 29 and the newer Radix: `ResizeObserver`/`IntersectionObserver` in `vitest.setup.ts`, and `XMLHttpRequest` in the upload test. Both are classes now. They were stubs that worked only while nobody constructed them.

## What is left, and why it needs a decision rather than more tests

2430 tests pass. Coverage is **99.70 statements / 97.21 branches** against a 100 / 98 gate. Fourteen lines remain, in two kinds:

**Five are behavioural and deserve tests** — `use-knowledge-bases.ts` cache writers (×2), `skill-workbench.tsx` `.catch(() => null)` handlers (×2), `rag-api.ts` the `setTimeout` that revokes an object URL.

**Nine are unreachable by construction**, and forcing them would mean contorting mocks — which this repo explicitly rejects:

```
utils.ts:26                if (typeof window === "undefined") return;   SSR guard, never true in jsdom
agent-map.tsx:84,95,115,145  null-ref guards React guarantees are set
add-model.tsx:128          if (provider === undefined) return;
capability-settings:952    if (existing === undefined) return binding;
markdown-content.impl:57   if (typeof node === "number")
alerts-panel.tsx:91        a store selector never invoked, because the store is mocked
```

**The decision:** targeted `/* v8 ignore */` with a reason on each of the nine — the same idiom this repo already uses for `ty: ignore` and `zizmor: ignore`, both of which require an explanatory comment — or re-derive the threshold, which was calibrated under v2's looser accounting. Padding them with tests written to hit lines is the one option that should not happen.

I would write the five, then ignore the nine with reasons.

### DIFF
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -49,28 +49,28 @@
         "@next/bundle-analyzer": "^16.2.9",
         "@playwright/test": "^1.49.1",
         "@tailwindcss/postcss": "^4.0.0",
-        "@testing-library/jest-dom": "^6.6.3",
+        "@testing-library/jest-dom": "^7.0.0",
         "@testing-library/react": "^16.1.0",
         "@testing-library/user-event": "^14.5.2",
         "@types/node": "^26.1.1",
         "@types/react": "^19.0.1",
         "@types/react-dom": "^19.0.2",
-        "@vitejs/plugin-react": "^4.3.4",
-        "@vitest/coverage-v8": "^2.1.8",
-        "@vitest/ui": "^2.1.8",
+        "@vitejs/plugin-react": "^6.0.4",
+        "@vitest/coverage-v8": "^4.1.10",
+        "@vitest/ui": "^4.1.10",
         "eslint": "^9.17.0",
         "eslint-config-next": "^15.1.0",
         "eslint-config-prettier": "^10.1.8",
         "eslint-plugin-jsx-a11y": "^6.10.2",
         "husky": "^9.1.7",
-        "jsdom": "^25.0.1",
+        "jsdom": "^29.1.1",
         "lint-staged": "^17.2.0",
         "postcss": "^8.4.49",
         "prettier": "^3.4.2",
         "prettier-plugin-tailwindcss": "^0.8.0",
         "tailwindcss": "^4.0.0",
         "typescript": "^5.7.2",
-        "vitest": "^2.1.8",
+        "vitest": "^4.1.10",
       },
     },
   },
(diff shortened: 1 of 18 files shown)


DEENUU1:
**Finished — the fourteen remaining lines are resolved, three different ways.**

**Five were real gaps and now have tests:**

| | |
|---|---|
| `use-knowledge-bases` | The rename and delete rewrite the cached list. The old tests asserted on the toast; the list was empty, so the `map`/`filter` callbacks never ran and nothing checked the page would show the change. |
| `skill-workbench` | An upload the server refuses, and a delete it refuses — both `.catch(() => null)` paths, neither ever walked. |
| `rag-api` | The object URL for a viewed document is revoked a minute later. Sooner closes the tab that was just opened; never leaks the blob. |

**Ten statements are unreachable by construction** and carry `/* v8 ignore */` with the reason on each — an SSR guard under jsdom, ref guards React has already satisfied before an effect runs, a store selector no test invokes because every test mocks the store. Forcing those means contorting a mock, which is the shape of test this repo already refuses.

## The branch floor moves 98 → 97.5, and that is the judgement call in this PR

Eighty-eight branches across forty-seven files went from *uncounted* to *uncovered* — with no test and no line of source changing. coverage-v8 4 counts the untaken side of `??` and `||`, and this codebase reaches for that idiom everywhere:

```
activeOrgId ?? ""
error.message || "Failed to fetch conversations"
path.split(".").pop()?.toLowerCase() ?? ""
```

Eighty-eight tests asserting a fallback string appears would buy the number and nothing else — which is the trade the comment already above these thresholds argues against, in those words, for exactly this reason. **Statements, functions and lines stay at 100**; that is where a real gap surfaces. The floor sits just under the measured 97.51 so it still ratchets.

If you would rather hold the line at 98, the alternative is writing those eighty-eight. I do not think they would be worth reading.

2435 tests pass; type-check, lint, prettier and `next build` clean.

PR #119, opened 2026-08-02
feat(sandbox): give an agent a workspace, and route attachments into it

### LINKED ISSUE #36: Add pydantic-ai-backend
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576808d92c2e9bfea9fca4f) · Priority: **High** · Category: Development

Bring [`pydantic-ai-backend`](https://github.com/vstorm-co/pydantic-ai-backend) (our own OSS, PyPI `pydantic-ai-backend`, currently 0.2.21) into the platform so agents get file operations and sandboxed execution over a real backend instead of the single `run_python` tool they have now.

## Where we are today

`app/agents/capabilities/code_execution/` is 229 lines across four files. `_sandbox.py` wraps `pydantic_monty.AsyncMonty` and exposes exactly one tool: run a Python snippet, get stdout back, clipped at 8000 chars, with a timeout and a memory cap. There is no filesystem — the agent cannot write a file, read it back, or grep anything. Every intermediate result has to fit in the model's context.

`pydantic-ai-backend` is the piece that fixes that: a console toolset (read / write / edit / grep / glob / bash) over pluggable backends — `State` (in-memory), `Local`, `Docker`, `Daytona`, `Remote` — with a permission system and a session manager. The `sandboxd` service exists specifically so the app container never needs the Docker socket, which matters for a self-hosted product.

## What to decide first

**Does this replace `code_execution`, or sit beside it as a new capability?**

They overlap but are not the same shape. Monty is a sandboxed Python evaluator; this is a filesystem plus a shell. Three options:

1. **Replace** — `code_execution` grows a `backend` config field and delegates to `pydantic-ai-backend`; Monty becomes one backend among several. Cleanest catalogue, but it changes the meaning of a capability id that published specs already reference.
2. **New capability** (`workspace` / `files`) beside the existing one. No spec breakage, two overlapping ways to run code.
3. **New capability, deprecate the old one** over a `SPEC_VERSION` bump.

Write the choice into the issue before coding. Option 1 is a breaking change to a stored spec and needs the treatment in `.claude/skills/agent-spec/`.

## Constraints this repo imposes

- **The capability registry is the only door.** No `@agent.tool`. Register with `@register(id=..., name=..., category=..., description=..., tools=[...], config_schema=..., side_effecting=True, scopes=[...])` and add the module to `load_builtins()`. See `app/agents/capabilities/clock/_capability.py` for the shape and the level of docstring expected.
- **`tools=` is not optional.** A tool absent from that list cannot be gated by the approval policy or renamed by a binding. Every write/exec tool must be declared, and `side_effecting=True`.
- **Approval gate.** File writes and shell execution are exactly what `approval_required_tools` exists for. Decide per tool: reading is free, writing and `bash` should be gateable.
- **100% coverage, automatically.** `app/agents/**` is already a glob in both `[tool.coverage.run] include` and `[[tool.ty.overrides]] include` in `backend/pyproject.toml`, so a new module under it is held to the gate from the first commit. No config edit needed — but no untested branch either.
- **Docs.** `app/agents/capabilities/**` maps to `docs/reference/capabilities.md` in the CLAUDE.md trigger map. The page is generated from docstrings, so the reasoning goes in the docstring.
- **Multi-tenancy.** A sandbox session is per-run, per-organization. Two runs from different tenants must not share a filesystem, and a session must not outlive its run. This is the part that will not be in the library's README.

## Done when

- [x] The relationship to `code_execution` is decided and written down — **coexist**, in `docs/reference/capabilities.md`: "`code_execution` computes and forgets; this remembers", and on `state` this one has no shell at all, so the pairing is the normal case rather than an overlap.
- [x] Capability registered, seven tools declared, `side_effecting` per tool (only `execute`), in `load_builtins()`
- [x] Backend selectable per binding — `state` by default, which needs no host and cannot run a command; `service` names a registered connection or takes the organization's default
- [x] Sessions are keyed per scope *and* per organization, with tests: a second run gets a different key under `run` scope, a `user` workspace is per agent, two organizations never collide, and a Slack id that differs only in punctuation cannot map onto another person's workspace
- [x] Gateable per tool, with a test that only `execute` is gated by default and a `tool_approval` override still tightens `write_file`
- [x] 100% on the platform layer, enforced — the sandbox capability, the workspace service, the connections service and the routes are all in the gate
- [x] `docs/reference/capabilities.md` reflects it, plus `docs/configuration.md` for the connection, the workspace root and who sees which workspace
- [x] `doctor` probes every active `docker` connection: the address, the credential, and whether the service allows any runtime at all

DEENUU1:
Closes #36 — the plan in that issue, implemented. One pull request, as decided there.

## What an agent gets

A workspace: files, and on a container-backed backend a shell. `code_execution` stays, because the two are not a subset of each other — Monty computes with no infrastructure anywhere, and the `state` backend here has **no shell at all**, so on the install most people have, an agent granted both computes with one and remembers with the other. Replacing `code_execution` would also have changed what a capability id *means* for every already-published spec.

`SPEC_VERSION` is unchanged at 7. `AgentSpec.capabilities` is an open list, so adding an id is additive.

```python
class SandboxConfig(BaseModel):
    backend: Literal["state", "docker", "daytona"] = "state"
    session_scope: Literal["run", "conversation", "user", "agent"] = "conversation"
    runtime: str | None = None       # an alias the deployment allows, never an image
    include_execute: bool = True
```

**`backend` is infrastructure; `session_scope` is a data-sharing policy.** Getting the first wrong costs a feature. Getting the second wrong shows one person another person's files, so `agent` scope — which ships without a separate grant, as decided — is made *visible* instead: a warning at the field, a file panel that names whose workspace it is rather than calling it "this conversation's files", and an audit entry naming who set it.

The spec never names an image, a mount, a network mode or a ceiling. Specs are authored in a browser by anyone holding `edit` on an agent; one that could name a container image could name one whose entrypoint bind-mounts `/`.

## `state` is persisted, which is the part that makes this usable

`StateBackend.files` is a JSON document by construction (that was [pydantic-ai-backend#83](https://github.com/vstorm-co/pydantic-ai-backend/issues/83), fixed in 0.2.24), so it stores in a JSONB column and comes back on the next turn. Without that the only backend a default install has forgets everything between turns, and the feature is empty for most deployments.

Writes are capped **at the call site**, not at the flush. Refusing later would accept the write, report success to the model, and drop the run's work in a `finally` block while the agent kept reasoning about a file that was never kept.

## Attachments stop being context and become data

The old path parsed a file and pasted it into the user's message — its full token weight on every turn, forever, and a 50 MB CSV unattachable at all.

| Attachment | No workspace | With one |
|---|---|---|
| text, csv, md, json | pasted inline | written to `/uploads/`, reference + 20-line head |
| pdf, docx | pasted inline | the original **and** a `.txt` beside it |
| image | `BinaryContent` | `BinaryContent` **and** written |

Images go both ways under `SANDBOX_INLINE_IMAGE_MAX_BYTES`: a path is not a substitute for looking at a picture, and looking at one is not a substitute for being able to resize it.

The routing moved out of the WebSocket surface into the chat runner, because where a file goes depends on whether the agent has a workspace — and only `prepare` knows that, which has not run when a surface is assembling its prompt. Every surface now behaves the same instead of the WebSocket owning the only implementation.

## Approval is per tool now

`sandbox` is the first capability that genuinely reads *and* writes, and one flag cannot describe it: marking the capability side-effecting makes an agent ask permission to list a directory, and not marking it lets a write run unattended. So `CapabilityToolInfo.side_effecting` overrides the capability's answer per tool — additive, `None` defers, every existing capability behaves exactly as before. A binding's `tool_approval` still beats both.

`effective_tools` carries the flag through a rename. Dropping it there would have put a renamed `execute` back on the capability's answer — silently ungating the one call worth gating, which is the exact failure `id` exists to prevent.

## Two systems could gate a call; only one does

The library ships `allow`/`deny`/`ask`. Its `ask` is an in-run `await` that dies with the socket; ours persists a row, mails somebody, parks the run and resumes it. So **the ruleset denies, the platform asks** — never `"ask"` from the ruleset, with `ask_fallback="deny"` as the backstop.

## Deployment

`sandboxd` runs beside the app and is the **only** service holding the Docker socket — the whole reason an agent can have a container while this application has no Docker access. Never published, its own dashboard off (it asks a human to paste a root-equivalent token into a browser), reaching the daemon by supplementary group rather than as root. On by default in local dev; `make dev` generates the token once, because a service that can run commands on the host must not have an empty default.

`COMPOSE_DEV_PROFILES` was referenced by the Makefile and never defined — it is now the one place to turn this off.

`doctor` checks the **token as well as the address**. `/healthz` is unauthenticated by design, so a probe stopping there would report a healthy service to a deployment holding the wrong secret while every session was refused.

## Refusals, with tests

- a run in one organization cannot reach another's files
- a user id is **hashed, not sanitised** — dropping the characters a session id forbids mapped `a.b` and `ab` onto one workspace, which is one person reading another's
- a scope with nothing to key on raises by name rather than falling back to a broader one, because the fallback would be silent
- changing the backend on a live conversation opens a new workspace rather than reattaching to one of a different shape
- `../../etc/passwd` becomes `etc_passwd`; two `report.csv` cannot overwrite each other
- publish refuses `docker` with no `SANDBOXD_URL`, and a `runtime` on a backend with no container
- `session_scope="user"` is *not* refused at publish: publishing cannot know which surfaces an agent will be reached from, and a web-only agent with a per-user workspace is a good configuration. The run refuses, by name, when it turns out there is nobody to attribute it to

## Verification

2007 tests pass; **100% coverage on every module this change adds or touches** (`app/agents/capabilities/sandbox/**`, `app/services/attachments.py`, `app/services/sandbox_workspace.py`, `app/repositories/agent_workspace.py`, and the changed lines in `agent_chat.py`, `agent_runner.py`, `agent_registry.py`). `ruff` and `ty` clean. All three compose files validate with the profile. `python3 scripts/docs_drift.py` reports no drift.

Three modules remain below 100% in a local run — `organization_secret` (repo and service) and `resource_grant` — and were equally uncovered before this branch; they are covered by `tests/integration/`, which skips without a database.

---

## Landed since this PR opened

- **Read-only workspace routes** on the conversation, authorised by fetching the conversation first, with the service token never leaving the process. A container-backed workspace is read off the host volume, so a week-old conversation lists its files after its session was reaped; a misconfigured service raises rather than returning an empty folder.
- **Deleting a conversation purges its workspace** on the service. The row cascades; a container would have sat on the host until its TTL.
- **The Builder gives the workspace a section of its own** — None / Files / Container / Daytona, the scope beside it, the runtime only where a container runs one. The generated form is suppressed for this one capability: a schema can render an enum, it cannot say that picking `agent` means a colleague reads your uploads. That warning is written out beside the field.
- **An audit entry when a workspace becomes shared**, which is the third half of the same decision (warning, ownership label, audit).
- **A Daytona workspace is billed to the organization that owns it.** The capability declared the secret and the service built the backend without it, so the key reached nothing and the SDK would have fallen back to `DAYTONA_API_KEY` from the environment — every tenant's sandboxes on whichever account the deployment configured, silently. Found while wiring the section; fixed with a test.

## Not in this PR

Real work, queued rather than dropped:

- **`sandbox:network` as an organization grant**, for a sandbox allowed to reach the
  internet. Today the network mode is the service's own configuration.
- **The `/dashboard` operator views**, which are @BartoszOchnik's ticket. The proxies
  they need are in this PR.
- **Content search in the workspace browser.** Names are searchable; content is not.
  Grepping a stored workspace is cheap - the document is already loaded - and grepping
  a container-backed one is a round trip per file, so one answer would be instant and
  the other unusable. A search box that behaves differently per host is worse than one
  that only matches names.
- **Downloading a binary from a container-backed workspace.**
  `WorkspaceArchive.read` returns `str`, so a PNG cannot be served faithfully and is
  refused rather than mangled. Filed as
  [pydantic-ai-backend#96](https://github.com/vstorm-co/pydantic-ai-backend/issues/96);
  a stored workspace serves any file today.
- **A `user`-scoped workspace created from Slack is invisible to the linked web
  account** - `owner_ref` is matched as a string. Filed as #131 with both ways out.

## Everything from the review on #36

The seven items from [this review](https://github.com/vstorm-co/agenticos/issues/36#issuecomment-5168354536), and what each turned into:

- **Approval defaults** — `write_file` and `edit_file` are no longer side-effecting;
  only `execute` is gated. Writing into a scratch space deleted with its conversation
  is not the same act as sending an email, and an agent that must ask before every
  write cannot do multi-step work at all.
- **Approval in `/chat` was a dead end** — nothing emitted `tool_approval_required`
  and the panel that handled it was unreachable template scaffolding. The event now
  carries the parked approval ids and the panel's buttons call the queue's own decide
  and resume endpoints, so the queue stays the source of truth.
- **A file panel in `/chat`** — beside the transcript, closed by default behind one
  icon, re-read when a turn ends rather than on a timer, and it names whose files
  these are.
- **More of what a turn cost** — the agent's own cap first (the organization's only
  past 80%), input and output split, and priced under each answer as well as under
  the input.
- **Add connection stopped asking for what the deployment knows** — `/local` probes
  `/healthz` and prefills the address, `/local/credential` stores this deployment's
  own token in the vault, `/runtimes` serves the library's whole catalog so the
  runtime field is a populated select, and `/probe` tests an address and a key before
  a row exists.
- **Workspaces is its own page, scoped per reader** — an operator sees the
  organization's; everybody else sees their own files, their own conversations, and
  the shared workspace of an agent they have talked to. **Open** goes to a real
  browser: folders, whole-tree search, previews, downloads.
- **"Runs on"** — the question, answered in the issue. Nothing to build.

Plus four defects found while using it: the tool-result event renamed under us (every
tool call in web chat answered with an `AttributeError`), a host that keeps no files on
disk reported as a 500 beside "nothing yet", sandbox tool calls rendered as raw JSON,
and a missing file reported as a broken host.

## CI

Green: `lint`, `test` (backend, coverage gate), `test-frontend` (100% frontend gate), `e2e`, `docs`, `Security Scan`. The `docker` job skips by design.

Two E2E specs have flaked on this branch and passed on a re-run of the same commit -
the vault rotation spec (#130) and the skill seed (#132). Both assert that a
just-created row is on screen and neither says anything useful when it is not; both are
pre-existing and filed.

Two e2e failures on the way were worth the trip. One was a flake that passed on re-run. The other was not: `sandbox` declares an api_key requirement for Daytona and registers before `web_research`, so the spec that drives the secret picker started finding it and then looked for its switch in a picker it no longer belongs to.


DEENUU1 (commented):
A read of this branch with fresh eyes, since the automated reviewer refused it on size and nobody else can hold 380 files in their head either.

The design holds up. Splitting `backend` (infrastructure) from `session_scope` (data-sharing policy) is the right seam, keeping the image and the mounts out of the spec is right, and `_workspace_bytes.py` is the best-reasoned file in the diff — refusing `.svg` and `.html` inline, `nosniff`, `filename*` only. `_identity._digest` hashing rather than sanitising, and `sessions()` filtering by tenant instead of proxying, are both the kind of thing that only gets written by someone who thought about the failure first.

Three things I'd fix before this lands, and they share a cause: **`RemoteSandbox` and `WorkspaceArchive` are synchronous `httpx.Client`, and the platform calls them straight from `async def`.** The library exports `ensure_async` for exactly this and its own toolset uses it, so the agent's tool calls are fine — it's only our own direct calls that block. That is nine call sites blocking the event loop, one leaked `httpx.Client` per workspace read (up to 25 on `GET /sandbox-workspaces/files`), and a lost update on the shared scopes where `version` claims to notice one and nothing reads it. Details inline.

Worth saying plainly: **the suite cannot catch any of this.** Every remote path is a hand-rolled fake class swapped in for `WorkspaceArchive`/`RemoteSandbox` (`tests/test_sandbox_workspace.py:712` and fourteen more), so the calls are instant, in-memory, and have no client to leak. 100% coverage here proves the lines ran, not that the I/O is safe — the fakes make both defects invisible by construction.

Two process notes. **This should have been several pull requests.** 57 commits and 380 files under a title about sandboxes, carrying an eleven-commit i18n sweep, usage reporting, skill proposals and a chat redesign — none of which a reader of "give an agent a workspace" expects. The reviewer refusing it is the system working. And the description is now stale in its first code block: `backend` is `Literal["state", "service"]` with a `connection_id`, not `Literal["state", "docker", "daytona"]`.

backend/app/services/sandbox_workspace.py
```diff
@@ -0,0 +1,920 @@
...
+        if row.backend == "state":
+            return WorkspaceContents(entries=stored_entries(dict(row.files or {})))
+        return await self._remote_entries(ctx, row)
+
+    async def _remote_entries(self, ctx: AuthContext, row: AgentWorkspace) -> WorkspaceContents:
+        try:
+            archive = await self._archive(ctx, row)
+            if archive is None:
+                return WorkspaceContents(entries=[])
+            return WorkspaceContents(entries=list(archive.ls(row.session_id or row.scope_key)))
```
DEENUU1: Blocking network I/O on the event loop. `WorkspaceArchive` wraps a synchronous `httpx.Client` (`remote/client.py:547`), so `archive.ls(...)` here blocks the whole worker for a round trip to sandboxd — not just this request, every request in the process.

The library already solved this and exports the fix: `ensure_async` is a public name in `pydantic_ai_backends.__init__`, and the library's own console toolset uses it on every call (`toolsets/console.py:301`). Our code is the only place reaching past it.

Same problem at nine sites:

| File | Line | Call |
|---|---|---|
| `sandbox_workspace.py` | 678 | `archive.ls` |
| `sandbox_workspace.py` | 729 | `archive.read` |
| `sandbox_workspace.py` | 354 | `stop(purge=True)` |
| `sandbox_workspace.py` | 387 | `RemoteSandbox(...).stop` |
| `attachments.py` | 191, 193, 220 | `exists`, `write` |
| `channels/attachments.py` | 214, 258 | `read_bytes`, `glob_info` |
| `skill_workspace.py` | 118, 192 | `write`, `read_bytes` |

The two that worry me most aren't this one. `attachments.py:193` writes a whole uploaded file over blocking HTTP — a 50 MB CSV stalls the worker for the entire upload, which is the exact case the PR description sells as the reason to have a workspace. And `skill_workspace.py:118` writes one file per skill resource inside `prepare`, so a run with five skills pays fifteen serial blocking round trips before the model sees a token.

`flat_files` is the one that compounds: 25 workspaces × one blocking `ls` each, serially, in one request.

Cheapest fix that covers all of it — wrap once where the backend is resolved rather than at each call site:

```python
from pydantic_ai_backends import ensure_async

# in _open_service, before returning
backend = ensure_async(backend)
```

That makes every method awaitable, so the call sites become `await archive.ls(...)` and the offload happens in the library's thread pool. Note its own docstring about sizing an executor: asyncio's default pool is `min(32, cpu+4)` and shared, so a few `npm install`-length `execute` calls will queue unrelated reads behind them. Worth passing an explicit `ThreadPoolExecutor` for the sandbox host.

backend/app/services/sandbox_workspace.py
```diff
@@ -0,0 +1,920 @@
...
+        distinct from the service being misconfigured: that raises.
+        """
+        if row.connection_id is None:
+            return None
+        from pydantic_ai_backends.remote import WorkspaceArchive
+
+        resolved = await self.connections.resolve(ctx, row.connection_id)
+        if resolved.kind != "docker" or not resolved.row.base_url:
+            return None
+        return WorkspaceArchive(resolved.row.base_url, token=resolved.token)
```
DEENUU1: This leaks a connection pool on every call. `WorkspaceArchive` builds and owns an `httpx.Client` when none is passed (`remote/client.py:547`, `_owns_client = True`) and exposes `close()` at `remote/client.py:585` to release it. Nothing here ever calls it.

One per `files_of`, per `read_text`, per `read_bytes` — and `flat_files` walks up to 25 workspaces, so a single `GET /sandbox-workspaces/files` opens and abandons up to 25 clients with their sockets. They only go back when the GC finalizer gets to them, which under load is later than you want.

Two fixes, and I'd take both. Close it:

```python
        archive = WorkspaceArchive(resolved.row.base_url, token=resolved.token)
        try:
            yield archive
        finally:
            archive.close()
```

(as an `@asynccontextmanager`, so the two callers get it for free) — or hand it a shared client the service owns.

And separately: `_archive` calls `self.connections.resolve(...)` per row, which is a query plus a vault unwrap per workspace. Inside `flat_files` that's a clean N+1 on top of the round trips. Resolving the distinct `connection_id`s once and reusing them would collapse both.

backend/app/db/models/agent_workspace.py
```diff
@@ -0,0 +1,136 @@
...
+    """Size of `files`, maintained on write so a cap can be enforced without
+    measuring the document on every read."""
+
+    version: Mapped[int] = mapped_column(Integer, nullable=False, default=0)
+    """Bumped on every flush.
+
+    Two turns of one conversation cannot normally run at once - the WebSocket
+    refuses a second while one is in flight - so this is not a retry loop. It is
+    how a concurrent flush is *noticed*: the loser logs rather than silently
+    overwriting the other turn's files.
```
DEENUU1: This docstring promises a safety property the code does not have, and `save_files` in the repository points at it as the explanation (`agent_workspace.py:75`, "The service logs the overlap; see ...version").

Nothing reads `version`. `_flush_state` (`sandbox_workspace.py:339`) writes the whole document and `save_files` bumps the counter unconditionally — there is no compare, so no loser, and nothing logs. `rg '\.version' app/services/` comes back empty. Since the reference docs are generated from these docstrings, we'd be publishing a page asserting detection that isn't there.

The premise is also narrower than the scopes. "The WebSocket refuses a second while one is in flight" is true per socket (`agent_session.py:96`) and covers `conversation` scope in web chat. It does not cover the scopes whose entire purpose is sharing:

- `agent` — one workspace across every user of the agent. Two people chatting at once are two sockets.
- `channel` — a Slack channel; two mentions land through the router, which has no such guard.
- `user` — one person on web and Slack at the same time.

Concretely, under `agent` scope: A opens with `{x}`, B opens with `{x}`, A writes `y` and flushes `{x, y}`, B writes `z` and flushes `{x, z}`. `y` is gone, no error, no log, and the agent already told A the file was saved.

Two honest ways out. Implement what the docstring says — read the version at open, compare at flush, log when it moved:

```python
    async def _flush_state(self, workspace: OpenWorkspace) -> None:
        ...
        if row.version != workspace.opened_at_version:
            logger.warning(
                "workspace_flush_overlapped",
                extra={"scope_key": workspace.scope_key, "expected": workspace.opened_at_version, "found": row.version},
            )
```

Or, if last-write-wins is the accepted trade for now, say that here instead and file the merge. What it must not do is describe a guard that doesn't exist — this is the docstring somebody will read in six months when a shared workspace loses a file.

backend/app/services/channels/attachments.py
```diff
@@ -0,0 +1,260 @@
...
+    Taken before the turn so what it added can be told from what it already had.
+    An unreadable workspace answers with an empty set, which makes the turn's
+    output look entirely new - the safe direction, because the alternative is
+    posting nothing at all.
+    """
+    try:
+        return _workspace_paths(backend)
+    except Exception:
+        logger.warning("workspace_snapshot_failed", exc_info=True)
+        return set()
```
DEENUU1: I think this fails in the unsafe direction, and the docstring above argues for the opposite. An empty `before` doesn't make "the turn's output look entirely new" — it makes *the whole workspace* look new, because `files_written` computes `paths - before`.

So when the snapshot round trip fails and the later one succeeds, a reply posts up to `MAX_OUTBOUND_FILES` files the turn never touched. `/uploads/` and `/skills/` are filtered, but nothing else is — and under `agent` or `channel` scope those are a colleague's files, posted into a Slack channel. "The alternative is posting nothing at all" is the better alternative here: nobody is harmed by a missing attachment.

It needs a transient failure — both functions call `_workspace_paths`, so a persistently broken glob fails both and posts nothing. Transient is exactly what a remote host over HTTP gives you, though.

Make the failure explicit rather than an empty set, and let `files_written` decline:

```python
def workspace_snapshot(backend: BackendProtocol) -> set[str] | None:
    """Every path the workspace holds right now, or `None` if it could not be read.

    `None` rather than an empty set: an empty one means "the workspace was empty",
    which would make every file already in it read as this turn's output.
    """
    try:
        return _workspace_paths(backend)
    except Exception:
        logger.warning("workspace_snapshot_failed", exc_info=True)
        return None
```

and in `files_written`, `if before is None: return DeliveredFiles([], [])`.

backend/app/services/sandbox_connection.py
```diff
@@ -0,0 +1,693 @@
...
+                details={"field": "secret_id"},
+            )
+        secrets = await self.secrets.resolve_for_bindings(ctx, [data.secret_id])
+        secret = secrets.get(data.secret_id)
+        if not isinstance(secret, ApiKeySecret):
+            raise BadRequestError(
+                message="That credential is not an API key, so it cannot authenticate a service",
+                details={"secret_id": str(data.secret_id)},
+            )
+        payload = await self._get_json(
```
DEENUU1, line 377: This is server-side request forgery with the response body handed back. `data.base_url` comes straight from the request, `_get_json` GETs it with a 5s timeout, and `probe_policy` returns `response.json()` to the caller. `SandboxProbeRequest.base_url` is `str = Field(min_length=1, max_length=512)` — no scheme check, no host check.

So a holder of `connections:manage` turns the API container into a fetch proxy for anything on its network that answers 200 with JSON on a GET: an internal admin API, an unauthenticated Kubernetes API, Elasticsearch, etcd, Prometheus. That matters here specifically because the network boundary is the thing this PR is built on — "`sandboxd` is the only service holding the Docker socket" is a statement about what the API container can reach, and this lets a tenant admin read through it. An org admin is not the deployment operator.

It also sends `X-Sandbox-Token` to whatever host is named, so a wrong or malicious address gets the organization's own service token.

Cloud metadata is partly self-defending (IMDSv2 needs a PUT, GCP needs a header), so I'd call this worth fixing rather than urgent. Validate the URL where it enters — a `field_validator` on `base_url` in the schema, since `create`/`update` store the same value and `_get_json` fetches that too:

```python
    @field_validator("base_url")
    @classmethod
    def _http_only(cls, value: str) -> str:
        parsed = urlparse(value)
        if parsed.scheme not in {"http", "https"} or not parsed.hostname:
            raise ValueError("A sandbox service address must be an http(s) URL with a host")
        return value
```

plus a deny list for link-local and loopback-outside-dev if you want the boundary to actually hold. `SANDBOXD_URL`-shaped allowlisting would be stronger still, but it costs the multi-host story this module exists for.

backend/app/services/sandbox_connection.py
```diff
@@ -0,0 +1,693 @@
...
+                details=details,
+            )
+        if response.status_code == 404:
+            raise NotFoundError(message="Sandbox session not found", details=details)
+        if response.status_code != 200:
+            raise BadRequestError(
+                message=f"The sandbox service answered {response.status_code}",
+                details=details,
+            )
+        payload: dict[str, Any] = response.json()
```
DEENUU1: `response.json()` is outside the `try`, so a 200 that isn't JSON raises `json.JSONDecodeError` and the operator gets a 500 — which is the one outcome this function was written to prevent. Pointing the probe at the wrong port is the commonest way to hit it: plenty of things answer 200 with HTML.

```suggestion
        try:
            payload: dict[str, Any] = response.json()
        except ValueError as exc:
            raise BadRequestError(
                message=f"The service at {base} answered, but not with JSON - check the address and the port",
                details=details,
            ) from exc
        return payload
```

backend/app/api/routes/v1/workspaces.py
```diff
@@ -0,0 +1,173 @@
...
+) -> Any:
+    """One file's text.
+
+    The path is a query parameter rather than part of the URL: workspace paths
+    contain slashes, and a path parameter would need either escaping the client
+    has to get right or a catch-all route that swallows the ones beside it.
+    """
+    content = await workspaces.read_file_of(ctx, workspace_id, path=path)
+    if content is None:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="No such file")
```
DEENUU1: This route and its twin on the conversation disagree about the error envelope for the same condition. Here it's `HTTPException`, so the body is `{"detail": "No such file"}`. `conversations.py:read_workspace_file` raises `NotFoundError` for the identical case, so that one gets `{"error": {"code": "NOT_FOUND", ...}}` through the handlers.

A client that reads `error.code` works against the conversation route and breaks against this one — and `_workspace_bytes.py`'s own module docstring is an argument that these two must not differ by surface. `/{workspace_id}/raw` beside it already raises `NotFoundError` via `_bytes_from`, so this file is inconsistent with itself too.

```suggestion
    content = await workspaces.read_file_of(ctx, workspace_id, path=path)
    if content is None:
        raise NotFoundError(
            message="No such file",
            details={"workspace_id": str(workspace_id), "path": path},
        )
    return WorkspaceFileContent(path=path, content=content)
```

(with `from app.core.exceptions import NotFoundError` at the top).

backend/app/api/routes/v1/_workspace_bytes.py
```diff
@@ -0,0 +1,65 @@
...
+            # `filename*` and nothing else: a workspace path can hold any UTF-8, and
+            # the bare `filename` form has no way to say so - a quote or a newline in
+            # it is a header-injection primitive rather than a filename.
+            "Content-Disposition": (
+                f"{'inline' if inline else 'attachment'}; filename*=UTF-8''{quote(name)}"
+            ),
+            # Everything off the list above is typed `application/octet-stream`, and
+            # this is what stops a browser deciding such a body is HTML after all -
+            # sniffing would hand back the inline-script hole the list refuses.
+            "X-Content-Type-Options": "nosniff",
```
DEENUU1, line 63: This file is the best-reasoned thing in the diff — refusing `.svg` and `.html` inline because "the agent wrote it" is not a trust boundary, `filename*` only so a newline in a path can't inject a header, `nosniff` so the octet-stream fallback can't be re-guessed. Nothing to change; I'd want this docstring quoted in `docs/`.

One cheap addition while you're here. `.pdf` served inline is a defensible call — modern viewers disable script and the reasoning above says why it's worth it — but the residual risk closes for free with a header:

```suggestion
            "X-Content-Type-Options": "nosniff",
            # Belt to the allowlist's braces: a document served inline runs with no
            # origin of its own, so script inside one an agent produced cannot reach
            # the application's DOM even if a viewer would have executed it.
            "Content-Security-Policy": "sandbox",
```

It has no effect on the raster types and doesn't stop a browser rendering a PDF.

backend/app/services/channels/attachments.py
```diff
@@ -0,0 +1,260 @@
...
+        except Exception:
+            logger.info("outbound_attachment_unreadable", extra={"path": path})
+            continue
+        if len(data) > MAX_OUTBOUND_BYTES:
+            refused.append(path)
+            continue
+        attachments.append(
+            OutgoingAttachment(
+                filename=path.rsplit("/", 1)[-1],
+                content=data,
```
DEENUU1, line 237: nit: every produced file goes out as `application/octet-stream`, including the chart the agent just rendered. Only Mattermost reads this field (`mattermost.py:94`), so the blast radius is small — but there it means a PNG posts as an unpreviewable blob, and the path already tells us better.

```suggestion
                mime_type=mimetypes.guess_type(path)[0] or "application/octet-stream",
```

backend/app/services/skill_workspace.py
```diff
@@ -0,0 +1,247 @@
...
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass, field
+from typing import Any
+
+from pydantic_ai_backends import BackendProtocol
+from pydantic_ai_skills._parsing import parse_skill_md
```
DEENUU1: nit: `_parsing` is private, so a patch release of `pydantic-ai-skills` can move it with no deprecation and this breaks at import time — the whole app, not one feature.

The reasoning for using their parser rather than writing a second one is right and I wouldn't change it. Since we own that library, the fix is on the other side: export `parse_skill_md` from `pydantic_ai_skills/__init__.py` and import the public name. Worth an issue there so this doesn't sit as a private import indefinitely.

backend/app/agents/capabilities/sandbox/_permissions.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Built fresh per call rather than shared as a module constant: a
+    `PermissionRuleset` is handed to a capability that lives for one run, and a
+    mutable default shared across every agent in the deployment is the kind of
+    thing that is fine until something writes to it.
+    """
+    off_limits = _deny_off_limits()
+    return PermissionRuleset(
+        default="allow",
+        read=OperationPermissions(default="allow", rules=off_limits),
```
DEENUU1, line 80: This ruleset is built correctly and then handed to something that never asks it the question.

`permissions=` reaches two places in `toolsets/console.py`: `_ruleset.requires_approval(...)` for the write/execute approval flags, and `_denied_tools(permissions)`, which unregisters a tool when `is_denied(ruleset, op)`. And `is_denied` is the whole of it:

```python
def is_denied(ruleset, operation) -> bool:
    """Whether `operation` is denied outright, so its tools must not exist."""
    if ruleset is None:
        return False
    return _default_action(ruleset, operation) == "deny"
```

It reads the operation's **default action**. Nothing anywhere reads `rules`. Since every operation below is `default="allow"`, the twenty patterns on each of `read` / `write` / `edit` are decoration.

Checked against this exact ruleset through `build_workspace(...)`, nothing else mocked:

```
read_file /.env              -> 'ROOT=sk-ROOTSECRET'
read_file /sub/.env          -> 'NESTED=sk-NESTEDSECRET'
read_file /credentials.txt   -> 'PASSWORD=sk-CREDSECRET'
read_file /etc/passwd        -> 'root:x:0:0'
write_file /sub/.env         -> 'Wrote 1 lines to /sub/.env'
```

The ruleset object is fine - `await PermissionChecker(workspace_ruleset()).check("read", "/sub/.env")` raises `PermissionDeniedError` exactly as intended. We just never run it.

The docstring above makes the claim I want to be true: "an agent that can `read_file` should not be able to exfiltrate `.env` by asking nicely". With `execute` on, path rules were never the boundary anyway - the docstring says so itself, and `cat` does not consult them. Which is precisely why the case to fix is `include_execute=False`: file tools, no shell, and this is the only thing between the agent and `**/.ssh/**`.

Two ways out. Enforce it on our side, where `CappedStateBackend` already sits and could carry it:

```python
class GuardedBackend:
    """Refuses a path before the backend sees it."""

    async def read(self, path: str, offset: int = 0, limit: int = 2000) -> str:
        await self._checker.check("read", path)   # raises PermissionDeniedError
        return self._backend.read(path, offset, limit)
```

Or push it into `pydantic-ai-backend`, since we own it, and have the toolset consult the checker per call - it is already async and already constructed there. I would do the second and file it, and until it lands, not let this module's docstring assert a boundary we do not have.

Either way this wants the assertion that would have caught it: `read_file` on a `SECRETS_PATTERNS` path is refused. `app/agents/**` is at 100% and this module is covered entirely by construction, because `workspace_ruleset()` is called from `build_workspace` - so every line runs and nothing ever asks what the rules do.

backend/app/services/sandbox_workspace.py
```diff
@@ -0,0 +1,957 @@
...
+        """Stop a run-scoped sandbox, as a courtesy rather than a guarantee.
+
+        `sandboxd` reaps idle sessions on its own, which is what makes this safe
+        to be best-effort: a run that crashes between opening a sandbox and
+        getting here leaves one behind for the idle timeout and no longer.
+        """
+        stop = getattr(workspace.backend, "stop", None)
+        if stop is None:
+            return
+        stop(purge=True)
```
DEENUU1: `DaytonaSandbox.stop()` takes no arguments, so this raises for every Daytona workspace:

```
>>> stop = getattr(sandbox, "stop")
>>> stop(purge=True)
TypeError: DaytonaSandbox.stop() got an unexpected keyword argument 'purge'
```

`close()` wraps this in `except Exception` and logs `workspace_close_failed`, so nothing surfaces but a warning - and the call that would have cleaned up is the one that crashes. `DaytonaSandbox.stop` is exactly the right call to make:

```python
def stop(self) -> None:
    """Delete the Daytona sandbox."""
    ...
    client.delete(sandbox)
```

So a run-scoped Daytona sandbox is never deleted, on the organization's own account, once per run. `RemoteSandbox.stop(purge: bool = False)` is the one with the parameter.

Since `getattr` already admits we do not know what we are holding, ask about the argument too:

```python
        stop = getattr(workspace.backend, "stop", None)
        if stop is None:
            return
        # `RemoteSandbox.stop` takes `purge`; `DaytonaSandbox.stop` takes nothing and
        # deletes the sandbox outright. Passing `purge` to the second raised a
        # `TypeError` that `close` swallowed, so a cloud sandbox on somebody else's
        # invoice was never deleted at all.
        if "purge" in inspect.signature(stop).parameters:
            await asyncio.to_thread(stop, purge=True)
        else:
            await asyncio.to_thread(stop)
```

This composes with the `to_thread` wrap - `purge=True` is the bug, not where it runs.

Why neither of us saw it: the test at `tests/test_sandbox_workspace.py:1411` fakes `class _Sandbox` with no `stop` attribute at all, and its docstring says "A Daytona sandbox exposes no `stop`, and `close` must not care." The real one does expose `stop`. That test proves the `stop is None` early return works, and the library never takes it.

backend/app/services/sandbox_workspace.py
```diff
@@ -0,0 +1,957 @@
...
+        return len(rows)
+
+    async def _purge_remote(self, ctx: AuthContext, row: AgentWorkspace) -> None:
+        if row.connection_id is None:
+            return
+        from pydantic_ai_backends.remote import RemoteSandbox
+
+        try:
+            resolved = await self._connection(ctx, row.connection_id)
+            if resolved.kind != "docker" or not resolved.row.base_url:
```
DEENUU1: The other half of the same bug: a Daytona workspace is never purged when its conversation is deleted. `resolved.kind` is `daytona`, so this returns before anything is called - the row goes and the sandbox stays.

`purge_for_conversation`'s own docstring is the argument for fixing it: "a container-backed workspace lives outside this database and would sit on the host until its TTL swept it". Daytona is the case with no TTL of ours *and* an invoice attached, which the module docstring already says out loud - "`daytona` is a cloud resource on the organization's own account, and the same applies with somebody else's invoice attached."

`_daytona` above already builds the object from the same two values this row holds:

```python
            if resolved.kind == "daytona":
                from pydantic_ai_backends import DaytonaSandbox

                sandbox = DaytonaSandbox(
                    api_key=resolved.token, sandbox_id=row.session_id or row.scope_key
                )
                await asyncio.to_thread(sandbox.stop)   # deletes it
                return
            if resolved.kind != "docker" or not resolved.row.base_url:
                return
```

The test at `tests/test_sandbox_workspace.py:1529` asserts today's behaviour as correct: "Its sandbox lives on the organization's own Daytona account, which keeps no session of ours to purge - so the row goes and nothing is called." The first clause is right and the conclusion does not follow. The reason not to reach for `RemoteSandbox` here is that it would send a Daytona key to a `sandboxd` that does not exist - which is an argument for calling `DaytonaSandbox` instead, not for calling nothing.

backend/app/services/sandbox_workspace.py
```diff
@@ -0,0 +1,957 @@
...
+        something nothing can outlive.
+        """
+        if scope == "run":
+            return None
+
+        existing = await workspace_repo.get_by_key(
+            self.db, organization_id=identity.organization_id, scope_key=key
+        )
+        if existing is not None:
+            return await workspace_repo.touch(self.db, workspace=existing)
```
DEENUU1, line 333: `touch` only moves `last_used_at`, so a reattached workspace keeps whatever `connection_id` and `session_id` it was created with - and `scope_key` cannot tell two connections apart.

`_BACKEND_LETTER` folds in the backend *kind* (`d` / `x`) and gives the reason: a stored document and a container "are not the same shape of thing". Two different hosts are not either, and both get `x`. So moving an agent from host A to host B, or from Docker to Daytona, computes the same key, finds the same row, and touches it. The run then writes to B while the row still names A - and every read resolves the host from the row: `_archive` (line 764), `_read_from`, `_remote_entries`, `_purge_remote`. The file panel lists A, which holds the old files or none, while the agent works in B; deleting the conversation purges A's session and orphans the real one on B.

The likeliest trigger is not a spec edit. `connection_id=None` means "the organization's default", so registering a second host and marking it default moves every existing workspace without anybody touching an agent.

Two shapes worth considering, and I would take the first:

- Fold the connection into the key for a service backend, the way the org prefix is already truncated into it. A moved agent then opens a *new* workspace instead of reattaching to one on a host it is no longer using - which is exactly what the backend letter already does for the other transition. Costs a key change and `scope_key` needing the resolved connection, so `open` would resolve before keying.
- Or keep the key and have `_row` re-point the row when `existing.connection_id != connection_id`, accepting that the old session is orphaned until its TTL. One comparison, and the row stops lying.

Either beats a row that quietly names the wrong host.

backend/app/services/attachments.py
```diff
@@ -0,0 +1,240 @@
...
+
+        if not backend.exists(path):
+            data = await get_file_storage().load(chat_file.storage_path)
+            result = backend.write(path, data)
+            if result.error is not None:
+                # A full workspace, most likely. The file is still worth
+                # mentioning - and for a parsed one the text is still usable -
+                # so this degrades to the inline path rather than vanishing.
+                logger.info("attachment_not_written", extra={"path": path, "reason": result.error})
+                return await self._without_workspace(chat_file)
```
DEENUU1: This degrades the wrong way round. A write only fails here because the file does not fit - 4 MB of JSONB by default - and `_without_workspace` answers by pasting the whole of `parsed_content` into the prompt. With `MAX_UPLOAD_SIZE_MB = 50`, that is up to 50 MB of text in one message.

A file small enough to paste safely would have fitted in the workspace. So this branch runs *only* for files too big to paste, which is the case the module opens by citing: "A fifty-megabyte CSV cannot be attached at all."

The comment says "the text is still usable", and it is - but the head is the usable part, not the whole file. What it cannot do is hand out `path`, because the write failed and `read_file` on it would error. So: the name, the size, the head, and a sentence saying it was not stored.

```suggestion
                return AttachmentPlan(reference=_unstored(chat_file), inline=None)
```

with something like:

```python
def _unstored(chat_file: ChatFile) -> str:
    """Named and sampled, but with no path offered.

    Never `_pasted`: this is only reached for a file too large to store, which
    makes it too large to put in a prompt. And no path, because the write failed -
    handing the model one it cannot read is worse than saying so.
    """
    parts = [
        f"\n---\nAttached file: {chat_file.filename} ({_size(chat_file)}, "
        f"{chat_file.file_type}) - too large for the workspace, so it was not stored"
    ]
    if chat_file.parsed_content:
        parts.append(f"\nFirst {HEAD_LINES} lines:\n```\n{_head(chat_file.parsed_content)}\n```")
    return "".join(parts)
```

Worth a glance while you are in here: `_without_workspace` inlines an image with no size check at all, while `_inline_image` immediately below it honours `SANDBOX_INLINE_IMAGE_MAX_BYTES`. Pre-existing on the no-workspace path, but this fallback is a new route into it.

backend/app/agents/capabilities/sandbox/_capped.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(self, backend: StateBackend, *, max_bytes: int) -> None:
+        self._backend = backend
+        self._max_bytes = max_bytes
+
+    @property
+    def files(self) -> dict[str, FileData]:
+        return self._backend.files
+
+    def _snapshot(self) -> dict[str, FileData]:
```
DEENUU1, line 57: Recording that I checked this one rather than believed it, because the argument is subtle and it holds. A one-level copy is only safe if the backend never mutates a `content` list in place, and the library cooperates on both paths: `StateBackend.write` builds a fresh `FileData` and rebinds `self._files[path]`, and `edit` does `stored["content"] = outcome.content.split("\n")` - a rebind, not a mutation. `.files` is the live dict too, so `_restore`'s `clear()` + `update()` reaches the backend rather than a copy. Nothing to change.

Two things that follow, worth a line here so the next reader does not have to derive them:

`_too_large` measures the absolute size, not the delta, so an over-cap document can only be brought back under in a single move - a partial cleanup is refused and rolled back. Reaching that needs `SANDBOX_STATE_MAX_BYTES` to be *lowered* under a document already stored, so I would not add a delta check; I would say so in the docstring.

And "Delete or shorten something first" names an act the `state` backend has no tool for. `StateBackend` exposes no `delete` and `WORKSPACE_TOOLS` declares none, so `edit` is the only way to shrink a document. Shortening is the whole of the available advice:

```suggestion
            f"The workspace is full: this would take it to {size} bytes, over the "
            f"{self._max_bytes}-byte limit. Shorten or overwrite something first, or "
            "keep large intermediate results out of the workspace."
```

(that suggestion belongs on `_too_large` below, not this line - GitHub will only let me anchor it here.)

DEENUU1 (commented):
A second pass, hunting in the places the first one skipped: the capability package, the tool declarations, and what the library actually *does* with the objects we hand it. Two things I'd fix before this lands, and they share a cause with the last three - the code is right about its intent and the library disagrees quietly.

**`workspace_ruleset()` is never evaluated.** `permissions=` reaches exactly two things in the console toolset, and neither reads `rules` - only each operation's `default` action. Every operation here is `default="allow"`, so all twenty deny patterns are decoration, and an agent reads `/etc/passwd`, `**/.env` and `**/*.pem` freely. The configuration where it matters most is the one where it is also inert: `include_execute=False` leaves the file tools and no shell, so this ruleset is the only thing standing there.

**A Daytona sandbox is never deleted, on any path.** `_release` calls `stop(purge=True)` on a `stop()` that takes no arguments, so it raises a `TypeError` that `close` swallows; `_purge_remote` returns early for anything that is not Docker. The call that would clean it up - `DaytonaSandbox.stop`, which does `client.delete(sandbox)` - is never made, and every sandbox stays on the organization's own account.

Two smaller ones inline: a workspace that moves to a different connection reattaches to a row still naming the old host, and the full-workspace fallback in `attachments.py` pastes the file it just refused to store.

Good news from the sweep, both checked rather than assumed. The invariant the registry docstring stakes out holds exactly - I enumerated what the toolset registers under our settings and it is the declared seven, no more, with `include_background=False` doing the work it claims. And `_capped._snapshot`'s one-level copy is correct for the right reason; note inline.

Anchors are against 2122ec3. I have a local commit and a dirty tree past it, so the `sandbox_workspace.py` lines will drift.

backend/app/agents/capabilities/sandbox/_permissions.py
DEENUU1: Fixed in 56a41d8. `GuardedBackend` wraps the backend at `build_workspace`, the one seam where a workspace meets the model's tools, and asks the checker on the way past.

Went slightly wider than the comment proposed, for one reason: `grep_raw` returns `GrepMatch` with the matching *line*, so a guard on `read` alone still handed over the contents of exactly the files the ruleset protects. It is filtered rather than refused, since a pattern over `/` legitimately covers the workspace. `exists`/`ls`/`glob` still delegate - names are not secret, only contents.

Ended up async rather than sync. A sync guard over an `AsyncBackendProtocol` returns un-awaited coroutines and `ty` refused it, correctly; `ensure_async` inside the guard covers both shapes and costs nothing already unpaid, since the toolset calls it per tool call anyway.

Verified through the real toolset - all four off-limits reads and the off-limits write now answer `Error: Permission denied: ...`, `grep PASSWORD` no longer matches `/credentials.txt`, and ordinary files are untouched. Tests assert the consequence rather than the ruleset's shape.

backend/app/services/sandbox_workspace.py
DEENUU1: Fixed in 752fe12, with the signature check rather than a type check - `getattr` was already duck-typing this and an `isinstance` would mean importing an SDK that `_daytona` deliberately imports lazily.

The old test is kept for the branch it does cover (a backend with no `stop` at all) with its docstring corrected, and two new ones assert each signature reaches its own call. Confirmed the Daytona one fails without the change and passes with it.

backend/app/services/sandbox_workspace.py
DEENUU1: Fixed in 9ec0608. `DaytonaSandbox(api_key=..., sandbox_id=...).stop()` built from the same two values `_daytona` uses when opening one, inside the existing `except` so a bad day at the provider still lets somebody delete their chat.

The test that asserted the defect as intended is replaced by one checking the sandbox is deleted with the organization's own key and the session id the row recorded, plus one for an unreachable host.

One thing fell out of it: with Daytona returning above, `resolved.kind != "docker"` became unreachable - a connection is one or the other - so the coverage gate caught it at 99.97%. Reduced to the address check, which is the half still worth refusing, and covered.

backend/app/services/sandbox_workspace.py
DEENUU1: Fixed in 113f334, taking the first of the two shapes - the connection is folded into the key.

It keeps the old workspace honest, which the re-pointing option does not: the files really are still on the old host, so the row naming it stays listed and readable, and a moved agent gets an empty workspace on the new one. That is what moving hosts means, and it is the same argument `_BACKEND_LETTER` already makes.

`open` now resolves the connection before building the key. The scope is still checked first, so an agent reached from a surface it cannot be keyed on says so rather than reporting whatever is wrong with a host it was never going to use - there is a test for that ordering. Keys are 53 characters against 44, both inside the 64 `sandboxd` allows.

backend/app/services/attachments.py
DEENUU1: Fixed in f105e7e. `_unstored` gives the name, size, type, head and a sentence saying it was not stored - and deliberately no path, since the write failed.

Kept the old route for an image, for the reason images go both ways at all: a path is no substitute for looking at the picture, and `_inline_image` has its own much smaller ceiling rather than the document's.

The existing test only asserted the head appeared, which the paste satisfies too; the new one fails without the change by asserting the tail of a 500-line file is absent.

The unbounded image inline in `_without_workspace` is noted in the commit body as not fixed here - happy to take it as its own change.

backend/app/agents/capabilities/sandbox/_capped.py
DEENUU1: Both notes written down in bd6b08f, and the message fixed.

"Delete or shorten" is now "Shorten or overwrite", since there is no delete to reach for - with a test asserting the advice names only moves the agent has, and one proving the overwrite it now recommends actually frees the room.

The absolute-size behaviour is documented rather than changed, with the reason: reaching it needs `SANDBOX_STATE_MAX_BYTES` lowered beneath an already-stored document, which is an operator's change and not something an agent can do to itself. And the `_snapshot` reasoning now names what makes it safe - `write` rebinds `files[path]`, `edit` assigns `stored["content"]`, `files` is the live dict - so the next person adding a backend method knows the property the rollback depends on.

Issue #37, opened 2026-08-01
Design and build the dashboard (HTML demo first, role-aware)


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576806d96daefd0a06fdeb1) · Priority: **Mid** · Category: Development

> **This ships in two stages. Stage 1 is an HTML demo, reviewed by @DEENUU1 before a line of React is written.** See *Deliverables* at the bottom.

## What the brief actually asks for

From the Notion card, verbatim in substance:

- The dashboard is a blank page today. **Analyse the available data, then design it.**
- **Assume a less technical reader.** This is not an operator's console.
- **Numbers *and* charts**, not one or the other.
- **It differs by role.** An app admin sees how many organizations exist; an org admin sees how many members are in theirs. Same page, different answers.
- **The statistics should show how people actually use the system** — adoption, not just health.
- **Start with an HTML demo that can switch between roles**, discuss it with @DEENUU1, then implement.

## Where we are

`frontend/src/app/[locale]/(dashboard)/dashboard/page.tsx` is 23 lines: a `PageHeader` and an `EmptyState` with a `Construction` icon reading *"the widgets that used to live here have been removed"*. It is first in the sidebar and it is where an app admin lands after signing in (`use-auth.ts:92`), so the first thing a deployment owner sees is a construction sign.

## Data inventory — do this before designing

**Roles that need an answer.** `app/core/permissions.py` defines six org roles — `owner`, `admin`, `builder`, `operator`, `member`, `viewer` — plus the deployment-level `is_app_admin` flag, which is not a role and is checked separately. Seven personas is too many pages; the design should collapse them into **three or four audiences** and say which role maps to which.

**What is answerable today, with no backend work:**

| Question | Endpoint | Scope |
|---|---|---|
| Users, orgs, agents, conversations, messages, active users 24h | `GET /api/v1/admin/stats` | app admin only |
| Every tenant with member + agent counts | `GET /api/v1/admin/organizations` | app admin only |
| Per-service health | `GET /api/v1/admin/system` | app admin only |
| Month-to-date spend, by agent / provider / key | `GET /api/v1/spend?days=` | org, `runs:view` |
| Runs, newest first, filterable by agent | `GET /api/v1/runs` | org, `runs:view` |
| Pending approvals | `GET /api/v1/approvals` | org, `approvals:decide` |

[...]

Issue #14, opened 2026-08-01
Stop putting tokens in the OAuth redirect URL


DEENUU1:
**Severity: High** · AUD-031

## What's wrong

```python
# backend/app/api/routes/v1/oauth.py:44-53
access_token = create_access_token(subject=str(user.id))
refresh_token = create_refresh_token(subject=str(user.id))
params = urlencode({"access_token": access_token, "refresh_token": refresh_token})
return RedirectResponse(url=f"{frontend}/auth/callback?{params}")
```

## Where

- `backend/app/api/routes/v1/oauth.py:44-53`
- Consumed at `frontend/src/app/[locale]/auth/callback/page.tsx:17-18`

## How it fails

A Google sign-in lands the user on `https://app/…/auth/callback?access_token=<jwt>&refresh_token=<jwt>`. Both tokens are therefore written to:

- the Next.js server's request log and any reverse proxy in front of it
- the browser address bar and session history, until `router.replace` runs at line 44
- the `Referer` of every same-origin request the page then makes — `Referrer-Policy: strict-origin-when-cross-origin` (`frontend/next.config.ts:47-49`) sends the *full* URL same-origin, so the subsequent `POST /api/auth/oauth-callback` carries both tokens in its `Referer`, and that request is logged too

The refresh token is good for seven days (`REFRESH_TOKEN_EXPIRE_MINUTES = 60 * 24 * 7`). Anyone with read access to an access log has a week of full account access.

Nothing in `docs/` records this as a decision — `docs/configuration.md:78-85` documents the OAuth settings and says nothing about token delivery.

## Fix

Have the backend set the cookies itself on the callback, or redirect with a single-use short-TTL exchange code that `/api/auth/oauth-callback` swaps for the token pair server-to-server. Either way the frontend page never sees a token.

## Related, worth doing in the same change

`frontend/src/app/api/auth/oauth-callback/route.ts` accepts an arbitrary token pair from the request body and mints cookies from it — a session-fixation shape. It could not be shown exploitable (a cross-origin JSON POST needs a preflight Next does not answer, and there is no CORS config on the frontend app), so it is not a finding on its own. But it is the route to revisit if this is fixed by any means other than removing the client-supplied tokens.

## Acceptance criteria

- [ ] The redirect `Location` contains no `access_token` or `refresh_token`
- [ ] Backend test asserting that
- [ ] Frontend test asserting the callback page posts a code, not tokens

PR #121, opened 2026-08-03
feat(auth): land every sign-in on the dashboard and honour deep links

### LINKED ISSUE #38: Set 'dashboard' as the landing page after sign-in
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680998f0ce61cfa1de2e6) · Priority: **Low** · Category: Development

> **Blocked by #37.** Do not do this first. `/dashboard` is currently an "Under construction" stub — making it the landing page for everyone would replace a working chat window with a construction sign for every user in the deployment.

## What is actually there today

`frontend/src/hooks/use-auth.ts:92`, on successful sign-in:

```ts
router.push(isAppAdmin(response.user) ? ROUTES.DASHBOARD : ROUTES.CHAT);
```

So app admins already land on `/dashboard` — and see the stub. Everyone else lands on `/chat`. The task is to collapse that fork: everyone lands on `/dashboard`.

Separately, `frontend/src/app/[locale]/page.tsx` redirects `/` unconditionally to the login form. That is deliberate and documented in the file; leave it alone unless you also want an authenticated visitor at `/` to skip the form, which is a different task.

## Do it properly, not with a one-line swap

Three call sites and one behaviour matter:

1. **Post-login redirect** — remove the `isAppAdmin` fork. Everyone goes to `ROUTES.DASHBOARD`.
2. **Post-OAuth callback** — `frontend/src/app/[locale]/auth/callback/page.tsx` has its own landing logic. Both paths must agree, or Google sign-in and password sign-in put people in different places. (This file is also touched by #14 — coordinate.)
3. **`returnTo`** — if the user was sent to the login form from a deep link, they should land back on that link, not on the dashboard. Check whether this exists today; if it does not, the default must not swallow it silently.

And the one that will be missed: **a user with no permission to see the dashboard's widgets still has to land somewhere useful.** After #37, a Viewer's dashboard may be nearly empty by design (no `runs:view` → no runs, no spend). Landing them on a blank page is worse than landing them on `/chat`. Either the dashboard has something for every role, or the redirect stays role-aware — decide, and put the decision in a comment so the next reader does not "simplify" it back.

## Done when

- [ ] #37 is merged first
- [ ] One landing destination, shared by password sign-in and OAuth callback
- [ ] A deep link survives the login round trip
- [ ] A Viewer landing on `/dashboard` sees something useful, or the fork is kept deliberately with a comment saying why
- [ ] A test covering the redirect for both an app admin and an ordinary member

OchnikBartek:
## Summary

Every sign-in path — password, the OAuth callback and the magic link — now resolves
its redirect through one shared destination and lands on `/dashboard`, replacing
three call sites that each decided on their own and disagreed (password sign-in
forked on `is_app_admin`, OAuth always went to `/dashboard`, the magic link always
to `/chat`). A deep link interrupted by the login form is now carried through
`?returnTo=` and resumed after signing in.

## Related Issue

Closes #38.

## Added

- `src/lib/auth-landing.ts` — `postSignInDestination(returnTo?)`, the single
  post-sign-in landing decision, with an open-redirect guard that refuses
  off-origin values (`//host`, `https://…`, `/\host`, scheme and relative
  paths); colocated `auth-landing.test.ts`.
- `returnTo` round trip: `AuthGuard` appends the refused path as `?returnTo=`
  when it sends a visitor to `/login`, the login form reads it at submit time,
  and `login()` forwards it to the destination helper.
- E2E spec `a deep link survives the login round trip` in `e2e/auth.spec.ts`.

## Changed

- `useAuth().login()` no longer forks on `isAppAdmin` — everyone lands on the
  shared destination; it now accepts an optional `returnTo`.
- The OAuth callback and the magic-link page redirect through
  `postSignInDestination()` instead of their hardcoded `/dashboard` and `/chat`.
- The mobile tab bar's Home tab targets `/dashboard` for every role; its unused
  `useAuth` dependency and the stale mock in `mobile-tab-bar.test.tsx` are gone.
- E2E landing assertions tightened from `/(chat|dashboard)/` to `/dashboard/`
  in `auth.setup.ts`, `auth.spec.ts` and `seed.setup.ts`.

## Checks

## Notes for Reviewers

- #38 is labelled "Blocked by #37" — waived deliberately: the deployment has no
  users yet and the dashboard stub links onward to chat and agents, so landing
  everyone there until #37 fills the page costs one extra click. The issue's
  "a Viewer sees something useful" checkbox is satisfied as its documented-
  decision variant: the one-landing-for-every-role rationale lives in the
  `postSignInDestination` docstring.
- A deep link through OAuth is deliberately out of scope — it needs the `state`
  parameter round trip, and #14 is rewriting that flow anyway.
- `window.location` instead of `useSearchParams`/`usePathname` in the login
  form and `AuthGuard` is intentional: the former would demand a Suspense
  boundary, the latter would tie the guard's verify effect to every navigation.
- Verified locally: `make check` green end to end (backend 2107 passed, platform
  gate at 100.00%; frontend gate 100/97.6/100/100 plus `next build`), and the
  E2E suite against a live stack — 88 passed; the 4 failures assert on the
  `openai default` bootstrap fixture a long-lived dev database does not have,
  so they fail there on any branch. CI seeds that fixture fresh.

### DIFF
--- a/frontend/src/components/auth/login-form.tsx
+++ b/frontend/src/components/auth/login-form.tsx
@@ -30,7 +30,10 @@ export function LoginForm() {
     setError("");
 
     try {
-      await login({ email, password });
+      // Read off the URL at submit time: useSearchParams would demand a
+      // Suspense boundary around a form that renders statically.
+      const returnTo = new URLSearchParams(window.location.search).get("returnTo");
+      await login({ email, password }, returnTo);
       toast.success(t("loginSuccess"));
     } catch (err) {
       const message = err instanceof ApiError ? err.message : t("loginFailedPleaseTry");
--- a/frontend/src/hooks/use-auth.test.tsx
+++ b/frontend/src/hooks/use-auth.test.tsx
@@ -286,6 +286,26 @@ describe("signing in", () => {
     expect(push).toHaveBeenLastCalledWith("/agents/a-1?tab=spec");
   });
 
+  it("resumes a deep link with a fragment by document navigation, not the router", async () => {
+    // The router cannot be trusted with a fragment: a soft navigation lands on
+    // /path#x#x in a production build (next@16.2 segment cache), so the
+    // fragment case must leave the SPA and load the document.
+    const assign = vi.fn();
+    vi.spyOn(window, "location", "get").mockReturnValue({
+      ...window.location,
+      assign,
+    } as unknown as Location);
+    vi.mocked(apiClient.post).mockResolvedValue({ user: user(), access_token: "t", message: "ok" });
+    const { result } = renderHook(() => useAuth(), { wrapper });
+
+    await act(async () => {
+      await result.current.login({ email: "a@example.com", password: "x" }, "/agents/a-1#monthly");
+    });
+
+    expect(assign).toHaveBeenCalledWith("/agents/a-1#monthly");
+    expect(push).not.toHaveBeenCalledWith("/agents/a-1#monthly");
+  });
+
   it("refuses a returnTo that leaves the origin", async () => {
     // ?returnTo= arrives in the URL, so anybody can mint a login link with it;
     // honouring an off-origin value would turn the form into an open redirect.
--- a/frontend/src/hooks/use-auth.test.tsx
+++ b/frontend/src/hooks/use-auth.test.tsx
@@ -73,6 +73,10 @@ beforeEach(async () => {
 
 afterEach(() => {
   vi.useRealTimers();
+  // `clearAllMocks` in beforeEach clears calls but does not undo a `spyOn`;
+  // without this, a test that spies on `window.location` freezes it for
+  // every test after it.
+  vi.restoreAllMocks();
 });
 
 describe("checking the session on load", () => {
--- a/frontend/src/hooks/use-auth.test.tsx
+++ b/frontend/src/hooks/use-auth.test.tsx
@@ -255,7 +255,7 @@ describe("signing in", () => {
     expect(useAuthStore.getState().accessToken).toBe("t-1");
   });
 
-  it("sends an app admin to the dashboard and everybody else to the chat", async () => {
+  it("lands an app admin and an ordinary member on the same dashboard", async () => {
     vi.mocked(apiClient.post).mockResolvedValue({
       user: user({ is_app_admin: true } as Partial<User>),
       access_token: "t-1",
@@ -272,7 +272,31 @@ describe("signing in", () => {
     await act(async () => {
       await result.current.login({ email: "b@example.com", password: "x" });
     });
-    expect(push).toHaveBeenLastCalledWith(ROUTES.CHAT);
+    expect(push).toHaveBeenLastCalledWith(ROUTES.DASHBOARD);
+  });
+
+  it("resumes the deep link the visitor was headed to", async () => {
+    vi.mocked(apiClient.post).mockResolvedValue({ user: user(), access_token: "t", message: "ok" });
+    const { result } = renderHook(() => useAuth(), { wrapper });
+
+    await act(async () => {
+      await result.current.login({ email: "a@example.com", password: "x" }, "/agents/a-1?tab=spec");
+    });
+
+    expect(push).toHaveBeenLastCalledWith("/agents/a-1?tab=spec");
+  });
+
+  it("refuses a returnTo that leaves the origin", async () => {
+    // ?returnTo= arrives in the URL, so anybody can mint a login link with it;
+    // honouring an off-origin value would turn the form into an open redirect.
+    vi.mocked(apiClient.post).mockResolvedValue({ user: user(), access_token: "t", message: "ok" });
+    const { result } = renderHook(() => useAuth(), { wrapper });
+
+    await act(async () => {
+      await result.current.login({ email: "a@example.com", password: "x" }, "//evil.example/a");
+    });
+
+    expect(push).toHaveBeenLastCalledWith(ROUTES.DASHBOARD);
   });
 
   it("does not ask who is signed in after a login that just said so", async () => {
--- a/frontend/src/hooks/use-auth.ts
+++ b/frontend/src/hooks/use-auth.ts
@@ -176,7 +176,15 @@ export function useAuth() {
         adoptUser(queryClient, setUser, response.user);
         useAuthStore.getState().setAccessToken(response.access_token);
         authChecked = true; // login already populated user + token; skip /auth/me
-        router.push(postSignInDestination(returnTo));
+        const destination = postSignInDestination(returnTo);
+        if (destination.includes("#")) {
+          // next@16.2's segment cache appends the fragment a second time on a
+          // soft navigation (/path#x becomes /path#x#x in a production build),
+          // so a destination with a fragment must load the document instead.
+          window.location.assign(destination);
+        } else {
+          router.push(destination);
+        }
         return response;
       } finally {
         setLoading(false);
--- a/frontend/src/hooks/use-auth.ts
+++ b/frontend/src/hooks/use-auth.ts
@@ -8,8 +8,8 @@ import { toast } from "sonner";
 import { resetSessionState, useAuthStore } from "@/stores";
 import { apiClient, ApiError } from "@/lib/api-client";
 import type { User, LoginRequest, RegisterRequest } from "@/types";
+import { postSignInDestination } from "@/lib/auth-landing";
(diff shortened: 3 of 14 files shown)


frontend/src/hooks/use-auth.test.tsx
```diff
@@ -272,7 +272,51 @@ describe("signing in", () => {
...
+
+  it("resumes a deep link with a fragment by document navigation, not the router", async () => {
+    // The router cannot be trusted with a fragment: a soft navigation lands on
+    // /path#x#x in a production build (next@16.2 segment cache), so the
+    // fragment case must leave the SPA and load the document.
+    const assign = vi.fn();
+    vi.spyOn(window, "location", "get").mockReturnValue({
+      ...window.location,
+      assign,
+    } as unknown as Location);
```
DEENUU1, line 301: **This spy is never restored, and it leaks into every test below it.**

`vitest.config.ts` does not set `restoreMocks`, and the `vi.clearAllMocks()` in `beforeEach` clears calls but does not restore a `spyOn`. So from this line to the end of the file, `window.location` is the frozen snapshot spread here, not the real one.

I reproduced it — appended this to the bottom of the file and ran it:

```ts
it("window.location still tracks navigation", () => {
  window.history.replaceState(null, "", "/probe-path");
  expect(window.location.pathname).toBe("/probe-path");
});
```

`AssertionError: expected '/' to be '/probe-path'`.

Nothing breaks today, because the tests below only need `origin` and the spread carried it over. But the next test somebody adds here that navigates will fail for a reason that has nothing to do with it — and this file is where auth tests go. One line in the existing `afterEach` at line 74 fixes it, and I checked the other 31 tests stay green:

```ts
afterEach(() => {
  vi.useRealTimers();
  vi.restoreAllMocks();
});
```

frontend/src/hooks/use-auth.ts
```diff
@@ -176,7 +176,15 @@ export function useAuth() {
...
-        router.push(isAppAdmin(response.user) ? ROUTES.DASHBOARD : ROUTES.CHAT);
+        const destination = postSignInDestination(returnTo);
+        if (destination.includes("#")) {
+          // next@16.2's segment cache appends the fragment a second time on a
+          // soft navigation (/path#x becomes /path#x#x in a production build),
+          // so a destination with a fragment must load the document instead.
+          window.location.assign(destination);
+        } else {
+          router.push(destination);
+        }
```
DEENUU1, line 191: **The destination is shared by three call sites. The knowledge of how to navigate to it is not.**

This is the same shape the PR is fixing, one level up. `postSignInDestination` is now the single answer to *where*, which is right. But "a destination with a fragment must load the document" is the single answer to *how*, and it only exists here — `auth/callback/page.tsx:52` and `auth/magic-link/page.tsx:47` both do a plain `router.replace(postSignInDestination())`.

Neither can produce a fragment today, because neither passes a `returnTo`. But the PR body says #14 is adding the OAuth `state` round trip: the day the callback starts forwarding a deep link, it hands a fragment to `router.replace` and hits exactly the bug this comment describes, with no test to catch it. Keeping the pair together closes it:

```ts
// auth-landing.ts
export function navigateAfterSignIn(
  router: { push: (href: string) => void },
  returnTo?: string | null,
): void {
  const destination = postSignInDestination(returnTo);
  // next@16.2's segment cache appends the fragment a second time on a soft
  // navigation (/path#x becomes /path#x#x in a production build), so a
  // destination with a fragment must load the document.
  if (destination.includes("#")) window.location.assign(destination);
  else router.push(destination);
}
```

Small thing while you are in here: `location.assign` is a full document load, so the fragment branch drops the in-memory `accessToken` and re-runs `/auth/me` on arrival, and the `toast.success(t("loginSuccess"))` at `login-form.tsx:37` fires into a page that is unloading. Both cosmetic, but the two branches are not equivalent and the comment currently reads as if they were — worth a sentence.

frontend/src/components/auth/login-form.tsx
```diff
@@ -30,7 +30,10 @@ export function LoginForm() {
     setError("");
 
     try {
-      await login({ email, password });
+      // Read off the URL at submit time: useSearchParams would demand a
+      // Suspense boundary around a form that renders statically.
+      const returnTo = new URLSearchParams(window.location.search).get("returnTo");
```
DEENUU1, line 35: **The OAuth gap is wider than the waiver suggests, and cheaper to close than it says.**

`OAuthBlock` sits at the bottom of this same form. So a visitor who arrives at `/login?returnTo=/agents/a-1` and signs in with Google loses the link silently — no message, they just land on `/dashboard`. Password sign-in now resumes and OAuth does not, which means *which button they pick* decides where they land. That is the drift this PR set out to remove, just moved from the roles axis to the provider axis.

The comment at `auth/callback/page.tsx:50` waives it on needing the OAuth `state` parameter, and I do not think that is right. `state` is for carrying data through an untrusted third party. Both ends of this round trip are the same tab on the same origin, so `sessionStorage` is enough and it does not touch the flow #14 is rewriting:

```ts
// login page, before redirecting to the provider
sessionStorage.setItem("returnTo", returnTo ?? "");

// auth/callback/page.tsx, after adoptSession
const returnTo = sessionStorage.getItem("returnTo");
sessionStorage.removeItem("returnTo");
router.replace(postSignInDestination(returnTo));
```

`postSignInDestination` already refuses anything hostile, so nothing new gets trusted. Entirely fine as a follow-up issue if you would rather keep this branch narrow — I would just fix the *reason* in that comment either way, since as written it tells the next reader the gap is blocked on something it is not blocked on.

DEENUU1 (commented):
Read this against the whole auth flow rather than the diff — the three sign-in paths, `AuthGuard`, the next-intl middleware, the persisted auth store, and every other site that redirects to `/login`.

The centralization is right, and the hardening you put on top of the bot's finding is genuinely good. I fuzzed `isSafeReturnPath` with control characters, `%09`, `%2F%2F`, `%5C%5C`, NUL/VT/FF and the backslash variants — nothing resolves off-origin, and nothing throws. The docstring explaining why *both* checks are load-bearing is the part that stops someone deleting one of them in six months. I also like that the e2e assertion is anchored (`#list$`), so it really does catch the `#list#list` regression the `location.assign` workaround exists for: the workaround is evidenced end to end against a real `next start` build, not just asserted in a comment. Tightening `/(chat|dashboard)/` to `/dashboard/` in three files was right too — the old regex passed either way.

Three non-blocking things below. One is a test-isolation bug I reproduced locally (the `window.location` spy is never restored, so every test after it in that file reads a frozen location). The other two are about scope: where the fragment workaround lives, and the OAuth `returnTo` gap.

One more, clearly **not** caused by this PR and better as an issue than a change here: `AuthGuard` returns early on the *persisted* `isAuthenticated` — `stores/auth-store.ts:82` keeps `user` and `isAuthenticated` in localStorage — so the `returnTo` path only engages for a visitor whose localStorage says signed-out. The commonest case, localStorage still saying authenticated while the cookie has expired, never reaches the `catch` at all: `lib/api-client.ts:96` refreshes once, fails, throws, and nothing redirects to login. So "a deep link survives the login round trip" holds for the fresh-visitor case #38 described, and not for the expiry case. Say the word and I will file it.

DEENUU1:
Rebased onto `main` and merging this for 0.0.5 — flagging what I touched, since I force-pushed your branch.

**The conflict was one hunk**, in `backend/pyproject.toml`: `main` gained `pydantic-ai-backend[console,remote]>=0.2.25` in #119 and your branch predated it. Kept main's block. Your `build(deps): bump cryptography to 50.0.0` commit dropped out of the rebase because that bump is already on `main` — so the branch is four commits now, not five, and `backend/` is untouched by this PR.

**I read the guard rather than taking the description's word for it**, since `?returnTo=` off a login form is the one thing here that could be a real hole. It holds. I threw fourteen bypasses your nine cases don't cover at it — vertical tab and form feed (which the URL parser does *not* strip, unlike tab/LF/CR), NUL, NEL, U+2028, `%2F%2F`, `https:/single-slash`, `/..//`, `/@host`, and tabs on both sides of the slashes — and every one lands back on our own origin. The two-check structure is doing exactly what its docstring claims.

Nothing changed in your code. `make check` green on the rebased tip: backend 2702 passed with the platform gate at 100%, frontend 100/97.58/100/100.

One note for later, not a blocker and not yours to fix here: the waiver on #38's "a Viewer sees something useful" is still a waiver. #149 is what fills that page, and it now has `main` to stand on.

Issue #45, opened 2026-08-01
Improve the Activity page (/runs) — plan + mockup first


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680e1af53c9a31c55199d) · Priority: **Mid** · Category: Development

> **This ships in two stages. Stage 1 is a plan plus an HTML mockup, reviewed by @DEENUU1 before implementation.** See *Deliverables*.

## What the brief asks for

From the Notion card: analyse the available data, then improve the page. **Decide whether to build the view on our own data or pull it from Logfire.** Produce a plan and take it to @DEENUU1.

## Where we are

"Activity" in the sidebar points at `ROUTES.RUNS` → `frontend/src/app/[locale]/(dashboard)/runs/page.tsx`, 342 lines, three tabs — **Approvals** (default), **Runs**, **Spend** — backed by `useApprovals()`, `useRuns(agentId?)` and `useSpend(30)`.

## The Logfire question — answered by the spec, not by preference

`AgentSpec.observability` exists (`app/agents/spec.py:311`) and its docstring settles this:

> Per agent rather than per deployment because the interesting case is a client's agent whose traces belong in **the client's own Logfire project**, not in the operator's.

Two consequences:

1. **An agent can redirect its traces away from us.** Build the Activity page on Logfire queries and it silently omits exactly the agents that matter most — a client's — while looking complete.
2. **Logfire is optional.** `LOGFIRE_TOKEN: str | None = None` in `app/core/config.py`. A self-hosted deployment without it would get an empty Activity page.

Meanwhile `agent_runs` is always written, is the row the budget guard and the audit trail key on, and already carries `logfire_trace_id` per run.

**So the answer is: our data is the page, Logfire is the drill-down.** The run row is the source of truth; the trace id is a link out for whoever wants the step-by-step. Anything else makes the page's completeness depend on a per-agent setting. Write this into the plan rather than re-deciding it later.

## What the run row can already answer

Per `agent_runs`: `surface`, `status`, `agent_id`, `agent_version_id`, `user_id`, `conversation_id`, `exposure_id`, `environment_id`, `provider`, `model_label`, `secret_id`, `input_tokens`, `output_tokens`, `cost_usd`, `cost_is_partial`, `started_at`, `ended_at`, `error`, `logfire_trace_id`, `paused_state`.

[...]

Issue #129, opened 2026-08-03
Sandbox views in /dashboard: capacity, sessions, activity


DEENUU1:
The read-only API this needs now exists on `feat/sandbox` (#36, `1ac763a`). This is
the `/dashboard` half, which #36 deliberately left out: the routes had to exist
before there was anything to render.

## What is already there

A minimal operator view ships on the **Sandboxes** page so nothing lands unread —
`src/components/sandboxes/sessions-panel.tsx` and `workspace-browser.tsx`. Read them
first; this ticket is about a real dashboard view, not about inventing the data
layer. The clients and hooks are done and typed:

| | |
|---|---|
| `src/lib/sandbox-connections-api.ts` | connections, `readSandboxPolicy`, `listSandboxSessions`, `readSandboxEvents` |
| `src/lib/sandbox-workspaces-api.ts` | `listWorkspaces`, `readWorkspaceFiles`, `readWorkspaceFile` |
| `src/hooks/use-sandbox-connections.ts` | `useSandboxConnections`, `useSandboxPolicy`, `useSandboxSessions`, `useSandboxEvents` |
| `src/hooks/use-sandbox-workspaces.ts` | `useSandboxWorkspaces`, `useWorkspaceFiles`, `useWorkspaceFile` |

Behind them, per sandbox connection:

- `GET /api/v1/sandbox-connections/{id}/policy` — the runtime allowlist with the
  ceilings actually in force per alias (memory, cpus, network mode, work dir).
- `GET /api/v1/sandbox-connections/{id}/sessions?usage=` — every sandbox **this
  organization** has open: runtime, alive, `running`/`hibernated`, created,
  last activity, idle seconds, and with `usage=true` memory against its own
  ceiling, CPU and pids. Plus `limit`, `open_limit`, `tenant_limit`.
- `GET /api/v1/sandbox-connections/{id}/sessions/{session_id}/events?after=` — the
  activity log: `seq`, `op`, `target`, `ok`, `detail`, `duration_ms`. `after` is
  the sequence you already hold, so polling does not re-read the whole log.
- `GET /api/v1/sandbox-workspaces` — every workspace in the organization with the
  agent, who shares it, size, version and last use. `/{id}/files` and
  `/{id}/file?path=` for one.

All gated on `connections:manage`, all read-only, and all proxied — the service
token never reaches a browser.

## What to build

1. **Capacity**, per connection: resident sessions against `max_sessions`, open
[...]

Issue #32, opened 2026-08-01
Let a failed query render an error, not an empty state


DEENUU1:
**Severity: Medium** · AUD-033

## What's wrong

`.claude/rules/frontend.md:50-54` names this exact defect ("an empty page is ambiguous"). Six call sites swallow a rejection into `[]` with no error state.

## Where

- `frontend/src/app/[locale]/(dashboard)/rag/page.tsx:171,179` (`setDocs([])`), `:191` (`setSyncLogs([])`), `:203` (`setSyncSources([])`), `:214` (`setConnectors([])`), `:303`
- `frontend/src/hooks/use-knowledge-bases.ts:160-170` — three `.catch(() => ({ items: [] }))` inside the `Promise.all`

## How it fails

- A 502 on `listTrackedDocuments` sets `docs = []` with no error state, and the page renders **"No documents — Upload PDF, DOCX, TXT, or MD"** (`:636-640`). A user whose documents failed to load is told they have none, next to a button inviting them to upload duplicates.
- A 500 on `listSyncSources` renders **"No sync sources configured"** (`:833-841`); on `listSyncLogs`, "No sync history yet" (`:933`).
- In `use-knowledge-bases.ts`, the three inner `.catch`es sit *inside* a `Promise.all` whose outer `catch` (`:178-179`) does set `error` — so the two calls that matter surface a failure and the three connector/sync-source calls silently do not. A failed `/kb/{id}/sync-sources/connectors` reaches `sync-source-wizard.tsx:447-450`, which renders **"No connectors enabled."** The user concludes Google Drive is not available in this deployment.

Unlike the deliberate `.catch`es elsewhere — `use-conversations.ts`, `use-invitations.ts`, `use-organizations.ts` and `use-members.ts` all toast on failure before returning `null`, and `use-knowledge-bases.ts:73-89` toasts too — none of these carries a comment justifying the swallow, and `useKBDetail` already has an `error` state at `:126` that they bypass.

## Fix

Let the rejection reach an error state and render an error component from `components/states/` instead of the empty state. In `useKBDetail`, either drop the three inner `.catch`es or record a per-section failure flag the section reads.

## Acceptance criteria

- [ ] Each section renders an error state on a 502, not its empty state
- [ ] For each, a test mocking the fetch as 502 and asserting the *error* state — explicitly not the empty state, and not only on chrome

PR #149, opened 2026-08-04
feat(dashboard): add the role-aware dashboard

### LINKED ISSUE #37: Design and build the dashboard (HTML demo first, role-aware)

OchnikBartek:
## Summary

The role-aware dashboard, both stages in one branch: the design artifact that
was reviewed, and the implementation written from it after review. One route,
one widget registry, twenty-seven cards; which cards a caller gets is decided
by the permissions they hold, never by their role name. Backend work is four
new answers on two new endpoints plus three filters on existing ones — every
one of them a question the run row could already answer and nothing was asking.

Review ran over three rounds and reversed two of stage 1's calls. Both
reversals are implemented here, and the section at the bottom closes out every
point from all three.

## Related Issue

Closes #37.
Refs #45, #208, #209.

## Added

**Endpoints.** `GET /api/v1/stats/usage` — inclusive `from`/`to` resolved to a
half-open UTC window, `scope=org` (needs `runs:view`) or `scope=own` (needs
only a signed-in membership), and one composed response so the eight cards
reading it share a single request, a single loading state and a single
failure. Two of the nine frozen `group_by` dimensions are implemented:
`version` for the builder's comparison card and `user` for the per-person
table. `GET /api/v1/ratings/summary` takes the same `scope`, and is distinct
from the deployment-wide app-admin summary it does not replace.

**Filters on what already existed.** `?status=failed,budget_exceeded` on
`GET /runs` (a list, because that is the operator's natural query);
`?shared_with_me=true` on the agents, skills and knowledge-base lists; and
`budget_monthly_usd` on the agent listing, read from the *published*
version's spec rather than the draft.

**Surface recording (#208).** `RunSurface.EMBED` stamped by
`embed_session.py` — an embedded-widget run was recorded as `web` — and
`mattermost` added to `_SURFACES`, where it fell through to `api`. No
migration and no backfill: historical rows keep the surface they were
recorded under, and `docs/channels.md` says so.

**The page.** `src/lib/dashboard/` holds the decision logic as pure functions
— registry, layouts, period, sections, the my-agents policy — with the React
in `src/components/dashboard/`. Period presets plus a two-month custom-range
calendar, built as a reusable `ui/date-range-picker` over `lib/calendar`
rather than as a dashboard-only control. A sections filter that travels in the
URL and can only ever hide. Every card has its own loading, empty and error
state, and its own Retry.

## Changed

- The dashboard route was an "under construction" stub; it now resolves an
  audience from the caller's role and app-admin flag, and holds a skeleton
  until `usePermissions()` settles rather than deciding the audience while
  `can()` still answers false.
- Home in the mobile tab bar always points at `/dashboard`; it used to fork
  on the app-admin flag.
- `AgentRead` carries the published budget cap, so the headroom card stops
  being the one card that cannot show per-agent numbers.
- The four aggregate queries refetch when the tab regains focus, against the
  app-wide default. Scoped to those hooks deliberately — the default is right
  for a page somebody is reading, and wrong for one they leave open.

## Testing

- Re-run in full after merging main at 0.0.30: `make lint` (ruff, `ty`, eslint,
  `tsc`, prettier, codespell, and the i18n guard), **3282 backend tests at 100%
  branch coverage on the platform layer**, 281 integration tests, **3303
  frontend tests** at 100% lines/statements/functions and 97.55% branches, the
  frontend build, `docs-build --strict` and the dependency audit. Two targets
  do not pass on a developer laptop for reasons that are not this branch — see
  Notes.
- Seven new integration tests drive the delegation split against real SQL,
  including that deleting a parent makes its orphaned child start counting
  (`ondelete=SET NULL`, so the cost is no longer inside anything).
- `make test-integration` — 240 tests against a real Postgres. The SQL is
  proven there rather than assumed: percentile interpolation, day buckets
  splitting on UTC midnight, both window edges, tenant isolation when the same
  person runs agents in two organizations, and the null-user exclusion that
  keeps the per-person table agreeing with the count above it.
- `dashboard.integration.test.tsx` covers nine scenarios across four
  audiences, and asserts the *absences*: a viewer's browser never requests
  `/stats/usage`, `/approvals`, `/spend` or the member list, and a member
  never asks the per-person question. A 502 on `/stats/usage` is asserted to
  cost the usage cards and nothing else, with Retry refetching.
- Playwright: `dashboard.spec.ts` and the `/dashboard` entry in
  `navigation.spec.ts`, 23 green against a container built from this branch.
  Verified there by hand as well — the per-person card renders real rows
  (`admin@example.com · 13 runs · $0.04 · 1h ago`).
- `python3 scripts/docs_drift.py` clean; `make docs-build --strict` green.

## Merged with main: which side of `include_delegations` each aggregate takes

#200 landed delegated runs while this was open. A parent's row already contains
its children's tokens, so an aggregate counting both bills one run twice — the
failure that would have shown up as an inflated spend tile and nowhere else.

`docs/governance.md` had already settled the rule — *the organization's question
excludes child rows, a question about one agent includes them* — so this follows
it rather than inventing a third convention.

| Aggregate | Child rows | Why |
|---|---|---|
| `cost.period_usd`, `cost.by_provider` | **excluded** | the double bill: the parent's row already holds these tokens |
| `total_runs`, `by_day`, `by_status` | **excluded** | the total the rest is a slice of, and the donut's segments must keep summing to it |
| `by_surface` | **excluded** | a delegated row *copies* its parent's surface — it is not a second arrival |
| `latency_ms` | **excluded** | the parent's window is what the person actually waited |
| `active_users`, `?group_by=user` | **excluded** | a delegated row copies its parent's `user_id`; counting it inflates a person by their agent's fan-out |
| `by_agent` | **included** | excluded, an agent running four hundred times a day as somebody's delegate has no row — and the adoption card names every published agent without one as forgotten and offers to archive it. Its bars can exceed the total beside them; nothing sums them |
| `?group_by=version` | **included** | a specialist that only ever runs as a delegate would otherwise compare nothing across its versions |
| pending-approval count | **no filter** | a parked child is a stuck parent. Today a delegation is written already finished and never parks, so this is a no-op that stays correct if that changes |

The invariant survives: the outcomes donut still sums to `total_runs` and its
`awaiting_approval` segment still counts the same rows as the approvals card,
because those three sit on the same side. `UsageStats`' docstring now says which
single block does not sum, instead of promising that they all do.

**Three conflicts were not mechanical**, and are worth a look rather than a
skim: main moved response building into `list_readable` on the skills and kb
services, and moved `GET /runs` onto `AgentRunnerService`. `shared_with_me` and
the status filter are threaded through the new shape rather than kept beside it
— which also puts the runs route back behind a service, as the layering rule
asks. Two new lint steps arrived with main that this branch had never seen:
prettier over `frontend/src` (17 files) and codespell.

## Review, point by point

All twelve inline threads are answered in place. Ten were already satisfied by
work on this branch; two needed code in the catch-up merge — the *Filtering*
section's stale claim that the page queries `?from=&to=` on `/spend`, and the
`lambda` nit below, which is refuted rather than applied.

### Reversed in review, and implemented

**A named per-person table is in** (was: "adoption is a count, not a table of
names"). It cost one dimension rather than an endpoint, because
`group_by=user` was already in the frozen vocabulary. The three things the
reversal had to settle:

- *Whose names, to whom.* The gate is `runs:view`, which builder and operator
  hold too. The card carries that sentence in its own copy, and the layouts
  feature it wherever usage is featured rather than on steward pages alone — a
  disclosure that is true on only some pages is not a disclosure.
- *Runs, with cost as a column.* Ordered by runs, then by email so equal
  counts do not reshuffle between two requests. Sorted by spend the same rows
  read as a league table.
- *The count stays.* "Active people" is the headline, the table is the
  evidence under it, and both come from the same rows — a run with no user is
  excluded from the table for exactly the reason `COUNT(DISTINCT user_id)`
  skips it.

**A user-arranged dashboard moved into stage 2.** Its shape is settled in the
note and four of its five prerequisites already hold in this code: span and
title ride the layout entry (not a widget-id lookup), `gate()` takes
`(can, isAppAdmin)` so the future catalog can run it over a list it is not
rendering, every widget declares a "see all" destination, and widget
definitions carry catalog metadata from day one. **What is deliberately not
here is the persistence and the arranging UI** — a preferences table,
migration, service and routes under the 100% gate, plus a catalog modal and
reorder controls. That roughly doubles a branch already at 121 files, and
`CLAUDE.md` asks for scoped diffs and one branch per pull request. It wants
its own branch, and I would rather it got a real one than a rushed corner of
this.

*One correction while moving it:* the note claimed the grid uses a closed
`s4/s6/s8/s12` span set. That is wrong about its own demo — the layouts pair
`s7`+`s5` and `s8`+`s4`. The set is `s3`–`s8` plus `s12`, and narrowing it
would leave somebody unable to rebuild the default they started from.

### Confirmed in review, and implemented

- **The org answer-quality card (#209)**, counts and trend only. A rating's
  free text is one person's opinion of one conversation; at `scope=org` it
  reaches an audience its author did not pick, so it stays behind the run
  detail. The endpoint returns no comments.
- **The headroom card links to the organization's settings.** Its "see all"
  is computed in the widget rather than declared in the registry — the page
  holding the cap is per-organization and its path carries an id the catalog
  cannot see. Raising a cap is still elsewhere; this is navigation.
- **Live updates, split.** Refetch on focus for the daily aggregates. The
  queue half needed no change: `useApprovals` already polls every 30s while
  focused.
- **Deployment-wide time series, the audit strip and sandbox stay out.**

### The widget-map correction pass

All four cells that disagreed with `main` are now true of the code:

| Was wrong | Now |
|---|---|
| The surface chart drew bars `agent_runs.surface` could not produce | #208 landed here — `embed` and `mattermost` are recorded |
| The outcomes donut drew 3 of `RunStatus`'s 6 values, contradicting the approvals card above it | Five segments covering all six, summing to `total_runs`, with the awaiting segment counting the same parked runs as the approvals card |
| `/runs?status=failed` marked as existing | Implemented, as a list |
| `from`/`to` on `/spend` marked as existing | Not built. The period total comes from the `cost` block on `/stats/usage`; `/spend` is untouched and its month-to-date line stays calendar-aligned, says so on the card, and deliberately ignores the filter |

Also from that round: **version comparison** is implemented
(`group_by=version` plus the compare card), the **`group_by` vocabulary is
fixed at nine words** for #45 to share rather than growing a second aggregate,
and the **registry makes adding a widget one entry** — adding the per-person
card in this branch was exactly that, plus a component and its copy.

### The five widgets whose data was already sitting there

Three are built: **budget headroom**, **integrations / MCP health** and
**knowledge freshness**. The budget reasoning in the note was wrong in both
halves and is corrected — `budgets:manage` does gate the `monthly_budget_usd`
field on `PATCH /orgs/{id}`, and the cap is stored on the organization and in
`AgentSpec.budget.monthly_usd`.

Two are **not** built, and I would rather say so than let them look shipped:

- **Cost coverage** (`agent_runs.cost_is_partial`) — how much of the spend
  figure is an estimate. Nothing on the page currently qualifies the number,
  which is a real gap.
- **Time to decision** (`tool_approvals.created_at` → `decided_at`) — the
  approvals card shows each row's age but there is no distribution and no
  oldest-item alarm. Your `ApprovalStatus.EXPIRED` observation is why it
  matters: nothing assigns that status, so the queue has no ceiling. Both are
  small and additive on top of what is here.

### The one bot finding not applied

`github-code-quality` flagged `lambda: AsyncMock()` in `test_runs_filter.py` as
an unnecessary wrapper. It is not: FastAPI introspects a dependency override's
signature, and `AsyncMock`'s is `(*args, **kwargs)`, which it turns into two
**required query parameters** — every request then answers
`422 … Some fields need fixing: args, kwargs`. Applying the suggestion failed
all six tests in the file. Reverted, with a comment saying why.

### Where I did not follow the review

**`docs/design/` is excluded from the docs site rather than added to
`mkdocs.yml`'s `nav`.** Round one offered either; round three asked for `nav`.
I went with `exclude_docs`, with the reason written above it in the file: these
two files are review material for a pull request, and a design note published
as a site page is stale the day after the code it describes merges — which is
the same failure mode ("an orphan page goes stale without anybody noticing")
read the other way. Happy to flip it if you still disagree; it is one line.

### Filed, not fixed here

**#207** (`RunSurface.SCHEDULE` defined and never assigned) and **#210**
(sorting run history by duration) are both out of this branch's scope. The
two-way deep link — Activity handing back "see this on the dashboard" with the
same window and dimensions — is recorded in the note; the dashboard already
carries its period in the URL, and the handoff itself belongs with #202.

## Notes for Reviewers

- **Two lint failures of my own, caught late and fixed in the last two
  commits — both would have been red in CI.** Five hardcoded strings that
  `check_i18n.py` refuses (in `widget-states.tsx` and `date-range-picker.tsx`),
  and a `ruff format --check` violation in `agent_registry.py` from the
  budget-cap commit. The skeleton's `aria-label` now reads
  `dashboard.states.loading`, whose English value is unchanged, so the e2e
  spec waiting on `role=status` still matches. Worth knowing when reading the
  earlier commit messages on this branch: their claims about a green gate were
  true of what they ran, not of the whole tree.
- **The gate placement is a third pattern** and worth a look: the stats routes
  carry no `require()`, because a route-level `require(runs:view)` would refuse
  a member's `scope=own` before the parameter was ever read. `StatsService`
  decides, the route sweep recognises it the way it recognises the grant-aware
  services, and `TestStatsScopeIsDecidedInTheService` proves the refusals —
  including that naming the organization's people is refused without
  `runs:view` while a member may still ask the same shape about themselves.
- **Two `make check` targets cannot pass on a developer laptop, and neither is
  this branch.** Both are pre-existing and invisible to CI; filed separately.
  `scripts/check_backticks.py` skips itself but not its copies, so a checkout
  with git worktrees under `.claude/worktrees/` gets false positives from other
  branches' files ([#225](https://github.com/vstorm-co/agenticos/issues/225)).
  And `make db-check` runs `alembic check` against whatever `backend/.env`
  points at: the vector store creates a `rag_<collection>` table at runtime for
  every collection, nothing models them, so on any database that has ever
  ingested a document `alembic check` proposes dropping them. CI points it at a
  throwaway database that has never ingested anything, which is why it is green
  there and has always been red for anyone who has used the RAG feature.
- Four Playwright specs fail on my machine (`agents` builder, `journey`,
  `models` ×2) because this dev database's seeded agent was repointed to
  OpenRouter `gpt-4o-mini` and has real run history. They are environment, not
  branch — the dashboard and navigation specs are green.

## Checks


docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+
+| Widget | Shown because | Fed by | Exists today |
+|---|---|---|---|
+| Platform at a glance | `is_app_admin` | `GET /api/v1/admin/stats` | yes |
+| Service health | `is_app_admin` | `GET /api/v1/admin/system` | yes |
+| Top organizations | `is_app_admin` | `GET /api/v1/admin/organizations` | yes |
+| Answer quality, deployment-wide | `is_app_admin` | `GET /api/v1/admin/ratings/summary` | yes |
+| Runs over time | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Outcomes (completed / failed / budget exceeded) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Where runs come from (by surface) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
```
DEENUU1: This is the widget you call the single best answer to "how do people use the system", so it is worth being precise about what `agent_runs.surface` can actually say today. The demo charts six surfaces (`dashboard-demo.html:473-477`) and three of them are not reachable:

- **Widget** — `app/services/embed_session.py:164` stamps `surface=RunSurface.WEB`. An embedded widget run is indistinguishable from a web chat run, so "Web chat 41%" silently contains both and the Widget bar is always 0.
- **Scheduled** — `RunSurface.SCHEDULE` appears exactly once in the codebase, in the enum that defines it (`app/db/models/agent_run.py:64`). Nothing assigns it.
- **Mattermost** is missing from the chart but does produce runs, and `app/services/channels/mentions.py:56-57` maps only `slack` and `telegram` — the comment right above it says anything else "is recorded as an API run rather than guessed at". So Mattermost lands in `api`, and "HTTP API 24%" contains it.

The failure is quiet, which is why I would rather settle it now: build `/stats/usage` exactly as specced and `by_surface` comes back looking correct with four rows, on a deployment that has a widget on its marketing site and a Mattermost bot. Nobody reading the card learns that.

Two honest ways out, and the review should pick one rather than leaving it to stage 2:

1. Widen the recording — a `RunSurface.EMBED`, pass it from `embed_session.py`, add `mattermost` to `_SURFACES`. Small, but it is backend work in `app/services/**` that "What stage 2 builds" does not mention, so it belongs in the estimate.
2. Draw the four surfaces that exist, and say in the card what `web` and `api` each include.

Either is fine. Shipping the six-bar chart against today's recording is the one option that is not.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+## Widget map
+
+| Widget | Shown because | Fed by | Exists today |
+|---|---|---|---|
+| Platform at a glance | `is_app_admin` | `GET /api/v1/admin/stats` | yes |
+| Service health | `is_app_admin` | `GET /api/v1/admin/system` | yes |
+| Top organizations | `is_app_admin` | `GET /api/v1/admin/organizations` | yes |
+| Answer quality, deployment-wide | `is_app_admin` | `GET /api/v1/admin/ratings/summary` | yes |
+| Runs over time | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Outcomes (completed / failed / budget exceeded) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
```
DEENUU1: Same class of thing as the surface chart, and this one puts two cards on the same page in disagreement. `RunStatus` has six values (`app/db/models/agent_run.py:37-52`): `running`, `completed`, `failed`, `cancelled`, `awaiting_approval`, `budget_exceeded`. The donut draws three, and computes its centre percentage and its "1 in N runs needed attention" line against the same `total` that feeds "Runs over time".

So on a steward's page: "Waiting on a decision" says three approvals are parked, and the donut immediately below sums to 100% with no parked runs in it. `cancelled` and `running` vanish the same way. In the demo it looks consistent only because `pct(0.91) + pct(0.06) + pct(0.03)` is 100 by construction.

Keep the three-way split — separating `budget_exceeded` from `failed` is right, and it is the reasoning `RunStatus`'s own docstring gives — but the donut needs somewhere for the other three to go. Simplest version that stays honest: four slices, with `awaiting_approval` as its own tone, since "waiting on us" is a third story again, and `running` / `cancelled` folded into a neutral segment the legend names. Whatever you pick, the invariant worth writing into the note is that the outcome segments sum to `total_runs` — that is what stops this card and the approvals card drifting apart again.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+| Where runs come from (by surface) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Agents: adopted and forgotten | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Latency (p50 / p95) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Active people (count, no names) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Spend, month to date | `runs:view` | `GET /api/v1/spend` | yes |
+| Models behind the runs | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Recent failures | `runs:view` | `GET /api/v1/runs?status=failed` | yes |
```
DEENUU1: Marked `yes`, but `/runs` has no `status` parameter — `list_runs` takes `agent_id`, `skip` and `limit` and nothing else (`app/api/routes/v1/runs.py:33-45`), and the repository underneath does not filter on status either (`app/repositories/agent_run.py:117-134`).

That matters more than a wrong cell, because "Recent failures" is in the first section of three of the four layouts. Without a `status` filter the only way to fill it is fetch the newest 100 runs and filter in the browser — exactly the "analytics over the most recent 100 rows, silently" the aggregation section rejects two screens later. And it fails the worst way: a busy organization racks up 100 runs since its last failure, the card renders "No failures — every run in this period finished cleanly", and that is a lie the user cannot see through.

Small work — a `status` query param on the route and a matching `where` in the repository — but it is work, so this row should read **no — needs a filter on an existing endpoint** and the route change should appear in "What stage 2 builds". Worth deciding whether it takes one status or a list, since the operator's version of this card arguably wants `failed` and `budget_exceeded` together.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+## Filtering
+
+One filter row sits above everything it scopes — never a filter inside a
+card. It is product UI, not demo chrome:
+
+- **Period** — a date-range button opening a picker: presets on the left
+  (Today, Last 7 / 30 / 90 days, This month, Last month) and a two-month
+  calendar for a custom range, future days disabled. Presets are UI sugar —
+  either way the page queries `?from=&to=` on `/stats/usage` and `/spend`,
+  and every card re-renders against the same slice.
```
DEENUU1: `/spend` cannot take `from`/`to` today — it takes `days: int = Query(30, ge=1, le=365)` (`app/api/routes/v1/runs.py:117-124`), a rolling window that always ends now. So two things in this sentence do not hold:

- A custom range is unrepresentable, not just awkward. `#period=2026-07-05..2026-07-20` ended two weeks ago; no `days` value expresses it.
- The card's headline is `month_to_date_usd`, and its docstring is explicit that month-to-date is calendar-aligned rather than rolling, precisely so it reconciles against an invoice. Pick "Last 7 days" and that number cannot move. The card sits inside a period-scoped section, under a filter, and quietly ignores it. The demo hides this because the spend figures are hardcoded constants (`dashboard-demo.html:505-508`), so the one card that should visibly disobey the filter appears to obey.

Fixable either way round and I do not mind which:

- add `from`/`to` to `/spend` and drop the month-to-date headline for "spend in this period" — then this row is no longer `exists today: yes`; or
- leave `/spend` alone and keep the card honest: title it "This month" and mark it as not scoped by the period.

The first is more work and a better page. The second is free. What I would not ship is a filter bar that claims to scope everything under it and one card opting out without saying so.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+  "by_agent": [{ "agent_id": "…", "name": "Support triage", "runs": 561 }],
+  "by_status": [{ "status": "completed", "runs": 1823 }],
+  "by_model": [{ "model_label": "claude-sonnet-5", "runs": 881 }],
+  "latency_ms": { "p50": 3200, "p95": 14800 },
+  "active_users": { "active": 14, "total_members": 23 }
+}
+```
+
+Everything above is answerable from `agent_runs` columns that already exist
+(`surface`, `status`, `agent_id`, `user_id`, `model_label`, `started_at`,
```
DEENUU1, line 294: This list is missing the three columns I would most want the dashboard to grow into, and all three are already on the row.

**`agent_version_id`** (`app/db/models/agent_run.py:89`) — and its comment is a hint about intent: *"Kept even if the version is later deleted: a run must not lose the record of what it executed."* That column exists so you can ask **did v4 actually behave better than v3**, and right now nothing anywhere asks it. For a builder that is a more useful question than any other on their page: they publish a new version, and the honest feedback loop is success rate, latency, cost per run and thumbs-up rate for the runs that executed *that* version versus the one before. I would add `by_version` to the `group_by` vocabulary and give the builder page a compare-two-versions card, even if v1 ships it as a plain table rather than a chart.

**`environment_id`** (`agent_run.py:121`) — which named environment resolved the version. Staging noise sitting in the same numbers as production is the sort of thing that makes a steward stop trusting the page. At minimum it is a filter; arguably it is a default (exclude non-production).

**`exposure_id`** (`agent_run.py:112`) — which binding admitted the run. This is the channel dimension, and it is *not* the same as `surface`: `ExposureSurface` is deliberately only `slack | telegram | mattermost` and says so in its docstring, because a run can come from the playground or a schedule and neither is somewhere an author exposes an agent. So `surface` answers "how did it arrive" and `exposure_id` answers "through which Telegram bot, which Slack workspace, which embed" — and with three bots in one org, the second is the question somebody actually has.

This is where #45 overlaps hardest. That page wants filters by version, channel and environment off exactly these rows, and its own note says if #37 adds `/stats/usage` then Activity should consume it rather than growing a second aggregate. So the `group_by` vocabulary is worth fixing once, here, with #45 in the room: `day | surface | agent | version | user | status | model | exposure | environment`. Adding a dimension later is a migration of the response shape; agreeing the list now costs a conversation.

DEENUU1 (commented):
This is the right artifact, and a genuinely good one. Computing visibility from a mirror of `ROLE_PERMS` instead of drawing seven screens is what makes the absences trustworthy — I can switch to viewer and believe what is missing. Gating "Your activity" and "Continue where you left off" on `agents:run` because a viewer's rows are structurally impossible is the best reasoning in the note, and it checks out against the catalog: a viewer really does hold only `agents:view` / `collections:view` / `skills:view`. `retryWidget` bringing one card back while the rest stay failed is the first mock here that demonstrates #32's point rather than asserting it. The token claims check out verbatim too — `--color-brand` is brand-700, `--color-chart` is brand-600, `--radius-card` is 0.5rem — and recharts is genuinely in the five files you name, already behind `next/dynamic`.

Two themes in the comments below.

**The widget map needs a correction pass before stage 2 is written from it.** I traced every row against `main` and four cells disagree with the code, two of them hiding backend work that is not in "What stage 2 builds". The pattern in the worst two is the same: the demo reads an enum more narrowly than it is. The surface chart draws three bars `agent_runs.surface` cannot produce today, and the outcomes donut draws three of `RunStatus`'s six values — including a split that contradicts the approvals card sitting right above it. Then `/runs?status=failed` and a `from`/`to` on `/spend` are both marked as existing and neither does.

**And I want the design to reach a bit further, because the run row already carries more than this uses.** Three things I would add or at least make room for: comparing agent **versions** (`agent_version_id` is on every run and the map never mentions it), a dashboard each user can **arrange themselves** rather than one layout per audience, and a widget registry where adding the next widget is one entry and not a refactor. Details in the comments on the aggregation section, the layout paragraph and the registry. Much of this overlaps #45 — that page wants filters by version, channel and environment off the same rows, and the two tasks should agree on one `/stats/usage` rather than growing a second aggregate each. Worth deciding the `group_by` vocabulary once, here, with #45 in the room.

One thing outside the diff: `docs/design/` is not in `mkdocs.yml`'s `nav`. `--strict` stays green because an unlisted page is an INFO rather than a warning, but mkdocs still builds and publishes both files, so the note ships to the docs site as a page nothing links to. Either add a `Design` section to `nav` or exclude `docs/design/**` — an orphan page is the one that goes stale without anybody noticing.

None of this is a rewrite. The four audiences, the builder page leading with their own agents, `scope=own` for members, and the aggregate-only adoption card are all right, and I would not soften the `agents:run` gate that empties the viewer's page. The org answer-quality card earns its endpoint — it is the only thing on the page that answers "is it any good", and `RatingSummary` already carries `ratings_by_day`, so the trend comes free.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+  a builder gets the same list with adoption numbers.
+- **It also decides what a card may claim about you.** "Your activity" and
+  "Continue where you left off" render only with `agents:run`: a run history
+  and a conversation list are made of rows the caller produced by running
+  agents, so for a viewer — who cannot run — both cards would be permanently,
+  structurally empty. Gating them on `agents:run` (not on the role name)
+  means a Phase-2 custom role that grants running gets them back
+  automatically.
+
+**One thing the permission rule deliberately does not decide: layout.** Each
```
DEENUU1, line 63: I would go one step further than this paragraph does. Four audience layouts are the right *default*, but they should not be the final answer — a user should be able to arrange their own dashboard: which widgets they want, in what order, and which they never want to see again. The people who will live on this page daily are the ones whose needs a four-way bucket fits worst, and "my page leads with the thing I actually check every morning" is cheap to give them once the registry exists.

The good news is the design is already most of the way there, because presence and layout are separate concerns — the audience layout supplies order and width, and `gate()` supplies presence. A per-user preference is a third layer over the same two, and it composes without touching either:

```
effective layout = user preference (if any) ?? audience default
visible widgets  = effective layout ∩ { w : w.gate() }
```

The one rule that has to survive is that a preference may only ever **reorder or hide** — never reveal. A stored preference naming a widget the caller cannot see must be dropped at render time, not trusted, or the preference becomes a way around the gate. Same for a preference stored under one role and read after a demotion, which is the realistic version of that bug.

For v1 I would not build the drag-and-drop. I would settle two things in stage 1 so it stays cheap later: that widget ids are stable and worth persisting (they are already the registry's keys), and where the preference lives — user-scoped, or user-and-organization-scoped. I think the latter, since the same person is a steward in one org and a member in another, and one saved layout across both would be wrong in one of them. Worth a line in "Deliberately out of v1" saying it is out but that the shape above is why it stays possible.

docs/design/dashboard-demo.html
```diff
@@ -0,0 +1,1161 @@
...
+    <span class="track"><span class="fill" style="width:${(i.v / max) * 100}%"></span></span>
+    <span class="bv">${fmt(i.v)}${unit}</span>
+  </div>`).join("");
+}
+
+/* ================================================================
+   Widget registry. gate() decides presence; source names the feed.
+   Order, width and heading come from the audience layout below.
+   ================================================================ */
+const WIDGETS = [
```
DEENUU1, line 765: The registry is the best structural decision in the demo — a widget is one entry with a `gate`, a `source` and a `render`, and the layouts only reference ids. That is what makes "add the sandbox section later" a believable claim rather than a hopeful one. Two things I would fix while porting it, both of which quietly cost you that property:

**A widget can only appear once per page.** `LIVE` is keyed by widget id (`renderSections`, and `chartW` reads it back), and the card's DOM id is `w-${id}`. List the same widget in two sections and the second entry overwrites the first's span and title, both cards render at the second's width, and `retryWidget`'s `getElementById` only ever repairs the first. No layout does this today, so it is latent — but "runs over time, once org-wide and once for me" is a natural thing to want, and the failure is silent. In React this stops being a hazard if the span and title ride the layout entry rather than a lookup table, which is the right shape anyway.

**`SPAN_W` hardcodes a 1180px page.** Fine for a mock, and you say so. Worth flagging only so it does not get ported: the recharts `ResponsiveContainer` measures, so the product should not carry a span→pixels table at all.

One more thing for the port: `gate()` closes over module-level `current`, so a widget's visibility is a function of global state. The React equivalent wants the gate to take what it needs — `(can, isAppAdmin) => boolean` — so a widget is testable on its own, which is what makes the `dashboard.integration.test.tsx` covering two roles cheap to write rather than a page-level fixture per role.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+| Outcomes (completed / failed / budget exceeded) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Where runs come from (by surface) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Agents: adopted and forgotten | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Latency (p50 / p95) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Active people (count, no names) | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Spend, month to date | `runs:view` | `GET /api/v1/spend` | yes |
+| Models behind the runs | `runs:view` | `GET /api/v1/stats/usage` | **no — new** |
+| Recent failures | `runs:view` | `GET /api/v1/runs?status=failed` | yes |
+| Waiting on a decision (approvals) | `approvals:decide` | `GET /api/v1/approvals` | yes |
+| Members | `members:manage` | `GET /api/v1/organizations/{id}/members` | yes |
```
DEENUU1: Two small things on this row.

The path is `GET /api/v1/orgs/{org_id}/members` — the members router is mounted at `/orgs`, not `/organizations` (`app/api/routes/v1/__init__.py:106`).

And the gate deserves a sentence, because it is the one place the "shown because" column is not mirroring the backend: `list_members` is `CurrentUser` and its docstring says "Any member may call this" (`app/api/routes/v1/members.py:19-28`). Gating the card on `members:manage` hides a headcount from a builder or operator who can fetch the same list from the members page. I am not asking you to change it — a headcount is context a steward wants and an operator does not, so it is a reasonable layout call — but the note's central rule is "widgets are keyed on permissions", and here the permission is doing layout work rather than marking a boundary. Say so, or the next reader concludes the endpoint is privileged when it is not.

Related: `active_users` in the `/stats/usage` response carries `total_members` and gates on `runs:view`, so a builder gets the headcount from that card anyway. Fine by me and consistent with decision 3 — a count, not names — but it means the headcount is not really withheld from anyone.

docs/design/dashboard-demo.md
```diff
@@ -0,0 +1,356 @@
...
+| Answer quality (org) | `runs:view` | `GET /api/v1/ratings/summary` | **no — new** |
+| Your agents | `agents:view` | `GET /api/v1/agents` | yes |
+| Continue where you left off | `agents:run` | `GET /api/v1/conversations` | yes |
+| Your activity | `agents:run` (own rows) | `GET /api/v1/stats/usage?scope=own` | **no — new** |
+| Agents you use most | `agents:run` (own rows) | `GET /api/v1/stats/usage?scope=own` | **no — new** |
+| Answer quality in your chats | `agents:run` (own rows) | `GET /api/v1/ratings/summary?scope=own` | **no — new** |
+| Shared with you | `agents:view` (shared scope) | `GET /api/v1/agents` · `/collections` · `/skills` | yes |
```
DEENUU1: `/collections` does not exist — collections are served under `/kb` (`app/api/routes/v1/__init__.py:110`, and their sharing router is mounted at `/kb` too, line 63).

More substantively, `exists today: yes` is optimistic for the counts this card shows. `GET /api/v1/agents` returns "Agents this member can see - their own, plus what was shared with them" as one list (`app/api/routes/v1/agents.py:129`), with no split and no owner filter. To get "Agents 4" meaning *shared with you*, the page has to fetch the list and count rows it does not own — over `limit ≤ 100`, the same silent-truncation trap as `/runs`. For a viewer it is near-harmless, since they cannot create agents and everything they see was shared; for a member owning a dozen of their own it is wrong as soon as the list pages.

Cheapest fix that keeps the card is probably a `shared_with_me` filter on the three list endpoints, or a count on the response. Worth its own line rather than folding into "yes".

docs/design/dashboard-demo.html
```diff
@@ -0,0 +1,1161 @@
...
+/* ================================================================
+   Permission mirror.
+   Must track backend/app/core/permissions.py ROLE_PERMS - only the
+   permissions the dashboard gates on. Scopes are omitted: a widget
+   asks "held at all?", and scope shows up as which rows the data
+   contains, not as whether the card renders.
+   ================================================================ */
+const ROLE_PERMS = {
+  owner:    ["runs:view", "approvals:decide", "members:manage", "agents:view", "agents:edit", "agents:run"],
+  admin:    ["runs:view", "approvals:decide", "members:manage", "agents:view", "agents:edit", "agents:run"],
```
DEENUU1: The mirror drops two permissions `owner` and `admin` really do hold. Both get every global permission via `_ALL_GLOBAL` in `app/core/permissions.py` — `mcp:manage` and `connections:manage` included — but here they are listed for `builder` only.

Nothing gates on either today, so no visible bug. The reason I would still fix it: "Deliberately out of v1" says the sandbox view "absorbs cleanly later: one more section, gated on `connections:manage`". Add that section against this mirror and it renders for a builder and hides from the owner and the admin — the two people most likely to be looking for it, and a bug that reads as a deliberate decision because it comes from the file claiming to be the source of truth.

```suggestion
  owner:    ["runs:view", "approvals:decide", "members:manage", "agents:view", "agents:edit", "agents:run", "mcp:manage", "connections:manage"],
  admin:    ["runs:view", "approvals:decide", "members:manage", "agents:view", "agents:edit", "agents:run", "mcp:manage", "connections:manage"],
```

Worth a line in the comment above it too: the mirror holds only what the dashboard gates on, so a permission listed here that nothing calls `can()` with is one somebody has to remember to keep in sync. Carry the whole catalog or carry only what is used — carrying two-thirds of it is how this drifted.

docs/design/dashboard-demo.html
```diff
@@ -0,0 +1,1161 @@
...
+  const atNow = first.getFullYear() === t.getFullYear() && first.getMonth() === t.getMonth();
+  const nav = side === "left"
+    ? `<button class="cal-nav prev" aria-label="Previous month">‹</button>`
+    : `<button class="cal-nav next" aria-label="Next month"${atNow ? " disabled" : ""}>›</button>`;
+  const dows = ["Mo", "Tu", "We", "Th", "Fr", "Sa", "Su"].map((x) => `<span class="dow">${x}</span>`).join("");
+  const title = first.toLocaleDateString("en-GB", { month: "long", year: "numeric" });
+  return `<div class="cal-month"><div class="cal-head">${nav}${title}</div><div class="cal-grid">${dows}${days}</div></div>`;
+}
+
+function retryWidget(id) {
```
DEENUU1, line 1256: This is the detail that makes the error state reviewable rather than asserted, and I want to say so explicitly: recovering one card to `normal` while every other card stays failed is the only way a reviewer can see "each card fails alone". Most mocks put the whole page in one state and print the claim in the copy. Keep this.

DEENUU1:
Following on from the review — two things I want to put in stage 1's scope rather than leave to stage 2, plus a correction to one of the "out of v1" calls.

## Let a user arrange their own dashboard

I would keep the four audience layouts as the **default** and let each person adjust it: pin a widget higher, widen it, remove one they never look at, add one from a list. The people who will live on this page daily are exactly the ones a four-way bucket fits worst.

The design already makes this cheap, because presence and layout are separate concerns — the layout supplies order and width, `gate()` supplies presence, and a stored preference is a third layer over the same two. Five things worth settling now, though, because each is expensive to retrofit:

1. **A preference may only reorder or hide — never reveal.** A stored layout naming a widget the caller cannot see has to be dropped at render time, not trusted. The realistic case is not an attack, it is a demotion: somebody saves a layout as an admin and later holds `member`.
2. **"Pin" and "expand" are two different things** — order and span. Span should be a closed set (s4 / s6 / s8 / s12), not a free resize. The demo already knows this: `SPAN_W` exists precisely because a chart generated at an arbitrary width breaks the "type inside a chart matches type outside it" rule.
3. **The "add a widget" catalog must itself pass through `gate()`.** Otherwise the list leaks what the page hides — a viewer never sees the Spend card but does see it offered. Same rule, one layer up, and easy to miss there.
4. **"Expand" is not a detail view.** Every widget should declare a "see all" destination pointing at a page that already exists — Activity for runs and approvals, the agent detail page, the members page. Without that rule this page slowly becomes a second Activity page with its own filters and its own bugs.
5. **Widget definitions should carry catalog metadata from day one** — description, permission, default span, which audiences show it by default. Adding those later means touching every widget. Plus a "reset to default", because somebody will hide everything.

For v1 I would not build drag-and-drop. I would just make sure widget ids are stable and worth persisting (they already are — they are the registry's keys), and decide where a preference lives. I think user-and-organization-scoped rather than user-scoped: the same person is a steward in one org and a member in another, and one saved layout across both is wrong in one of them.

## Five widgets whose data is already sitting there

Ordered by how much I would want them, not by effort. None of these needs a new column.

| Widget | Answers | Data |
|---|---|---|
| **Budget headroom** | how much of the monthly cap is left, org-wide and per agent | `organizations.monthly_budget_usd` + `AgentSpec.budget.monthly_usd`; the spend half is already aggregated by `/spend` |
| **Integrations that stopped working** | which MCP server stopped answering, and since when | `mcp_connections.last_status` / `last_error` / `last_checked_at` / `is_enabled` |
| **Knowledge freshness** | which collection stopped syncing | `sync_source.last_sync_at` / `last_sync_status` / `last_error`, `sync_log.failed` |
| **Cost coverage** | how much of that figure is an estimate rather than a fact | `agent_runs.cost_is_partial` |
| **Time to decision** | how long approvals sit, and what the oldest one is | `tool_approvals.created_at` → `decided_at` |

### The budget one is worth reopening, because the reason for dropping it does not hold

The note says (`dashboard-demo.md:325-327`): *"`budgets:manage` gates no route today; there is no budgets endpoint at all, so this is its own design task, not a dashboard line item."* Both halves are wrong:

- `budgets:manage` **does** gate something — the `monthly_budget_usd` field on `PATCH /orgs/{id}` (`app/services/organization.py:137`). The docstring above it is explicit that it was wired up so the permission would not be *"a row that means nothing"*.
- The cap **is** stored: `organizations.monthly_budget_usd`, with a CHECK constraint keeping it positive (`app/db/models/organization.py:62,80`), and per agent as `AgentSpec.budget.monthly_usd` (`app/agents/spec.py:175-185`).

So *managing* budgets is fair to call its own design task. *Showing the headroom* is reading two stored values against a number `/spend` already returns.

And it is a real gap on the steward's page as designed: the outcomes donut draws `budget_exceeded` — the **symptom** — while nothing anywhere shows the **cause**. A steward sees runs failing and cannot see that the org is at 94% of its cap. That is the one card I would trade something else for.

The MCP one is second for a similar reason: when a client's OAuth token expires, every agent using that server quietly loses its tools, and nothing on this page would say so. `last_status`, `last_error` and `last_checked_at` are all there and nothing reads them.

## One observation about the approvals card, for context only

While checking the approvals widget I noticed `ApprovalStatus.EXPIRED` is defined (`app/db/models/agent_run.py:193`) and permitted by the status CHECK constraint (`:243`), but nothing in `backend/app` ever assigns it — those two lines are its only occurrences. A parked tool call stays `pending` indefinitely.

Pre-existing and well outside this PR, so I have not touched it and I am not asking you to. It only matters here because it shapes the card: "Waiting on a decision" has no ceiling, so the oldest row could be from months ago with nothing escalating it. Worth knowing when you design what that card sorts by and whether it needs an age threshold of its own.

DEENUU1:
## Decisions on this design — including two that reverse it

Reviewed alongside #45 / PR #202, which shares this design's `group_by` vocabulary and its deep-link contract. The note is the best map in the repo and I am not asking for it to be rewritten — but four of its open calls now have answers, and two of them go against what it argued.

### Reversed

**Decision 3 is reversed: a per-person breakdown is in.** The note says "adoption is a count, not a table of names — 14 of 23 members ran an agent answers the steward's question without shipping a surveillance table", and #202's §6 repeats it as the reason a per-person spend table stays out. The call went the other way.

Cheaper than either document implies: **`group_by=user` is already in the frozen `/stats/usage` vocabulary**, so nothing was refused at the endpoint — only the rendering. What genuinely needs deciding, and belongs in this design rather than in code review:

- **Whose names, to whom.** `runs:view` is held by builder and operator as well as the stewards; it is a wider audience than the person being listed may assume. If the answer is "any `runs:view` holder", say so on the card.
- **Cost or usage or both.** "Who ran the most agents" and "who spent the most money" read very differently to the person on the list.
- **The aggregate card stays.** The count answers the steward's question at a glance; the table is the evidence under it. Do not replace one with the other — that is the same summary-and-its-evidence pattern the rest of this page follows.

**A user-arranged dashboard is in, and moves into stage 2.** The note puts it out of v1 but settles the entire shape, which is why this is the cheapest of the reversals: `effective layout = preference ?? audience default`, then `visible = effective ∩ gate()`; may only ever **reorder or hide, never reveal**; a stored layout naming a hidden widget is dropped at render time; spans from the closed s4/s6/s8/s12 set; the "add a widget" catalog itself passes through `gate()` so the list cannot leak what the page hides; widget definitions carry catalog metadata from day one; and preferences stored **per user and per organization**, because the same person is a steward in one org and a member in another.

Implement that shape. Do not redesign it, and do not let it weaken the gate: the four audience layouts stay the default, and a preference is a third layer over the two that already exist.

Two things in the note's own "must not be ported as-is" list get load-bearing here rather than staying advice: span and title must ride the **layout entry** rather than a widget-id lookup (otherwise "runs over time, once org-wide and once for me" is a latent collision the moment somebody arranges their own page), and `gate()` must take `(can, isAppAdmin)` rather than closing over module state.

### Confirmed

- **Decision 4: the org-level answer-quality card is in.** It is the only widget answering "is it any good", and message ratings carry a **comment** as well as a score — nothing below `admin_ratings.py` surfaces either. Filed as **#209**, which also carries the half that makes the number actionable: the 👎 on the run row in Activity, and the comment in the run detail view. `GET /ratings/summary` with the same `scope` parameter, per-day series as well as the headline, and one guardrail the note should state — a free-text comment shown at `scope=org` is user-written text about a specific conversation reaching a wider audience than its author expected. Decide whether the org card is counts-only with the text behind the run detail.
- **Managing budgets stays out**, as the note has it. One addition: the outcomes donut draws `budget_exceeded` and Activity shows the same symptom, and neither offers a way out. A link to the organization's settings from the headroom card is not a budget-request flow and is worth having.
- **Live updates stay out for the aggregates, in for the queue.** Refetch on focus is right for daily aggregates. Approvals and `running` runs are the two places where state changes in minutes rather than days — the approvals count in the quick-actions row is a number somebody acts on, and a stale one sends them to decide something already decided.
- **Deployment-wide time series, the audit strip, and sandbox** all stay out. #129 owns the sandbox view and has its data layer already.

### Shared with #202

- **Both design files need adding to `mkdocs.yml`'s `nav`** — flagged by the reviewer on both PRs, and #202's plan explicitly says these "should inherit whatever #149 decides". This is that decision: both go in, same section, in the same change as whichever lands second.
- **#208 is filed** for the surface-recording widening — `RunSurface.EMBED` stamped by `embed_session.py` (today an embedded run is recorded as `web`) and `mattermost` added to `_SURFACES` in `mentions.py` (today it falls through to `api`). Both designs priced it and both said "whichever branch lands it first ships it", which is how work with two owners ends up with none. It is what makes "Where runs come from" truthful; the card is not blocked from shipping without it. **#207** covers the sibling: `RunSurface.SCHEDULE` is defined and never assigned, which is why both designs had to write a paragraph explaining an omission.
- **#210** — sort and filter run history by duration. This page reports p50/p95 and there is no way to reach *those runs*; every other figure here has its rows one click away.
- The deep-link contract works in one direction only. Activity should be able to hand back: "see this on the dashboard", carrying the same `from`/`to` and the same nine `group_by` dimensions.

docs/design/dashboard-demo.md
OchnikBartek: **Option 1, widened.** `RunSurface.EMBED` exists and `embed_session.py` stamps it; `mattermost` is in `_SURFACES` and no longer falls through to `api`. Both have regression tests — a parametrized row in `test_channel_mentions.py` and `TestTheRunRecordsItsSurface` asserting the runner is called with `RunSurface.EMBED`.

Deliberately **no migration and no backfill**: rows written before the cutover keep the surface they were recorded under, so an old period still folds widget runs into `web`. `docs/channels.md` says that, and `SurfaceCount`'s docstring repeats it where somebody reading the API would hit it.

`SCHEDULE` is still dead and still not charted — it is #207 now rather than a paragraph explaining an omission.

docs/design/dashboard-demo.md
OchnikBartek: **Fixed, and the invariant is now enforced rather than only written down.**

Five segments covering all six values: `completed`, `failed`, `budget_exceeded`, `awaiting_approval` on its own tone, and `running`/`cancelled` folded into one neutral segment the legend names. The three-way failure split stayed.

The invariant you asked for — the segments sum to `total_runs`, and the awaiting segment counts the same parked runs as the approvals card — is in the note, in `UsageStats`' docstring, and is now a test: `test_the_status_split_still_sums_to_the_total`. That test earned its place immediately, because merging #200 forced a decision about whether delegated runs count, and a half-applied filter would have broken exactly this.

docs/design/dashboard-demo.md
OchnikBartek: **Both halves done.** The row reads `no — needs a status filter on /runs`, and the filter is built.

It takes a list, as you suspected the operator's version wants: `?status=failed,budget_exceeded`. An unknown value is a **422 naming the vocabulary**, not an empty 200 — the specific lie you described ("No failures — every run in this period finished cleanly") is what `test_an_unknown_status_is_refused_not_matched_against_nothing` exists to prevent.

Since #200 moved run history behind `AgentRunnerService`, the filter is threaded through the service rather than the route reaching for the repository, and it composes with `include_delegations` rather than replacing it — two tests pin that pair.

docs/design/dashboard-demo.md
OchnikBartek: **Your second option, and thank you for the correction — `days: int` is right and the note was wrong.**

`/spend` is untouched. The period total moved to a `cost` block on `/stats/usage` (period, previous period, split by provider), so the card obeys the filter using an endpoint that can express a window. The month-to-date line stayed, and says in its own copy that it is a calendar month, reconciles against an invoice, and is deliberately not scoped by the filter.

The sentence you were commenting on still claimed the page queries `?from=&to=` on `/spend`. That is fixed in this push — the *Filtering* section now states outright that `/spend` takes `days: int`, that a range ending two weeks ago is unrepresentable, and why the total moved instead of the endpoint widening.

docs/design/dashboard-demo.md
OchnikBartek: **Vocabulary frozen exactly as you wrote it**, in the note and in the endpoint's own docstring:

`day | surface | agent | version | user | status | model | exposure | environment`

v1 implements two of the nine. `version` feeds the builder's compare card — `agent_version_id` LEFT JOINed so a run whose version was deleted survives as a row rather than vanishing, which is the property that column's comment exists to protect. `user` was added by your later review and feeds the per-person table.

`exposure` and `environment` stay in the contract unimplemented, for #45 to land. An unimplemented word is a **422**, not an empty answer — the vocabulary is a contract, so asking for a dimension that does not answer yet has to say so.

docs/design/dashboard-demo.md
OchnikBartek: **Settled here, then promoted by your own follow-up into stage 2.** The note carries the shape verbatim: `effective = preference ?? audience default`, then `∩ gate()`; reorder-or-hide-never-reveal, with a stored layout naming a hidden widget dropped at render; the catalog itself passing through `gate()`; preferences per user **and** per organization, for the steward-here-member-there reason you gave.

Four of the five prerequisites already hold in the shipped code, so it stays cheap: span and title ride the layout entry, `gate()` is injected as `(can, isAppAdmin)`, every widget declares a "see all" destination, and widget definitions carry catalog metadata from day one.

What is **not** in this branch is the persistence and the arranging UI — a preferences table, migration, service and routes under the 100% gate, plus a catalog modal. Reasoning is in the PR body; short version, it roughly doubles a branch already at ~200 files and wants its own review.

One correction while moving it: the note claimed the grid uses a closed `s4/s6/s8/s12` span set. That was wrong about its own demo — the layouts pair `s7`+`s5` and `s8`+`s4`. The real set is `s3`–`s8` plus `s12`, and narrowing it would leave somebody unable to rebuild the default they started from.

docs/design/dashboard-demo.html
OchnikBartek: **All three honoured in the port.**

Span and title ride the **layout entry** (`LayoutEntry { widget, span, titleKey? }`), not a lookup keyed by widget id — so the same widget can appear twice with different widths and headings. That stopped being latent within this branch: the viewer's layout overrides the my-agents title, and the per-person card renders beside the count it belongs under.

No span→pixels table anywhere; `ResponsiveContainer` measures. `SPAN_CLASS` maps a span to a literal Tailwind class and nothing else.

`gate()` takes `(can, isAppAdmin) => boolean` and closes over nothing. That is what made `dashboard.integration.test.tsx` cheap — nine scenarios across four audiences by swapping one `can`, including the assertions that a viewer's browser never *requests* `/stats/usage`, `/approvals`, `/spend` or the member list, and a member never asks the per-person question.

docs/design/dashboard-demo.md
OchnikBartek: **Path corrected** to `GET /api/v1/orgs/{org_id}/members`.

**And the sentence written.** The widget map's fine print now says outright that Members is the one row where the permission does layout work rather than marking a boundary — the endpoint is open to any member, the card is merely *featured* on steward pages, and nothing is actually withheld. Your related point is in the same paragraph: `active_users` carries `total_members` under `runs:view`, so a builder gets the headcount from that card anyway.

Worth flagging how far that reasoning travelled. Your later review reversed the aggregate-only adoption call, and "the permission is wider than the card's subjects would assume, so say so on the card" became the rule for the per-person table too — its copy names `runs:view` and states that builders and operators see it, and the layouts feature it wherever usage is featured, because a disclosure true on only some pages is not one.

docs/design/dashboard-demo.md
OchnikBartek: **Path corrected** to `/kb`, and the `exists today` cell no longer reads `yes` — it says the lists return own+shared as one page and names what was missing.

`?shared_with_me=true` is built on agents, skills and kb. Predicate: org-visible **or** explicitly granted, **and** not owned by the caller.

The trap worth naming, because it is the one that would have shipped silently: a role with `Scope.ALL` gets `None` back from `visible_resource_ids` meaning "no filtering needed", so the grant list has to be fetched *explicitly* for those roles. Without that, "shared with me" degenerates to "the whole organization minus mine" for exactly the people most likely to look at it. `TestSharedWithMeIsWhatWasDeliberatelyShared` covers it against a real database, and kb additionally excludes PERSONAL and APP scopes.

docs/design/dashboard-demo.html
OchnikBartek: **Applied.** `owner` and `admin` now carry `mcp:manage` and `connections:manage`, matching `_ALL_GLOBAL`.

The comment above the mirror also says what you asked it to: the mirror holds only what the dashboard gates on, and `connections:manage` is the single deliberate exception, reserved by the sandbox section the note names — so the next reader knows a listed-but-uncalled permission is intentional rather than drift.

Your reason for fixing it was the right one, and it was about to bite: the sandbox section is still the named example of "absorbs cleanly later", so against the old mirror it would have rendered for a builder and hidden from the owner.

docs/design/dashboard-demo.html
OchnikBartek: Kept — and carried into the product rather than left in the mock.

`WidgetErrorBody`'s Retry calls `refetch()` on that widget's own query, so one card recovers while the others stay failed. It is asserted rather than claimed: `dashboard.integration.test.tsx` returns a 502 for `/stats/usage` only, then checks the usage cards show the error body, the approvals and agents cards still render their data, and Retry re-issues the request. The e2e spec takes the other side — after load, zero cards showing "Couldn't load this".

backend/tests/api/test_runs_filter.py
```diff
@@ -0,0 +1,82 @@
...
+
+pytestmark = pytest.mark.anyio
+
+
+@asynccontextmanager
+async def _client() -> AsyncIterator[AsyncClient]:
+    """A caller holding runs:view, with the database stubbed out."""
+    ctx = AuthContext(user_id=uuid4(), organization_id=uuid4(), role="owner")
+    app.dependency_overrides[deps.get_auth_context] = lambda: ctx
+    app.dependency_overrides[deps.get_db_session] = lambda: AsyncMock()
```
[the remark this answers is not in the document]
OchnikBartek: Not applied — the lambda is load-bearing here, and swapping it for the class breaks all six tests in this file.

FastAPI introspects a dependency override's signature. `AsyncMock`'s is `(*args, **kwargs)`, which it turns into two **required query parameters**, so every request answers 422:

```
VALIDATION_ERROR: Some fields need fixing: args, kwargs
```

`lambda: AsyncMock()` has an empty signature, which is what makes it a zero-dependency override that constructs a fresh mock per resolution. I tried the suggestion before rejecting it; the file now carries a comment saying why, so the next person does not have to.

Issue #199, opened 2026-08-05
A run count is built as "{n} runs", and check_i18n.py does not notice


DEENUU1:
## What breaks

Two spend breakdowns on Activity build a count the way only English builds one, and the
i18n guard lets them through.

`frontend/src/app/[locale]/(dashboard)/runs/page.tsx`:

```tsx
<span className="text-muted-foreground text-xs">{entry.run_count} runs</span>
```

once in the `spendByAgent` list and once in `SpendBreakdown`. `.claude/rules/frontend.md`
is explicit: "A count is an ICU `plural`, never a ternary", and a text node mixing words
with an interpolation "is one message with a named parameter, not English with a hole in
it". The catalog already holds the right shape elsewhere -
`pages.sandboxes.registeredCount` is `{count, plural, =1 {1 connection} other {#
connections}}`.

In Polish it renders "1 runs" and "3 runs" - the word never declines, and in a locale
without a bare plural it cannot be translated at all.

The second half is more interesting than the strings: **`scripts/check_i18n.py` passes
these.** It is the guard whose whole job is refusing exactly this, so whatever it looks
for does not see `{expr} word` inside a JSX text node. Worth fixing there as well as
here, or the next one lands the same way.

## Sequence

`cd frontend && make lint` - clean. Switch the locale to `pl` and read the spend tab.

## How you would know it was fixed

Both counts pass `count` to an ICU `plural` message, and `check_i18n.py` fails on a
reintroduced `{expr} runs` text node.

## Where it sits

Pre-existing; found implementing #181, which rewrote the run table on the same page and
deliberately did not touch the spend breakdowns beside it.

Cluster: #139 (the one table primitive) is where these two lists eventually go, so a
fix here should not be a rewrite. Related: #181.

Issue #198, opened 2026-08-05
Activity counts runs for all time next to spend for one month


DEENUU1:
## What breaks

The three figures at the top of Activity read as one row of related numbers, and two of
them measure different windows:

| Card | Window |
|---|---|
| Spend this month | calendar month to date (`CostSummary.month_to_date_usd`) |
| Runs | **all time** - `AgentRunList.total`, which `list_runs` counts with no time filter |
| Waiting on a person | current queue |

So an organization three years old shows "8,412 runs" beside "$31.20", and the obvious
reading - that those runs cost that money - is wrong by three years.

`frontend/src/app/[locale]/(dashboard)/runs/page.tsx:88` renders the count;
`backend/app/repositories/agent_run.py` `list_runs` builds its `count_query` from
`organization_id` alone, while `sum_cost_since` takes a `since`.

## Sequence

Open Activity on any deployment with more than a month of history. The two numbers
cannot be reconciled and nothing on the page says why.

## How you would know it was fixed

Either the count is scoped to the same calendar month as the spend beside it - which
needs `list_runs` (or a new aggregate) to accept `since` - or the cards are labelled so
each says its own window and neither invites the comparison.

## Where it sits

Pre-existing; found implementing #181. That fixed the *other* half of the same card: the
count included delegated rows while the money excluded them, and it was rendering
`runs.length` (a page of fifty) rather than `total`. The window mismatch is orthogonal
and was left alone, with the card captioned about delegations only.

Cluster: #170 (the backend aggregates for run history and spend) is where a windowed
count belongs. Related: #181.

DEENUU1:
**#45 stage 2 should close this, rather than rediscovering it.**

The Activity design (PR #202) reaches the same conclusion independently — its §3 item 5 prices "the Runs figure renders `runs.length`, at most 50, as the organization's run count, while the hook already returns an unused `total`" — without knowing this issue exists. The window half described here is the part that plan does *not* cover: its mockup's Runs figure follows the runs tab's date range while the Spend figure stays calendar-aligned, so two of the three cards still measure different things and only one says which.

Signed off for stage 2 as: each figure names its own window in its caption, and the Runs count is scoped to a window rather than counting all time. Recorded on #45 ([comment](https://github.com/vstorm-co/agenticos/issues/45#issuecomment-5190861177)).

Cluster note: #170 is where a windowed count belongs on the backend, as this issue already says. #181 fixed the delegation half of the same card.

Issue #197, opened 2026-08-05
Run history routes call the repository directly, which the layering rule forbids


DEENUU1:
## What breaks

`app/api/routes/v1/runs.py` imports `agent_run_repo` and calls it from two handlers:

- `list_runs` -> `agent_run_repo.list_runs` (`backend/app/api/routes/v1/runs.py:47`)
- `get_run` -> `agent_run_repo.get_run` (`backend/app/api/routes/v1/runs.py:62`)

`CLAUDE.md` and `.claude/rules/architecture.md` both state it plainly: "Routes call
services only - **never** import or call a repository directly." Every other handler
in the same file goes through `AgentRunnerSvc` or `ApprovalSvc`.

It is not only tidiness. `AgentRunnerService` is where tenant scope, spend arithmetic
and `resolve_access` live for this domain, so a rule that belongs to run history has
two places it could be added and one of them is invisible to the service tests. The
sweep in `tests/api/test_platform_routes.py` checks the permission gate on these
routes, not the layer they call.

## Sequence

Read the file. Both handlers take `db: DBSession` rather than a service, which is
itself the tell - no other platform route does.

## How you would know it was fixed

`rg 'agent_run_repo' backend/app/api/` returns nothing, and the two reads are methods
on `AgentRunnerService` with their own service tests.

## Where it sits

Pre-existing; found implementing #181, which added a `parent_run_id` and an
`include_delegations` parameter to `list_runs` and had to thread both through the
route rather than through a service. Deliberately not folded into that branch - it is
a refactor of its own and #181 is about a surface reading columns nobody read.

Cluster: nothing quite matches. Nearest is #170 (the backend aggregates for run
history and spend), which is the other issue about this module's read path.

Issue #139, opened 2026-08-04
One table primitive, with sortable headers and filters, everywhere


DEENUU1:
Every table in this product is built twice over, and only two of them can be sorted. This is the cross-cutting issue behind several page-level complaints — filed once rather than repeated on each page.

## Two primitives, and neither one sorts

**`components/ui/data-table.tsx`** — a `Column<T>[]` config with built-in loading skeletons, an empty state and responsive column hiding. Used by four pages: `admin/users`, `admin/ratings`, `admin/conversations`, `orgs/{id}/members`, `kb/{id}`.

**`components/ui/table.tsx`** — the shadcn `<Table>/<TableHead>/<TableRow>` primitives, hand-assembled. Used by `sandboxes/connections-table.tsx`, `sandboxes/sessions-panel.tsx`, `sandboxes/policy-panel.tsx`, `sandboxes/workspace-browser.tsx`, `teams/members-table.tsx`, `vault/secrets-table.tsx`.

They do not look the same. `DataTable` headers are `font-mono text-[11px] uppercase tracking-wider`; the shadcn ones are not. `DataTable` wraps itself in `rounded-xl border bg-card`; the others are dropped inside a `Card` by the caller, with its own padding. Row height, empty state and loading state are per-caller in the second group.

**Neither supports sorting.** Where sorting exists it was written locally, twice:

- `admin/conversations/page.tsx:72-95` — a `SortButton` with `ArrowUpDown`/`ArrowUp`/`ArrowDown`, wired to the backend's `sort_by`/`sort_dir`
- `admin/users/page.tsx` — a `SortHeader` doing the same thing, separately

Neither is exported. Every other table in the product — sandbox connections, running sessions, vault secrets, org members, workspaces, kb documents — has fixed-order rows and no header controls at all.

Filtering is the same story: `useListControls` + `SearchInput` (`components/ui/list-controls.tsx`) exist and are good, and they are used by five *galleries* and by zero tables.

## What to build

**One table.** `DataTable` is the right base — it already has the config-driven columns that make this possible. Extend `Column<T>`:

```ts
sortable?: boolean;          // renders the header as a sort control
filter?: "text" | "select";  // a per-column filter in the header row
```

and give `DataTable` two modes for both, because the two kinds of list in this codebase are genuinely different and `list-controls.tsx:14-26` already draws the distinction:

- **Client-side** — the caller holds every row. Sorting and filtering are `useListControls`' problem, extended with a comparator.
[...]

PR #200, opened 2026-08-05
feat(runs): tell a delegated run from one somebody started

### LINKED ISSUE #181: Run history cannot tell a delegated run from one somebody started: the columns exist and nothing reads them
**Caused by #164** (delegation) in the sense that it created the rows; the surface work was
scoped out of that branch deliberately. Filing so the other half is not lost.

## What breaks

Delegation writes an `agent_runs` row per delegation to a published delegate, and the backend
was given everything a surface needs to present it. Nothing consumes any of it:

- `AgentRunRead` sends `parent_run_id` and `subagent_task_id`
  (`backend/app/schemas/agent_run.py`), the first with a description explaining that "a
  surface that sums a page of rows double-counts every delegation".
  `frontend/src/types/runs.ts` declares neither — `rg parent_run_id frontend/src` is empty.
  Extra keys on a JSON response are invisible to `tsc`, so nothing catches it.
- `list_runs` (`backend/app/repositories/agent_run.py`) applies no filter, so delegated rows
  interleave into run history pixel-identical to runs a person started. One fan-out turn of
  three delegations reads as **4 runs** at `runs/page.tsx`, and the cost column reads as
  additive down the page — next to a `month_to_date_usd` that correctly excludes children.
  So the page contradicts itself.
- `SubagentCompleteFrame.run_id` (`frontend/src/types/chat.ts`) is declared and read nowhere;
  the reducer copies five of the six fields and drops it. `DelegationRecorder`'s docstring says
  the id travels back "so a surface can link a delegation panel to the run history entry it
  produced."
- `agent_runs_parent_run_id_idx` exists for what migration `0007` calls "the run-detail
  question — what did this run delegate", and no query selects by `parent_run_id`. There is no
  run detail page.

Three docstrings promise a link no code makes, and the index is speculative weight on the
hottest insert table.

- Run history distinguishes the two kinds: either delegated rows are nested under their parent,
  or badged and excluded from the top-level count. The count next to the month-to-date figure
  agrees with it.
- Clicking a delegation panel in chat reaches that delegation's run, using the `run_id` the
  frame already carries.
- A query exists that the `parent_run_id` index serves — or the index is dropped and the
  migration's docstring stops claiming it.

## Smaller, adjacent

`agent_runs.parent_run_id` is `ON DELETE SET NULL`, which is right — deleting a parent orphans
the child rather than cascading, and the child correctly starts counting toward the
organization's bill. But `subagent_task_id` is left populated, so the row becomes a top-level
run carrying a delegation handle that points at a transcript nobody can reach. Null it in the
same statement, or make the surface ignore it when `parent_run_id IS NULL`.

Related: #40 and #164 (delegation), #170 (the backend aggregates, fixed there), #139 (the one
table primitive — if that lands first, nesting belongs in it rather than in a bespoke run
table). Cluster and spine on #168.

DEENUU1:
Delegation (#164) writes an `agent_runs` row per delegation, and #192 put
`parent_run_id` and `subagent_task_id` on `AgentRunRead` and split the cost so a
parent's row already contains its children's. What was still missing: run history
and the Activity page read none of that, so a delegated run and a run somebody
started looked identical and summed together. This closes that.

## The contradiction, in numbers

`sum_cost_since` counts only rows where `parent_run_id IS NULL`, because every run
shares one spend ledger and a child row is the same money written down twice.
`list_runs` applied no filter at all. So one turn that fanned out to three delegates
read like this on Activity:

| | Before | After |
|---|---|---|
| Rows in the run table | 4 | 1 |
| Cost column, read down the page | $1.00 + $0.40 + $0.40 + $0.40 = **$2.20** | **$1.00** |
| "Runs" figure | **4** | **1** |
| "Spend this month" beside it | **$1.00** | **$1.00** |

Two smaller things in the same figure, found on the way:

- The card rendered `runs.length`, a page of fifty, not `total`. An organization with
  213 runs read as 50.
- It followed `?agent=`, so arriving from the Builder put **one agent's** run count
  beside the **organization's** month. It now reads the organization's total
  unconditionally; the table below keeps the filter and says so.

## Badge, not nest — and why

**#139 proposes one table primitive for the whole product**, and a nesting tree built
into a bespoke run table is one that gets thrown away when that lands. So: delegated
rows are badged and excluded from the top-level list, and the drill-down is a query
rather than a widget. The nesting a reader actually wants — a specialist's own
delegations under it — already exists in the chat delegation panel.

`list_runs` now takes the same two-sided arithmetic `sum_cost_since` already had:

| Ask | Answer |
|---|---|
| `GET /runs` | Runs somebody started. `parent_run_id IS NULL` |
| `GET /runs?agent_id=&include_delegations=true` | One agent's own history |
| `GET /runs?parent_run_id=` | What that run delegated. Takes precedence over both |
| `GET /runs/<id>` | One run, delegated or not |

`include_delegations` is not decoration. Narrowed to one agent the arithmetic inverts —
a delegate's rows are the only record of what it itself did — and without it an agent
that only ever runs as somebody's delegate would show an **empty history beside a spend
figure of forty dollars**. That is why `useRuns(agentId)` sends the flag.

`?parent_run_id=` is also the query `agent_runs_parent_run_id_idx` was created for in
`0007` and had never had. The migration's docstring claimed the index was speculative
weight; it now names the caller.

## Reaching a delegation's run from chat

The chat delegation panel already holds the `run_id` its terminal frame carries (#164);
nothing consumed it. Now:

`?run=<id>` on Activity shows one run, its delegations under it each badged
`Delegated · 4f2a1b8c`, and a link **up** from a delegated run to the one that was
charged for it. The delegation panel links **down** to it — the only place a
delegation's own cost, model and tokens are recorded as its own rather than folded into
the parent's.

No run detail page. The narrowed table answers the question with the components that
already exist, and a second partial copy of Activity is one that drifts.

- The link is **not rendered** without `runs:view`, rather than rendered and then 403'd.
- It is absent for an inline specialist, which gets no run row at all — the same shape as
  an unlinked delegate in the approval queue.
- `FocusedRun` tells a run that is not there (404, which is also the answer for another
  tenant's run) from one that could not be read. Every page here draws its empty state
  when a query fails, and "no such run" for a refused permission sends somebody looking
  for a run that is there.

## The orphaned task id: the schema, not a migration

`parent_run_id` is `ON DELETE SET NULL`, which is the right arithmetic. But a foreign
key can only null its own column, so an orphan kept a `subagent_task_id` naming a
transcript that went with the parent. **`AgentRunRead` withholds it whenever
`parent_run_id` is null. No migration.**

A trigger would be the only database-side answer, because the parent is usually deleted
by a cascade from its agent with no application code in the transaction — and a trigger
on every insert into the hottest table in the schema is a permanent cost for a case that
happens when somebody deletes an agent. Every surface reads this schema, so one guard
here is the same guarantee for a fraction of the weight. The stored value stays wrong
and unread, which the migration's docstring now says.

The row itself still lists, badged `Delegated` without a handle: it counts toward the
bill now, so it belongs on the page.

## How it was verified

- `make check` green end to end — `make lint`, `make test`, `make test-frontend-cov`,
  `build-frontend`, `docs-build` (`--strict`), audit.
- Backend: platform layer at **100% statements and branches**. New integration tests
  against real Postgres — a fan-out listing as one run, the per-parent query,
  `include_delegations` both ways, `parent_run_id` winning over it, tenant isolation on a
  guessed parent id, and an orphan listing top-level with its handle withheld.
- Frontend: `bun run test:coverage` (**not** `test:run`) — **100%**
  lines/statements/functions, **≥97.5%** branches, the gate CI applies.
- One e2e guard is worth reading on its own: `runCount` in `delegation.spec.ts` asserts
  the delegate has never run *before* the delegation, so the row found at the end must be
  this turn's. It asked `/api/runs?agent_id=`, which now excludes exactly the delegated
  rows a delegate's history consists of — so it would have answered 0 whether or not a
  leftover row existed. Fixed by threading `include_delegations`; nothing about it was
  visible from a green suite.

## Seen and deliberately not fixed

- **#197** — `GET /runs` and `GET /runs/{id}` call `agent_run_repo` directly from the
  route, which the layering rule forbids ("routes call services only"). Pre-existing on
  `main`; this branch threaded two more parameters through the existing call rather than
  fold in a refactor that #197 owns for all five repository-calling routes.
- **#198** — the "Runs" figure is all-time while "Spend this month" is calendar-month.
  They agree on delegations now, which was the reported half; the window is a separate
  question the card is honest about.
- **#199** — `{n} runs` pluralisation elsewhere on this page is English-only, which
  `scripts/check_i18n.py` does not catch. Untouched here; the new copy uses ICU `plural`.

Closes #181


DEENUU1:
Cross-linking: **#202** (stage-1 design for the Activity page, #45) was written without knowing about this branch, and the two collide on four items — the run detail view, the top-level-only run list, `RunTable` as its own component, and the `total` fix on the Runs figure. Its mockup's approval card also has no delegate, so building the Approvals tab to it as drawn would delete `ApprovalDelegate` and its integration test.

Raised there as the one blocking item ahead of stage 2 ([review](https://github.com/vstorm-co/agenticos/pull/202#pullrequestreview-4863429187)); this branch is closer to landing, so #202 reconciles to it rather than the other way round. Nothing is asked of this PR.

Two details from here that the Activity design has to absorb, both from `record_delegated_run`: `environment_id` is deliberately NULL on a delegated run, so an Environment filter narrowed to `production` silently drops every delegation; and a per-version cost summary has to pick a side of `include_delegations` and say which, or it counts a delegate's spend twice inside its parent's.

PR #217, opened 2026-08-05
chore: cut 0.0.7

DEENUU1:
Two gates that existed and did not run. No behaviour change, no schema change,
`SPEC_VERSION` unchanged at **7**.

Cut now rather than batched with the delegation branch, because this repo's own
release argument says so: the point of a release commit is that the version
literals and both lockfiles move together, and #190 and #204 were already sitting
on `main` unreleased. Delegation is large and still taking review rounds; holding
a release for it is how a lock ends up disagreeing with its manifest.

## What is in it

- **#190** — `make check` now runs every job CI runs. It was documented as "what
  CI runs" and ran about half: `bun run build`, `pip-audit` and `mkdocs --strict`
  had no local equivalent at all, and eslint, prettier and `tsc` sat outside
  `make lint`. One divergence ran the other way and is the sharper one — the i18n
  check was local-only, so a pull request could merge an untranslated string.
- **#204** — spelling is checked over the tree rather than over the files a commit
  happens to touch. One misspelling was sitting on `main`, waiting for whoever next
  opened that file for an unrelated reason.

Both are fixed structurally: the workflow calls the Makefile's targets, and
`backend/tests/test_ci_parity.py` asserts both directions, so a job added to one
and not the other fails the suite rather than drifting.

## Verified

`make check` green on this commit — and this is the first release where that
command means what it says. 2714 passed, the platform layer at 100% of statements
and branches, the frontend gate, `next build`, `mkdocs --strict`, `pip-audit`.

The changelog's reference links were checked in both directions: every `## [x]`
heading has a link and every link a heading.

One thing worth recording: the new spelling gate went red on the changelog entry
itself, which had quoted the misspelling in order to explain it. It no longer
quotes it, and says why. The gate earned its place inside an hour of existing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,57 @@ Two things are versioned separately from this file and worth knowing about:
 
 Nothing yet.
 
+## [0.0.7] - 2026-08-05
+
+Two gates that existed and did not run. No behaviour change, no schema change,
+`SPEC_VERSION` unchanged at 7 — cut now rather than left to accumulate, because
+the point of a release commit here is that the version literals and both
+lockfiles move together.
+
+Both changes are the same defect wearing two costumes: a check that reports
+green while the thing it checks is unchecked. The interesting part of each is not
+the wiring but what was found once it ran.
+
+### Fixed
+
+- **`make check` now runs every job CI runs.** It was documented as "what CI
+  runs" and ran about half of it. Missing entirely: `bun run build`, `pip-audit`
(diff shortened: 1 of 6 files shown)


DEENUU1:
Fixed in `a979e00` — the probe now reports `app.__version__`, and `test_liveness_probe_reports_the_build` asserts it rather than asserting everything except it. The changelog carries the entry, because a release that fixes the one thing a release claims is worth recording. Good catch: this was wrong from 0.0.1 and the `getattr` is what kept it quiet.

Issue #148, opened 2026-08-04
rag: KB stats count status "completed", but ingestion writes "done", so indexed_count is always 0


OchnikBartek:
**What breaks:** `indexed_count` on every knowledge base is always `0`, however many documents finished ingesting.

**Sequence:**
1. Ingest any document; the pipeline marks the row `status = "done"` (backend/app/services/rag_document.py:276).
2. `GET /api/v1/kb` or `GET /api/v1/kb/{id}` → `indexed_count: 0`.
3. `SELECT status, count(*) FROM rag_documents GROUP BY status;` on a live deployment returns only `done` — no row has ever held `completed`.

**Why:** the stats aggregation filters on a status value nothing writes — backend/app/repositories/rag_document.py:189:

```python
func.count().filter(RAGDocument.status == "completed"),
```

The documented status set is `processing` / `done` / `error` (docs/file-processing.md, RAGDocument model), so the filter can never match.

**Fixed when:** the filter counts `done`, a KB with one finished document reports `indexed_count == 1`, and a test pins the status literal to the one the pipeline writes.

Pre-existing on `main`; found exercising the #61 branch against a live stack. Sibling of the `chunk_count` omission (#147) — both make the same stats row lie, but the fixes are independent.

DEENUU1:
Cross-linking from the #168 map rebuild: **#147 is the line above this one.** `repositories/rag_document.py:188` sums `chunk_count` values that ingestion never records; `:189` filters on a status nothing writes. Independent fixes, one function — worth one pull request, which the map records as a pair.

Neighbours: #548 (PR #670) and #566 on the document-lookup helpers, #707 on the CLI write path.

DEENUU1:
Confirmed live while fixing the replace-mode count on #834. `test_re_ingesting_a_document_does_not_count_it_twice` prints `CollectionCounts(documents=1, chunks=12, indexed=0)` for a collection holding one `status='done'` document - the `FILTER (WHERE status = 'completed')` in `counts_by_collection` (repositories/rag_document.py:179) matches nothing production ever writes.

Worth noting the integration test hides it: `test_only_completed_documents_count_as_indexed` in `tests/integration/test_platform_flows.py` seeds `status = 'completed'` and `'failed'`, neither of which exists in the code - `update_status` writes `'done'` and `'error'`. Fixing the query without rewriting that test would leave it green for the wrong reason.

Issue #147, opened 2026-08-04
rag: chunk_count is never recorded, so every document claims 0 chunks


OchnikBartek:
**What breaks:** every ingested document permanently records `chunk_count = 0`, so knowledge-base stats (documents/vectors on the KB cards and the detail header) report empty collections that in fact answer searches. The parsed-content view computes `len(chunks)` live from the vector store and shows the real number, so the product contradicts itself.

**Sequence:**
1. Upload any file to a KB (`POST /api/v1/kb/{id}/documents`), wait for `status: done`.
2. `GET /api/v1/kb/{id}/documents` → `"chunk_count": 0`.
3. `POST /api/v1/rag/search` against the same collection returns its chunks; `GET .../documents/{doc}/parsed` reports the true count.

**Why:** `IngestionResult` (backend/app/services/rag/models.py:98) carries no chunk count, and every `complete_ingestion` call site omits the argument, so the default `0` is written:
- backend/app/worker/tasks/rag_tasks.py:224 (upload flow)
- backend/app/worker/tasks/rag_tasks.py:326 (sync flow)
- backend/app/worker/background/rag.py:38
- backend/app/commands/rag.py:214

`complete_ingestion` defaults the parameter at backend/app/services/rag_document.py:269, and the stats aggregation sums the zeros at backend/app/repositories/rag_document.py:188.

**Fixed when:** after ingesting a document, its `rag_documents` row records the same chunk count the vector store holds for its `vector_document_id`, and a regression test covers the upload path.

Pre-existing on `main`; found exercising the #61 branch against a live stack. Related but distinct from the `indexed_count` status-value mismatch, filed separately.

DEENUU1:
Cross-linking from the #168 map rebuild: **#148 is the adjacent line of the same function.** This one is `repositories/rag_document.py:188` summing zeroes; #148 is `:189` filtering on a status literal (`"completed"`) that the pipeline never writes. Both make one stats row lie, the fixes are independent — but they are two lines of one aggregation, so one pull request and one regression test fixture covers both. The map records them as a pair.

Neighbours in the same cluster: #548 (PR #670) and #566 on the document-lookup helpers, #707 on the CLI write path.

Issue #141, opened 2026-08-04
check_i18n.py passes over most real sentences, including the 404 page


DEENUU1:
Found while scoping the `/sandboxes` and chat tool-rendering work, not caused by it.

`scripts/check_i18n.py` reports **"No hardcoded copy in frontend/src. 2552 messages in en.json."** It is wrong, and by more than a handful of strings.

> **Corrected 2026-08-04.** This issue first named interpolation as the blind spot. That is one of two, and the smaller one — the larger is a text node that does not share a line with its tags, which is what Prettier produces for any sentence longer than a few words. The examples below replace the original three.

## The 404 page is in English, and it imports the translator

```tsx
// app/not-found.tsx
export default function NotFound() {
  const t = useTranslations("pages.root");
  return (
    …
      <h1 className="…">
        Page not found
      </h1>
      <p className="text-muted-foreground mt-4">
        The page you&apos;re looking for doesn&apos;t exist or has been moved.
```

`useTranslations` is called and then not used for the two strings on the page. Nothing complains.

`app/global-error.tsx:50,59` is the same: "Something went wrong" and "An unexpected error occurred. Please try again or contact support if the problem persists." `app/[locale]/(auth)/magic-link-sent/page.tsx:49` — "We sent a sign-in link". `app/[locale]/shared/[token]/page.tsx:33,71`. `components/chat/tool-results/web-search.tsx:51` — "No web results found".

A crude scan for word-bearing lines carrying no tags, braces or punctuation of code turns up **229 candidates**; most are block-comment continuations, but the residue is well into the dozens and includes the error page, the not-found page and a step in the sign-in flow.

## Why the checker misses them

`JSX_TEXT = re.compile(r">\s*([^<>{}\n][^<>{}\n]*?)\s*<")` — the pattern requires the `>` of the opening tag and the `<` of the closing tag **on the same line as the text**. Prettier puts a sentence on its own line as soon as the element does not fit in 100 characters, and at that point the line the text is on contains neither.

The character class also excludes `{` and `}`, so a text node with any interpolation in it is skipped whole — which is the original finding:

```
components/sandboxes/sessions-panel.tsx:85    Running on {connection.name}
components/dashboard/active-sessions.tsx:181  Last active {timeAgo(session.last_used_at)}
components/sharing/sharing-panel.tsx:215      Access for {name}
components/chat/tool-results/web-search.tsx:61  {n} web result{n !== 1 ? "s" : ""}
```

[...]

PR #221, opened 2026-08-05
feat(rag): search every readable base from one knowledge page

### LINKED ISSUE #31: Gate the /rag write controls on collections:edit
**Severity: Medium** · AUD-032

## What's wrong

`frontend/src/app/[locale]/(dashboard)/rag/page.tsx` has no `usePermissions` import anywhere in the file, and renders every write control unconditionally.

## Where

- Controls at `:285` (create collection), `:516` (Upload files), `:528-544` (Delete collection), `:703-721` (Delete document), `:894` (delete sync source)
- Offered in the sidebar on `Perm.collectionsView` (`frontend/src/components/layout/app-sidebar.tsx:92`)
- Backend gates: `backend/app/api/routes/v1/rag.py` — `require(Perm.COLLECTIONS_EDIT)` on twelve routes (`:145,172,277,301,389,406,470,509,533,559,579,595`)

## How it fails

Viewer, Operator and — outside their own rows — Member hold `collections:view` without `collections:edit` (`backend/app/core/permissions.py:210-236`). The sidebar therefore shows them "RAG Search", and the page offers Upload / Delete collection / Delete document / Add sync source. Clicking any of them produces a 403 toast.

`.claude/rules/frontend.md:44-45`: a control the caller may not use is **not rendered** — not rendered and then 403.

The sibling page already does it right: `frontend/src/app/[locale]/(dashboard)/kb/[id]/page.tsx:86-88` computes `mayEdit = can(Perm.collectionsEdit)` with a comment naming Viewers specifically. `/rag` was never brought along.

## Fix

`const mayEdit = can(Perm.collectionsEdit)` and gate the five control sites, mirroring `kb/[id]/page.tsx`.

## Acceptance criteria

- [ ] A Viewer sees no write control on `/rag`
- [ ] `rag-page.integration.test.tsx` — Testing Library against a mocked `/me/permissions` returning `collections:view` only, asserting the upload and delete controls are absent. The rule requires an integration test, not Playwright

## Note

There is **no drift in the permission catalog itself** — checked specifically. `hooks/use-permissions.ts:32` fetches `/me/permissions` from the server rather than deriving anything; `types/permissions.ts:11` deliberately types `Permission` as an open template literal so the UI cannot hardcode the list; and the 21 constants at `:40-62` match `Perm` in `backend/app/core/permissions.py:37-63` exactly. The only role→behaviour mapping left in TS is `ASSIGNABLE_FALLBACK` (`use-permissions.ts:86`), a pre-catalog placeholder replaced as soon as `/roles/catalog` answers.

### LINKED ISSUE #32: Let a failed query render an error, not an empty state

### LINKED ISSUE #61: RAG search: fix the page, then improve the search
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680e3825cc2bc57101949) · Priority: **Low** · Category: Development

RAG search already exists on both sides. The Notion card said only "RAG search", so this issue records what is actually there and what is worth doing to it.

**Backend:** `POST /api/v1/rag/search` (`app/api/routes/v1/rag.py:237`, `search_documents`), gated on `Perm.COLLECTIONS_VIEW`, resolves the collection through `CollectionAccessService`, supports multi-collection search and an optional `use_reranker` query flag.

**Frontend:** the `search` tab of `frontend/src/app/[locale]/(dashboard)/rag/page.tsx` — a 1009-line file with three tabs (`documents` / `search` / `sync`), tab state mirrored into the URL.

## Read these two first — they land on the same file

This page carries two confirmed audit findings, and doing this task without fixing them means touching the same 1009 lines twice:

- **#31** — every write control (upload, delete collection, delete document, add sync source) is rendered unconditionally. A Viewer holds `collections:view` without `collections:edit`, sees all of them, and gets a 403 toast on click. The sibling page `kb/[id]/page.tsx:86-88` does it correctly.
- **#32** — a 502 on `listTrackedDocuments` sets `docs = []` and the page renders "No documents — Upload PDF, DOCX, TXT, or MD". The user whose documents failed to load is told they have none.

**Do #31 and #32 as part of this task.** They are the same file, the same afternoon, and they are why the page is untrustworthy today.

## What is worth improving in search itself

1. **The reranker is invisible.** `use_reranker` is a query parameter the UI never sets, so the better retrieval path exists and nobody can reach it. Either surface it as a toggle or decide it is always on — but an unreachable flag is dead weight.
2. **Results carry no provenance a human can act on.** A chunk without its source document, page and score is hard to judge. Check what `RAGSearchResult` actually returns and render all of it — this page's job is letting someone answer "why did the agent say that".
3. **Multi-collection search is supported by the API and probably not by the UI.** `searchDocuments` is called with a single `selected` collection (`page.tsx:390`). If the backend takes a list, the UI should too — "search everything I can read" is the common case.
4. **No empty-query guard beyond `.trim()`**, and no result-count or timing feedback. A search that returns nothing looks identical to a search that failed — the same ambiguity as #32, one layer down.

## Constraints

- `.claude/rules/frontend.md`: controls the caller may not use are not rendered; an empty state must not stand in for an error; prove both with `*.integration.test.tsx`, not Playwright.
- 1009 lines in one file is the real problem behind all of this. Extracting the three tabs into components is not scope creep here — it is what makes the rest reviewable.
- Every string through `next-intl`.

## Done when

- [x] #31 fixed: no write control rendered without `collections:edit`, with an integration test
- [x] #32 fixed: each section renders an error state on a 502, asserted on the error and not on chrome
- [x] The reranker is either reachable from the UI or deliberately removed from the API surface
- [x] A result shows its source document and score
- [x] "No results" is distinguishable from "the search failed"
- [x] The three tabs are separate components; `page.tsx` is no longer four figures long
      — the hub is 280 lines with `SearchTab` beside it, and `rag/[id]/page.tsx` is
      382 with nine new component files under `components/rag/`
- [x] `bun run type-check && bun run lint && bun run test:coverage` clean
      — `test:coverage`, not `test:run`: the run named here measures no coverage,
      and the coverage one is what CI gates on

### LINKED ISSUE #476: ruff format fails on main: a missing blank line in test_check_i18n_nodes.py
## What breaks

`make lint-backend` fails on `main` — `ruff format --check` wants a reformat of
`backend/tests/test_check_i18n_nodes.py`, which needs a second blank line before
`test_masking_a_generic_is_what_keeps_a_whole_file_quiet` at line 155.

```
uv run --directory backend ruff format . ../scripts --check
unformatted: File would be reformatted
   --> tests/test_check_i18n_nodes.py:153:1
    |
152 |     assert _offences(tmp_path, source) == []
153 +
154 +
155 | def test_masking_a_generic_is_what_keeps_a_whole_file_quiet(tmp_path: Path) -> None:
    |

1 file would be reformatted, 543 files already formatted
make: *** [lint-backend] Error 1
```

Every branch cut from `main` inherits it, and every branch's `lint` job is red for a
reason that is not theirs.

## Sequence that triggers it

```bash
git worktree add --detach /tmp/mainchk origin/main
cd /tmp/mainchk/backend && uv run ruff format . ../scripts --check
```

Reproduced exactly that way on a throwaway worktree of pristine `origin/main` at
`eed9eed` — `1 file would be reformatted, 543 files already formatted`. Nothing of a
feature branch is involved.

## Where it came from

`backend/tests/test_check_i18n_nodes.py:155`, added by
**#403** (`test(frontend): pin mask_generics with a case that fails without it`,
merged as `8e176f5`).

This is the **second** time today `main` has shipped a red `lint-backend`. #451
(`36c6119`) had just fixed the previous pair — `RET501` in
`test_vector_store_reserved_names.py` and `RUF100` in `test_migrations.py`, both from
merges made during the GitHub Actions outage. #403 landed after that fix, so the tree
was green in between and is red again.

Worth noting because the cause is the same shape both times: **a `test`-only change
still has to pass `lint`**, and `ruff format` is the step that catches the
one-blank-line kind of slip nothing else does.

- `cd backend && uv run ruff format . ../scripts --check` reports `544 files already
  formatted` and exits 0 on `main`.
- `make lint` is green on a branch that changes nothing in `backend/`.

The fix is one blank line; `ruff format` writes it. Worth doing by hand rather than
letting `--fix` run over the tree, for the same reason #426 gave.

## Related

Third in the run of "`main` merged red" this week, after #407 (fixed in #426) and
#451. Adjacent to **#203**, which is the standing complaint that the pre-commit basics
read only changed files and no gate reads the whole tree — this is exactly the class of
break that argument predicts. Not a duplicate of #203: that asks for a tree-wide gate,
this is one file that needs fixing now. Checked the map in #168; the CI-hygiene cluster
is where this belongs.

Found while merging `main` into `feat/rag-one-knowledge-page` (PR #221).

OchnikBartek:
## Summary

The knowledge pages were two: `/kb` listed the bases, and `/rag` was a 1009-line file whose search tab could reach one collection at a time. Everything a base needs is configured in one place now — `/rag` lists them, `/rag/[id]` is the base itself, and `/kb` only redirects. Search defaults to every base the caller can read, and each result carries its document, page, score and the base it came from. Neither page is four figures long any more: the hub is 280 lines and the detail page 382, with nine new component files beside them under `components/rag/`.

## Related Issue

Closes #61
Closes #31
Closes #32
Closes #476

## Added

- `/rag` hub with **Knowledge bases** and **Search** tabs, the active tab mirrored into `?tab=`.
- `SearchTab` (`frontend/src/components/rag/search-tab.tsx`) — searches every readable base by default, deduplicating the physical collections several bases may share, with a scope selector, per-result provenance, a result count and the elapsed time.
- `KBSectionFailures` in `useKBDetail`, so the sync sources, the org integrations and the connectors each report their own failed request.
- A visible refresh-failure alert on `/rag/[id]`: the sections below keep the last good answer and say it may be stale.
- `rag-page.integration.test.tsx` (10 tests) and `kb-detail-sections.integration.test.tsx` (6 tests), plus three backend tests in `test_capability_edges.py` for the retrieval fixes below.
- Message keys for all of the above under `pages.kb`, plus `rag.search` and two wizard strings under `rag`, in both `en.json` and `pl.json`.

## Changed

- `kb/[id]/page.tsx` moved to `rag/[id]/page.tsx` with `git mv`, so its history survives; `/kb` and `/kb/[id]` are redirect stubs.
- `rag/page.tsx` went from 1057 lines to 280. Its `isAppAdmin` redirect is gone — the sidebar had been offering the page to anyone holding `collections:view`.
- **`rag/[id]/page.tsx` went from 1000 lines to 382, in four commits**, one section at a time: `KBDetailSkeleton`, `FileDropZone`, `UploadProgressList`, `KBDetailHeader`, `KBStatsStrip`, `DocumentsTable`, `document-cells.tsx`, `SyncSourcesSection` and `SyncSourceRow`, all under `components/rag/`. Each takes the state only it reads — `downloadingId` with the table, the collapse toggle with the sync sources, the drag counter with the drop zone. What coordinates two sections stays on the page and goes down as props: the three confirmations, the file viewer, and the parse-override every upload path has to carry.
- Write controls on the hub are gated on `collections:edit`; a control the caller may not use is not rendered.
- A failed list renders an error state with a retry instead of "No knowledge bases yet" — on **both** tabs, see Checks below.
- `PgVectorStore.search` reports a collection whose table does not exist yet as empty instead of letting an `UndefinedTableError` become a 500, checked before embedding so an empty collection costs no metered embedding call.
- `RetrievalService.retrieve_multi` no longer swallows a collection's failure: it used to log and skip, answering 200 with a partial result. `retrieve` now tags `metadata.collection`, so provenance survives a single-collection search and reaches the knowledge capability's citations.
- `SearchTab.runSearch` refuses to start while a search is in flight, so holding Enter cannot race the request the button already disabled itself for.
- `use_reranker` removed from `POST /api/v1/rag/search`, along with `RerankService` and `app/services/rag/reranker.py`. `retrieval.py` keeps the ×2 overfetch that min-score filtering and dedup need.
- `ROUTES` gained `RAG_DETAIL` and `RAG_SEARCH` and lost `KB` / `KB_DETAIL`; the sidebar, command palette, breadcrumbs, mobile tab bar and collection picker follow.
- `rag-api.ts` went from 341 to 140 lines — about twenty unused wrappers deleted, along with `qk.rag.*` and `drag-drop-overlay.tsx`.
- `docs/file-processing.md`: the "RAG is Global" section claimed any authenticated user could search any collection and that only admins manage them. Replaced with what the permission catalog actually does.

## Checks

- [x] `make check` passes

Verified on the merged tree the review sees: `make lint-backend` and `make lint-frontend` green (the backend half runs `check_i18n.py`); `tsc` clean; frontend **214 files, 3450 tests** with the coverage gate met (100% statements/functions/lines, 97.58% branches); backend **3662 tests, coverage 100%** with `POSTGRES_PASSWORD` supplied. The three gates that were red on `main` are all resolved now — see the Notes.

## Notes for Reviewers

- **The decomposition #61 asked for landed here, and it is four commits rather than one** — `refactor(rag): lift …`, each with the section's state moving with it and the integration tests re-run in between. The risk in a split like this is losing #32 on the way down: `sectionFailures` is what makes each section say "could not load" instead of rendering an empty state as a fact, and a child handed `syncSources` already flattened to `[]` puts that defect straight back. So the flags go down as `syncSourcesFailed` and `connectorsFailed`, each required, and it was worth proving the suite notices — passing `false` for both fails exactly the three specs that assert a 502 renders an error, including the one where sources already exist. Restored, all 37 detail-page tests pass.
- **Five gates were red on `main` while this branch tracked it; all five are resolved now.** Each was reproduced on a throwaway worktree of pristine `origin/main` and filed before being touched, because an unrelated fix does not belong in a branch about something else. They share one shape worth naming — a gate the change's own test run does not exercise, so a test-only PR sails through green:
  - **#476** — `ruff format` wanted two blank lines in `backend/tests/test_check_i18n_nodes.py` (from #403). **Fixed here** in its own `style:` commit, because it was the last thing between this branch and a green `lint`; `Closes #476` above.
  - **#485** — two integration tests from #428 drove the real application engine, and the app defaulted `POSTGRES_PASSWORD` to `""` while `tests/integration/conftest.py` defaulted it to `postgres`, so every worktree without a `backend/.env` failed exactly those two. Fixed on `main` by **#491** and merged in.
  - **#487** — `tsc` found a duplicate `usePermissions` key in `chat-model-picker.test.tsx` (from #360). Fixed on `main` and merged in.
  - **#407** (a `RET501`/`RUF100` pair) and the **`h2` CVE** in `Security Scan` were reported earlier and are fixed on `main` by #451 and #462.

  Removing the raw-collections page also orphaned 69 message keys — `main`'s #447 taught `check_i18n.py` to fail on a key nothing reads, and this branch is the first to trip it. Those translations died with the page; pruned from both catalogs. `main`'s new dashboard reads three RAG things earlier cleanups here had removed (`listSyncSources`, `qk.rag.syncSources`, and a `see all` that pointed at the deleted `ROUTES.KB`); all three restored.
- **`Closes #32` stands, and here is where.** The review found the same defect inside the tab this branch adds: the search tab was handed the base list as a plain array, so a 502 on `GET /kb` reached it as zero bases and it said there was nothing to search. Fixed in `rag/page.tsx` — the tab now renders the list's loading, failed and empty states separately — and pinned by `the search tab says the list failed rather than that there is nothing to search`. I reverted the fix to confirm that test fails without it, and did the same for the Enter-race guard.
- **`main` has been merged in repeatedly, and one trap accounts for most of the work.** `rag/[id]/page.tsx` and `rag/page.tsx` exist only on this branch — created by moving and rewriting the `kb/` pages — so **git never conflicts them** while `main` keeps changing the files they mirror. Every one of `main`'s changes had to be carried across by hand, and none of them announced itself as a conflict:
  - the i18n pass (49 translated strings that taking "our" side would have deleted), #246's ICU-plural counts, the three-step delete confirmation, the loaded-vs-total counts, `SCOPE_META` on keys, and deletion moving off the card;
  - and most recently **#445**, which is the one worth reading. It replaced the two status badges with one, because two of them is what let them disagree: the document badge mapped `completed`/`pending`/`failed` and the sync badge tested `failed`, and the worker writes none of those four (#356). This branch's decomposition had faithfully carried both forward into separate components — faithful to the page, and faithful to the defect. They are now one `RagStatusBadge` asking `lib/rag-status.ts`, and `main`'s seven new tests for that behaviour moved to `rag/[id]/` and pass unchanged, which is the evidence the port is right rather than merely green.

  Both pages were rebuilt from `main`'s versions rather than merged into, which is also why `check_i18n.py` — tightened three times on `main` since, by #315, #363 and #447 — finds nothing here.
- **Two of `main`'s brand-new specs were deleted, deliberately.** `rag/page.integration.test.tsx` and `rag/create-collection.integration.test.tsx` mount the 1158-line raw-collections page at `/rag` that this branch replaces with the hub — replacing it *is* #61. The behaviour behind them is not lost: `MAX_COLLECTION_NAME_LENGTH` was a `maxLength` on the raw collection-name input, and this branch has no such input because `POST /kb` derives the name. #431's actual fix — claim and bound, server side — is untouched. Every consumer of the `rag-api.ts` wrappers restored by that merge was checked and is one of those two specs or the page itself.
- **The two hardcoded English strings @DEENUU1 found were mine**, and `check_i18n.py` missed them because its JSX pattern needs the tags on the text's own line — already filed as #141. Fixed here; the guard is not this branch's to fix.
- **Removing `use_reranker` cannot break a caller.** FastAPI ignores query parameters a route does not declare, so a client still sending it gets the same 200 and the same results — which it already did, since `RerankService` was never constructed and had no concrete implementation.
- **The e2e specs were rewritten for the `/rag` URLs but not run locally.** Playwright's `reuseExistingServer` points at `:3000`, and locally that was a container serving `main`, so a local run would have exercised the wrong build. CI runs them against its own server.
- **Two pre-existing defects were filed rather than fixed**, both in the detail page's stats row: `chunk_count` is never recorded (#147) and the indexed count filters on a status value the pipeline never writes (#148). Together they make a base report "0 documents indexed · 0 vectors" while its search answers correctly.
- **`sync-source-wizard.tsx` is 761 lines and was left alone.** It has the same problem this branch just fixed next door, and it is not this issue's — filed as #461 rather than folded in, and added to the map in #168.
- Verified by hand against a live stack as well as by the suite: a query spanning two bases ranks the right chunk from each, narrowing the scope to one base returns nothing from the other, and an agent bound to both answers from them with citations.


frontend/src/components/rag/sync-source-wizard.tsx
```diff
@@ -215,6 +221,15 @@ export function SyncSourceWizard({
             </div>
           )}
 
+          {orgIntegrationsFailed && step === "source" && (
+            // Unsaid, a failed org-integrations request is indistinguishable from
+            // an organization that has none: the "Use existing" toggle never appears.
+            <p className="text-destructive mt-2 text-xs">
+              Existing workspace integrations could not be loaded; you can still create a new
```
DEENUU1: This message ships in English even in the Polish UI — it's a hardcoded text node, not a `t()` key. And `check_i18n.py` misses it only because the text sits on its own line: `JSX_TEXT` needs the `>`…`<` on the same line as the text, so a text node on a line of its own slips straight through and `make check` passes. That's why the gap is silent. Same issue with the connector message below at line 482.

The wizard already does `useTranslations("rag")`, so move both into the catalog:

```suggestion
            <p className="text-destructive mt-2 text-xs">{t("orgIntegrationsLoadFailed")}</p>
```

and add `orgIntegrationsLoadFailed` (and `connectorLoadFailed` for line 482) to both `en.json` and `pl.json` under `rag`.

frontend/src/components/rag/sync-source-wizard.tsx
```diff
@@ -453,7 +475,13 @@ function ConnectorStep({
         <Label className="text-foreground/80 text-xs font-medium tracking-wider uppercase">
           {t("connector")}
         </Label>
-        {connectors.length === 0 ? (
+        {connectorsFailed ? (
+          // "No connectors enabled" is a statement about the deployment; a failed
+          // request has not made it.
+          <p className="border-destructive/30 text-destructive rounded-xl border px-4 py-3 text-sm">
+            The connector list could not be loaded. Close the dialog and try again.
```
DEENUU1: The other half of the same thing — hardcoded English that the i18n guard doesn't see because it's on its own line. Same fix:

```suggestion
            <p className="border-destructive/30 text-destructive rounded-xl border px-4 py-3 text-sm">{t("connectorLoadFailed")}</p>
```

with `rag.connectorLoadFailed` in both catalogs.

frontend/src/components/rag/search-tab.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  return (
+    <div className="space-y-4">
+      <div className="border-border bg-card rounded-xl border p-4">
+        <div className="flex flex-col gap-2 sm:flex-row">
+          <Input
+            placeholder={t("placeholder")}
+            value={query}
+            onChange={(e) => setQuery(e.target.value)}
+            onKeyDown={(e) => e.key === "Enter" && runSearch()}
```
DEENUU1, line 125: The Search button disables while `status === "searching"`, but Enter doesn't — `runSearch` only guards `!trimmed || names.length === 0`. So holding Enter fires concurrent searches, and whichever request resolves last wins; change the query between presses and you can end up painting results for the earlier query. Cheap to make Enter match the button:

```suggestion
            onKeyDown={(e) => e.key === "Enter" && status !== "searching" && runSearch()}
```

DEENUU1 (commented):
Nice consolidation — one `/rag` page that lists the bases and searches all of them is clearly the right shape, and the tenant-switch guarding throughout `useKnowledgeBases`/`useKBDetail`/`SearchTab` is genuinely careful work (clearing state during render *and* re-checking the active org after every await). The per-section failure states on the detail page are exactly the "an empty page is ambiguous" rule done properly.

I went through the automated reviewer's five findings and they hold up — in particular #2 is real and worth prioritising: `PgVectorStore.search` is the one store method that doesn't call `_collection_exists`, while `get_collection_info`, `get_documents` and `get_document_chunks` all guard. Teaching `search` to return `[]` on an absent table resolves both #2 and the "absent collection" half of #3 at the source, and leaves `retrieve_multi`'s `except Exception` for genuine failures only (which is the point of #3 — right now a real vector-store error in one base is logged and dropped, and the surviving bases are returned as if the search were complete).

Two things the reviewer didn't catch, both about the wizard's new error copy — see inline.

frontend/src/app/[locale]/(dashboard)/rag/page.tsx
```diff
@@ -1,1057 +1,297 @@
...
       </div>
 
-      {!selected ? (
-        <div className="border-border bg-card text-muted-foreground flex flex-col items-center justify-center rounded-xl border py-16 text-center">
-          <Database className="mb-3 h-8 w-8" />
-          <p className="text-sm">{t("selectCreateCollectionGet")}</p>
-        </div>
+      {tab === "search" ? (
+        <SearchTab kbs={sorted} />
       ) : (
```
Comment (OchnikBartek) [inline frontend/src/app/[locale]/(dashboard)/rag/page.tsx:153]: Fixed in 3e2df25, taking this shape. The tab now renders the list's three states separately and the failed one carries a retry.

This is the finding that bore on the PR's `Closes #32`, so I checked the test is not passing vacuously: reverted the fix, ran the file, got `1 failed | 8 passed`, restored it. The claim in the description now points here.

frontend/src/components/rag/search-tab.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const trimmed = query.trim();
+    const names = collectionNames(scope);
+    if (!trimmed || names.length === 0) return;
+    const startedIn = orgId;
+    setStatus("searching");
+    const startedAt = performance.now();
+    try {
+      const data = await searchDocuments({
+        query: trimmed,
+        collection_names: names,
```
OchnikBartek, line 88: Confirmed and fixed in 3e2df25, in `PgVectorStore.search`. `get_collection_info` right below it already documents this exact failure - it used to let the `UndefinedTableError` become a 500 - and `search` had never learned it.

The guard sits before `_for_collection`, so an absent collection also costs no embedding call, which would have been metered. Regression test in `test_capability_edges.py` asserts both: empty result, and that neither the embedder nor a session is reached.

frontend/src/components/rag/search-tab.tsx
```diff
@@ -0,0 +1,203 @@
...
+        limit: RESULT_LIMIT,
+      });
```
OchnikBartek, line 90: Confirmed and fixed in 3e2df25. The `except Exception` is gone, so a collection that fails takes the search with it - which is what the access layer already does with a name it cannot reach, and what `docs/file-processing.md` says about one unreachable collection refusing the whole search.

An absent lazy table is handled separately, as you suggested: the store now reports it as no results, so an empty collection merges as none rather than as an error. Two tests cover the pair - one failing collection alongside a healthy one raises; one empty collection alongside a healthy one returns just the healthy results.

frontend/src/components/rag/search-tab.tsx
```diff
@@ -0,0 +1,203 @@
...
+          <Search className="text-muted-foreground mb-3 h-8 w-8" />
+          <p className="text-foreground text-sm font-medium">{t("noResults")}</p>
+          <p className="text-muted-foreground mt-1 text-xs">{t("noResultsHint")}</p>
+        </div>
+      )}
+
+      {results.length > 0 && (
+        <div className="space-y-2">
+          {results.map((result, i) => {
+            const source = baseFor(result.metadata?.collection);
```
OchnikBartek, line 181: Fixed in 3e2df25, but at the other end from the suggestion - worth saying why.

Tagging `metadata.collection` in `retrieve` rather than falling back in the component removes the asymmetry that caused this, instead of compensating for it: one code path, and it covers every caller. The knowledge capability formats citations from the same field, so a single-collection agent search was losing attribution too.

With that, the component needs no change and no guess about what the backend searched. Covered by `test_every_result_names_the_collection_it_came_from`.

frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx
```diff
@@ -0,0 +1,807 @@
...
+          )}
+        </div>
+
+        {sectionFailures.syncSources ? (
+          <ErrorState
+            title={t("syncSourcesFailedTitle")}
+            description={t("syncSourcesFailedDescription")}
+            cta={{ label: t("retry"), onClick: () => refresh() }}
+          />
+        ) : syncSources.length === 0 && sectionFailures.connectors ? (
```
Comment (OchnikBartek) [inline frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx]: Confirmed and fixed in 3e2df25. The notice is now its own element above the list rather than a branch of the states below it, because the connector list failing is orthogonal to whether any sources loaded - and it is what hides the Connect button, so the two disappeared together and read as the product not offering connectors at all.

Sources that did load stay on screen, the retry stays, and when the list is genuinely empty the empty state still appears. Test renders a non-empty sync-source response with a failed connectors response and asserts both the notice and the surviving source.

frontend/src/components/rag/search-tab.tsx
OchnikBartek: Fixed in 3e2df25, though guarded inside `runSearch` rather than on the key.

Matching Enter to the button leaves the same hole for the next caller - the error state's retry already calls `runSearch` too - so the in-flight check lives with the function. A ref rather than `status`, because two keydowns can land before a re-render commits, and released in a `finally` so the tenant-switch early returns cannot leave the flag raised and the search dead for the session.

With one request ever in flight, the stale-overwrite half of what you describe cannot happen either. Test holds Enter twice against a promise that has not resolved and asserts one call; reverting the guard makes it fail.

frontend/src/components/rag/sync-source-wizard.tsx
OchnikBartek: Fixed in 3e2df25 - `rag.orgIntegrationsLoadFailed`, in both `en.json` and `pl.json`. Mine to begin with, so thank you for catching it.

Your diagnosis of why the guard missed it is exact: `JSX_TEXT = re.compile(r">\s*([^<>{}\n][^<>{}\n]*?)\s*<")` excludes the newline from both classes, so Prettier wrapping the sentence onto its own line takes it out of scope. A probe allowing the text on its own line finds about twenty more on `main`, including `global-error.tsx` and `not-found.tsx`.

I did not file that: it is already #141, corrected on 2026-08-04 to name this case as the larger of its two blind spots. `Showing {documents.length} of {documentsTotal}` in the detail page is hidden the same way, and I left it as `main` writes it rather than fix one instance of a repo-wide gap in this branch.

frontend/src/components/rag/sync-source-wizard.tsx
OchnikBartek: Fixed in 3e2df25 - `rag.connectorLoadFailed`, both catalogues. Same cause and same note as the one above.

Issue #220, opened 2026-08-05
CodeQL blocks merges on a false positive: bare `await <task>` reads as a statement with no effect


DEENUU1:
## What breaks

CodeQL's `py/ineffectual-statement` ("Statement has no effect") fires on a bare
`await <task>` expression statement. The query does not treat `Await` as
side-effecting, so **every** statement of that shape is flagged. Awaiting a task
suspends until it completes and re-raises whatever it raised — the effect is the
entire point of the line.

That would be tolerable as noise. It is not tolerable because
`github-code-quality` posts each alert as a **pull request review thread**, and
the `main` ruleset requires every thread resolved before a merge. So each false
positive is a merge blocker that has to be read and answered by hand.

## The sequence that triggers it

1. Write an async test that awaits a task it created — draining a cancelled task
   so the loop does not report an unretrieved exception, or running a coroutine to
   completion so a done callback can be given a genuinely-finished task.
2. Open a pull request.
3. `github-code-quality` opens one unresolved review thread per occurrence.
4. `mergeStateStatus` is `BLOCKED` until each is replied to and resolved.

Concretely on #196, which added the first bare `await <task>` statements in the
repository — `git grep -nE '^\s+await [a-z_][a-zA-Z0-9_]*$' -- backend` on
`main` before it returns nothing:

| `backend/tests/test_agent_session.py` | Statement |
|---|---|
| `:295` | `await turn_task` |
| `:311` | `await first` (two threads on the same line) |
| `:319` | `await second` |
| `:593` | `await task` |
| `:644` | `await task` |
| `:665` | `await later` |

Seven threads, one alert, none of them a defect. Each needed a hand-written
reply naming the line and the reason before #196 could merge.

There is no CodeQL configuration in the repository at all — no
`.github/codeql/codeql-config.yml`, and the `Analyze (python)` job runs the
default query suite unfiltered. So nothing records that this alert has been
adjudicated, and the next pull request that awaits a task pays the same cost from
scratch.

## The fix

> **Corrected 2026-08-05 — the fix below does not work. See
> [this comment](https://github.com/vstorm-co/agenticos/issues/220#issuecomment-5191791828)
> for the checks.** Both CodeQL analyses run on **default setup**: GitHub
> generates the workflow, passes its configuration to `codeql-action/init`
> inline with the comment `queries: "" # No query customization supported`, and
> never passes `config-file`. So `.github/codeql/codeql-config.yml` has no
[...]

PR #224, opened 2026-08-05
probe: which code-quality alerts a comment can suppress (do not merge)

DEENUU1:
Throwaway probe for #220, to be closed without merging.

Default setup for Code Quality passes `queries: "" # No query customization supported` to `codeql-action/init`, so a `.github/codeql/codeql-config.yml` in the repository is not read at all. Before writing a fix, this branch asks the only mechanism that could work under default setup whether it works: inline suppression comments.

`backend/tests/test_codeql_probe.py` writes four shapes:

1. a bare `await task` with no suppression - the control;
2. the same with `# codeql[py/ineffectual-statement]` on the line before;
3. the same with a trailing `# lgtm[py/ineffectual-statement]`;
4. a `Protocol` body of `...` and a loop whose fall-through is `pytest.fail`.

Whichever of these draws a thread from `github-code-quality` is the answer.

Refs #220

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,5 @@
+name: "AgenticOS CodeQL configuration (probe)"
+
+query-filters:
+  - exclude:
+      id: py/ineffectual-statement
--- a/backend/tests/test_codeql_probe.py
+++ b/backend/tests/test_codeql_probe.py
@@ -1,0 +1,50 @@
+"""Throwaway probe for #220. Not for merging.
+
+Each function writes one of the shapes `github-code-quality` flags, so the pull
+request answers three questions at once: which of them actually reach the review
+thread path, and whether either inline suppression syntax (`# codeql[...]` on the
+line before, `# lgtm[...]` on the line itself) stops one arriving.
+"""
+
+import asyncio
+from typing import Protocol
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


DEENUU1:
Answered, so this closes without merging. Results, all four from `github-code-quality[bot]` on this branch:

| Probe | Line | Outcome |
|---|---|---|
| bare `await task`, no suppression | 23 | flagged — the control works |
| `# codeql[py/ineffectual-statement]` on the preceding line | 29 | **flagged anyway** |
| trailing `# lgtm[py/ineffectual-statement]` | 34 | **flagged anyway** |
| `...` as a `Protocol` method body | 43 | flagged — the second shape confirmed |
| loop whose fall-through is `pytest.fail` | 46 | flagged as `py/mixed-returns` — the third shape confirmed |
| bare `await task` **with** `.github/codeql/codeql-config.yml` excluding the query | 39 | **flagged anyway** |

And the mechanism, from the `Analyze (python)` log of the second push, which is the configuration CodeQL actually received with our config file sitting in the checkout:

```yaml
disable-default-queries: true
queries:
  - uses: code-quality
query-filters:
  - exclude:
      tags: exclude-from-incremental
```

GitHub's own incremental filter, and nothing of ours. Written up in #220 and in `docs/code-review.md`.

Refs #220

Issue #299, opened 2026-08-06
make check fails in a fresh checkout: nothing creates backend/.env


OchnikBartek:
## What breaks

`make check` in a fresh checkout fails at `db-check`, and the error names neither
the cause nor the fix:

```
cd backend && uv run alembic check
...
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server
at "localhost" (::1), port 5432 failed: fe_sendauth: no password supplied
make: *** [db-check] Error 1
```

`backend/.env` is gitignored, nothing in `make install` creates it, and
`POSTGRES_PASSWORD` defaults to `""` in `app/core/config.py:50`. So on a checkout
where nobody has copied `.env.example` yet, `alembic check` dials the laptop's
Postgres with an empty password and is refused — while `docs/install.md` says
"backend/.env is pre-configured for development", which reads as "there is
nothing to do here".

It costs what #227 cost: `db-check` sits third in `check`'s prerequisite list,
after `lint` and `test`, so the failure lands four minutes in and takes
`test-frontend-cov`, `build-frontend`, `docs-build` and `audit` with it — none of
which need a database at all.

## The sequence that triggers it

1. `git clone`, or `git worktree add` (each worktree has its own `backend/`, and
   `.env` is not shared).
2. `make install`.
3. `make check`.
4. `db-check` fails with `fe_sendauth: no password supplied`, regardless of
   whether Postgres is running.

## How you would know it was fixed

`git clone && make install && make check` reaches `test-frontend-cov` on a
machine with no `backend/.env`. Candidate fixes:

- `make install` copies `backend/.env.example` to `backend/.env` when there is
  none — which is what `make sandbox-token` already half-does by appending to it,
  and would make `docs/install.md`'s claim true; or
- `db-check` says "no `backend/.env` — `cp backend/.env.example backend/.env`"
  instead of surfacing a psycopg2 traceback.

The first is preferable for the same reason it was in #227: a setup command that
reports success while half the environment is absent is the misleading part.

## Where it sits

**Tests, CI, dependencies** cluster of #168. Same target and same blast radius as
**#288** (`db-check` fails on any database that has ingested a document, and
blocks the four targets after it), but a different cause — #288 needs a database
that has been used, this needs no database at all. Local-environment kin to
**#227** (`make install` never ran `bun install`) and **#223** (hardcoded E2E
ports), both of which are "the repository is harder to run truthfully on a laptop
[...]

Issue #288, opened 2026-08-05
make db-check fails on any database that has ingested a document


OchnikBartek:
`make db-check` runs `alembic check` against whatever `backend/.env` points at.
The vector store creates a table per collection **at runtime**, and nothing
models or migrates those tables, so on any developer database that has ever
ingested a document `alembic check` reports them as tables to drop and exits
non-zero. Since `db-check` sits in `make check` ahead of the frontend targets,
`make check` also never reaches `test-frontend-cov`, `build-frontend`,
`docs-build` or `audit`.

## What breaks

```
$ make db-check
INFO  [alembic.autogenerate.compare.tables] Detected removed table 'rag_company_handbook_12db7e'
INFO  [alembic.autogenerate.compare.tables] Detected removed table 'rag_engineering_runbook_c5105b'
ERROR [alembic.util.messaging] New upgrade operations detected: [('remove_index', ...), ('remove_table', ...)]
FAILED: New upgrade operations detected
make: *** [db-check] Error 255
```

## Why

`app/services/rag/vectorstore.py:271` issues `CREATE TABLE IF NOT EXISTS
{table}` for each collection, named `rag_<slug>_<hash>`. Nothing in
`app/db/models/**` or `alembic/versions/**` mentions them - which is correct,
they are per-tenant runtime objects rather than schema - but `alembic check`
compares the whole database against the metadata and so reads every one of them
as a table the models dropped.

The result is inverted: the check is meant to catch "edited a model, forgot the
migration", and instead it fails for a reason that has nothing to do with
models, on exactly the machines where somebody is editing them.

## Why CI is green

`.github/workflows/ci.yml:125-143` runs `make test-migrations` first, which
cycles a throwaway `test_db` to head, then `db-check` against that. It has never
ingested anything, so it has no `rag_*` tables. The workflow comment already
notes the local/CI divergence for `test-migrations` ("on a laptop it points at
whatever `backend/.env` says"); this is the same divergence biting the step
after it.

## Reproducing

Upload one document to any knowledge base on a local stack, then `make db-check`.

## How you would know it was fixed

`make check` runs to completion on a developer machine that has ingested
documents, and still fails when a model gains a column with no migration.

`alembic check` honours `include_object` / `include_name` in `env.py`; excluding
names matching the vector store's table prefix would scope the comparison to the
schema alembic actually owns. Worth pinning the prefix in one place shared with
`vectorstore.py` rather than repeating the pattern.

## Scope

[...]

Issue #143, opened 2026-08-04
make check does not run the frontend coverage gate, so a branch reads green while CI is red


DEENUU1:
`make check` is documented as "What CI runs. Run this before opening a pull
request." It runs `lint test test-frontend`, and `test-frontend` is
`bun run test:run` — **no coverage**. CI's `test-frontend` job runs
`bun run test:coverage` (`.github/workflows/ci.yml:202`), which carries the
100% statements/lines/functions threshold in `frontend/vitest.config.ts`.

So a branch can pass `make check` locally and fail CI on the one gate the
frontend actually has, and the failure is only visible after a push. That is
what happened on `feat/sandbox`: every CI run since `832d6477` (3 Aug, 23:29)
has failed with

```
ERROR: Coverage for lines (99.83%) does not meet global threshold (100%)
ERROR: Coverage for functions (99.67%) does not meet global threshold (100%)
ERROR: Coverage for statements (99.81%) does not meet global threshold (100%)
ERROR: Coverage for branches (97.09%) does not meet global threshold (97.5%)
```

while `make check` was reported green in the pull request and in #36's comments.
All 2909 tests pass; only the gate fails.

## To reproduce

```bash
make check          # green
make test-frontend-cov   # red, and this is what CI runs
```

## How you would know it was fixed

`check` depends on `test-frontend-cov` rather than `test-frontend`
(`Makefile:295`), and a branch that fails CI's frontend job fails locally
first. `test-frontend` stays as the fast loop, which is what it is for.

## Second, related finding, in the same job

The shortfall on `feat/sandbox` is reported against
`frontend/src/hooks/use-mcp-tool-servers.ts` (14.28% statements, 0% functions,
lines 30-44 — the whole hook body). The file *has* a test,
`use-mcp-tool-servers.test.tsx`, it passes, and running that file alone with
coverage drops the file from the report entirely rather than showing it at
100%. So the number is likely a v8 attribution problem between the worker that
executes the hook and the worker that loads it through the mocked `@/hooks`
barrel (`tool-call-card.test.tsx`), not an untested hook. Worth settling before
anyone widens the `include` list further — a gate that can be wrong about a
covered file will be argued with rather than obeyed.

Pre-existing, found while fixing two `/chat` defects on `feat/sandbox`.

PR #300, opened 2026-08-06
ci(deps): install the frontend toolchain in make install

### LINKED ISSUE #227: make check fails in a fresh checkout: nothing ever runs bun install
## What breaks

`make check` — the command `CLAUDE.md` names as the thing to run before opening a
pull request — fails in any checkout where `frontend/node_modules` does not exist,
and the error names neither the cause nor the fix:

```
cd frontend && bun run lint
$ eslint . --max-warnings 0
/bin/bash: eslint: command not found
error: script "lint" exited with code 127
make: *** [lint-frontend] Error 127
```

`make install` does not install them: it runs `uv sync` for the backend and
installs the pre-commit hooks, and nothing in the Makefile ever runs
`bun install`. So the documented setup path leaves `make check`, `make lint`,
`make lint-frontend`, `make test-frontend` and `make build-frontend` all broken
until somebody works out that a `bun install` is owed.

It costs more than it looks, because the failure lands **at the end**: `check`
runs `lint-backend` first, so on this machine it spent 4 minutes getting through
the backend half before failing on a missing `node_modules`. Then the whole run
starts again.

1. Clone the repository, or create a worktree under `.claude/worktrees/` (each
   one has its own `frontend/`, and `node_modules` is not shared).
2. `make install`, as `docs/install.md` says.
3. `make check`.
4. `lint-frontend` fails at step 1 of 3 with `eslint: command not found`.

`git clone && make install && make check` reproduces it from nothing.

## Why now

Worktrees make a fresh checkout routine rather than a once-a-laptop event — this
is the same species as #223 (`playwright.config.ts` hardcodes ports, so the E2E
suite cannot run beside another checkout), and both are "the repository is harder
to run truthfully on a laptop than CI makes it look". CI never sees this one: its
`test-frontend` job runs `bun install --frozen-lockfile` explicitly, which is
exactly the step `make install` is missing.

`git clone && make install && make check` is green on a machine with no
`frontend/node_modules`. Two candidate fixes, either acceptable:

- `make install` also runs `cd frontend && bun install --frozen-lockfile`, which
  makes the documented setup path complete; or
- the frontend targets depend on a `frontend/node_modules` guard that installs it
  (or fails with "run `bun install` in `frontend/`" rather than
  `command not found`).

Prefer the first: `make install` claiming "✅ Installation complete!" while half
the toolchain is absent is the part that misleads.

## Where it sits

**Tests, CI, dependencies** cluster of #168. Nearest kin is the #143/#190 pair —
"`make check` is CI, not an approximation of it" — of which this is the inverse:
the parity test compares the *steps*, and cannot see that one of them cannot run
at all in a fresh checkout. Local-environment pair with #223. Not a duplicate of
either.

Found while running `make check` for #220 in a fresh worktree. Pre-existing —
nothing in that branch touches the Makefile or the frontend.

OchnikBartek:
## What this changes

`make install` now installs `frontend/node_modules` as well as the backend
virtualenv, so the documented setup path leaves a checkout able to run the
commands the repository documents. Before this, `make install` finished with
"✅ Installation complete!" while eslint, prettier, tsc, vitest and next were all
absent, and `make lint`, `make lint-frontend`, `make test-frontend`,
`make build-frontend` and `make check` failed with `eslint: command not found` —
in `check`'s case four minutes in, because the backend half runs first.

`backend/tests/test_ci_parity.py` grows the mirror of the rule it already holds,
so this cannot come back: every *setup* command a gating CI job runs must be part
of `make install`, just as every *check* it runs must be reachable from
`make check`.

## Why this way

#227 offered two fixes and preferred this one, and so do I: a guard on the
frontend targets would turn `command not found` into a better error message, but
`make install` claiming success while half the toolchain is missing is the part
that misleads. `docs/install.md` names `make install` and nothing else, so it has
to be true.

The parity test is the interesting half. #143/#190 compared `make check` to CI
step for step and `test_ci_parity.py` keeps that true — but neither compares a
step to the tree it runs in, which is why a `check` that ran *exactly* what CI
runs still could not run at all in a fresh checkout. `ENVIRONMENT_SETUP` already
listed CI's setup commands (it was the allow-list for "not a check, so not in
`check`"); the new test reads the same list from the other end. Two entries are
exempt, each with the reason in the code: `uv python install 3.12`, which
`uv sync` does itself from `backend/.python-version`, and the docs group, which
`uv run --group docs` resolves at the point of use rather than into a synced
virtualenv.

## Checks

      reverting the Makefile line fails
      `test_every_setup_command_a_gating_job_runs_is_part_of_install` with
      `CI runs ['bun install --frozen-lockfile'] … and make install does not`
- [x] `make check` passes — with the exception below
- [ ] n/a — no migration

Verified from nothing, in a worktree with no `frontend/node_modules`:
`make install` (1014 packages plus the hooks) then `make check` — lint, test,
`test-frontend-cov` (100% lines/statements/functions, 97.55% branches),
`build-frontend`, `docs-build` and `audit` all green.

## Notes for the reviewer

**`make check` still stops at `db-check` in a fresh checkout, for an unrelated
reason, and that is filed rather than fixed here.** `backend/.env` is gitignored,
nothing creates it, and `POSTGRES_PASSWORD` defaults to `""`, so `alembic check`
is refused with `fe_sendauth: no password supplied` — #299, which is #288's twin
on the same target from the opposite direction. Out of scope for a branch about
`bun install`; the four targets after `db-check` were run individually to confirm
they pass.

Docs: `docs/install.md` and `docs/commands.md` both described `make install` as
uv + pre-commit, and now describe what it does.

Closes #227. Refs #299, #288.


Makefile
```diff
@@ -177,8 +177,16 @@ prod-logs:
...
+# Both halves of the toolchain, because `check` runs both. eslint, prettier, tsc,
+# vitest and next all live in `frontend/node_modules` and nowhere else, so a setup
+# that installs only the backend leaves `lint`, `lint-frontend`, `test-frontend`,
+# `build-frontend` and therefore `check` failing with `eslint: command not found`
+# - and `check` runs the backend half first, so it says so four minutes in.
+# `node_modules` is per-checkout and not shared between worktrees, which is why
+# this is owed on every clone rather than once a laptop.
 install:
 	uv sync --directory backend --dev
+	cd frontend && bun install --frozen-lockfile
```
DEENUU1: Ordering: `bun install` now sits **in front of** the pre-commit block, and make stops at the first line that fails.

So on a machine where the frontend install fails — no network, a corporate proxy, a partial bun install, a lockfile that needs a newer bun — `make install` dies here and **never installs the git hooks**. That developer then has no `commit-msg` hook and, more to the point, no hook refusing a commit on `main`. They also have no reason to suspect it, because the error they saw was about bun.

The hooks are the cheapest step in the target and the one with the worst failure mode if missing, so I would put the flakiest step last:

```make
install:
	uv sync --directory backend --dev
	@if git rev-parse --git-dir > /dev/null 2>&1; then \
		uv run --project backend pre-commit install --hook-type pre-commit --hook-type commit-msg; \
	else \
		...
	fi
	cd frontend && bun install --frozen-lockfile
```

Same set of commands, same guarantee for `check`, and a partial failure leaves you protected instead of unprotected. The new parity test does not care about order, so it stays green.

docs/commands.md
```diff
@@ -12,7 +12,7 @@ Run these from the project root directory.
 | Command | Description |
 |---------|-------------|
 | `make quickstart` | Install deps, start Docker, run migrations, create admin user |
-| `make install` | Install backend dependencies with uv + pre-commit hooks |
+| `make install` | The whole toolchain: backend dependencies with uv, `frontend/node_modules` with bun, and the pre-commit hooks. Both halves, because `make check` runs both — eslint, prettier, tsc, vitest and next live only in `node_modules`, which is per-checkout and not shared between worktrees |
```
DEENUU1, line 15: nit, but it is the row directly above the one you fixed and it says the same untrue thing #227 is about: `make quickstart` does **not** install deps. It is `quickstart: dev`, and `dev: sandbox-token` — nothing in that chain reaches `install`.

So somebody following this table for a fresh checkout runs `make quickstart`, gets a stack, and hits `eslint: command not found` on their first `make check` — which is exactly the journey this PR is closing off, one line up.

```suggestion
| `make quickstart` | Start Docker, run migrations, create admin user. Does **not** install dependencies — run `make install` first |
```

DEENUU1 (commented):
This is the right fix and the parity test is the best part of it — holding the *setup* commands to the mirror-image rule, so the next toolchain CI adds has to land in `install` or be exempted with a reason, is what stops #227 happening again with something other than bun.

Two things below. The first is a real ordering problem I would fix before merging; the second is a stale line in a file this PR is already editing.

Checked and it's fine, for the record: bun is already listed as a prerequisite at `docs/install.md:13`, so `make install` gaining a hard dependency on it is not a new demand on anybody. And the substring match in the new test is right — CI's bare `bun install --frozen-lockfile` does sit inside `cd frontend && bun install --frozen-lockfile`.

DEENUU1:
Pushed both points from my review rather than leaving them for a round trip.

The ordering one mattered more than it looks: with `bun install` first, any machine where it fails ends up with no git hooks at all — including the one that refuses a commit on `main` — and the error on screen is about bun, so nobody would think to check. Reordered, with the reason in the comment above the target so it does not get shuffled back.

`make check` reaches `test_ci_parity.py` and it is green (17 passed).

Issue #189, opened 2026-08-05
Two integration runs on one machine drop each other's tables


DEENUU1:
**Pre-existing, found running the integration suite while implementing #179.** Not caused
by that branch; it is a property of the fixture.

## What breaks

`tests/integration/conftest.py:96-101` — the `engine` fixture is **function-scoped** and
does `drop_all` + `create_all` on every test, against the database named by `POSTGRES_DB`,
which `tests/conftest.py` pins to `agenticos_test`. The name is constant and the drop is
unconditional, so two runs of the suite on one machine are dropping and recreating each
other's tables mid-test.

That is not hypothetical any more: this repository's own workflow now puts several agents in
`.claude/worktrees/*`, each with its own `.venv`, all pointing at the one `agenticos_db`
container. Two of them ran `pytest tests/integration` at the same time and both degraded.

## The sequence

1. `make docker-db` once (one container, one `agenticos_test` database).
2. From worktree A: `uv run pytest tests/integration -q --no-cov`.
3. From worktree B, while A is still running: the same command.
4. Both slow to a crawl, then report errors and failures that have nothing to do with either
   branch. `pg_stat_activity` during it:

   ```
    pid  | state  | wait_event_type |  q
   48885 | active | Client          | CREATE INDEX knowledge_bases_organization_id_idx ON knowledg…
   48877 | active | Lock            | INSERT INTO users (id, email, hashed_password, …
   48886 | active | Lock            | DROP TABLE tool_approvals
   48890 | active | Lock            | DROP TABLE tool_approvals
   ```

   One session mid-`CREATE INDEX`, two blocked on `DROP TABLE`, an `INSERT` blocked behind a
   relation lock. Observed result in one of the two runs: `E` on 20 of the first 144 tests,
   two `F`, in an order that changes from run to run.

## Why it matters beyond being slow

The suite exists to answer questions a mock cannot - constraints, cascades, tenant
isolation - so a failure in it is exactly the kind a person is meant to trust. A failure
that came from another checkout's `drop_all` reads identically to a broken cascade, and
`.claude/rules/testing.md` already warns that a test failing for the wrong reason is worse
than no test. Right now the honest advice is "do not run the integration suite while another
worktree is", which is not something a fixture should require a human to remember.

## How you would know it was fixed

- Two concurrent `pytest tests/integration` runs from different worktrees both pass, with no
  ordering dependence between them.
- The obvious shapes: derive the database name per run (`agenticos_test_<pid>` or
[...]

Issue #183, opened 2026-08-05
alembic check reports drift on main, so it cannot be used as a gate


DEENUU1:
**Pre-existing, found reviewing #164** while verifying that branch's two migrations.

`uv run alembic check` fails on `main`, and not because of anything recent: it reports index
naming drift from migrations `0002`–`0004` on `agent_workspaces`, `sandbox_connections` and
`skill_proposals`. The models and the migrations disagree about index names, so autogenerate
wants to "fix" them on every run.

Two consequences:

- **The command is useless as a gate.** It cannot be added to `make check` or CI while it fails
  for reasons unrelated to the change under test, so the one tool that would catch "somebody
  edited a model and forgot the migration" is unavailable. `tests/test_migrations.py` runs the
  chain forwards and back, which is valuable and a different question.
- **It hides real drift.** Verifying #164's two migrations meant reading the diff by hand to
  confirm the noise was only the pre-existing three — exactly the work a green `alembic check`
  would have done. The next person will either repeat that or skip it.

## How you would know it was fixed

`uv run alembic check` exits 0 on `main`, and is wired into `make check` so a model change
without a migration fails before review.

The fix is either a migration that renames the three index sets to what the models declare, or
`Base.metadata`'s naming convention adjusted to produce what the migrations already created —
whichever matches what is actually in production databases. That question has to be answered
before the edit, since renaming an index that exists under a different name is a no-op that
leaves the drift in place.

Related: #143 (`make check` not running the frontend coverage gate — the same class: a check
that exists and does not run). Cluster and spine on #168.

PR #302, opened 2026-08-06
test(migrations): create the database the chain is tested on

### LINKED ISSUE #234: tests/test_migrations.py skips in CI: nothing creates the database it needs
**Pre-existing, found fixing #183.** Every test in `backend/tests/test_migrations.py`
skips in CI, and has always skipped there. Nothing in the repository creates the database
the module needs.

## What breaks

`backend/tests/test_migrations.py:26` pins the module's own database:

```python
MIGRATION_DATABASE = "agenticos_migrations_test"
```

and `_db_available()` (`:52`) turns "that database does not exist" into a module-level
skip:

```python
pytestmark = pytest.mark.skipif(
    not _db_available(),
    reason="No live database available - skipping migration tests",
)
```

The docstring is explicit that the module will not create it — "creating a database is a
privileged act and a test module is the wrong place to perform one silently" — and names
`make test-migrations` as what does. But `make test-migrations` runs `alembic upgrade
head` against whatever `POSTGRES_DB` says, which in CI is `test_db`. It never touches
`agenticos_migrations_test`.

`rg agenticos_migrations_test` over the tree returns exactly one hit: the constant
itself. So the name exists on a laptop only because somebody once typed
`CREATE DATABASE agenticos_migrations_test` by hand.

## The sequence

1. CI's `test` job starts with `POSTGRES_DB: test_db` and a fresh pgvector service.
2. `make test-migrations` cycles the chain — against `test_db`.
3. `make test COV_REPORT=xml` collects `tests/test_migrations.py`.
4. `_alembic("current")` runs with `POSTGRES_DB=agenticos_migrations_test`, which does
   not exist, so it exits non-zero.
5. All five tests skip. The job is green and reports nothing about it, because a skip is
   not a failure.

## Why it has not bitten

`make test-migrations` covers the same ground in the step before, and after #183 `make
db-check` covers the drift assertion — so the *questions* are asked in CI even though
this module is not. What is lost is the module itself: `test_downgrade_base`,
`test_upgrade_downgrade_cycle`, `test_current_matches_head` and the new
`test_the_chain_builds_the_schema_the_models_declare` have never run on a runner, so a
regression in one of them shows up as a green build and a local skip nobody reads.

It is also the trap where somebody adds a test here believing CI runs it. That is what
happened in #183: the drift assertion belongs in this module because only this module has
a database built by nothing but the chain, and it had to be gated separately through
`make db-check` to actually run anywhere.

`tests/test_migrations.py` reports 5 passed rather than 5 skipped in CI's `test` job, and
a deliberately broken `downgrade()` turns that job red. Either the module is allowed to
create its own database when the server answers and the name is safe (which
`tests/integration/conftest.py` already does, and which would settle the "privileged act"
argument by precedent), or CI creates it in a step before `make test`.

Whichever way it goes, it should not silently skip. A skip that means "your environment
cannot answer this" and a skip that means "nobody has ever answered this" look identical
in pytest output, and the second is the one worth failing on in CI.

## Where it sits

**Tests, CI, dependencies** on #168 — the same shape as #143, #188, #203 and #190: a
check that exists and does not run. Closest relative is #188 (codespell read only the
files a commit touched, so the tree was gated by nothing); here the module is gated by a
database nothing creates. Not a duplicate of any of them — #203 is the remaining
pre-commit hooks, and none of the four is about `test_migrations.py`.

Refs #183, #168.

OchnikBartek:
## What this changes

`tests/test_migrations.py` runs. It has never run on a runner: it needed a database
called `agenticos_migrations_test`, a missing one became a module-level skip, and
nothing in the repository ever created it — so the only assertions in this project
that `downgrade()` works at all have always reported "5 skipped" into a green build.

The module now creates that database before its first test and drops it after its
last, the way `tests/integration/conftest.py` already does, with the process id in
the name. The remaining skip means one thing only — no Postgres answered — and under
`CI` it is not a skip at all but a `RuntimeError`, because a declared service
container that did not come up is not a laptop without Docker.

## Why this way

#234 offered two fixes: let the module create its own database, or have CI create it
in a step before `make test`. The first, for two reasons. It fixes the module
everywhere rather than on one runner — a laptop with `make docker-db` up now runs
these tests too, where before it needed somebody to have typed `CREATE DATABASE` by
hand at some point. And the "privileged act" argument the old docstring refused on
does not survive contact with the outcome: the alternative to creating it carefully
was not somebody else creating it carefully, it was nobody creating it, and
`tests/integration/conftest.py` had already settled the precedent.

The pid in the name is not decoration. This module drops the database it names,
twice, so a constant name is #189 again — two worktrees, or a `make test` beside
this one, dropping each other's mid-upgrade. The name is a literal rather than an
environment-derived value, so it needs no `_refuse_a_real_database` guard of its own;
the new test asserts it would pass that guard anyway, since "obviously a test
database, and a plain identifier" is what makes an unconditional drop safe.

Two notes on the issue text, which is slightly stale: the module has four tests, not
five — `test_the_chain_builds_the_schema_the_models_declare` is not in `main`, #183
landed its drift assertion through `make db-check` instead. And I removed
`test_current_matches_head`'s `alembic heads` round-trip: a `pytest.skip("No
migration revisions found")` that cannot fire with nine migrations, in the one module
whose defect was a skip that could not be told apart from a real one. It was also the
only subprocess in the file that ran without the forced `POSTGRES_DB`.

## Checks

      `tests/test_migration_suite_database.py` covers both halves of the skip
      decision and the naming; the same `pytest tests/test_migrations.py` reports
      **4 passed** on this branch and **4 skipped** on `origin/main`
- [x] `make check` passes — with the exception below
- [x] Platform-layer coverage is still 100% (3195 passed, 6 skipped)
- [ ] n/a — no migration added

Verified against a real Postgres 16:

- `pytest tests/test_migrations.py tests/test_migration_suite_database.py` → 10
  passed in 9s, and `agenticos_migrations_test_p*` is gone from `pg_database`
  afterwards.
- Breaking `0009_align_index_names.py`'s `downgrade()` with a bad index name turns
  `test_downgrade_base` and `test_upgrade_downgrade_cycle` red — the issue's "a
  deliberately broken `downgrade()` turns that job red".
- `make lint`, `make test`, `make test-frontend-cov`, `make build-frontend`,
  `make docs-build`, `make audit` all green.

## Notes for the reviewer

**`make check` fails at `db-check` on this machine and it is not this branch.**
`alembic check` runs against the development database `backend/.env` names, which is
currently at `0010_message_run_id` — a revision from another branch in flight. Same
target and same blast radius as #288 and #299. The four targets after it were run
individually.

Worth a thought, not done here: CI's `test` job still runs `make test-migrations`
against `test_db` in the step before `make test`, which now asks the same question
less safely and less specifically. It stays for now — it is in
`CI_ONLY_TARGETS` in `tests/test_ci_parity.py` and removing a CI step belongs in its
own change.

Docs: `docs/testing.md` gains the third test database beside the other two,
`docs/commands.md` points at the module from the `test-migrations` row, and both the
`alembic-migration` and `backend-tests` skills said `make test-migrations` was "the
only thing that proves `downgrade()` is real", which was true only because this
module never ran.

Closes #234, #346. Refs #183, #189, #288, #299.

### DIFF
--- a/backend/tests/test_migrations.py
+++ b/backend/tests/test_migrations.py
@@ -84,13 +84,20 @@ def _alembic(*args: str) -> subprocess.CompletedProcess[str]:
     )
 
 
-def _server_answers(url: str) -> bool:
+def _why_the_server_did_not_answer(url: str) -> str | None:
+    """`None` when the server answers, otherwise the reason it did not.
+
+    The reason is carried out rather than flattened to a bool because a server
+    that is up and refusing - a wrong password, a database still in recovery -
+    is a third case, and reporting it as the two below sends the reader to the
+    wrong place. Telling those apart is the whole job of `_demand_a_server`.
+    """
     engine = create_engine(url)
     try:
         with engine.connect():
-            return True
-    except Exception:
-        return False
+            return None
+    except Exception as exc:  # noqa: BLE001 - any failure to connect is the answer
+        return str(exc).strip()
     finally:
         engine.dispose()
 
@@ -118,15 +125,16 @@ def _demand_a_server(url: str) -> None:
     over a migration chain nobody applied, which is exactly the failure this
     module was already reporting silently.
     """
-    if _server_answers(url):
+    reason = _why_the_server_did_not_answer(url)
+    if reason is None:
         return
     if os.getenv("CI"):
         raise RuntimeError(
-            f"No database reachable at {url} but CI is set. The Postgres service "
-            "container did not come up; refusing to skip the migration suite and "
-            "report a green build."
+            f"No database reachable at {url} but CI is set, so the Postgres "
+            f"service container did not come up or is refusing us: {reason}. "
+            "Refusing to skip the migration suite and report a green build."
         )
-    pytest.skip("No database reachable - start one with `make docker-db`")
+    pytest.skip(f"No database reachable ({reason}) - start one with `make docker-db`")
 
 
 @pytest.fixture(scope="module", autouse=True)
--- a/backend/tests/test_migrations.py
+++ b/backend/tests/test_migrations.py
@@ -12,18 +12,55 @@ developer's working database. An ordinary `pytest` run emptied it: model tables
 gone, `alembic_version` recreated by the upgrade that followed, and nothing in
 the output to say so. The name below is the fix, and it is passed explicitly to
 each subprocess rather than left to the environment to get right.
+
+**That database is created here, and dropped again when the module is done.** It
+used to have to exist already: `alembic current` was run for its side effect of
+failing, and a missing database became a module-level skip. Nothing in the
+repository ever created it - the docstring saying so named `make test-migrations`,
+which cycles the chain against whatever `POSTGRES_DB` says and never touches this
+name - so every test here skipped on every CI run this project has ever had, and a
+build stayed green over four assertions nobody was making (#234). A skip meaning
+"no Postgres on this laptop" and a skip meaning "nobody has ever answered this"
+read identically in pytest's output, and only one of them is acceptable on a
+runner; `_demand_a_server` below is where they are told apart.
 """
+
+from __future__ import annotations
 
 import os
 import subprocess
 import sys
+from collections.abc import Iterator
 
 import pytest
+from sqlalchemy import create_engine
 
-# Owned by this module alone. Not the integration suite's database either: that
-# one builds its schema from the models with `create_all`, and alembic finds
-# those tables already present and fails on the first `CREATE TABLE`.
-MIGRATION_DATABASE = "agenticos_migrations_test"
+from app.core.config import settings
+
+# Owned by this module alone, and by this run of it. Not the integration suite's
+# database either: that one builds its schema from the models with `create_all`,
+# and alembic finds those tables already present and fails on the first `CREATE
+# TABLE`. The process id carries the same weight it does in `tests/conftest.py` -
+# this module drops the database it names, so a constant name would mean two runs
+# on one machine (two worktrees, or a `make test` beside this one) dropping each
+# other's mid-upgrade (#189).
+MIGRATION_DATABASE = f"agenticos_migrations_test_p{os.getpid()}"
+
+# Where `CREATE DATABASE` and `DROP DATABASE` are issued from, neither of which
+# can run from inside the database it names. Every Postgres image this project
+# uses ships it.
+_MAINTENANCE_DATABASE = "postgres"
+
+
+def _url(database: str) -> str:
+    """The DSN alembic builds, pointed at a different database on that server.
+
+    Derived from `settings` rather than from `os.environ` so this module and the
+    subprocesses it starts cannot disagree about the host, the user or the
+    password: `alembic/env.py` builds its URL from the same place, and a laptop
+    keeps those three in `backend/.env` rather than in the environment.
+    """
+    return f"{settings.DATABASE_URL_SYNC.rsplit('/', 1)[0]}/{database}"
 
 
 def _env() -> dict[str, str]:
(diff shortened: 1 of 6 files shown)


backend/tests/test_migrations.py
```diff
@@ -47,20 +84,80 @@ def _alembic(*args: str) -> subprocess.CompletedProcess[str]:
...
 
 
-def _db_available() -> bool:
-    """Whether this module's own database exists and answers.
+def _server_answers(url: str) -> bool:
+    engine = create_engine(url)
+    try:
+        with engine.connect():
+            return True
+    except Exception:
```
DEENUU1: This catches everything and answers "no server", so a Postgres that *is* up but refuses the connection — wrong password, wrong user, database in recovery — comes out the far end as `The Postgres service container did not come up`.

That is the one diagnosis this PR is otherwise careful not to make. The whole point of `_demand_a_server` is that two failures which look alike get told apart, and this collapses a third one into the wrong bucket. Somebody would go and stare at the service container definition, which would be fine.

Cheapest fix is to carry the reason out rather than a bool:

```python
def _server_answers(url: str) -> str | None:
    """`None` when the server answers, otherwise why it did not."""
    engine = create_engine(url)
    try:
        with engine.connect():
            return None
    except Exception as exc:  # noqa: BLE001 - any failure to connect is the answer
        return str(exc)
    finally:
        engine.dispose()
```

and then `_demand_a_server` puts it in both messages. The skip gets more useful too — "no database reachable: password authentication failed" on a laptop is a better hint than pointing at `make docker-db`.

Your call whether it is worth it here; the tests would need `lambda url: None` / `lambda url: "boom"` instead of the bools.

backend/tests/test_migrations.py
```diff
@@ -12,18 +12,55 @@
...
+from app.core.config import settings
+
+# Owned by this module alone, and by this run of it. Not the integration suite's
+# database either: that one builds its schema from the models with `create_all`,
+# and alembic finds those tables already present and fails on the first `CREATE
+# TABLE`. The process id carries the same weight it does in `tests/conftest.py` -
+# this module drops the database it names, so a constant name would mean two runs
+# on one machine (two worktrees, or a `make test` beside this one) dropping each
+# other's mid-upgrade (#189).
+MIGRATION_DATABASE = f"agenticos_migrations_test_p{os.getpid()}"
```
DEENUU1, line 47: This is the bit I would flag to anyone reading later: it also fixes **#346**, which I filed this afternoon after two `make check` runs in different worktrees dropped each other's chain mid-upgrade. Same defect, same reasoning as #189, and you got there first.

Worth putting `Closes #346` in the footer so the issue closes with the merge instead of sitting open behind a fix that already shipped.

DEENUU1 (commented):
Good catch, and the right fix of the two the issue offered. "Four skipped" reading as green for the entire life of the only tests that prove `downgrade()` works is exactly the shape this repo keeps getting bitten by, and fixing it in the module rather than in a CI step means a laptop with `make docker-db` up gets it too.

Two things I like specifically: `_demand_a_server` making the laptop case and the CI case *different outcomes* rather than the same skip, and the pid in the name — that second one closes an issue I filed a few hours ago (#346) without knowing this branch already had it, so it is worth adding to the footer.

One real comment below, on the error path of the probe. Nothing blocking.

DEENUU1:
Pushed the probe change from my review directly, since it was small and I wanted it in before the merge — `_why_the_server_did_not_answer` carries the reason out and both messages quote it.

It earned itself immediately. Running the suite in a worktree with no `backend/.env`, the old message said to start a database with `make docker-db` while that database was already up and listening; the real answer was `fe_sendauth: no password supplied`. Exactly the third case the bool was collapsing.

Verified against the real chain: `pytest tests/test_migrations.py tests/test_migration_suite_database.py` → 10 passed, and the temporary database is gone afterwards.

Also adding `Closes #346` to the body — the pid in the name fixes an issue I filed this afternoon without knowing this branch already had it.

Issue #232, opened 2026-08-05
Four more route modules call repositories directly


DEENUU1:
## What breaks

Four more route modules do what #197 describes: they import a repository and
read rows from a handler, bypassing the service that owns the entity.

| Route | Call |
|---|---|
| `backend/app/api/routes/v1/audit.py:32,35` | `audit_log_repo.list_for_org`, `audit_log_repo.count_for_org` |
| `backend/app/api/routes/v1/model_providers.py:144` | `organization_secret_repo.list_secrets`, inside a module-level `_listing_key` helper that also unseals the secret |
| `backend/app/api/routes/v1/org_integrations.py:152` | `sync_log_repo.get_all` |
| `backend/app/api/routes/v1/knowledge_bases.py:415` | `sync_log_repo.get_all` |

`CLAUDE.md` and `.claude/rules/architecture.md` both state it plainly: "Routes
call services only - **never** import or call a repository directly."

None of the four is a tenant leak today - `audit.py` and `model_providers.py`
scope on `ctx.organization_id`, `org_integrations.py` resolves the source through
`CollectionAccessService` first, and `knowledge_bases.py` filters the rows it
read against the collection it did check. The cost is the one #197 names: the
scope is a keyword argument a handler fills in itself, no service test can see
it, and the next filter added to the entity is added in the service while the
route keeps its own answer.

`model_providers.py` is the widest of the four - a route module reading sealed
secrets out of the vault and unsealing them - and is worth taking first.

## Sequence

```
rg -n 'from app.repositories' backend/app/api/routes/
```

Four hits, plus `from app.repositories.skill import SkillSort` in `skills.py`,
which is a type alias rather than data access and is fine.

## How you would know it was fixed

`backend/tests/test_route_layering.py` holds these four in `_KNOWN_VIOLATIONS`
so that the *next* one fails instead of being found by reading a file. Fixing a
module means deleting its entry - the same test fails if an allowlisted module
has stopped violating, so the list cannot rot in either direction.

`audit.py` needs a service that does not exist yet (`AuditService`); the other
three have one already (`OrganizationSecretService`, `SyncSourceService` /
`CollectionAccessService`, `KnowledgeBaseService`).

## Where it sits

Pre-existing; found fixing #197, which is the same boundary in
`app/api/routes/v1/runs.py` and left these four behind on purpose - one branch,
one module, and `audit.py` needs a new service of its own.

Cluster: nothing exactly. Nearest is #197 (the same rule, the run-history
[...]

PR #318, opened 2026-08-06
fix(rag): resolve a sync source before reading its logs

### LINKED ISSUE #233: KB sync-source logs page before they filter, and answer empty instead of 404
## What breaks

`GET /kb/{kb_id}/sync-sources/{source_id}/logs`
(`backend/app/api/routes/v1/knowledge_bases.py:404-436`) reads sync logs by
`source_id` with no scope on them at all, then filters the rows in Python
against the collection of the KB it *did* check:

```python
kb = await service.get(kb_id, ctx=ctx)
logs = await sync_log_repo.get_all(db, sync_source_id=source_id, limit=limit)
# Verify source belongs to this KB's collection (security: don't leak other KBs' logs).
items = [... for log in logs if log.collection_name == kb.collection_name]
```

Nothing leaks - the filter does catch a source belonging to another KB or another
organization. Two things are still wrong:

1. **`limit` is applied before the filter, in SQL.** A caller asking for 20 logs
   gets however many of *those* 20 rows survive the comparison, so a source whose
   history interleaves with another's answers with fewer rows than it holds and no
   way to page further. `total=len(items)` then reports the filtered page as the
   whole history.
2. **A source that is not this KB's answers `200 []`, not `404`.** Every other
   per-resource read in this surface reports a row it may not have as missing.
   "Empty" and "not yours" are the same response here, which is exactly the
   ambiguity `.claude/rules/frontend.md` warns about: the page renders "no syncs
   yet" for a request that should have failed.

`GET /org/integrations/{source_id}/logs` (`org_integrations.py:152`) is the same
query done correctly - it resolves the source through `CollectionAccessService`
first and then reads its logs - and is the shape to copy.

## Sequence

1. Two KB collections in one organization, each with a sync source that has run.
2. `GET /kb/{a_id}/sync-sources/{source_of_b}/logs` -> `200 {"items": [], "total": 0}`.
3. Run enough syncs on B that its logs fill the first 20 rows by `started_at`,
   then `GET /kb/{b_id}/sync-sources/{source_of_b}/logs?limit=20` while A is also
   syncing -> fewer than 20 rows, and `total` disagrees with what the source ran.

The scope is in the query rather than in a list comprehension: the sync source is
resolved for this KB first (404 if it is not this KB's), and the logs are read for
that resolved source, so `limit` and `total` describe the same set of rows. A test
asserts the 404 and that a full page comes back full.

## Where it sits

Pre-existing; found fixing #197 while sweeping the other routes that call a
repository directly. Filed separately because it is a defect rather than the
layering refactor: the layering issue for this file is the companion issue above,
and fixing that one is the natural moment to fix this.

Cluster: RAG and knowledge (#147, #148 - "the stats lie" about ingestion counts;
this is the same complaint about a sync-log listing).

OchnikBartek:
## What this changes

`GET /kb/{kb_id}/sync-sources/{source_id}/logs` now resolves the sync source
against the knowledge base before reading anything, instead of reading every log
row with that source id, cutting the page in SQL, and thinning the survivors in
Python.

Two answers change:

- **A source that is not this base's is `404`, not `200 []`.** "This source has
  never run" and "this source is not yours" used to render the same page, and one
  of them is a request that should have failed.
- **A full page comes back full.** `SyncSourceUpdate` carries `collection_name`,
  so a source can be repointed at another base and its earlier runs keep the name
  they ran against. Those rows were dropped *after* `limit` had already cut the
  page: asking for three runs answered with two, `total` described the survivors
  rather than the source, and nothing let a caller page past the gap.

Nothing leaked either way — the comparison did catch a foreign source — so this
was a wrong answer rather than an exposure.

## Why this way

The issue names the shape and it is already in the tree twice:
`GET /org/integrations/{source_id}/logs` resolves the source and then reads its
logs, and `trigger_kb_sync_source` — directly above this handler — does the
resolve-or-404 against `kb.collection_name`. Copying the neighbour keeps the two
sibling routes reading the same way, and it puts the scope in the query rather than
in a list comprehension, which is what makes `limit` and `total` describe one set
of rows.

The `if source.collection_name != kb.collection_name: raise NotFoundError` block is
now the sixth copy in this module (three for documents, three for sources). That is
deliberate rather than lazy: **#232 moves these route bodies into services**, and
extracting a helper here would be a second, conflicting refactor of the same lines.
`total` is still the page size, matching `list_org_integration_logs`; a real count
query is not what this issue asks for and would change the response's meaning on
both routes at once.

## Checks

- [x] Tests cover the new behaviour, and would fail without the change — both fail
      on the previous code with `assert 200 == 404` and `assert 2 == 3`
- [x] `make check` passes — except `db-check`, below
- [x] Platform-layer coverage is still 100% (3189 passed)
- [ ] n/a — no migration

The tests are integration rather than unit on purpose: the defect was the order of
a query and a filter, and a mocked session cannot show that. They live in
`tests/integration/test_platform_flows.py` beside the other `/kb` sync-source
tests, and use the `rag_estate` fixture; `_sync_log_row` gained an optional
`sync_source_id`, because the seeded rows deliberately have none.

Verified: `make lint`, `make test`, `make test-frontend-cov` (3175 passed),
`make build-frontend`, `make docs-build`, `make audit` — all green. `db-check`
fails on this machine for want of a `backend/.env`, which is
[#299](https://github.com/vstorm-co/agenticos/issues/299), the twin of #288.

## Notes for the reviewer

**Heads-up for #232**, which touches these same lines: the two behaviours above
have to survive the move into a service, and the two tests here should keep passing
untouched. If they need editing, the refactor changed behaviour.

One thing the issue claims that the code does not: it describes the short page as
two sources' histories interleaving. They cannot interleave — the SQL already
filters by `sync_source_id`. The reachable version is the repointed source above,
which is what the test builds.

Docs: `docs/file-processing.md`'s "Sync Operations" section described only the
collection-wide `GET /rag/sync/logs`; it now also describes the per-source endpoint
and what it answers for a source that is not that base's.

Closes #233. Refs #232, #299.

### DIFF
--- a/backend/app/api/routes/v1/knowledge_bases.py
+++ b/backend/app/api/routes/v1/knowledge_bases.py
@@ -406,14 +406,30 @@ async def list_kb_sync_source_logs(
     kb_id: UUID,
     source_id: UUID,
     service: KnowledgeBaseSvc,
+    sync_source_svc: SyncSourceSvc,
     db: DBSession,
     ctx: Auth,
     limit: int = Query(20, ge=1, le=100),
 ) -> Any:
-    """List sync run history for a specific KB sync source."""
+    """List sync run history for a specific KB sync source.
+
+    The source is resolved against this KB before its logs are read, rather than
+    the rows being filtered afterwards. Filtering afterwards leaked nothing, but
+    `limit` was applied in SQL to logs of *every* source with that id and then
+    thinned in Python - so a page came back short whenever another collection's
+    history interleaved with this one's, `total` described the survivors rather
+    than the source, and there was no way to page past the gap. A source that is
+    not this KB's is now missing rather than empty, which is what every other
+    per-resource read on this surface answers.
+    """
     kb = await service.get(kb_id, ctx=ctx)
-    logs = await sync_log_repo.get_all(db, sync_source_id=source_id, limit=limit)
-    # Verify source belongs to this KB's collection (security: don't leak other KBs' logs).
+    source = await sync_source_svc.get_source(str(source_id))
+    if source.collection_name != kb.collection_name:
+        raise NotFoundError(
+            message="Sync source not found in this knowledge base",
+            details={"kb_id": str(kb_id), "source_id": str(source_id)},
+        )
+    logs = await sync_log_repo.get_all(db, sync_source_id=source.id, limit=limit)
     items = [
         RAGSyncLogItem(
             id=str(log.id),
@@ -431,7 +447,6 @@ async def list_kb_sync_source_logs(
             completed_at=log.completed_at,
         )
         for log in logs
-        if log.collection_name == kb.collection_name
     ]
     return RAGSyncLogList(items=items, total=len(items))
 
--- a/backend/tests/integration/test_platform_flows.py
+++ b/backend/tests/integration/test_platform_flows.py
@@ -538,13 +538,16 @@ async def _sync_source_row(db, *, tenant: Tenant, collection_name: str) -> SyncS
     return source
 
 
-async def _sync_log_row(db, *, collection_name: str) -> SyncLog:
+async def _sync_log_row(
+    db, *, collection_name: str, sync_source_id: uuid.UUID | None = None
+) -> SyncLog:
     log = SyncLog(
         id=uuid.uuid4(),
         source="gdrive",
         collection_name=collection_name,
         mode="full",
         status="running",
+        sync_source_id=sync_source_id,
     )
     db.add(log)
     await db.flush()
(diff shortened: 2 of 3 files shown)


DEENUU1 (commented):
Good fix, and the reasoning in the docstring is the part worth keeping: "filtering afterwards leaked nothing, but the page was cut before the thinning" is a much more precise account than #233's title, and it is the kind of thing the next reader needs because the old code *looked* correct — the comment above it even claimed the security property it did have.

The two integration tests are doing real work. `test_a_full_page_of_one_sources_history_comes_back_full` fails against the old code for exactly the stated reason, and `200 []` versus `404` is the right distinction to pin: both render "no syncs yet", and only one of them is a request that should have failed.

**One thing about sequencing rather than the code.** `git merge-base --is-ancestor` says **#321 contains this branch in full** — the same route change and both of these tests. So the two cannot both land as-is; whichever goes second is either a no-op or a conflict.

I would rather keep both than fold this one away, because they are two different claims and the squashed history reads better with them apart: this branch is *the bug*, #321 is *the layering*. So the plan is this first, then #321 rebased — which drops these hunks and leaves it as what it actually is, moving the reads into services and finding the pagination bug on the way.

Nothing to change here. One note below on a detail I checked.

backend/app/api/routes/v1/knowledge_bases.py
```diff
@@ -406,14 +406,30 @@ async def list_kb_sync_source_logs(
...
+    The source is resolved against this KB before its logs are read, rather than
+    the rows being filtered afterwards. Filtering afterwards leaked nothing, but
+    `limit` was applied in SQL to logs of *every* source with that id and then
+    thinned in Python - so a page came back short whenever another collection's
+    history interleaved with this one's, `total` described the survivors rather
+    than the source, and there was no way to page past the gap. A source that is
+    not this KB's is now missing rather than empty, which is what every other
+    per-resource read on this surface answers.
+    """
     kb = await service.get(kb_id, ctx=ctx)
```
DEENUU1, line 425: Checked this rather than assuming it, since the new refusal rests on it: `get_source` is **not** scoped by organization — it resolves any source by id — so the whole tenancy check here is the `collection_name` comparison on the next line.

That is the same basis the old Python filter used, so this branch changes nothing about the exposure. Worth knowing anyway: `collection_name` carries no unique constraint, and its collision-resistance comes from the random suffix `_derive_collection_name` appends — which only applies when the caller lets it choose. **#367** is open and `severity:critical` precisely because `POST /kb` accepts an explicit `collection_name` and never calls `CollectionAccessService.claim`.

So this line means what its docstring says once #367 is closed, and not before. Recording it here so whoever takes #367 knows this call site is downstream of it.

Issue #317, opened 2026-08-06
CI burns ~8,900 billed Actions minutes a week, a fifth of it on runs nothing cancels


DEENUU1:
**Found reviewing the org's August billing, 2026-08-06.** Cluster: **Tests, CI,
dependencies** (#109 #130 #143 #154 #155 #156 #158 #162 #188 #203 #215 #220 #223 #234).
Nearest relative is **#215** — the integration suite's 92 seconds, which is one job
inside the largest line below. Not a duplicate: #215 makes one suite faster, this is
about how many times the whole matrix runs and what each run pays for. Nothing else open
covers CI minutes.

## First, the honest part: nothing is being billed today

`agenticos` shows **$71.33** gross metered Actions usage for 2026-08-01 → 08-06, of
$90.81 across the org. **Billable spend is $0** — public-repository Actions usage is
fully discounted, and the `641 min / 2,000 min included` counter on the billing page is
other repositories' private usage, not this one's.

So this is not "we are losing money". It is three other things:

1. **The discount is a policy, not a guarantee**, and it evaporates the day this
   repository goes private. At the current rate that is ~13,000 billed minutes a month,
   which is 6.5× the 2,000 included and about **$90–100/month** at the Linux 2-core rate.
2. **About 1,800 of those minutes in six days were provably wasted** — see the
   concurrency section. That is bad at any price.
3. **Every wasted job holds a slot in the org's concurrent-job limit**, so the cost is
   already being paid in other repositories' queue time even while the invoice reads $0.

Filing it now because the fix is small, mostly mechanical, and gets much less appealing
once it is urgent.

## Where the minutes go

Method — the `/timing` endpoint reports `0` billable on a public repository, so this is
computed from each job's own `started_at`/`completed_at`, rounded up per job the way
GitHub bills:

```bash
gh api 'repos/vstorm-co/agenticos/actions/workflows/ci.yml/runs?per_page=100&page=N'
gh api "repos/vstorm-co/agenticos/actions/runs/$id/jobs?per_page=100"
```

**369 completed `ci.yml` runs since 2026-08-01** — 316 success, 51 failure, **2
cancelled**. Over a 124-run sample of them, **24.1 billed minutes per run**, so `ci.yml`
alone is on the order of **8,900 billed minutes in six days**. Per job, from that same
sample:

| Workflow / job | billed min | billed/job | wall/job | note |
|---|---|---|---|---|
[...]

Issue #165, opened 2026-08-04
agent_session.py carries the whole chat wire format and is in neither gate (31% covered)


DEENUU1:
**Pre-existing, found implementing #40.** This branch did not cause it.

`app/services/agent_session.py` is the module that decides every frame the dashboard
chat WebSocket sends and every frame it accepts. It is in **neither**
`[tool.coverage.run] include` nor `[[tool.ty.overrides]] include` in
`backend/pyproject.toml` — only its sibling `app/services/agent_chat.py` is.

Measured on `feat/subagents`:

```
$ uv run pytest tests/test_agent_session.py -q --cov=app.services.agent_session --cov-report=term-missing
app/services/agent_session.py   195   122   76   2   31%
Missing: 80-101, 105-112, 116-127, 131-136, 140, 160-301, 310-317, 352-355, 365-380, 390-434
```

So neither the 100% gate nor strict `ty` applies to it, and `tests/test_agent_session.py`
covers under a third of it. What is uncovered is not incidental: `160-301` is
`process_message` — the whole turn, including the `tool_approval_required` frame and both
terminal `complete` frames — and `390-434` is the delta/tool event translation.

## Why it matters more than a coverage number

Every surface reads this format. A frame renamed here is a frontend branch that silently
stops matching, which is exactly the failure #144 describes for `web_search` and
`create_chart`: *"the frontend matches tool names the backend stopped emitting"*. That
class of bug is what a gate on this file would catch.

It also came within one line of mattering on `feat/subagents`. The delegation work adds
six event types here; the brief for that work asserted this module was gated, it is not,
and the new code was only held to 100% because the author checked rather than because CI
would have.

## How you would know it was fixed

- `app/services/agent_session.py` appears in **both** lists in `backend/pyproject.toml`,
  verbatim and in the same order (`tests/test_coverage_gate.py` enforces the pairing).
- `make test` passes with it there, which means the module is at 100% and `ty`-clean.

That is real work — roughly 120 statements and 74 branches, most of them behind an
`AgentSession` needing a fake WebSocket and a driven agent run — which is why this is an
issue rather than a fix folded into an unrelated branch.

Refs #40, #144

DEENUU1:
[...]

PR #319, opened 2026-08-06
ci: stop paying for superseded runs and suites a change cannot affect

DEENUU1:
## What this fixes

CI spent about **8,900 billed Actions minutes** in the first six days of August —
369 `ci.yml` runs at 24.1 minutes each. Three of the four things #317 asked for.

Refs #317 — the issue stays open for the fourth (see *Deliberately not done*).

### 1. A concurrency group, which the workflow simply did not have

`.github/workflows/ci.yml:20` — `ai-review.yml` and `docs.yml` both carry one;
`ci.yml` did not. Every push started a fresh matrix and left the previous one
running to completion. **75 of those 369 runs were superseded while still in
flight — about 1,800 billed minutes**, a fifth of the total, and only 2 runs in
the window were ever `cancelled`, which is what nothing-cancels looks like from
the outside.

Keyed on `github.ref`, exempting `main`: cancelling a merge's run leaves a commit
whose CI never finished and a badge that means nothing.

### 2. Caching, absent at all seven install sites

- `setup-uv` × 5 → `enable-cache: true`. It was called with only
  `version: "latest"`, so every job re-resolved and re-downloaded all **278**
  locked packages.
- `~/.bun/install/cache` × 2 → `setup-bun` caches the binary, not the packages,
  so `bun install --frozen-lockfile` refetched everything twice per run.
- `~/.cache/ms-playwright` → `e2e` downloaded ~170 MB of Chromium every run.

All three keyed on the lockfile that pins them, so a dependency change
invalidates them and nothing else does.

### 3. Skipping the suites a change set cannot affect

`test`, `test-frontend` and `e2e` are 8.2, 5.3 and 5.1 billed minutes, and a
backend-only pull request paid for all three. A `changes` job now decides, and
the decision lives in `scripts/ci_changed_scope.py` rather than in a glob, so it
is testable.

**It is a job-level `if:`, deliberately not a `paths:` filter.** GitHub satisfies
a required status check with `success`, `skipped` **or** `neutral`, so a skipped
job is a pass. A filtered-out workflow never posts its checks at all, and `main`'s
ruleset would wait forever on six contexts that never arrive — the trap named in
#317.

**The classifier is written the timid way round.** A job is skipped only when
*every* changed path is provably irrelevant to it, so anything unrecognised runs
everything. The permissive spelling — *run when a path I recognise as relevant
changed* — would let a new directory silently stop a suite, which is not a red
build but a green one with a gate missing from it. That is #143 and #165. Two
exemptions only, both checked rather than assumed:

- `docs/**`, `mkdocs.yml`, a top-level `*.md` — no test under `backend/tests/`
  reads any of them; the one mention is a docstring naming `docs/channels.md`.
- the opposite half of the tree, for each unit suite.

`e2e` is exempted from neither half. `lint` is not gated at all, because
`make lint-spelling` is the only thing that reads every tracked file.

`changes` is its own job and costs the billed minute GitHub rounds it up to.
Folding it into `lint` would be free but would put all three suites behind
`needs: lint`, so a ruff error would skip them and hide whatever else was broken
until the next push.

## The guard

`backend/tests/test_ci_changed_scope.py` — 25 tests, same shape as
`test_ci_parity.py` and `test_coverage_gate.py`, because both halves of this
change fail *quietly*. It asserts the timid direction explicitly, that each gated
job reads its own output with `!= 'false'` and not `== 'true'` (which would skip
it on a push, where the classifier does not run), that `lint` is never gated, and
that `main` is never cancelled.

## One correction to #317

**It overstated the saving, and the arithmetic is worth having straight.** #317
claimed a release pull request would stop paying for the whole matrix. It will
not: the version lives in `backend/pyproject.toml`, `backend/uv.lock` and
`frontend/package.json`, so every filter matches — now asserted in
`test_a_release_bump_runs_everything`.

Measured over the 97 pull requests merged since 2026-07-25, the available skips
are 27 `test-frontend`, 15 `test`, 4 `e2e` — about 286 minutes over first runs,
~850 across re-runs. Minus the `changes` job's own minute per run, **the path gate
nets around 5%** of the total. The concurrency group is 20% for four lines. These
were not three equal wins and the issue read as though they were.

## Deliberately not done

Folding `lint`, `docs` and `Security Scan` into one job. They bill a whole minute
each for 22–28 seconds of work — 372 minutes for about 150 minutes of work — and
merging them recovers most of it. But **all three are required status checks by
name**, so it needs `main`'s ruleset edited in the same breath or every pull
request is blocked; it costs three independent failure signals; and it means
editing `test_ci_parity.py`'s `GATING_JOBS`, which is the one thing keeping
`make check` honest (#143). That is a merge-gate change, so it wants a decision
rather than a commit. Left on #317 with the trade written down.

CodeQL's 625 `Analyze` jobs (1,118 minutes) are also untouched: default setup
generates those workflows and gives us no surface to tune, as `docs/code-review.md`
already records.

## Testing

- `make test` — 3216 passed, 6 skipped, **100%** on the platform layer
- `make lint-backend`, `make lint-spelling` — clean
- `make docs-build` under `--strict` — clean; `scripts/docs_drift.py` — no drift
- yamlfmt + zizmor over the workflow, and the full pre-commit set over the diff
- `scripts/ci_changed_scope.py` run as the workflow runs it, against a docs-only
  list, a backend-only list and this change's own file list

**Not verified, and not verifiable from this branch:** that a skipped required
check satisfies the ruleset. This pull request touches `.github/`, `scripts/` and
`backend/`, so by its own rule it runs everything — there is no skip to observe.
The behaviour is documented by GitHub, a failure would show immediately as a merge
button that never goes green, and the revert is the three `if:` lines. **The first
docs-only pull request after this merges is the real proof, and is worth watching.**

## Docs

`docs/branching.md` gains *A required check may legitimately report `skipped`* and
*One run per branch*; `docs/testing.md` says CI may run fewer jobs than
`make check` and that this is not drift; `CLAUDE.md`'s **The loop** says a push now
cancels the run in flight, so the answer you were waiting on disappears rather
than arriving about a commit you already replaced.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,21 @@ on:
   pull_request:
     branches: [main, master]
 
+# One run per branch. `CLAUDE.md` asks for a commit and a push per finished piece,
+# which is right - but with nothing cancelling, each push started a fresh 24-minute
+# matrix and left the previous one running to completion, answering a question
+# about a commit nobody was waiting on. 75 of the 369 runs in the first six days of
+# August were superseded while still in flight, about 1,800 billed minutes, and
+# only 2 runs in that window were ever `cancelled`. See #317.
+#
+# `main` is deliberately excluded: the push run is what makes the merge history and
+# the badge mean anything, and cancelling it would leave commits whose CI never
+# finished. `github.ref` separates them without a second condition - a pull request
+# is `refs/pull/N/merge`, a merge is `refs/heads/main`.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 # Least privilege for every job. Nothing in this workflow writes anything back:
 # the docker job builds with `push: false`, and Codecov uses its own token. A
 # job that later needs more should declare it on itself, not here.
@@ -17,7 +32,55 @@ permissions:
   contents: read
 
 jobs:
+  changes:
+    # Which of the three expensive suites this pull request can provably not
+    # affect. `scripts/ci_changed_scope.py` decides, timidly - a path it does not
+    # recognise runs everything - and its docstring says why that direction is the
+    # only safe one. A skipped job satisfies a required status check (GitHub counts
+    # `success`, `skipped` and `neutral`), which is what makes this legal against
+    # the six contexts `main`'s ruleset requires; a workflow-level `paths:` filter
+    # would instead leave those checks never reporting and block every merge
+    # forever, which is the trap named in #317.
+    #
+    # A job of its own rather than a step inside `lint`, at the cost of the one
+    # billed minute GitHub rounds it up to: folding it into `lint` would be free,
+    # but it would put `test`, `test-frontend` and `e2e` behind `needs: lint`, so a
+    # ruff error would skip all three and hide whatever else was broken until the
+    # next push.
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      pull-requests: read
+    outputs:
+      backend: ${{ steps.scope.outputs.backend }}
+      frontend: ${{ steps.scope.outputs.frontend }}
+      e2e: ${{ steps.scope.outputs.e2e }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Work out which suites the changed paths require
+        id: scope
+        # Nothing is emitted on a push, so every downstream `if:` compares against
+        # an empty string and runs. A merge to `main` runs the whole matrix, and
+        # so does a re-run of this job that fails - `!= 'false'` is the safe
+        # direction in both cases.
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          gh api --paginate \
+            "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/files?per_page=100" \
+            --jq '.[].filename' > "$RUNNER_TEMP/changed.txt"
+          python3 scripts/ci_changed_scope.py "$RUNNER_TEMP/changed.txt" >> "$GITHUB_OUTPUT"
+
   lint:
+    # Never gated. It is one billed minute, `make lint-spelling` reads every
+    # tracked file rather than a subtree, and a suite that skips is only defensible
+    # while something still looks at the whole tree.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -28,6 +91,10 @@ jobs:
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
         with:
           version: "latest"
+          # Caches `~/.cache/uv`, keyed on `**/uv.lock`. Without it every job
+          # re-resolved and re-downloaded all 278 locked packages, five times per
+          # run across this workflow. See #317.
+          enable-cache: true
 
       - name: Set up Python
         run: uv python install 3.12
@@ -65,6 +132,10 @@ jobs:
     # runs in the `test` job, where a database exists.
 
   test:
+    needs: changes
+    # `!= 'false'` rather than `== 'true'`: on a push the `changes` step does not
+    # run and the output is empty, which has to mean *run*.
+    if: needs.changes.outputs.backend != 'false'
     runs-on: ubuntu-latest
     services:
       postgres:
@@ -115,6 +186,7 @@ jobs:
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
         with:
           version: "latest"
+          enable-cache: true
 
       - name: Set up Python
         run: uv python install 3.12
@@ -164,6 +236,8 @@ jobs:
           fail_ci_if_error: false
 
   test-frontend:
+    needs: changes
+    if: needs.changes.outputs.frontend != 'false'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -172,6 +246,18 @@ jobs:
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+
+      # `setup-bun` caches the bun binary, not the packages, so
+      # `bun install --frozen-lockfile` re-downloaded every dependency on every
+      # run. Bun's global install cache is what makes the second run cheap; the
+      # key is the lockfile, so a dependency change invalidates it and nothing
+      # else does.
+      - name: Cache the bun install cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.bun/install/cache
+          key: bun-${{ runner.os }}-${{ hashFiles('frontend/bun.lock') }}
+          restore-keys: bun-${{ runner.os }}-
 
       - name: Install dependencies
         run: bun install --frozen-lockfile
@@ -200,6 +286,10 @@ jobs:
     # uv, a migrated database, a seeded organization and a running backend;
     # linting needs none of that, and making a formatting error wait on a
     # database pull — or hiding it behind a browser timeout — helps nobody.
+    needs: changes
+    # Exempted from neither half: this job drives the frontend against the
+    # backend, so it is only skippable for a change that touches neither.
+    if: needs.changes.outputs.e2e != 'false'
     runs-on: ubuntu-latest
     services:
       postgres:
@@ -253,6 +343,7 @@ jobs:
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
         with:
           version: "latest"
+          enable-cache: true
 
       - name: Set up Python
         run: uv python install 3.12
@@ -313,6 +404,13 @@ jobs:
       - name: Setup Bun
         uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
 
+      - name: Cache the bun install cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.bun/install/cache
+          key: bun-${{ runner.os }}-${{ hashFiles('frontend/bun.lock') }}
+          restore-keys: bun-${{ runner.os }}-
+
       - name: Install frontend dependencies
         run: bun install --frozen-lockfile
         working-directory: frontend
@@ -324,7 +422,19 @@ jobs:
         run: bun run build
         working-directory: frontend
 
+      # Chromium is ~170 MB downloaded on every run otherwise. Keyed on the
+      # lockfile, because that is what pins the Playwright version the browser
+      # has to match - a browser cached against a different version is the kind
+      # of failure that reads as a flaky spec.
+      - name: Cache the Playwright browsers
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('frontend/bun.lock') }}
+
       - name: Install Playwright browsers
+        # `--with-deps` still runs: the apt packages Chromium links against live
+        # outside the cached directory, and installing them on a hit is seconds.
         run: bunx playwright install --with-deps chromium
         working-directory: frontend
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -10,14 +10,77 @@ on:
...
+# August were superseded while still in flight, about 1,800 billed minutes, and
+# only 2 runs in that window were ever `cancelled`. See #317.
+#
+# `main` is deliberately excluded: the push run is what makes the merge history and
+# the badge mean anything, and cancelling it would leave commits whose CI never
+# finished. `github.ref` separates them without a second condition - a pull request
+# is `refs/pull/N/merge`, a merge is `refs/heads/main`.
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
DEENUU1: `cancel-in-progress: false` does not mean "leave `main` alone" — it means "queue". A push to `main` still joins the group `ci-refs/heads/main`, so it waits for the run in flight, and GitHub's documented rule is that **any previously pending run in the group is cancelled when a newer one is queued**.

So merge A is running, B lands and goes pending, C lands and B is cancelled outright. B's commit never gets CI at all — which is exactly the thing the comment above says the `main` exemption exists to prevent. With 14 releases in six days plus feature merges, and a `main` run taking ~10 minutes wall (the `docker` job is 5.9 of it), two merges inside one window is not a rare shape.

The fix is to keep `main` out of the group entirely rather than out of the cancelling:

```suggestion
  group: ci-${{ github.ref }}-${{ github.event_name == 'push' && github.run_id || 'pr' }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

`github.run_id` is unique per run, so every push to `main` gets a group of its own and neither cancels nor queues. Pull requests are unaffected — they all resolve to `-pr` and cancel each other per `github.ref` as before. Worth an assertion in `TestTheConcurrencyGroup` too: something like `assert "github.run_id" in workflow["concurrency"]["group"]`, with the reason, since nothing else would catch this coming back.

.github/workflows/ci.yml
```diff
@@ -10,14 +10,77 @@ on:
...
+    #
+    # A job of its own rather than a step inside `lint`, at the cost of the one
+    # billed minute GitHub rounds it up to: folding it into `lint` would be free,
+    # but it would put `test`, `test-frontend` and `e2e` behind `needs: lint`, so a
+    # ruff error would skip all three and hide whatever else was broken until the
+    # next push.
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      pull-requests: read
```
DEENUU1, line 69: A job-level `permissions:` block **replaces** the workflow-level one rather than adding to it, so declaring `pull-requests: read` here silently sets `contents` to `none` for this job — and this job runs `actions/checkout`.

It passes today because the repository is public and public content is readable by any token. But the reason #317 exists at all is that the Actions discount "evaporates the day this repository goes private", and on that day this is the first job that breaks — with a checkout failure that reads like a credentials problem rather than a permissions one.

```suggestion
    permissions:
      contents: read
      pull-requests: read
```

.github/workflows/ci.yml
```diff
@@ -10,14 +10,77 @@ on:
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Work out which suites the changed paths require
+        id: scope
+        # Nothing is emitted on a push, so every downstream `if:` compares against
+        # an empty string and runs. A merge to `main` runs the whole matrix, and
+        # so does a re-run of this job that fails - `!= 'false'` is the safe
+        # direction in both cases.
```
DEENUU1: This is the one I'd want fixed before merge, because the comment states a protection the workflow does not have.

`!= 'false'` covers the case where the **step** is skipped and the output is empty. It does not cover the case where the **job fails** — `gh api` returning a 502, a rate limit, the checkout failing, `set -euo pipefail` catching anything at all. When a job in `needs:` fails, GitHub marks every dependent job `skipped` without evaluating its `if:` at all.

And by this pull request's own (correct) premise, `skipped` satisfies a required status check. So the sequence is:

1. `changes` fails for a transient reason.
2. `test`, `test-frontend` and `e2e` are skipped.
3. All three report `skipped`, which the ruleset accepts.
4. `changes` itself is **not** one of the six required contexts — I checked `main — protected`, they are `lint`, `test`, `test-frontend`, `e2e`, `docs`, `Security Scan` — so nothing blocks on its failure.
5. The merge button goes green on a branch where no suite ran.

That is #143/#165 again, arriving through the mechanism built to avoid them.

The in-workflow fix is one function per gated job, and it keeps the timid direction rather than adding a second rule (suggestions on each `if:` below):

```yaml
if: ${{ !cancelled() && needs.changes.outputs.backend != 'false' }}
```

A failed `changes` then leaves the output empty, `'' != 'false'` is true, and the suite runs — which is what "timid" is supposed to mean here. Adding `changes` to the ruleset would also work, but that is a merge-gate edit and this pull request deliberately avoids those.

Worth extending the comment too: the honest version is that `!= 'false'` handles the *empty* output and `!cancelled()` handles the *failed job*, because they are two different mechanisms.

.github/workflows/ci.yml
```diff
@@ -10,14 +10,77 @@ on:
...
+        if: github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          gh api --paginate \
+            "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER/files?per_page=100" \
+            --jq '.[].filename' > "$RUNNER_TEMP/changed.txt"
```
DEENUU1: `.filename` is the *new* path, so a rename is invisible on the side it left. Renaming `backend/app/services/foo.py` to `frontend/src/foo.ts` reports one changed path under `frontend/`, and the classifier answers `{'backend': False, ...}` — I ran it to be sure. The backend suite is then skipped for a change that deleted a backend module.

Small window, but it breaks the script's actual claim, which is not "these paths changed" but "every changed path is provably irrelevant". A rename means the proof is missing half its input, and the endpoint already hands you the other half:

```suggestion
            --jq '.[] | .filename, (.previous_filename // empty)' > "$RUNNER_TEMP/changed.txt"
```

`previous_filename` is only present on renames, so `// empty` keeps every other entry a single line. Worth a case in `TestNothingUnrecognisedIsEverSkipped` — the classifier can't be tested for it, so it has to be asserted on the jq.

.github/workflows/ci.yml
```diff
@@ -65,6 +132,10 @@ jobs:
     # runs in the `test` job, where a database exists.
 
   test:
+    needs: changes
+    # `!= 'false'` rather than `== 'true'`: on a push the `changes` step does not
+    # run and the output is empty, which has to mean *run*.
+    if: needs.changes.outputs.backend != 'false'
```
DEENUU1:
```suggestion
    if: ${{ !cancelled() && needs.changes.outputs.backend != 'false' }}
```

.github/workflows/ci.yml
```diff
@@ -164,6 +236,8 @@ jobs:
           fail_ci_if_error: false
 
   test-frontend:
+    needs: changes
+    if: needs.changes.outputs.frontend != 'false'
```
DEENUU1:
```suggestion
    if: ${{ !cancelled() && needs.changes.outputs.frontend != 'false' }}
```

.github/workflows/ci.yml
```diff
@@ -200,6 +286,10 @@ jobs:
     # uv, a migrated database, a seeded organization and a running backend;
     # linting needs none of that, and making a formatting error wait on a
     # database pull — or hiding it behind a browser timeout — helps nobody.
+    needs: changes
+    # Exempted from neither half: this job drives the frontend against the
+    # backend, so it is only skippable for a change that touches neither.
+    if: needs.changes.outputs.e2e != 'false'
```
DEENUU1:
```suggestion
    if: ${{ !cancelled() && needs.changes.outputs.e2e != 'false' }}
```

backend/tests/test_ci_changed_scope.py
```diff
@@ -0,0 +1,180 @@
...
+    def test_each_gated_job_reads_its_own_output(
+        self, workflow: dict[str, Any], job: str, output: str
+    ) -> None:
+        definition = workflow["jobs"][job]
+        assert "changes" in definition["needs"], f"`{job}` does not depend on `changes`"
+        assert definition["if"] == f"needs.changes.outputs.{output} != 'false'", (
+            f"`{job}` must gate on `needs.changes.outputs.{output} != 'false'`. "
+            "`== 'true'` would skip it on a push to `main`, where the classifier does "
+            "not run and the output is empty."
+        )
```
DEENUU1: This assertion is doing the right job — it is the only thing that would notice someone "simplifying" the gate to `== 'true'`. It needs to move with the `!cancelled()` fix, and while it is being edited it may as well assert both halves separately, since they guard two different failure modes:

```suggestion
        expected = f"${{{{ !cancelled() && needs.changes.outputs.{output} != 'false' }}}}"
        assert definition["if"] == expected, (
            f"`{job}` must gate on `{expected}`. `== 'true'` would skip it on a push "
            "to `main`, where the classifier does not run and the output is empty; "
            "and without `!cancelled()` a *failed* `changes` job skips this suite "
            "entirely, which a required status check accepts as a pass."
        )
```

DEENUU1 (commented):
The measurement work behind this is the good part, and so is correcting #317's own arithmetic instead of letting the 340 minutes stand. A few of the small decisions are subtle and right: `changes` runs unconditionally with only its *step* gated on `pull_request` (gating the whole job would have skipped it, and a skipped `needs` skips every dependent), `!= 'false'` rather than `== 'true'`, and a guard test in the same shape as `test_ci_parity.py`.

Three things I'd fix before this merges, all the same class as the bug the script's docstring is written to avoid — a gate that reads green because the thing is not in it:

1. **A `changes` job that *fails* skips all three suites, and skipped is a pass.** `changes` is not one of the six required contexts (I checked the ruleset), so one 502 out of `gh api` merges a branch nothing tested. The comment at `ci.yml:66` claims `!= 'false'` covers this; it does not, because a failed `needs` skips the dependent before the `if:` is ever evaluated.
2. **On `main` the group now queues instead of cancelling**, and GitHub cancels a *pending* run when a newer one is queued into the same group. That is the "commit whose CI never finished" the comment is there to prevent — reintroduced by the mechanism meant to avoid it.
3. **The `changes` job's `permissions:` block drops `contents: read`.** Job-level permissions replace the workflow-level ones entirely, so `actions/checkout` there runs with `contents: none`. It passes today only because the repository is public, and #317's premise is the day it isn't.

Plus a rename hole in the classifier and one nit. Details inline.

scripts/ci_changed_scope.py
```diff
@@ -0,0 +1,96 @@
...
+    paths = Path(argv[1]).read_text().splitlines()
+    decided = scope(paths)
+    for job, required in decided.items():
+        print(f"{job}={'true' if required else 'false'}")
+
+    skipped = sorted(job for job, required in decided.items() if not required)
+    print(
+        f"{len(paths)} changed path(s); skipping {', '.join(skipped) if skipped else 'nothing'}",
+        file=sys.stderr,
+    )
```
DEENUU1, line 102: nit: `len(paths)` counts the raw lines, including the blank ones `scope()` then filters out, so the log line can say "4 changed path(s)" for three files. One-line fix, and it makes the number match the thing being decided on:

```suggestion
    skipped = sorted(job for job, required in decided.items() if not required)
    counted = sum(1 for path in paths if path)
    print(
        f"{counted} changed path(s); skipping {', '.join(skipped) if skipped else 'nothing'}",
        file=sys.stderr,
    )
```

DEENUU1:
All six taken, in `52ae7dc`. Four were defects and three of those four were the same
mistake the gate exists to prevent — a green build with a suite missing from it —
arriving through the mechanism built to avoid it. Each verified before changing
anything rather than taken on the strength of the argument.

**The failed-`changes` hole was the worst of them, and the reasoning is exactly
right.** A failed `needs` skips its dependents without reading their `if:`, the three
report `skipped`, a skipped required check is a pass, and `changes` is not one of the
six required contexts — so one 502 from `gh api` and the merge button goes green over
a branch where nothing ran. `!cancelled()` on each gated job, and the comment now
says plainly that `!= 'false'` handles the empty output and `!cancelled()` handles the
failed job, because they are two different mechanisms.

**`cancel-in-progress: false` queues rather than exempts** — confirmed in GitHub's
own words before touching it: "any existing pending job or workflow in the same
concurrency group will be canceled and the new queued job or workflow will take its
place." So the exemption I wrote produced the precise outcome its comment said it
prevented. `github.run_id` on a push, as suggested.

**The rename**, checked against PR #101, which renames
`framework-freshness.yml` → `dependency-freshness.yml`:

```
$ gh api repos/vstorm-co/agenticos/pulls/101/files \
    --jq '.[] | .filename, (.previous_filename // empty)'
.github/dependabot.yml
.github/workflows/dependency-freshness.yml
.github/workflows/framework-freshness.yml     <- the previous name
Makefile
backend/pyproject.toml
docs/branching.md
```

Both ends of the rename, one line for everything else. The script's docstring now
says the caller owes both paths and why the script itself cannot notice when it does
not get them.

**The `permissions:` block** — right, and the failure mode you name is the one that
would waste the most time, since a checkout refusal reads as a credentials problem.
`contents: read` restored, with the reason.

**The counting nit** — fixed; it now counts the way `scope()` counts.

Three of the four are now asserted rather than argued: `github.run_id` in the group,
`previous_filename` in the command, and the full `if:` expression with both halves
explained in the failure message. The fourth is `zizmor`'s. `docs/branching.md` also
had to change — it described a `main` exemption the workflow did not actually have.

One thing I did not fold in: the suggestion to assert on `TestTheConcurrencyGroup` is
there, but as `test_a_push_gets_a_group_of_its_own` rather than an added line in the
existing test, so the failure message can carry the pending-cancellation rule. That
rule is the entire reason the group looks strange.

PR #321, opened 2026-08-06
refactor(api): move the last four route reads into services

### LINKED ISSUE #232: Four more route modules call repositories directly

OchnikBartek:
## What this changes

The four route modules that imported a repository no longer do, and a test keeps it
that way.

| Was | Now |
|---|---|
| `audit.py` held both queries and the 7-field mapping | `AuditService.list_for_organization` |
| `model_providers.py` read sealed rows out of the vault and unsealed them | `OrganizationSecretService.listing_key` |
| `knowledge_bases.py` read one source's logs, with its own copy of the mapping | `SyncSourceService.list_logs` |
| `org_integrations.py` read the same rows, with a second copy of the same mapping | the same method |

`backend/tests/test_route_layering.py` fails on any import from `app.repositories`
under `app/api/routes/` — on the previous tree it names all four — and fails equally
if its allowlist keeps an exemption that no longer applies, so the list cannot rot in
either direction. The single exemption is `SkillSort`, a `Literal` of sort orders
imported as a type rather than as data access.

**Behaviour is unchanged. That is the point**, and it is what the tests are for.

## Why this way

**`AuditService` takes the auth context, not an organization id.** The route was
passing `organization_id=ctx.organization_id` to two repository calls, and the value
was right. The problem is where the rule lived: "an entry belongs to exactly one
organization" was a property of that handler rather than of the log, so no service
test could see it and the next reader — an export, a channel command, an admin page —
would have had to know to pass the same thing. With `ctx` there is nowhere to pass
another tenant.

**`listing_key` is deliberately not narrowed the way `list_secrets` is.** Visibility
decides which keys a caller may *see and account for*; a member who cannot see the
organization's OpenAI key is still shown the same model list as everybody else,
because what leaves the service is a catalog and the plaintext goes to the provider.
Narrowing it would make a form's suggestions depend on who is looking at the form.
This preserves exactly what the route did — the move is the point, not a change of
policy.

**`list_logs` takes an id that has already been resolved.** Scope is not that
method's to add: an org integration resolves the source through
`CollectionAccessService`, and the `/kb` route resolves it against the base in the
path (that resolve-or-404 is #233, in the commit below). Putting a second scope
inside the shared method would mean two answers to the same question.

**`app/services/audit.py` joins the platform layer** — both `pyproject.toml` lists
and `PLATFORM_MODULES` in `tests/test_coverage_gate.py`, at 100%. It is one of the
services AgenticOS adds, and the one wrong argument in it would show another tenant's
actions.

## Checks

      `test_route_layering.py` fails on the pre-change tree with all four modules
      named; `test_audit_service.py` (4) and the `listing_key` class in
      `test_secrets.py` (4) are new
- [x] `make check` passes — except `db-check`, below
- [x] Platform-layer coverage is still 100% — 3199 passed, 107 files at 100%
- [ ] n/a — no migration

`make lint`, `make test`, `make test-frontend-cov` (3175 passed), `make
build-frontend`, `make docs-build`, `make audit` all green. `db-check` fails on this
machine for want of a `backend/.env` —
[#299](https://github.com/vstorm-co/agenticos/issues/299).

## Notes for the reviewer

**Depends on #318** (issue #233). This branch is cut from that one, not from `main`,
because #233 rewrites the same handler — `list_kb_sync_source_logs` — and cutting
from `main` would have meant silently reverting its fix while moving the code. Merge
#318 first; this then applies cleanly. **The two integration tests from #233 pass
untouched here**, which is the evidence the refactor preserved the behaviour they
pin: the source is still resolved against the base, a foreign source is still 404,
and a full page still comes back full.

The two sibling log routes now share one mapping instead of holding a copy each,
which is the only place this change removes code rather than moving it.

`app/core/audit.py` (the *write* half) still builds the model directly instead of
going through the repository. Out of scope here — it is not a route, so the rule this
branch is about does not reach it — and worth a look by whoever next touches the
trail.

Closes #232. Refs #197, #233, #299.


backend/app/api/deps.py
```diff
@@ -719,3 +719,12 @@ def get_secret_service(db: DBSession) -> OrganizationSecretService:
 
 
 SecretSvc = Annotated[OrganizationSecretService, Depends(get_secret_service)]
+from app.services.audit import AuditService
+
```
DEENUU1, line 723: This import is at the bottom of the file, wedged between `SecretSvc` and the new factory, rather than in the import block at the top.

`.claude/rules/code-style.md` asks for imports strictly ordered in one block at the top, and every other service in this file follows that. Worth fixing because **the linter will not tell you**: I checked, `ruff check app/api/deps.py` is clean — E402 is not in the selected rules — so this survives `make lint` and the next person adding a service copies the nearest example.

Move it up with the other `app.services` imports and leave the factory here:

```suggestion
```

While you are there, the docstring is the generated-template one:

```python
def get_audit_service(db: DBSession) -> AuditService:
    """Create AuditService instance with database session."""
```

Every other factory in this file either says something specific or says nothing. This one restates the signature, and the reference docs are built from these.

DEENUU1 (commented):
The pagination bug you found on the way through is worth more than the layering fix it came with. `limit` applied in SQL across every source with that id and *then* filtered in Python meant a short page whenever another collection's history interleaved, `total` describing the survivors rather than the source, and no way to page past the gap — that is a real defect nobody had reported, and resolving the source first is the right shape. `list_logs` having one home instead of two copies of the same twelve-field mapping is the other good half.

Two things below. The first is small and the linter cannot see it; the second is a note rather than a change.

I also traced `get_source` since the new code trusts it: it is unscoped by organization, which is fine here because both callers gate before it — the org route through `CollectionAccessService`, this one through the `collection_name` comparison — and your `list_logs` docstring says exactly that. No complaint, just confirming it was checked rather than assumed.

DEENUU1:
One more, and it is a note rather than a change to this branch.

The new refusal in `list_kb_sync_source_logs` decides "is this source this KB's?" by comparing `source.collection_name` against `kb.collection_name`. That is the same basis the old Python filter used, so this branch does not make anything worse — but it is worth writing down that the basis itself is only as strong as **#367**, which is open and `severity:critical`: `POST /kb` takes an explicit `collection_name` and never calls `CollectionAccessService.claim`, so a member of one organization can point a knowledge base at another's collection name deliberately. `collection_name` carries no unique constraint; the collision-resistance comes from the random suffix in `_derive_collection_name`, which only applies when the caller leaves the name alone.

So: same exposure as before, tied to #367, and closing #367 is what makes this refusal mean what its docstring says. Recording it here so whoever picks up #367 knows this call site depends on it.

DEENUU1:
Correction on my inline comment about the `AuditService` import — I was wrong on both halves of it, so ignore it.

I checked whether ruff flags the placement and stopped there, without checking what `deps.py` already does. It has around thirty `from app.services` imports scattered down the file — lines 88, 103, 112, 135, 279, 630, 687, 695, 705 — each sitting directly above the factory that uses it. Your placement is the file's convention, not a departure from it.

Same for the docstring: `"""Create XService instance with database session."""` is what every factory in that file says. Copying it is right.

Nothing to change here. The pagination fix and `list_logs` stand as the substance of the branch.

PR #327, opened 2026-08-06
ci: create backend/.env in make install

### LINKED ISSUE #299: make check fails in a fresh checkout: nothing creates backend/.env

OchnikBartek:
## What this changes

`make install` creates `backend/.env` from `backend/.env.example` when there is
none, and never touches it again.

Before this, `make check` in a fresh checkout got through `lint` and `test` and then
stopped at `db-check` with a psycopg2 traceback ending in `fe_sendauth: no password
supplied` — because `POSTGRES_PASSWORD` defaults to empty and nothing in the
repository had ever created the file that holds the real value. Everything running
on the host reads it through `app.core.config`: `db-check`, `db-upgrade`, `run`, and
pytest. Four minutes in, and it took `test-frontend-cov`, `build-frontend`,
`docs-build` and `audit` with it — four targets that need no database at all.

## Why this way

**Copied from the example, not generated.** One definition of the development
defaults, and they already match `docker-compose.yml`'s (`postgres` / `postgres` /
`agenticos`), so a seeded file and the Docker stack agree.

**Never overwritten.** `docs/install.md` calls every step idempotent, and this is
the one where idempotent and destructive are one character apart: the file that
exists holds somebody's provider keys and their generated `SANDBOXD_TOKEN`. The
guard is what the second new test pins, and it is the more important of the two.

**The example needed one fix before it was safe to copy at all**, and this is the
part worth a reviewer's attention. It set `REDIS_PASSWORD=change-me-in-production`
in its Traefik/production section — a *second*, live definition of a key already
present and commented out under `# === Redis ===`. `docker-compose.yml` feeds
`backend/.env` to the app container via `env_file`, the local redis runs without
`requirepass`, and `config.py` puts the password in the URL whenever it is set. So
copying the example verbatim would have broken every dev request with `Client sent
AUTH, but no password is set`. It is now one commented-out definition, which also
means the deployed stacks — which pass `--requirepass ${REDIS_PASSWORD:?…}` — refuse
to start until somebody sets a real password instead of inheriting a placeholder
from an example file. Anybody who followed `docs/install.md:38` or `docs/deploy.md:20`
and copied it by hand was hitting this already.

## Checks

      only the `Makefile` fails both new tests in
      `TestInstallPreparesWhatTheChecksNeed`
- [x] `make check` passes — with the one exception below, which is this machine
- [x] Platform-layer coverage is still 100% (3188 passed)
- [ ] n/a — no migration

Verified from nothing, in a worktree with no `backend/.env`:

1. `make db-check` → `fe_sendauth: no password supplied`.
2. `make install` → `▶ Created backend/.env from backend/.env.example`; the file has
   `POSTGRES_PASSWORD=postgres` and no live `REDIS_PASSWORD`.
3. `make install` again → `backend/.env already exists - leaving it alone`.
4. `db-check` against a database created for the purpose → `No new upgrade
   operations detected`.

`make lint`, `make test`, `make test-frontend-cov` (3175 passed), `make
build-frontend`, `make docs-build` and `make audit` are green.

## Notes for the reviewer

**Depends on #300** (issue #227). This branch is cut from `ci/bun-install-in-setup`,
not from `main`, because both edit the `install` recipe and cutting from `main` would
guarantee a conflict. Merge #300 first.

**`db-check` still fails on my machine, for a third reason that is neither this nor
#288**: the shared development database is at `0011_drop_expired_approval`, a
revision from another branch in flight, so alembic cannot locate it from a branch
that does not have that migration. A database migrated by somebody else's branch is
not something a setup target can fix — worth remembering when #288 (runtime `rag_*`
tables) is looked at, since "run `db-check` against a database this branch built" would
answer all three.

Docs: `docs/install.md` (both the Docker note and the host section),
`docs/commands.md` (the `install` row) and `docs/configuration.md` — whose "Getting
Started" told you to run `cp .env.example .env` from `backend/`, which is now done for
you, and which pointed at the wrong path anyway. That section now names the two
placeholder secrets (`SECRET_KEY`, `VAULT_MASTER_KEY`) instead, since generating those
is the part a person still has to do.

Closes #299. Refs #227, #288.


DEENUU1 (commented):
Third of the three things a fresh checkout is missing, and the guarded copy is the right shape — idempotent where idempotent and destructive are one character apart, as your own test docstring puts it. The `REDIS_PASSWORD` cleanup is a good catch on the way past: commented out in both directions, so dev does not send AUTH to a redis with no password and prod refuses to start rather than inheriting a placeholder from an example file.

Two things I checked before worrying about them, since "`make install` now writes a credentials file by itself" is the obvious thing to be nervous about. Both fine:

- **the default `SECRET_KEY` cannot reach production** — `backend/app/core/config.py:84` refuses it outright when `ENVIRONMENT=production`, same for `API_KEY` at `:111`. Auto-copying the example does not make shipping a known key easier.
- **CI is unaffected** — no job runs `make install`, and real environment variables beat a `.env` in pydantic-settings anyway.

Two things below. Neither is in the diff, so they are here rather than inline.

---

### `backend/.env.example:96` — the empty `SANDBOXD_TOKEN=` and `make dev` disagree once this copy is automatic

`sandbox-token` greps for `^SANDBOXD_TOKEN=.` — a value with at least one character. An empty assignment does not match, so it **appends a second one**, and every fresh checkout ends up with the key twice:

```
SANDBOXD_TOKEN=            # line 96, from the example
...
SANDBOXD_TOKEN=hR3k…       # appended by make dev
```

It works — I checked rather than assumed: `dotenv_values` on a duplicate resolves to the last, and compose's `env_file` does the same. So this is untidy rather than broken. But it is untidy in a credentials file, it works only by last-wins, and anything reading it first-wins gets an empty token and a sandbox that refuses every session.

Dropping the empty assignment and keeping the comment above it fixes both: `sandbox-token` then appends the only copy, which is what it was written for.

Nobody hit this before the branch — with no `.env` at all, `make dev` created the file containing just the token. Now the file always exists, so the append always lands beside an empty one.

### `Makefile:180` — this forked from #300 before my change there, so it carries the old step order

This branch left #300 at `02e1de3`, ahead of the commit I pushed there this afternoon, so it still has `bun install` in front of the pre-commit block. That ordering is the one thing I asked to change on #300: make stops at the first failing line, so a machine where the frontend install fails ends up with no git hooks — including the one that refuses a commit on `main` — behind an error message about bun.

#300 lands first, so this wants a rebase. The order that should come out is `.env` (a local `cp`, nothing to fail on), `uv sync`, the hooks, then `bun install` last. I'll do that as part of the merge rather than bouncing it back to you.

DEENUU1:
Rebased onto `main` (which now has #300) and merged both review points in rather than bouncing them back.

- Step order is now `.env` → `uv sync` → hooks → `bun install`. The `.env` copy is a local `cp` with nothing to fail on, so it goes first; `bun install` is the one most likely to fail on a given machine, so it goes last and a partial failure still leaves the git hooks installed.
- Dropped the empty `SANDBOXD_TOKEN=` from `.env.example`, keeping the comment. Verified end to end on a scratch copy: create the file from the example, run the `sandbox-token` step, and you get **one** `SANDBOXD_TOKEN` line instead of two.
- `docs/commands.md` keeps main's corrected `quickstart` row alongside your `install` row; `docs/install.md` keeps your three-part version plus the reloader note that landed in 0.0.41.

`test_ci_parity.py` green, 19 tests.

PR #343, opened 2026-08-06
ci(lint): skip a nested checkout rather than a directory name

### LINKED ISSUE #225: check_backticks.py walks into git worktrees and fails make lint
`scripts/check_backticks.py` skips itself but not its copies, so a checkout
that has git worktrees under `.claude/worktrees/` gets three false positives
from the script's own error-message string and `make lint` exits 1.

## What breaks

```
$ make lint
python3 scripts/check_backticks.py
.claude/worktrees/issue-45-review-e89fb6/scripts/check_backticks.py:177: ...
.claude/worktrees/issue-61-planning-bea4f4/scripts/check_backticks.py:177: ...
.claude/worktrees/task-dependencies-order-7227f1/scripts/check_backticks.py:170: ...

3 line(s) use ``double backticks``, which render literally. Run with --fix.
make: *** [lint] Error 1
```

Each hit is line 170/177 of the *copy* of this script in another branch's
worktree — the f-string that reports the error, which necessarily contains
double backticks.

## Why

`scripts/check_backticks.py:44` sets `SELF = Path(__file__).resolve()` and
excludes that one absolute path. `SKIP_DIRS` (`:47`) lists `.git`, `.next`,
`.venv`, `__pycache__`, `node_modules`, `htmlcov`, `dist`, `build`,
`.pytest_cache`, `.ruff_cache` — no `.claude`, and nothing that skips a
nested checkout. Excluding one path rather than the pattern means every
copy of the script is a hit.

Worse than the false positive: a worktree is a full checkout, so the script
also scans another branch's markdown and TypeScript. A `make lint` here can
go red because of a file that is not on this branch at all.

## Reproducing

```bash
git worktree add .claude/worktrees/scratch -b throwaway
make lint
```

`make lint` is green with worktrees present. Verified locally by moving
`.claude/worktrees/` aside — `python3 scripts/check_backticks.py` then
reports "No double backticks in Markdown, TypeScript comments or Python
files" and exits 0 — so the branch content itself is clean and the
worktrees are the only cause.

Adding `.claude` to `SKIP_DIRS` fixes the reported case; skipping any
directory containing a `.git` entry that is not the repository root would
also cover worktrees placed elsewhere.

## Scope

Pre-existing and invisible to CI, which has no worktrees. Found while
running `make check` on `feat/dashboard-page-demo` (#149); that branch did
not cause it and does not fix it.

OchnikBartek:
## What this changes

`scripts/check_backticks.py` no longer decides where to walk by directory name. It
skips a directory that is **another checkout** — a git worktree carries a `.git` file,
a clone a `.git` directory — and walks everything else.

Read the issue's symptom first, because it is already gone: `worktrees` was added to
`SKIP_DIRS` at some point after #225 was filed, so `make lint` is green in a checkout
with worktrees today. What is left is the trade that made:

- a `docs/worktrees/` that is only a directory with a name is **dropped from the
  scan**, and nothing says so — a guard reporting "no double backticks" about files it
  never opened is the failure mode this script exists to prevent;
- a worktree placed anywhere else is still walked, which is the original bug;
- and none of it was pinned by a test, so the fix was a string in a set with a comment
  as its only protection.

## Why this way

**Detected, not named.** `.claude/worktrees/` is where this repository happens to put
them; `git worktree add scratch` puts one at the root. The property that matters is
"these files belong to another branch", and the `.git` entry is exactly that property.

**`os.walk` with in-place pruning of `dirnames`, not `rglob`.** Skipping a checkout
means not descending into it, rather than filtering its files out one at a time —
which is also why pointing the script at a tree holding thirteen worktrees takes 0.4s.

**The self-exemption is the file's name now, not one absolute path.** Every copy of
this script contains the pattern it detects, in the error message it prints; the issue
names this as the reason each copy was a finding. Pointing the script directly at a
copy used to report it, and `--fix` would have rewritten the regex it exists to run.

## Checks

      only the script fails `test_a_directory_merely_called_worktrees_is_still_walked`
      and the exemption test
- [x] `make check` passes
- [x] Platform-layer coverage is still 100% (3221 passed)
- [ ] n/a — no migration

`backend/tests/test_check_backticks.py` asserts **both** halves, because either alone
looks like success: a worktree's copy is not walked, and a directory merely called
`worktrees` still is. The first passes on `main` too — it is there to say what must not
regress, since that is the behaviour somebody tidying `SKIP_DIRS` would take out.

Verified against the checkout #225 was filed from — thirteen worktrees present:
`python3 scripts/check_backticks.py` is green, walks 1365 files, and none of them is
under a worktree. `make lint`, `make test`, `make test-frontend-cov` (3175 passed),
`make build-frontend`, `make docs-build` and `make audit` all green.

## Notes for the reviewer

One thing the test file has to do, and it looks odd until you see why: it builds its
offending line from `"`" * 2` instead of quoting a span. A literal one there is a
finding in `make lint`, for exactly the reason the script exempts itself — the test
that proves the guard works cannot be written in prose the guard rejects.

No docs page describes the guard's traversal (it appears as "the two guards" in
`make lint`'s description and as a pre-commit hook), so nothing under `docs/` is owed;
`scripts/docs_drift.py` agrees.

Closes #225.

### DIFF
--- a/scripts/check_backticks.py
+++ b/scripts/check_backticks.py
@@ -14,7 +14,9 @@ untouched either way. Whole lines are scanned, string literals included: a
 runtime string holding a double-backtick span ends up in a chat message or a
 tool description, which is Markdown territory again.
 
-This file exempts itself — it has to be able to name the pattern it detects.
+This file exempts itself — it has to be able to name the pattern it detects — and
+skips any nested checkout, because a git worktree holds another branch's files plus
+a copy of this script, whose error message spells the pattern out.
 
 Usage::
 
@@ -27,6 +29,7 @@ Exits 1 when anything is found (so it can gate CI), 0 when clean.
 from __future__ import annotations
 
 import argparse
+import os
 import re
 import sys
 from collections.abc import Iterator
@@ -41,20 +44,15 @@ TYPESCRIPT_SUFFIXES = {".ts", ".tsx", ".js", ".jsx"}
 PYTHON_SUFFIXES = {".py"}
 
 # This script's own docstring, regex and messages must be able to spell the
-# pattern out.
-SELF = Path(__file__).resolve()
+# pattern out - and so must every copy of it. Matched by name rather than by
+# absolute path: a copy under a worktree is a different path and was reported as
+# three findings on line 170 of a file nobody had edited.
+SELF = Path(__file__).name
 
 SKIP_DIRS = {
     ".git",
     ".next",
     ".venv",
-    # A git worktree is another checkout of this repository, so scanning one
-    # reports every finding twice - and `SELF` only excludes *this* copy of this
-    # script, not the copies of it living under each worktree, each of which
-    # contains the literal double backticks in its own error message. So with any
-    # worktree checked out the hook fails on itself and nothing else can be
-    # committed. Found the hard way, with six worktrees open at once.
-    "worktrees",
     "__pycache__",
     "node_modules",
     "htmlcov",
@@ -102,7 +100,7 @@ def typescript_comments(text: str) -> Iterator[tuple[int, str]]:
 
 def scan(path: Path) -> list[tuple[int, str]]:
     """Every offending line in one file, as (line number, line)."""
-    if path.resolve() == SELF:
+    if path.name == SELF:
         return []
     if path.suffix in MARKDOWN_SUFFIXES:
         lines = markdown_prose(path.read_text(encoding="utf-8"))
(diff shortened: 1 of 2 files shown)


DEENUU1 (commented):
Yes — "do not descend into another checkout" is what the rule always meant, and detecting it beats naming it. The `docs/worktrees/` case in the tests is the one that would have bitten later and quietly: an exclusion by directory name makes the guard read less of the tree than it reports on, which is the same class of defect as the thing it is guarding.

I checked it against the real thing rather than trusting the unit tests: `walk()` over my main checkout with twelve worktrees open walks 1311 files and **zero** under `.claude/worktrees/`. `pass_filenames: false` on the hook means it walks the tree rather than a staged list, so this is the path #225 was actually failing on. And `fix()` goes through `scan()`, so a copy of the script is safe from `--fix` too, which is what the docstring claims.

One behaviour change worth a line below. Nothing blocking.

scripts/check_backticks.py
```diff
@@ -130,18 +128,41 @@ def fix(path: Path) -> int:
...
+    way the files under it belong to some other branch. Scanning one reports findings
+    twice, reports them against paths that are not on this branch, and - since a copy
+    of this script lives there too - fails on the error message it is about to print.
+
+    Detected rather than named: `.claude/worktrees/` is where this repository puts
+    them, but skipping every directory called `worktrees` both misses one placed
+    anywhere else and silently stops reading a `docs/worktrees/` that is only a
+    directory with a name.
+    """
+    return (directory / ".git").exists()
```
DEENUU1, line 144: Small thing that comes with the move off `rglob`: this `stat`s a directory before `os.walk` has a chance to handle it, so an unreadable one takes the whole run down with a traceback.

```
PermissionError: [Errno 13] Permission denied: 'docs/locked/.git'
```

I checked what the old code did with the same tree, because "it got worse" and "it got louder" are different verdicts: `rglob` skipped the unreadable directory **silently** and reported on the rest. So this is strictly the better half of the trade — the guard no longer claims a clean tree it could not fully read.

It is still a raw traceback out of a pre-commit hook, and the reader cannot tell it from a bug in the script. One line:

```python
    try:
        return (directory / ".git").exists()
    except OSError:
        # Unreadable: not a checkout we can descend into either way, and the
        # walk below reports it rather than dying on it.
        return True
```

with `os.walk(root, onerror=...)` printing what it could not read. Genuinely optional — a permission-denied directory inside a checkout is rare, and today's outcome is the safe one. Mentioning it only because a silent partial scan is the exact failure this file exists to prevent, and the reasoning is worth having written down either way.

Issue #353, opened 2026-08-06
A write is answered before its transaction commits, so the next read can miss it


DEENUU1:
**Severity: high** · pre-existing, found diagnosing #335

## What breaks

Every write in this API is acknowledged **before its transaction commits**, so a
client that acts on a 2xx — the way every client does — can issue its next
request against a database the write has not been applied to. It is not a
theory and it is not about concurrency: a single sequential client, one
connection, one worker, reproduces it.

Measured against a local backend, straight at `:8000` with no Next in the way:

```
an acceptance answered 204, and the membership row it created was
invisible to the very next read for 21.7ms
```

## Why

`get_db_session` is a dependency with `yield`, and the commit is in its exit
code (`backend/app/db/session.py:34-43`):

```python
async with factory() as session:
    try:
        yield session
        await session.commit()      # the exit code
```

FastAPI unwinds that exit stack **after the response has been written to the
socket**. From `fastapi/routing.py` (`request_response`, pinned 0.141.1):

```python
async with AsyncExitStack() as request_stack:      # dependencies with yield land here
    scope["fastapi_inner_astack"] = request_stack
    async with AsyncExitStack() as function_stack:
        response = await f(request)
    await response(scope, receive, send)           # the client is answered HERE
    response_awaited = True
# request_stack unwinds HERE: `await session.commit()` runs now
```

`get_request_handler` puts every `Depends` on `fastapi_inner_astack`
(`routing.py:477`), which is the outer of the two — the inner `function_stack`
does unwind before the send, but nothing in FastAPI's public surface puts a
dependency on it.

A minimal app on this repo's own pinned FastAPI, with a 200ms sleep standing in
for the commit so the ordering is unambiguous:

```
     0.0ms  client got 204
   256.4ms  commit finished
```

## Seen in the product's own SQL log

`DB_ECHO=1`, one client, requests strictly sequential. uvicorn serves this on
one event loop, so a second request can only be picked up while the first is
suspended — and between `await response(...)` and the commit is exactly such a
suspension. Two of the nine overlaps in one probe run:

```
DELETE /api/v1/orgs/<org>/members/<user>
  BEGIN 15:33:56.538, still uncommitted when
  POST /api/v1/orgs/<org>/invitations arrived at 15:33:56.548

POST /api/v1/orgs/<org>/invitations
  BEGIN 15:34:00.257, still uncommitted when
  POST /api/v1/invitations/<token>/accept arrived at 15:34:00.291
```

[...]

Issue #346, opened 2026-08-06
Two concurrent make check runs fight over one migrations database


DEENUU1:
**Severity: medium** · found while running `make check` on PR #334 (#308), 2026-08-06

## What's wrong

`tests/test_migrations.py` cycles the whole Alembic chain against a database whose
name is a **constant**:

```python
# backend/tests/test_migrations.py:26
MIGRATION_DATABASE = "agenticos_migrations_test"
```

`make check` runs `make test`, which runs `pytest tests/` — so that file runs on
every `make check`. Two concurrent runs on one machine therefore point at one
database and drop each other's schema mid-chain. The loser fails with an error
that names neither cause nor culprit, on a branch that touched no model and no
migration.

This is **the same defect as #189, one file over**, and `.claude/rules/testing.md`
already records the fix for the other half:

> a constant name meant two runs on one machine dropping each other's tables
> mid-test (#189)

`tests/conftest.py` points `POSTGRES_DB` at `<base>_p<pid>` for exactly this
reason. `test_migrations.py` forces its own name instead, and forcing it is
deliberate and correct — a default is something an environment can override —
but the name it forces is shared.

## How it reproduces

1. Two checkouts (or two worktrees) of this repository on one machine, against
   one Postgres.
2. `make check` in both, overlapping.
3. One of them fails in `tests/test_migrations.py`. Which one depends on timing.

Observed today with several worktrees in flight: the failure appeared on two
consecutive runs of a compose-only branch, passed in isolation, and passed in CI.

## Why it is worth fixing rather than documenting

CI is safe — each job gets its own Postgres — so this only ever bites locally,
which is precisely where it is most expensive: the run takes about five minutes,
the failure looks like a real migration regression, and the branch it appears on
has nothing to do with migrations. The cost is paid in investigation, twice, by
whoever happens to be second.

It also erodes the claim `make check` exists to make. A gate that fails for
reasons unrelated to the diff trains people to re-run it rather than read it,
which is the habit #143 and #165 are both about.

## The fix

Scope the name the way the unit suite already does — `agenticos_migrations_test_p<pid>`
— and keep it forced rather than defaulted. The session-scoped create/drop in
`tests/integration/conftest.py` is the shape to copy, including its refusal of any
database whose name contains neither `test` nor `ci`, since this file also calls
`drop_all` unconditionally.

Worth checking in the same pass whether any other test forces a constant database
name.

## How you would know it was fixed

[...]

Issue #311, opened 2026-08-06
The ai-review reviewer has produced nothing since 2026-08-05, and reports success


DEENUU1:
**Severity: high** · found on PR #309, 2026-08-06

## What's wrong

The `ai-review` reviewer has produced no review since **2026-08-05 evening**, and
fails in a way that reads as a normal outcome. Every run since then posts:

> No review: the reviewer did not produce a result. See the workflow run.

The workflow itself concludes **success** — all three jobs (`Resolve pull request
context`, `Run the reviewer`, `Publish the review`) are green — so nothing on the
pull request page says the reviewer is broken rather than merely quiet. The check
is advisory and not required, which is correct, but it means the only signal is a
sentence in a comment that looks like it could be about the diff.

## When it started

Last working review: **PR #265**, merged 2026-08-05 17:43 UTC.
First failure: **PR #268**, merged 2026-08-05 18:33 UTC.

| PR | merged | result |
|---|---|---|
| #240 | 08-05 15:14 | real review |
| #260 | 08-05 16:21 | real review |
| #265 | 08-05 17:43 | real review |
| #268 | 08-05 18:33 | no result |
| #270, #272, #274, #275 | 08-05 evening | no result |
| #289, #291, #294 | 08-06 | no result |
| #309 | 08-06 | no result, twice (re-labelled) |

So **eleven consecutive pull requests have merged with no automated review**,
including three releases, while the process in `CLAUDE.md` treats the reviewer as
the thing you ask when you believe a branch is finished.

## Where it fails

`.github/workflows/ai-review.yml`, the `Review the diff` step. The Codex CLI dies
about 12 seconds in:

```
/home/runner/work/_actions/openai/codex-action/…/dist/main.js:23546
  reject(new Error(`${program2} exited with code ${code}`));
Error: codex exited with code 1
##[error]Process completed with exit code 1.
```

Twelve seconds is too fast to be the model working and failing — it is the shape
of an authentication, quota or entitlement refusal at the first API call. The
step's own output is not surfaced beyond the Node stack trace, so the actual
reason is not in the log we can read. The `Normalize the result` step then
catches the missing artefact and writes the generic sentence.

Worth checking in order: the OpenAI credential's validity and quota, whether the
`openai/codex-action` pin (`52fe01e`) still matches the CLI's expectations, and
whether anything changed on 2026-08-05 evening between #265 and #268.

## Why severity:high

Not because a build is red — nothing is red. Because the repository's stated
[...]

Issue #230, opened 2026-08-05
A stored secret can stay off the vault list forever: the invalidation refetch is answered with the pre-write list


DEENUU1:
Store a secret and the vault sometimes goes on showing the list without it — indefinitely, until something reloads the page. The row is in the database, the API answers a list containing it, and the page keeps rendering the one that does not.

This is the root cause of [#132](https://github.com/vstorm-co/agenticos/issues/132), found while fixing it. #132's PR stops the E2E seed from depending on the buggy behaviour; **it does not fix this.**

## What was measured

Reproduces locally about once in eight runs of `frontend/e2e/seed.setup.ts` — against a `next start` standalone build, a host uvicorn and a `pgvector/pgvector:pg16` container, one Playwright worker.

The failing run's screenshot shows the vault reading **"4 keys stored"**, dialog closed, and `e2e-capability-token` absent. From its trace, the four `/api/secrets` exchanges on that page:

| browser time | | status | body |
|---|---|---|---|
| 24.219 | `GET /api/secrets` | 200 | sha1 `2830db9d…`, 1479 bytes, 4 items |
| 24.338 | `GET /api/secrets` | 200 | **same sha1** |
| 24.783 | `POST /api/secrets` | **201** | `e2e-capability-token`, `created_at 24.787058` |
| 24.798 | `GET /api/secrets` | 200 | **same sha1 again — the 4-item list** |

The GET at 24.798 is the refetch TanStack fires from `useSecrets`' `invalidateQueries`, awaited inside the mutation's `onSuccess`. It answered the pre-write list, so the dialog closed onto a list missing the row it had just made, and nothing refetches again — no timeout is long enough.

## Where it is not

Three places ruled out by measurement rather than by reading:

- **Not the commit ordering.** The backend's own SQL log for that exact window: `POST /api/v1/secrets` at `.786`, `BEGIN` `.787`, `COMMIT` `.797`; then `GET /api/v1/secrets` at `.801`, `BEGIN` `.802`. The read's transaction opened 5ms after the write committed, so under `READ COMMITTED` it had to see the row.
- **Not a backend read-after-write.** POST-then-GET against `127.0.0.1:8000` directly: **0/60** on one connection, **0/80** with the read on a second connection, **0/200** from eight concurrent workers.
- **Not Next's proxy.** The same hammer through `localhost:3100/api/secrets` with the browser's own cookies, four concurrent workers: **0/100**.

[...]

PR #355, opened 2026-08-06
fix(e2e): poll for the colleague's membership, and say what was seen

### LINKED ISSUE #335: e2e: the seed's membership step asserts once and takes the whole suite down when it loses
**Severity: medium** · pre-existing, found while watching CI on PR #333 (#310)

## What breaks

The `[seed]` project's last step fails intermittently, and because it is a
project dependency the whole Playwright suite is reported red having run **no
product spec at all**:

```
✘  8 [seed] › e2e/seed.setup.ts:244:6 › a colleague is a member of the organization (2.8s)
   Error: the colleague accepted the invitation but is not a member
1 failed
87 did not run
A FIXTURE FAILED, SO THE SUITE DID NOT TEST THE PRODUCT.
```

Seen twice within an hour, on two different heads, so it is not one branch's
doing:

| Run | Branch | When |
|---|---|---|
| [31099267560](https://github.com/vstorm-co/agenticos/actions/runs/31099267560) | `main` | 11:58 UTC, 2026-08-06 |
| [31103106471](https://github.com/vstorm-co/agenticos/actions/runs/31103106471) | `fix/prefect-runner-has-its-own-health-probe` (#333) | 12:52 UTC, 2026-08-06 |

Neither branch touches invitations, membership or the frontend. PR #333 is
compose and Dockerfile only; it **found** this, it did not cause it.

## The sequence

`frontend/e2e/seed.setup.ts:244-265`:

1. the owner sends an invitation from the members page and keeps the token;
2. `acceptAsColleague` signs the colleague in, opens `/invitations/<token>`,
   clicks **Accept invitation** and waits for `You joined the organization!` —
   so the server has answered the acceptance before anything else happens;
3. the owner reads `GET /api/orgs/<id>/members` **once** and asserts the
   colleague's email is in it.

Step 3 is the one that fails.

## Where

- `frontend/e2e/seed.setup.ts:257-264` — the single read.
- `frontend/e2e/seed.setup.ts:285-305` — `nowThere`, the polling helper every
  *other* step in this file uses, added precisely because "a read straight after
  a write can be answered with the pre-write list (#230)". This step is the only
  one that does not use it.

Two candidates, and the log does not settle which:

- **The fixture.** A single read where the rest of the file polls. If the
  membership row lands a beat after the acceptance response, this is a fixture
  asserting too early and the fix is `nowThere`.
- **The product.** If the acceptance page can say `You joined the organization!`
  before the membership is committed and visible to the inviting owner — or if
  the `/api/orgs/<id>/members` BFF route serves a cached response to the second
  reader — then polling would only paper over it. Worth confirming which before
  changing the spec: this is the same family as #230, whose whole point is that
  the stale read is the product's and not the test's.

The `[seed]` project passes on 20 consecutive runs, and — if the cause turns out
to be the product — a backend or BFF test covers "a member accepted; the next
read by the inviting owner lists them", so a regression fails somewhere cheaper
than the e2e suite.

## Related

Cluster: **e2e reliability** — #154 (sharing and skills specs fail when a
mutation's refetch lands stale), #162, #130, and the product bug behind them,
#230. Closest to #154: same "a read after a write was answered stale" shape,
different step, and this one is worse because it is in a fixture project and
therefore takes all 87 specs down with it. Not a duplicate of #154, which names
two product specs.

DEENUU1:
The `[seed]` project's last step read `/api/orgs/<id>/members` once, right after
the colleague's acceptance was acknowledged, and asserted the row was in it. When
it lost, Playwright skipped all 87 product specs behind the fixture project and
the job read like a product regression. It now polls, like its four siblings.

## The diagnosis, because the next person to see a red `e2e` will read this

#335 asked which of two things this was. **It is both, and they compose.**

### The proximate cause is a missed conversion

PR #222 introduced `nowThere` and converted four call sites in `seed.setup.ts` —
skills, knowledge bases, secrets, MCP connections. `gh pr diff 222` shows the
membership step appearing only as context: it was never converted, and it went on
reading once. That is why this one step, out of eight, is the one that goes red.

### The reason a single read can lose is a product defect — filed as #353

Writes here are **acknowledged before they commit.** `get_db_session` puts the
commit in a `Depends`-with-`yield` exit code
(`backend/app/db/session.py:34-43`), and FastAPI unwinds that stack *after* the
response has been written to the socket. From `fastapi/routing.py`
(`request_response`, pinned 0.141.1):

```python
async with AsyncExitStack() as request_stack:      # dependencies with yield land here
    scope["fastapi_inner_astack"] = request_stack
    async with AsyncExitStack() as function_stack:
        response = await f(request)
    await response(scope, receive, send)           # the client is answered HERE
    response_awaited = True
# request_stack unwinds HERE: `await session.commit()` runs now
```

`get_request_handler` puts every `Depends` on `fastapi_inner_astack`
(`routing.py:477`) — the outer of the two.

**Evidence, three kinds:**

1. A minimal app on this repo's own pinned FastAPI, with a 200ms sleep standing in
   for the commit: `0.0ms client got 204` / `256.4ms commit finished`.
2. Against the real backend on a local host uvicorn, straight at `:8000` with no
   Next in the way: an acceptance answered 204 whose membership row was **invisible
   to the very next read for 21.7ms** — 1 miss in 45 acceptances.
3. The product's own SQL log with `DB_ECHO`, one sequential client. uvicorn runs
   one event loop, so a second request can only be picked up while the first is
   suspended, and between `await response(...)` and the commit is such a
   suspension. Nine requests in one probe run arrived there. Two of them:

```
DELETE /api/v1/orgs/<org>/members/<user>
  BEGIN 15:33:56.538, still uncommitted when
  POST /api/v1/orgs/<org>/invitations arrived at 15:33:56.548

POST /api/v1/orgs/<org>/invitations
  BEGIN 15:34:00.257, still uncommitted when
  POST /api/v1/invitations/<token>/accept arrived at 15:34:00.291
```

The client had been answered, had the invitation token in its hand, and was
spending it 34ms before the transaction that created it committed. Those two are
ordinary product calls, not test scaffolding.

So the state the fixture reads is genuinely eventually consistent — by tens of
milliseconds, which is wide enough for a loaded two-core runner to lose. Polling
here is not papering over a broken write: the colleague *does* become a member.
What is broken is the acknowledgement contract, and a fixture project that takes
87 specs down is the worst possible place to discover it. **#353 is that defect,
filed with the evidence above and a "how you would know it was fixed" that puts a
regression test somewhere cheaper.** #353 is also a strong candidate root cause for
**#230** and therefore **#154** — #230 ruled the ordering out on one trace where
the commit won by 5ms, which is what most traces look like at 1-in-8.

### What it is not

- **Not another branch's doing.** Three sightings on three unrelated heads on
  2026-08-06: run 31099267560 on `main`, 31103106471 on #333, and again on #323.
  None touch invitations, membership or the frontend. #322, #326 and #334 went
  green through the same job, so it is intermittent rather than broken.
- **Not #346's shape.** A shared unscoped test database is a proven failure mode
  here, but each CI job gets its own Postgres service container and its own
  database, and `playwright.config.ts` runs `workers: 1` with `retries: 0`. Two
  runs cannot reach each other. Ruled out rather than assumed.
- **Not a stale read in the BFF.** `backendFetch` sets no `cache`, Next 16 defaults
  fetch to `no-store`, the route handler reads cookies so it is dynamic, and
  `page.request` is Playwright's own context, which implements no HTTP cache. The
  failing read reached the backend and the backend answered honestly.

## The change

`nowThere` polls the *list* and matches on it rather than polling a boolean, so a
failure prints what was actually there instead of only that it lost:

```
the write was accepted, but /api/orgs/<id>/members never listed a row
whose email is colleague@example.com
Received array: ["admin@example.com"]
```

`alreadyThere` and `nowThere` now share `valuesAt`, so the two checks read a list
the same way. Both docstrings, the file's header, `.claude/rules/testing.md`, the
`e2e-tests` skill and `docs/testing.md` now carry the half of the rule that was
missing: a fixture asserts through the API **and, after a write, polls it**,
because a 2xx means the request was answered and not that the write is readable.
The `alreadyThere` guard each step opens with stays a single read — it runs
*before* the write, where a stale answer only costs doing the work twice.

## The siblings, checked

- Four steps already poll (`nowThere`). Unchanged.
- `ensurePendingInvitation` revokes a leftover invitation and immediately posts a
  new one. That is the same #353 window in the delete direction — and it is exactly
  the overlap in the SQL log above. It can only bite a re-run against a dirty
  database, never CI, which starts fresh. Left alone; noted here.
- `a draft agent exists` still proves itself with a page heading rather than
  through the API, which the file's header does not quite admit. It auto-waits, so
  it is not the shape that failed, and taking the wait out risks the next step's
  navigation. Left alone; worth a look by whoever is next in this file.
- `registerColleague` and `revokeAnyPendingInvitation` assert the write's own
  response, which is right.

## Verified

- **`[seed]`, 31 consecutive local runs, 31 passed** — 20 in one go, which is the
  bar #335 set, then 8 after the `valuesAt` refactor and 3 after the review pass
  below. Against a host uvicorn, a `pgvector/pgvector:pg16` container and a
  `start:standalone` build, deleting the colleague's membership and invitations
  between runs so every run took the full create path.
- **The new assertion is known to be able to fail.** Pointed at an address that
  could never be a member, it failed in 27.3s with the message and `Received
  array` shown above.
- CI green on `7d0a502`, `e2e` included, before the review pass.
- **`make check` — "All checks passed — every CI job except e2e."**

## Not done, and one thing to know

- The diff has **not** been through an automated reviewer: `ai-review` no longer
  runs on a pull request (PR #313 removed the trigger, #311). I reviewed it as a
  maintainer would instead, which found three things and produced the last
  commit: three prose copies had grown a universal ("every check is a poll") that
  a `grep` disproves — `alreadyThere` is a single read by design and
  `a draft agent exists` does not use the API at all, so both are now named where
  the rule is; the FastAPI claim is pinned to the version it was read at, since
  `fastapi>=0.135.3` is a floor and that teardown has moved before; and `members`
  became `membersPath`, because it holds a URL.
- If `tests/test_migrations.py` is red locally it is **#346**, not this branch — a
  constant database name and several worktrees on one Postgres. It passes in
  isolation and in CI, where each job gets its own database.

Closes #335
Refs #353, #222, #230, #154

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -71,7 +71,10 @@ Traps, each of which has cost a red job here:
   `element(s) not found` for a refusal it never looked at (#132). And a **fixture** step
   asserts through the API, never on the row appearing - the refetch after a write is
   sometimes answered with the pre-write list (#230), which is a product bug and must not
-  be reported as a broken fixture.
+  be reported as a broken fixture. It asserts by **polling** that API, because a 2xx
+  from this backend says the request was answered and not that the write is readable:
+  the commit runs after the response goes out (#353). One step read once instead and
+  took all 87 specs down three times in a day (#335).
 - **Coverage instrumentation slows tests enough to trip a 5s `testTimeout`.** A
   heavy spec that passes under `test:run` can time out under `test:coverage`; re-run
   before believing it.
--- a/.claude/skills/e2e-tests/SKILL.md
+++ b/.claude/skills/e2e-tests/SKILL.md
@@ -63,10 +63,17 @@ It does **not** promise the row is rendered, and that is not a shortcut — the 
 refetch is sometimes answered with the pre-write list even though the row is committed
 and both server layers return it (**#230**, about one run in eight). Two consequences:
 
-- **A fixture step asks the API.** Every step of `seed.setup.ts` asserts through
-  `/api/…` now, because its job is that the fixture exists. Whether it renders is a
-  product claim, and `vault.spec.ts` / `skills.spec.ts` make it against the rows the
-  seed left, on pages they loaded themselves.
+- **A fixture step asks the API, and asks more than once.** Every step of
+  `seed.setup.ts` asserts through `/api/…` now, because its job is that the fixture
+  exists. Whether it renders is a product claim, and `vault.spec.ts` /
+  `skills.spec.ts` make it against the rows the seed left, on pages they loaded
+  themselves. Asking the API is not enough on its own: **a 2xx from this backend
+  means the request was answered, not that the write is readable.** The commit sits
+  in a `Depends`-with-`yield` exit code, which FastAPI unwinds after the response
+  has gone out (**#353**) — measured at 21.7ms on one acceptance. So a fixture check
+  is `nowThere`, a poll, never a single read. The membership step was the one site
+  #222 missed when it converted the file, and it cost 87 skipped specs three times
+  in a day (**#335**).
 - **A product spec about the rendering reloads first**, marked `#230`, until that
   issue closes.
 
--- a/.claude/skills/e2e-tests/SKILL.md
+++ b/.claude/skills/e2e-tests/SKILL.md
@@ -70,10 +70,13 @@ and both server layers return it (**#230**, about one run in eight). Two consequ
   themselves. Asking the API is not enough on its own: **a 2xx from this backend
   means the request was answered, not that the write is readable.** The commit sits
   in a `Depends`-with-`yield` exit code, which FastAPI unwinds after the response
-  has gone out (**#353**) — measured at 21.7ms on one acceptance. So a fixture check
-  is `nowThere`, a poll, never a single read. The membership step was the one site
-  #222 missed when it converted the file, and it cost 87 skipped specs three times
-  in a day (**#335**).
+  has gone out (**#353**) — measured at 21.7ms on one acceptance. So a check that
+  *follows a write* is `nowThere`, a poll, never a single read; the `alreadyThere`
+  guard at the head of each step stays one read, because it runs before the write.
+  The membership step was the one site #222 missed when it converted the file, and
+  it cost 87 skipped specs three times in a day (**#335**). One step is still the
+  exception in either direction: `a draft agent exists` waits on the Builder's
+  heading.
 - **A product spec about the rendering reloads first**, marked `#230`, until that
   issue closes.
 
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -193,9 +193,15 @@ pre-write list even though the row is committed and both server layers return it
 ([#230](https://github.com/vstorm-co/agenticos/issues/230), about one run in
 eight). So:
 
-- **A fixture step asks the API.** Every step of `seed.setup.ts` asserts through
-  `/api/…`, because its job is that the fixture exists — and a fixture step that
-  fails takes every product spec with it.
+- **A fixture step asks the API, and keeps asking.** Every step of
+  `seed.setup.ts` asserts through `/api/…`, because its job is that the fixture
+  exists — and a fixture step that fails takes every product spec with it. It
+  asks by polling (`nowThere`), never with a single read: a 2xx from this
+  backend means the request was answered, not that the write is readable, because
+  the commit runs in a dependency FastAPI unwinds after the response has gone out
+  ([#353](https://github.com/vstorm-co/agenticos/issues/353)). The one step that
+  read once cost 87 skipped specs three times in a day
+  ([#335](https://github.com/vstorm-co/agenticos/issues/335)).
 - **A product spec that is about the rendering says so**, and reloads first if it
   needs a list it can trust. `vault.spec.ts` has three `page.reload()` calls
   marked `#230`; when that issue closes, they come out.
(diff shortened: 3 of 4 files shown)


bll-kacper-wlodarczyk:
**Addendum on the run history** — two cancelled runs and a force-push against an unchanged tree. Both are mine and neither says anything about the diff.

The first run was superseded by the review-pass commit. The second I cancelled myself, having convinced myself the `test` job had hung on `make coverage-all`. It had not: `started 14:18:38Z → completed 14:26:27Z`, 7m49s, against 8m04s the run before. I had read a local process's elapsed time as that job's. The amend that followed is the same tree with a later commit timestamp, pushed only to get a fresh run after the cancel. I filed #364 on that misreading and have since rewritten it — what survives there is only the checkable part, that `ci.yml` gives `changes` a `timeout-minutes` and no other job one.

**Final state, run [31110850073](https://github.com/vstorm-co/agenticos/actions/runs/31110850073), all green:** `e2e` 4m06s · `test` 7m43s · `test-frontend` 5m09s · `lint` · `docs` · `changes` · `Security Scan` · CodeQL. Plus `make check` locally: "All checks passed — every CI job except e2e."

Issue #395, opened 2026-08-06
check_i18n.py misses a one-word template literal, and the reading model is why


OchnikBartek:
> **Rewritten 2026-08-06, and narrowed.** Filed while measuring #314, against a
> tree that has since moved: the multi-line text-node half of this is
> [#141](https://github.com/vstorm-co/agenticos/issues/141) and was already filed,
> and the interpolation rules on `main` now read the file joined back together. What
> is left is one blind spot with a measured backlog, plus the argument about the
> mechanism under both. Numbers below are re-measured against `main` at 0.0.59.

## What breaks

A template literal holding **one word beside its interpolation** is below
`TEMPLATE`'s two-word threshold, so it is never reported:

```tsx
frontend/src/components/kb/reusable-integrations.tsx:187
aria-label={`Remove ${source.name}`}
```

The threshold exists because `` `audience${key}Hint` `` builds a catalog key rather
than a sentence. It is the wrong test: what separates the two is **whitespace**. A
word with a space between it and the interpolation is prose; an identifier assembled
by interpolation has none.

**51 strings in 30 files**, all rendering in English under `pl`: `aria-label`s
(`Open ${org.name}`, `Remove ${m.full_name || m.email}`, `Delete ${skill.name}`,
`Toggle ${tool.name}`) and toasts (`${name} updated.`,
`${connection.name} disconnected.`). A further **14** are number-and-unit formatters
(`${Math.round(bytes / 1024)} KiB`, `${seconds / 60} min`, `HTTP ${r.status}`), which
want a units rule or an exemption rather than a message — that decision is part of
this.

## Why one more regex is the wrong answer, and this is the fourth time

`scripts/check_i18n.py` has been patched for a new shape four times: #199, #246,
#249, #314. Each was a real defect and each fix was correct; the pattern is the
problem. The script reads a `.tsx` file as text and decides, per candidate, whether it
is looking at TypeScript or JSX — so every rule carries a threshold or an exclusion
standing in for a parse, and the next shape falls between two of them.

Measured, not asserted: a prototype that reads text nodes over the whole file finds
the copy **and** about 90 false positives, in the two shapes a regex cannot see —
generics spanning lines
(`React.forwardRef<React.ElementRef<typeof X>, React.ComponentPropsWithoutRef<typeof Y>>`,
fourteen `components/ui/*.tsx` files) and a ternary between JSX branches
(`) : error ? (`, ten files). Excluding those means more patterns beside the ten
[...]

Issue #249, opened 2026-08-05
check_i18n.py misses word-first copy and lambda-in-count interpolations


DEENUU1:
## What breaks

`scripts/check_i18n.py` still reads past two shapes of hardcoded copy, so a
string in either form ships un-translated and the Polish UI renders it in
English — the exact failure #199 was about, one class over.

**1. Word-first copy — an interpolation with the words in front of it.** The
`MIXED` sweep counts the words left after removing the interpolation and fires at
two or more; `COUNT` (added in #199 / #246) catches an interpolation *followed*
by a single word. Neither sees a single word *before* one:

- `frontend/src/components/vault/secret-dialog.tsx:440` — `Rotate {secret?.name}`
- `frontend/src/components/chat/tool-results/rag.tsx:152` — `chunk {chunk.chunk}`
- `frontend/src/app/[locale]/(dashboard)/orgs/[id]/members/page.tsx:400` —
  `Invited {formatDate(inv.created_at)}`

Each is one message with a named parameter (`Rotate {name}`, `Chunk {n}`,
`Invited {date}`), and each renders its English word verbatim under `pl`.

**2. A count whose expression holds a lambda.** `COUNT` disallows an angle
bracket inside the interpolation so `{cond && <span/>} more` stays a conditional
— but that also means the `>` in a `=>` breaks the match, so a count built from a
`.reduce(...)` or `.filter(...)` escapes. #246 hit this at the KB `vectors` node
(`{documents.reduce((sum, d) => sum + d.chunk_count, 0)...} vectors`) and fixed it
by hand; a reintroduced one with a lambda would pass the guard.

## Sequence

`cd frontend && make lint` — clean, with all three word-first strings and a
lambda-count present. Switch the locale to `pl` and open the vault rotate dialog,
a RAG tool result, or the org members list.

## How you would know it was fixed

`check_i18n.py` fails on `>Word {expr}<` (a single word before an interpolation)
and on a `{expr} word` count whose expression contains a lambda, and the three
strings above read from `messages/en.json`.

## Where it sits

Pre-existing; found extending the guard in #199 / #246, which deliberately scoped
its `COUNT` rule to the interpolation-first count shape. Same guard, same
`.claude/rules/frontend.md` rule ("every user-facing string through `next-intl`").

Cluster: Frontend surfaces (#168). Sibling of #199; the durable fix lives in the
same script.

PR #396, opened 2026-08-06
fix: read a text node beside an inline handler

### LINKED ISSUE #314: check_i18n.py hides copy on any line holding an arrow, and misses text nodes split across lines
## What breaks

`scripts/check_i18n.py` reads a `.tsx` file **one line at a time**, and decides
per line whether it is TypeScript rather than JSX. Both shortcuts hide copy that
is sitting in the tree right now, in English, on a page a Polish user opens.

**1. A line holding `=>` is skipped entirely, so copy beside an inline handler is
invisible.** `offences()` sets `typescript = "=>" in line or "Promise<" in line`
and skips the text-node sweep for those lines, because
`onTest: (() => Promise<void>) | null` reads as a text node to a regex. An inline
handler is the single most common thing on a JSX line, so the exemption is far
wider than the problem:

- `frontend/src/components/mcp/mcp-server-list.tsx:968` —
  `<DropdownMenuItem onSelect={() => onTools(connection)}>Check connection</DropdownMenuItem>`
- `frontend/src/components/mcp/mcp-server-list.tsx:969` —
  `<DropdownMenuItem onSelect={() => onEdit(connection)}>Settings</DropdownMenuItem>`

Both sit between two siblings that read `t("finishSign")` and `t("disconnect")`,
which is what makes them a slip rather than a decision. Ungating the sweep on its
own is not the fix: it reports eleven false positives in the current tree (ten
`Promise`, one `(await apiClient.get`), so the discriminator has to get better -
the `>` that opens a candidate text node must not be the `>` of an arrow or of a
closing generic.

**2. A text node spanning two lines matches nothing.** Every rule anchors on `>`
and `<` in one line, so a node broken across lines by the formatter is read by
none of them:

```tsx
frontend/src/app/[locale]/(dashboard)/orgs/[id]/members/page.tsx:396
    {inv.used_count ?? 0} of {inv.max_uses ?? "∞"} used
    {inv.email_domain && <> · @{inv.email_domain} only</>}
```

`{n} of {m} used` is two words with two interpolations - exactly what `MIXED` is
for - and it passes because neither `>` nor `<` is on that line. `· @{domain} only`
is the `COUNT` shape with punctuation in front of the interpolation, which the rule
does not step over the way `LEAD` now does for a leading word.

**3. A one-word template literal is below `TEMPLATE`'s two-word threshold.** The
threshold is there because `audience${key}Hint` builds a catalog key, but it also
exempts real copy:

- `frontend/src/components/kb/reusable-integrations.tsx:187` —
  `aria-label={`Remove ${source.name}`}`, two lines above the title #249 fixed
- `frontend/src/components/admin/user-detail-drawer.tsx:187` —
  `` ` · ${c.message_count} msg` ``, which is also a count, so it wants an ICU
  `plural` and not only a message

## Sequence

`cd frontend && make lint` is clean with all six strings present. Switch the locale
to `pl` and open the MCP server list's row menu, the org members list with a
pending invitation, or the reusable-integrations row.

`check_i18n.py` fails on each of the six above, and still passes on
`onTest: (() => Promise<void>) | null`, on `(await apiClient.get(...))` and on
`audience${key}Hint`. The catalog holds all six, and the two counts hold them as
ICU `plural`.

Worth doing together: (1) and (2) both come from reading the file line by line, so
the honest fix is probably to join a component's JSX into one string per element
before matching, rather than three more regexes with three more thresholds.

## Where it sits

Third in the line that is #199 → #246 → #249, all in `scripts/check_i18n.py`, all
"the guard is narrower than the rule it enforces". Not a duplicate: #249 is the
word-first shape and the lambda count, both of which are fixed and held by tests
(`tests/test_check_i18n_word_first.py`, `tests/test_check_i18n_count.py`). These
three shapes survive that change, and item (1) is the one that hides the most,
because an inline handler is everywhere.

Cluster: **Frontend surfaces / data layer** on #168, beside #249.

Pre-existing. Found while extending the guard for #249 — measured by ungating each
shortcut in turn and reading what it then reported.

OchnikBartek:
## What this changes

The i18n guard reads a text node that sits beside an inline handler.

It used to skip any line holding `=>`. The reason was real — `onTest: (() =>
Promise<void>) | null` puts a `>`, a word and a `<` on one line and reads as a text
node to a regex — but an inline handler is the most common thing on a JSX line, so the
exemption was far wider than the problem. `<DropdownMenuItem onSelect={() =>
onTools(c)}>Check connection</…>` and the `Settings` beside it were invisible for that
reason, sitting between two siblings that read from the catalog. Both now read from
`mcp`.

The line-level guess is gone. What separates the two cases is the `>` itself — an
arrow's, a comparison's, or a tag's — so the rule is a lookbehind on the character
before it rather than a decision about the whole line.

## Why only one of #314's three items

**Because I measured the other two before writing them, and each is a copy backlog
rather than a rule.** A prototype that reads whole-file text nodes instead of lines,
diffed against what the guard reports today:

| #314 item | What it actually finds |
|---|---|
| 1 — a line holding `=>` hides a text node | **2** strings, 1 file — this PR |
| 2 — a text node split across lines | **61** strings, 31 files |
| 3 — a one-word template literal | **53** strings, 32 files, plus 14 unit formatters |

Item 2 is `global-error.tsx` ("Something went wrong", "Try again"), `not-found.tsx`
("Page not found"), `shared/[token]/page.tsx`, `auth/callback/page.tsx` and a dozen
explanatory paragraphs — all rendering in English under `pl` right now. Item 3 is
mostly `aria-label`s and toasts.

And the prototype also produces **~90 false positives**, in exactly the two shapes a
regex cannot see: TS generics spanning lines
(`React.ComponentPropsWithoutRef<typeof X>>` in fourteen `components/ui/*.tsx` files)
and a ternary between JSX branches (`) : error ? (` in ten). Excluding those means
more patterns beside the ten already there — the fifth and sixth shape this script has
been patched for, after #199, #246 and #249.

So both are [#395](https://github.com/vstorm-co/agenticos/issues/395), which names the
mechanism instead: `ts.createSourceFile` from a bun script — `frontend/node_modules`
already holds `typescript` and `make lint` already runs bun — where a multi-line node
*is* one node, a generic is not JsxText at all, and the thresholds standing in for "is
this a sentence" can be deleted rather than ported. The 114 strings are that issue's
part 2, with the counts and the file list.

**This is the last regex patch to this file**, and the PR says so on purpose.

## Checks

      `scripts/check_i18n.py` fails `test_copy_beside_an_inline_handler_is_refused`
- [x] `make check` passes
- [x] Platform-layer coverage is still 100% (3196 passed; frontend 3175)
- [ ] n/a — no migration

`backend/tests/test_check_i18n_arrow_lines.py` pins both directions: copy beside a
handler is refused, and the four shapes the old skip was covering for — a
`Promise`-returning signature, a generic call inside an arrow, a `>=` comparison, and
the `t()` call that replaces the copy — still pass. Verified over the whole tree before
writing the rule: two findings, no false positive.

## Notes for the reviewer

**Depends on #315** (issue #249). Cut from `fix/i18n-guard-word-first-copy` rather than
`main`, because both rewrite the same rules in the same file. Merge #315 first.

`.claude/rules/frontend.md` now says the guard is quiet about those two shapes and how
many are behind them, because a developer reading that rule needs to know the silence
is not approval.

Refs #314, #395. Does not close #314 — its items 2 and 3 are #395's part 2, and
whoever takes that should decide whether #314 closes with it.


DEENUU1 (commented):
The reasoning in this PR is the best thing in it, and it is going to outlive the code: measuring items 2 and 3 before writing them, finding 61 and 53 strings plus ~90 false positives, and concluding that the answer is `ts.createSourceFile` rather than an eleventh regex — that is the right call and #395 is the right place for it. "This is the last regex patch to this file" is worth keeping as a line in the sand.

**But the branch has been overtaken, and merging it as-is would undo a day of work.** It was cut from #315, which merged this afternoon as v0.0.51, and then #363 merged as v0.0.52 and did item 1 by a different route. Against today's `main` this branch is **147 files and 9,376 deletions** — it would delete `provider-row.tsx`, `reload_supervisor.py`, `vector_tables.py`, `services/audit.py`, `ci_changed_scope.py` and revert `check_backticks.py` and `check_i18n.py`.

### What I checked rather than assumed

- **Your two MCP strings are already fixed on `main`** — `t("checkConnection")` and `t("settings")`, landed with #363.
- **`main` passes all five of your tests.** I copied `test_check_i18n_arrow_lines.py` onto `main` unchanged and ran it: 5 passed. So the detection is there, reached through `mask_generics` on the joined source instead of your lookbehind on the `>`.
- **Your guard finds nothing on `main`'s tree** that `main`'s guard does not — both report clean over 2,849 messages.

Two mechanisms, same outcome; the one that landed first won on timing, not on merit. Yours is arguably the more surgical of the two.

### What was genuinely left, and where it went

One thing, and it is a finding about `main` rather than about your branch. None of the ten cases in `test_check_i18n_nodes.py` actually test `mask_generics` — **stub it to `return text` and all ten still pass**, while the guard then reports three false positives over the real tree. It is load-bearing and was untested, so a refactor could have broken it with only a tree-wide `make lint` to notice.

That is #403, with a case built from `brand-icon.tsx`: one offence without the mask, none with it. Credited to this review, because chasing your tests is what turned it up.

### What I would do with this one

Close it, and keep #395. Everything it fixes is fixed, its tests are covered, and the measurement work — the table, the false-positive analysis, the argument for the compiler API — belongs on #395 where somebody will act on it, not on a branch that cannot merge. I have not closed it myself; that is yours.

If you would rather keep something, the only merge-able residue is a rebase onto `main` carrying `test_check_i18n_arrow_lines.py` alone, and #403 already covers the gap that was worth covering.

OchnikBartek:
Superseded: main already carries the same rule — `(?<!=)` on `JSX_TEXT` with the per-line TypeScript gate gone — and both `mcp-server-list.tsx` strings are already in the catalog, landed while this was open. Nothing here to add. The backlog measurement stays useful and is on #395.

Issue #404, opened 2026-08-06
make test is red on main: PgVectorStore calls a resolver it documents as optional


OchnikBartek:
## What breaks

`make test` is red on `main` for anybody with a database. One integration test
crashes in the store rather than asserting anything:

```
tests/integration/test_vector_store_reserved_names.py::
  test_a_collection_beside_it_is_created_and_dropped_for_real
E   TypeError: 'NoneType' object is not callable
app/services/rag/vectorstore.py:278: TypeError
```

Line 278 is `PgVectorStore._for_collection`:

```python
resolved = await self._resolver(name)
if resolved is None:
    return self.embedder, self.dim
```

The `None` check is one line too late. `__init__` declares
`resolver: "EmbeddingResolver | None" = None` and its docstring says what that
means — "None keeps the deployment defaults for everything, which is the
pre-resolver behaviour and still what a collection outside the KB table gets" —
so a store built without one is a documented, supported shape. It is also the
shape that raises: the attribute is called before it is checked.

## Why the test hits it and nothing else does

Every production caller passes a resolver: `app/main.py:103`,
`app/api/deps.py:653`, `app/agents/capabilities/knowledge/_search.py:31`,
`app/commands/rag.py:57`. So the crash is not reachable from a running
deployment today — what it breaks is the gate, and the contract.

The integration test builds the documented shape deliberately
(`store._resolver = None`, with a docstring explaining it is not using the
deployment's engine), and creating a collection needs the vector width, which
goes through `_for_collection`.

## Reproducing

Any database will do — it fails before pgvector is asked for anything:

```bash
cd backend && POSTGRES_PASSWORD=postgres uv run pytest \
  tests/integration/test_vector_store_reserved_names.py -q --no-cov
```

## Two ways to fix it, and they say different things

Whoever takes this should decide which contract is true, rather than silencing
the test:

1. **`None` is supported** — the constructor and its docstring say so, and the
   guard moves up: `if self._resolver is None: return self.embedder, self.dim`.
   The declared type stays as it is.
2. **`None` is not supported** — a store always resolves per collection, the
   parameter loses its default and its `| None`, and the test constructs one with
   a resolver that answers `None`. That is a smaller surface, and it makes the
   four production call sites the only shape.

Either way the unit half (`tests/test_reserved_collection_names.py`) and this
integration half both have to keep passing, and the fix wants a test that builds
a store the documented way and asks it for a width.

## How you would know it was fixed

[...]

Issue #306, opened 2026-08-06
Ingestion ignores the collection's embedding key and model: the worker's store has no resolver


DEENUU1:
Uploading a document to a collection that chose an organization vault key for
its embeddings fails in the worker with advice that is wrong:

```
RuntimeError: Ingestion failed for 253 Stories To Learn About Data Analysis HackerNoon.md:
No embedding credential is configured, so documents cannot be indexed or searched.
Set OPENROUTER_API_KEY in the backend environment and restart.
```

A credential *is* configured — it was picked in **Create knowledge base →
Embeddings → Key** and is stored in the vault. The ingestion path never reads
it.

## Sequence

1. Deployment has no `OPENROUTER_API_KEY` (the normal case when an organization
   pays for its own embeddings).
2. Create a collection and choose an OpenRouter key from the vault. It is
   validated at creation — `KnowledgeBaseService._check_embedding_secret`
   (`app/services/knowledge_base.py:259-289`) refuses a key the organization
   does not hold or one of the wrong purpose — and stored as
   `embedding_secret_id`.
3. Upload a document. The Prefect flow fails as above.

## Cause

`app/worker/tasks/rag_tasks.py:52-53`:

```python
embed_service = EmbeddingService(settings=rag_settings)
vector_store = VectorStore(settings=rag_settings, embedding_service=embed_service)
```

No `resolver=`. `PgVectorStore._for_collection`
(`app/services/rag/vectorstore.py:222-241`) short-circuits to the deployment
embedder whenever `self._resolver is None`, so `embeddings_for_collection` is
never called and the collection's `embedding_secret_id` is never looked at.
`EmbeddingService` then falls back to `app_settings.OPENROUTER_API_KEY`
(`app/services/rag/embeddings.py:124`), which is empty, and
`OpenAIEmbeddingProvider.client` raises the `ConfigurationError` above
(`embeddings.py:68-77`).

**Every other construction passes the resolver** — `app/main.py:103-107`,
`app/api/deps.py:653-655`, `app/services/rag/ingestion.py:38-42`,
`app/commands/rag.py:56-58`,
`app/agents/capabilities/knowledge/_search.py:31-33`. The one that does not is
the one that ingests documents in the background, i.e. the only path a document
upload actually takes.

## Two consequences, not one

**The key.** The organization's chosen key is silently unused. Where the
deployment key is empty this is the crash above. Where both are set, it is
worse than a crash: embeddings are billed to the deployment's account while the
[...]

Issue #154, opened 2026-08-04
e2e: sharing and skills specs fail when a mutation's refetch lands stale


DEENUU1:
## What breaks

Two Playwright specs fail intermittently, both with the same shape: the
mutation reached the backend and committed, the invalidation refetch went
out, and the panel kept the pre-mutation snapshot for the whole 15s
timeout.

- `e2e/sharing.spec.ts:55` "shares an agent, changes the level, then revokes it"
  - run 30887160085: after `Remove colleague@example.com`, the row is still
    there — `34 × locator resolved to 1 element` on
    `expect(panel.getByLabel('Access for …')).toHaveCount(0)` (`sharing.spec.ts:80`).
  - run 30854869368, **same test, earlier assertion**: the level change never
    showed — `Expected "Can use", Received "Can view"` (`sharing.spec.ts:76`).
- `e2e/skills.spec.ts:84` "a save survives a reload"
  - run 30887160085: `POST /api/v1/skills` succeeded, the list refetched, and
    the gallery still reads `1 skill` — `skillCard` never appears
    (`skills.spec.ts:99`).

## The sequence

From the backend log of run 30887160085 (`agenticos:cache:` is unused, no
`@cache` on these routes, and the proxy sends `cache: "no-store"` — the read
really did hit Postgres after the write committed):

```
07:21:11.053 POST /api/v1/skills
07:21:11.057   BEGIN;
07:21:11.069 COMMIT;            <- the write is committed
07:21:11.075 GET /api/v1/skills ? skip='0' & limit='50' & sort='name'
07:21:11.079   BEGIN;
07:21:11.089 COMMIT;            <- the read is after it
07:21:26.908 (nothing else asked; the test had already timed out)
```

Same for the revoke: `DELETE …/sharing/grants/<id>` at `07:20:44.435`, the
sharing `GET` at `.452`, and no further request in the next fifteen seconds.
So exactly one refetch happens per mutation, it is ordered after the commit,
and the UI still shows the old value — which points at the client dropping
the response rather than at the read being stale.

## Not caused by the branch it was found on

`feat/sandbox` / #119 found it; the sharing spec failed the same way at
`832d647`, before that branch's i18n work, and passed at `c655e08` and
`a13c2a3`. Nothing outside components changed between the green and red runs
(`git diff a13c2a3..c927fd0 -- frontend/src/lib frontend/src/hooks
frontend/src/app/api` is only `src/i18n.ts`).

## How you would know it was fixed

[...]

PR #405, opened 2026-08-06
fix(api): declare no-store on responses the backend leaves unmarked

### LINKED ISSUE #230: A stored secret can stay off the vault list forever: the invalidation refetch is answered with the pre-write list

OchnikBartek:
## What this changes

Every JSON response the platform proxy returns now carries a cache policy. It
carried none: no `Cache-Control`, no `ETag`, no `Last-Modified`, on every mutable
collection on the surface. `platform-proxy.ts` forwards a policy when the backend
sets one, and for JSON the backend sets none — so what reached the browser was a 200
it may reuse on its own judgement. **Silence is not "do not cache".**

That is the leading suspect in #230: a vault list refetched immediately after a write
is answered with the pre-write body, byte for byte, and nothing refetches again, so a
stored secret can stay off the list until something reloads the page.

## Why this way

**Two layers ruled out by reading before touching anything.** The backend does not
cache this list — no Redis, no cache header anywhere near
`app/api/routes/v1/secrets.py`, and the only endpoints that set `Cache-Control` at all
are the catalog icons, the embed bundle and two RAG binaries. And the proxy's own
outbound `fetch` already says `cache: "no-store"`. The undeclared hop was the response
the proxy hands back, which is also the one place that can declare it for the whole
surface.

**The header is right whether or not it closes the race.** Every answer here depends
on a cookie, a permission set and an organization header; there is nothing on this
surface a shared or heuristic cache may keep. One line in `describingHeaders` covers
`/api/secrets`, `/api/skills`, `/api/agents`, `/api/kb` and every other list a mutation
invalidates. A backend that *does* name a policy still wins — there is a test for each
direction, because a proxy that overrode the icons' `immutable` would be inventing a
policy rather than forwarding one.

**Not `refetchInterval`, not a delay, not a `staleTime`.** Those move the race or hide
it. The only thing changed is a statement of fact about what may be cached.

## Checks

      `platform-proxy.ts` fails `refuses to let a list be cached when the backend
      named no policy`
- [x] `make check` — see the exception below, which this diff cannot reach
- [x] Frontend coverage gate: 3247 passed, 100% lines/statements/functions
- [ ] n/a — no migration

`make lint`, `make test-frontend-cov`, `make build-frontend`, `make docs-build` and
`make audit` are green.

**`make test` was red on one integration test when this branch was cut, and it was not
this branch** — that was `test_vector_store_reserved_names.py` setting `_resolver = None`
against a contract #306 had already made required. Fixed on `main` in v0.0.55 (#397),
and #404 is closed as a duplicate of it. Rebased onto `main`, so this branch no longer
carries it.

## Notes for the reviewer

**This closes #230 on the fix, not on a measurement, and that is worth stating.** The
issue is explicit that its own suspect is unproven — the backend log shows the refetch
*did* reach the server, which a pure cache hit would not have produced — and settling
it needs the reproduction the issue describes: a seeded stack (postgres, host uvicorn,
production frontend build) and fifty-odd Playwright seed runs at about one failure in
eight. That has not been run here.

The header is right either way. An undeclared cache policy on every mutable collection
the proxy serves is a defect on its own terms, whatever it did or did not cause in the
vault, and it is the fix the issue asks for first.

The three `page.reload()` calls in `frontend/e2e/vault.spec.ts` marked `#230` stay in
this diff. Whoever takes them out is running the real test of this fix: the flake was
measured at roughly one failure in eight, so it wants repeated runs of the suite rather
than the single green one this branch would give it. If it flakes then, that is
information — `no-store` was not the whole cause, and #230 comes back with a run count
behind it.

One thing measured for whoever does: the issue's second suspect — "why did the page
issue two `GET /api/secrets` on load" — is narrower than it looked. The vault page holds
exactly **one** `useSecrets()`, so the second request is not a second consumer of the
hook. With `staleTime` unset on that list, a second observer mounting after the first
fetch resolves is enough to explain it, and that is worth confirming before anything is
changed about it.

`.claude/rules/frontend.md` now states the policy, because the hand-rolled binary
routes under `/api/files` and `/api/users/avatar` do not go through this proxy and owe
the header themselves.

Closes #230
Refs #154

### DIFF
--- a/.claude/rules/frontend.md
+++ b/.claude/rules/frontend.md
@@ -60,6 +60,12 @@ There is no `(marketing)` route group.
   translator, so it holds *keys* and the component translates at the point of use;
   a pure helper either answers with a key or takes `t`.
 - Keep components under ~100 lines; extract when they grow.
+- **A response with no `Cache-Control` is not a response nobody caches.**
+  `platform-proxy.ts` stamps `no-store` on anything the backend left unmarked: every
+  answer on this surface depends on a cookie, a permission set and the organization
+  header, and a list refetched right after a write must reach the server. A
+  hand-rolled route file owes the same header - the proxy is the only place that
+  applies it for you.
 - Do not hand-edit `src/lib/mcp-logos.generated.ts` — run `bun run gen:mcp-logos`.
 
 ## Permissions
--- a/frontend/src/lib/platform-proxy.test.ts
+++ b/frontend/src/lib/platform-proxy.test.ts
@@ -177,6 +177,30 @@ describe("platformProxy", () => {
     expect(response.status).toBe(204);
     expect(await response.text()).toBe("");
   });
+
+  it("refuses to let a list be cached when the backend named no policy", async () => {
+    // Silence is not "do not cache". A 200 with no `Cache-Control`, no `ETag` and
+    // no `Last-Modified` is one the browser may reuse on its own judgement - so a
+    // list refetched right after a write could be answered without the server
+    // being asked, and the page kept rendering the row it had just created as
+    // absent (#230). Every answer here depends on a cookie, a permission set and
+    // an organization header, so there is nothing to cache in the first place.
+    backendReplies('{"items":[],"total":0}');
+
+    const response = await platformProxy().GET(request("/api/secrets"));
+
+    expect(response.headers.get("Cache-Control")).toBe("no-store");
+  });
+
+  it("still lets the backend choose a policy where it has one", async () => {
+    // The catalog icons and the embed bundle are cacheable and say so. A proxy
+    // that overrode them would be inventing a policy rather than forwarding one.
+    backendReplies("{}", { headers: { "cache-control": "public, max-age=86400, immutable" } });
+
+    const response = await platformProxy().GET(request("/api/catalog/icons/github.svg"));
+
+    expect(response.headers.get("Cache-Control")).toBe("public, max-age=86400, immutable");
+  });
 });
 
 // -- the sweeps ---------------------------------------------------------------
--- a/frontend/src/lib/platform-proxy.ts
+++ b/frontend/src/lib/platform-proxy.ts
@@ -45,11 +45,25 @@ const WITH_BODY = new Set(["POST", "PUT", "PATCH"]);
  * Response headers that describe the body and must survive the hop.
  *
  * A proxy forwards what the origin said rather than inventing a policy of its
- * own: `Content-Disposition` is how a download gets its filename, and caching a
- * response the backend never marked cacheable is the backend's decision to make,
- * not this file's.
+ * own: `Content-Disposition` is how a download gets its filename, and a
+ * `Cache-Control` the backend chose - the catalog icons, an embed bundle - is
+ * the backend's decision to make.
  */
 const DESCRIBES_THE_BODY = ["content-type", "content-disposition", "cache-control"];
+
+/**
+ * What a response gets when the backend named no policy at all.
+ *
+ * Silence is not the same as "do not cache". A 200 with no `Cache-Control`, no
+ * `ETag` and no `Last-Modified` is a response the browser may reuse on its own
+ * judgement, and every mutable collection on this surface arrived that way - so a
+ * list refetched immediately after a write could be answered without the server
+ * being asked, and the page went on rendering the row it had just created as
+ * absent (#230). Nothing here is cacheable by default anyway: every answer
+ * depends on the caller's cookie, their permissions and the organization header,
+ * which is the same reason the outbound `fetch` below is `no-store`.
+ */
+const NO_POLICY_MEANS = "no-store";
 
 type Handler = (request: NextRequest) => Promise<Response>;
 
@@ -70,6 +84,7 @@ function describingHeaders(source: Headers): Headers {
     if (value) headers.set(name, value);
   }
   if (!headers.has("content-type")) headers.set("content-type", "application/json");
+  if (!headers.has("cache-control")) headers.set("cache-control", NO_POLICY_MEANS);
   return headers;
 }
 


DEENUU1 (commented):
Right change, right place, and "silence is not do-not-cache" is the sentence that makes it obvious in hindsight. Declaring it once in `describingHeaders` rather than per route is what stops the next hand-rolled route file from having the same hole.

### What I checked rather than took on trust

**Nothing loses caching.** Four backend endpoints set `Cache-Control` — `knowledge_bases.py:217`, `rag.py:381`, `catalog_icons.py:28`, `embed.py:94` — and the first three are proxied and forwarded, since `cache-control` is already in `DESCRIBES_THE_BODY`. The embed bundle has no Next route at all, so it never reaches this code. `/api/files/[id]` sets its own `private, max-age=3600` in the route file and keeps it, because your line only fires when the header is absent.

**The test bites.** Removed the one line from `platform-proxy.ts` and `refuses to let a list be cached when the backend named no policy` fails, 18/19. Both directions covered.

### The part I would change before merging

**The `make test` note and #404 are both stale, and the branch is 14 commits behind `main`.**

That `TypeError: 'NoneType' object is not callable` was mine — a test helper setting `_resolver = None` against a contract that #306 had already made required. It shipped in v0.0.45 and I fixed it in **v0.0.55** (#397) about an hour before #404 was filed.

More to the point, **#404's premise no longer holds**: it says the constructor "documents `resolver=None` as supported", and on today's `main` the signature is `resolver: "EmbeddingResolver"` with a comment explaining at length why it is required and not defaulted. The choice #404 asks you to make has already been made — the store's contract stands, the test was wrong. Worth closing it with a pointer to #397 rather than leaving an open issue describing a fixed bug and a contract that reads the other way.

So: rebase onto `main`, drop that paragraph, and this is ready.

### On #230, and this is worth more than the header

You are right that the cache is unproven, and I think today gives you the actual answer. **#353**: this backend answers a write *before* its transaction commits — `get_db_session` commits in a `Depends`-with-`yield` exit, and FastAPI unwinds that stack after `await response(...)` has already gone out. It was found chasing #335, the e2e seed flake, and measured three ways: a minimal app on the pinned FastAPI (`0.0ms client got 204` against `256.4ms commit finished`), the real backend (a row invisible to the very next read for 21.7ms, 1 miss in 45), and the product's own SQL log showing nine requests arriving while the previous transaction was still open.

That is #230's shape exactly, and it explains the detail you flagged as inconvenient for the cache theory: **the refetch did reach the server.** It reached it and was answered honestly with a list that did not yet contain the row.

So this header is correct on its own merits and should land — but I would put #353 in the body as the likely cause, so whoever picks up #230 starts there instead of re-testing the cache.

DEENUU1:
Correction to my review, and it matters because I put it in your PR body.

I said **#353** was the likely cause of #230 and suggested whoever picks up #230 should start there rather than re-testing the cache. **That was wrong**, and the work on #353 is what disproved it.

PR **#428** fixes the commit ordering, and in doing so it read #230's own trace: **the commit wins by 5ms there.** On the old ordering it sometimes lost; on the new one it always wins. Either way, #230's refetch was answered after the row was committed — so the write being acknowledged early is not what produced the stale list.

Which puts your `Cache-Control` hypothesis back as the live one. Your original reasoning was better than my correction to it: you flagged that the refetch reaching the server sat awkwardly with a pure cache hit, and you declined to close #230 on the strength of it. That caution was right, and I overrode it with a theory that has now failed.

I have edited the PR body to remove the `Refs #353` framing I added. The header change stands on its own merits regardless — every answer on that surface depends on a cookie, a permission set and an organization header, and silence is not "do not cache".

Issue #410, opened 2026-08-06
make lint fails on main: two ruff errors merged during the CI outage


DEENUU1:
**Severity: low** · found running `make lint` while working #358/#366, not caused by that branch

## What's wrong

`make lint-backend` fails on `main`. `ruff check . ../scripts` reports two
errors, so the `lint` job is red for every branch cut from `main` until they
are cleared:

```
RET501 Explicit `return None` in function annotated `-> None`
  --> tests/integration/test_vector_store_reserved_names.py:42
RUF100 Unused `noqa` directive (non-enabled: `BLE001`)
  --> tests/test_migrations.py:99
```

Neither is interesting in itself. What is interesting is how they landed:
both arrived during the GitHub Actions outage of 2026-08-06, in #397 and #302
respectively, and were merged with the `lint` job queued rather than run. This
is the first thing to check when the outage clears - a red `main` looks like
"the branch broke it" to whoever cuts the next one.

## Where

- `backend/tests/integration/test_vector_store_reserved_names.py:42`
- `backend/tests/test_migrations.py:99`

## How it reproduces

```bash
git switch main && make lint-backend
```

## How you would know it was fixed

`make lint-backend` passes on `main` with nothing staged.

## Related

Cluster: **CI, gates and the check parity** (#143, #311). Adjacent to #311 -
the automated reviewer being off is the other half of "nothing looked at this
diff"; here it was the gate itself that never ran. Not a duplicate of anything
open: checked the list, no issue names either file.

Fixed on `fix/wedged-worker-beyond-the-local-stack` rather than separately,
because that branch cannot be verified locally until `make lint` is green.

DEENUU1:
Duplicate of **#407**, which is fixed in **#426** (commit `fb1b6f2`).

Five of us filed this within minutes of each other — #406, #407, #408, #409, #410 — because every agent working a branch today cut it from a `main` that does not pass `make lint-backend`, and each hit the same two ruff errors before touching anything. That is the correct instinct and the wrong outcome five times over, so consolidating on #407.

The cause is worth recording: both errors (`RET501` in `test_vector_store_reserved_names.py`, `RUF100` in `test_migrations.py`) came from PRs #397 and #302, which I merged during the GitHub Actions outage with every check still pending. Neither would have reached `main` with CI running. #426 fixes them by hand rather than with `--fix`, because the autofix deletes the reasoning along with the line.

Closing as duplicate.

DEENUU1:
[...]

Issue #406, opened 2026-08-06
ruff fails on main: two RET501 violations in the new migration and reserved-name tests


OchnikBartek:
**What breaks:** `make lint` fails on `main` itself, so every branch that merges it inherits a red `lint` check.

```
RET501 Do not explicitly `return None` in function if it is the only possible return value
  --> backend/tests/integration/test_vector_store_reserved_names.py:42:5
  --> backend/tests/test_migrations.py:99:31
Found 2 errors. [*] 2 fixable with the `--fix` option.
```

**Verified against a pristine checkout**, not a working tree: `git worktree add --detach <tmp> origin/main`, then `uv run ruff check app tests cli` in `<tmp>/backend`, reports the same two errors. Both files are byte-identical to `main`'s on the branch that found this, and `pyproject.toml` is `main`'s too, so it is not a config difference.

**Where they came from:** `test_vector_store_reserved_names.py:42` is the resolver stub added by #397, whose body is `return None`. `test_migrations.py:99` is the same rule.

**Fixed when:** `uv run ruff check app tests cli` is clean on `main`. `ruff check --fix` resolves both — the first becomes a bare `return`, which reads the same for a stub whose whole point is answering "nothing recorded".

Found merging `main` into #221, which is where the red `lint` shows up. Not fixed there on purpose: two lint violations in migration and reserved-name tests have nothing to do with that branch, and carrying somebody else's fix would have made its diff harder to review. Related to #397, which introduced the first of the two.

DEENUU1:
Filed five times over: this, #407, #408, #409 and #410 are all the same two ruff findings — the `RET501` in `tests/integration/test_vector_store_reserved_names.py` from #397 and the `RUF100` in `tests/test_migrations.py` from #302. Parallel branches all hit the red tree during the Actions outage and each filed one. Worth closing four of them as duplicates of this.

Fixed in #418 (commit `e37e47f`), as its own commit so it can be dropped if another branch lands the same fix first.

DEENUU1:
Duplicate of **#407**, which is fixed in **#426** (commit `fb1b6f2`).

Five of us filed this within minutes of each other — #406, #407, #408, #409, #410 — because every agent working a branch today cut it from a `main` that does not pass `make lint-backend`, and each hit the same two ruff errors before touching anything. That is the correct instinct and the wrong outcome five times over, so consolidating on #407.

[...]

Issue #370, opened 2026-08-06
A Drive file named with a path escapes the sync directory


DEENUU1:
## What breaks

A Google Drive file's name is used as a path component with no sanitizing, so a
name containing `/` or `..` writes outside the directory the sync created for
it:

```python
# app/services/rag/connectors/google_drive.py:179
dest_path = dest_dir / file.name
```
```python
# app/services/rag/sources/google_drive.py:137
dest_path = dest_dir / file.name
```

`file.name` is the Drive object's name taken verbatim
(`connectors/google_drive.py:126`, `name = f.get("name", "")`). Drive names are
not filesystem paths and may contain separators. `dest_dir` is a
`TemporaryDirectory` created by the worker (`app/worker/background/rag.py:119`,
`app/services/rag/sources/base.py:97-98`), so `../../..` escapes it and the file
is written wherever the worker's uid can write — and then ingested from that
path.

The S3 connector does not have this: `Path(key).name` reduces the key to one
component (`connectors/s3.py:110`, `sources/s3.py:61,89`). Two connectors
implementing the same `BaseSyncConnector.download_file` contract disagree about
who sanitizes, which is why this is worth fixing in the base rather than in one
connector.

## The sequence that triggers it

1. A sync source points at a Drive folder — commonly one shared with people
   outside the tenant, since that is what folder sharing is for.
2. Anyone who can add a file to that folder names it
   `../../../../home/app/.ssh/authorized_keys` (or any path the worker can
   write).
3. The next sync writes it there. The attacker is not necessarily the tenant.

## How you would know it was fixed

- A remote file named `../evil` lands inside the sync directory as one
  component, or the download is refused.
- The check is in `BaseSyncConnector` / `sources/base.py`, so a connector added
  later inherits it — the pattern `app/services/file_storage.py:57-68,104-110`
  already uses (sanitize to one component, then assert containment after
  `resolve()`).
- A test names a remote file with a traversal and asserts nothing is written
  outside the temporary directory.

## Where it sits

**RAG and knowledge · security** on #168, beside the Google Drive query
injection filed in the same sweep. Related but distinct: that one widens what is
listed, this one controls where a listed file is written.

Pre-existing. Found sweeping for places a caller-supplied name reaches a path or
an identifier while fixing #345.

DEENUU1:
[...]

PR #418, opened 2026-08-06
fix(rag): trust no name or id a sync source hands us

### LINKED ISSUE #369: A sync source's folder_id is interpolated into the Drive query unescaped
## What breaks

A sync source's `folder_id` is interpolated into a Google Drive query with no
escaping, and the value is a free-form config field:

```python
# app/services/rag/connectors/google_drive.py:102
query = f"'{folder_id}' in parents and trashed = false"
```
```python
# app/services/rag/sources/google_drive.py:71
query_parts.append(f"'{path}' in parents")
```

`config` is `dict[str, object]` (`app/schemas/sync_source.py:37`) and
`BaseSyncConnector.validate_config` (`app/services/rag/connectors/__init__.py:49-54`)
only checks that required fields are *present* — `CONFIG_SCHEMA["folder_id"]`
declares `"type": "string"` and no pattern
(`connectors/google_drive.py:65-69`). A `'` closes the literal, so
`x' in parents or name contains 'salary` is a valid, wider query.

## How far it actually reaches

Worth stating plainly, because it changes the severity: the credential is
normally the tenant's **own** service account JSON, supplied per source and
required by the schema (`connectors/google_drive.py:57-63`). Widening a query
run under your own credential is not a boundary crossing.

The case that matters is the fallback. `_get_drive_service`
(`connectors/google_drive.py:78-90`) falls back to
`settings.GOOGLE_DRIVE_CREDENTIALS_FILE` — a deployment-wide credential — when
the config field is absent, which is how sources created before the field
existed still work. On a deployment with that set, a tenant's `folder_id`
widens a query running under the operator's service account, and
`drive.readonly` on it is whatever the operator shared with it.

`POST /rag/sync/sources` or `POST /kb/{kb_id}/sync-sources` with
`collections:edit`, `connector_type: "gdrive"`, and a `folder_id` containing a
quote. Also reachable from `rag-sync-gdrive --folder-id`
(`app/commands/rag.py:461`), which takes the value straight from the shell.

- A `folder_id` containing a quote is refused at source creation, or escaped
  where the query is built — not both places disagreeing about which.
- A test asserts the query built from a hostile `folder_id` still constrains to
  one parent.
- Ideally the deployment-wide fallback goes as well: a per-tenant connector
  quietly running on the operator's credential is the part that turns this from
  a self-inflicted query into an escalation.

## Where it sits

**RAG and knowledge · security** on #168, in the sync-source family with #233
(sync-source logs paging before they filter) and #356 (a failed source drawn as
a success). Nothing there covers connector input validation, and this is not a
duplicate of #345 — same sweep, different sink: a query language rather than a
table name.

Pre-existing. Found sweeping for places a caller-supplied string reaches an
interpolated query while fixing #345.

DEENUU1:
## What this changes

The Google Drive sync connector took two caller-supplied strings at face value:
a file's **name**, which it used as a path component, and a source's
**`folder_id`**, which it interpolated into the Drive query. Both are closed
here, in one module — `backend/app/services/rag/remote_names.py` — because they
are one defect with two sinks.

Worth stating first, because it decides how bad each one is: **a remote file
name is attacker-controlled from this system's point of view.** Sharing a Drive
folder with people outside the organization is what folder sharing is *for*, so
anyone who can drop a file into a synced folder chooses the name, and the
attacker is not necessarily the tenant.

### #370 — a name is a label, not a path component

`dest_dir / file.name`, and `../../../../home/app/.ssh/authorized_keys` is a
legal Drive name. The file left the `TemporaryDirectory` the worker made for it,
landed wherever the worker's uid could write, and was then ingested from there.

`destination_within` reduces the name to its final component and then
**resolves the result and confirms it is a child of the directory**, rather
than cleaning the spellings we happened to think of. `..`, its
percent-encodings, its homoglyphs and a symlink already sitting in the
directory are one question after `resolve()`; an enumeration of separators is a
list that is always one entry short. A name that is no component at all (`..`,
`.`, `/`, empty, a NULL byte) is refused. Anything else lands inside as one
file with its spaces and its suffix intact — the pipeline routes on the suffix,
and a mangled name is a worse answer than a contained one.

**The destination is now `BaseSyncConnector.download_file`'s answer, not a
connector's.** It computes the path and hands it to a new abstract `_fetch`.
The two shipped connectors had decided this separately and disagreed — S3
reduced its key to one component, Drive did not — which is the argument for the
base owning it: a connector added later inherits an answer it cannot forget to
ask for. `tests/test_untrusted_remote_names.py` asserts no registered connector
overrides `download_file`.

### #369 — a folder id reaches a query language

`f"'{folder_id}' in parents"`, so `x' in parents or name contains 'salary`
closed the literal and appended its own clauses: a well-formed, wider query.
The id is now checked against what Google can issue — `[A-Za-z0-9_-]` — **where
the query is built**, which is the one funnel the configured folder and every
recursed sub-folder pass through, so a row written before the check is covered
too. `validate_config` asks the same function, so the route that accepted the
value is the one that answers rather than a sync log an hour later; the two
cannot disagree, because a value the allowlist accepts is a value nothing has
to escape. An allowlist rather than an escape for that same reason — an escape
leaves the next sink to remember.

### The escalation the issue asked about, taken as well

`_get_drive_service` fell back to `GOOGLE_DRIVE_CREDENTIALS_FILE` whenever
`service_account_json` was absent. That is what turned "widening your own query"
into a boundary crossing: a tenant's `folder_id` choosing what was listed under
the **operator's** service account, whose `drive.readonly` covers whatever the
operator was shared. The fallback is gone. `service_account_json` is required by
`CONFIG_SCHEMA`, so it only ever covered rows written before that field existed,
and this deployment has none. The setting stays for `rag-sync-gdrive`, which an
operator runs from their own shell. The `ValueError` that function raised is now
a `BadRequestError` naming the field.

## The siblings, swept

| | |
|---|---|
| `connectors/s3.py` | Had neither defect — `Path(key).name` at list time, no query interpolation — but now inherits the containment check, which is what refuses a key ending `/..` |
| `sources/google_drive.py` | **Had both.** `rag-sync-gdrive --folder-id` takes the id from the shell, and `download_file` wrote `dest_dir / name` |
| `sources/s3.py` | `Path(file_id).name` — contained already, now uses the same helper so there is one answer rather than two that agree by luck |

`sources/` is the legacy CLI pair. Its contract discovers the name mid-download
and so cannot be handed a path the way `BaseSyncConnector` can, so there each
implementation calls the helper and the base documents the requirement.

## Found while reviewing my own diff, fixed here

- **A `folder_id` that is not a string at all.** A source's config is
  `dict[str, object]`, so `{"folder_id": 1234}` is a body the route accepts.
  `validate_config` was stringifying before checking, and the same value
  reaching `_list_folder` was handed to `re.fullmatch` unstringified — a
  `TypeError`, not a refusal. `checked_drive_folder_id` now takes `object` and
  refuses a non-`str`. Commit `eeda0bd`.
- **Four more pages teaching the old contract** (`patterns.md`,
  `howto/add-rag-source.md`, two places in `architecture.md`) — each showed
  `dest_dir / file.name`, which is the defect written down as guidance. Commit
  `b5e1a97`.

## The one thing here that is not this branch's work

`make lint` is **red on `main`** at `16a31b9`: a `RET501` in the resolver stub
from #397 and a `RUF100` in the connectivity probe from #302, both merged during
the Actions outage. That is **#406**, and it has been filed five times over
(#406–#410) by parallel work hitting the same tree. It is fixed in its own
commit (`e37e47f`) because a branch cannot claim `make lint` otherwise — **if
another branch lands that fix first, drop this commit rather than resolving the
conflict.**

## How it was verified — CI could not run

**GitHub Actions is in a major outage** (incident opened 15:22 UTC): 150+ runs
queued, none in progress. Nothing below was answered by CI, and a pending check
on this PR is not a signal.

| | |
|---|---|
| `make lint-backend` | Green — ruff, ruff format, `ty check` (0 errors), both guards |
| `make db-check` | `No new upgrade operations detected` — against the real pgvector |
| `make docs-build` (`--strict`) | Green |
| `tests/test_untrusted_remote_names.py` | 45 tests, green |
| Coverage of `remote_names.py` | **100%**, branches included |
| `test_commands.py`, `test_coverage_gate.py`, `test_route_layering.py`, `test_supported_formats.py`, `test_ci_parity.py` + the new file | 104 passed — the blast radius of a changed connector contract |
| `python3 scripts/docs_drift.py` | No drift owed |
| **The whole backend suite** | **3449 passed, 6 skipped** in 17m47s. Slow because five agents were running suites in sibling worktrees at a load average of 40, not because of anything here |
| The integration suite | **260 passed** in 16m00s, against `pgvector/pgvector:pg16`. Slow for the same reason plus #215 — eight concurrent runs each rebuilding the schema per test, starved on I/O against the one Postgres |

The tests assert **the consequence** — what exists on disk afterwards, and
which query string the Drive client was handed — never that a helper was
called. They were confirmed to fail without the fix: with the containment check
stubbed out, 14 fail; with the identifier check stubbed out, the sub-folder test
recurses without bound, which is the defect seen from the other side.

`remote_names.py` is in `[tool.coverage.run] include` **and**
`[[tool.ty.overrides]] include`. The rest of `app/services/rag/` came with the
template and is reported rather than gated; this module is ours and is nothing
but refusals, so the branch never taken in normal operation is the whole of it.

## Review

The automated reviewer does not run (#311/#313) and the label does nothing, so
**this diff has not been through one.** I swept it myself: every place in the
RAG subsystem where a remote-supplied string reaches a path, a query or an
identifier, plus every doc page that names the changed contract. The two
findings above came out of that sweep.

## Checks

- [x] Tests cover the new behaviour, and were confirmed to fail without the change
- [x] Everything CI would run for this diff passes locally: lint, backend suite, integration suite, `db-check`, `docs-build`. CI itself could not answer
- [ ] n/a — no frontend file changed, so `test-frontend` and `build-frontend` are the jobs CI would skip for this branch anyway
- [ ] n/a — no migration; nothing about this touches a column

Docs updated in the same change: `file-processing.md` gains the section on what
a sync source is not allowed to decide, `configuration.md` and
`howto/configure-sync-sources.md` say the Drive credential is per-source and the
setting is the CLI's, and `howto/add-sync-connector.md`, `patterns.md`,
`howto/add-rag-source.md` and `architecture.md` teach `_fetch` rather than
`download_file`.

Closes #369, #370

### DIFF
--- a/.claude/rules/frontend.md
+++ b/.claude/rules/frontend.md
@@ -60,6 +60,12 @@ There is no `(marketing)` route group.
   translator, so it holds *keys* and the component translates at the point of use;
   a pure helper either answers with a key or takes `t`.
 - Keep components under ~100 lines; extract when they grow.
+- **A response with no `Cache-Control` is not a response nobody caches.**
+  `platform-proxy.ts` stamps `no-store` on anything the backend left unmarked: every
+  answer on this surface depends on a cookie, a permission set and the organization
+  header, and a list refetched right after a write must reach the server. A
+  hand-rolled route file owes the same header - the proxy is the only place that
+  applies it for you.
 - Do not hand-edit `src/lib/mcp-logos.generated.ts` — run `bun run gen:mcp-logos`.
 
 ## Permissions
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,11 +4,30 @@ name: CI
 # Work reaches `main` by pull request from a short-lived branch, so the pull
 # request is where these run. The push trigger covers the merge itself - a
 # branch CI ignores is a branch whose status badge means nothing.
+#
+# **`pull_request` carries no `branches:` filter, and that is the point of #359.**
+# That filter matches on the *base*, so a pull request stacked on another feature
+# branch - which is what two branches touching one file are told to do here - matched
+# no trigger at all and ran nothing. Not red, not pending: `gh pr checks` answered
+# "no checks reported" and the rollup was empty, which reads as "still starting" to
+# anybody who does not go looking. Four pull requests merged that way on 2026-08-06,
+# each verified only on somebody's laptop, and the gap closed only when the child was
+# retargeted to `main` after the parent merged - the moment nobody waits seven minutes
+# for a branch that has looked settled for a day.
+#
+# Two things make this cheap rather than a doubling of the bill. The `changes` job
+# below already skips the three expensive suites a diff cannot affect, and it reads
+# `pulls/{n}/files`, which is the diff against *that* pull request's own base - so a
+# stacked child is classified on its own changes rather than on its parent's. And the
+# `concurrency` group is keyed on `github.ref`, so a stacked branch cancels its own
+# superseded runs like any other.
+#
+# The `push` trigger stays pinned to `main`: a branch push is already covered by its
+# pull request, and widening it would run everything twice.
 on:
   push:
     branches: [main, master]
   pull_request:
-    branches: [main, master]
 
 # One run per branch. `CLAUDE.md` asks for a commit and a push per finished piece,
 # which is right - but with nothing cancelling, each push started a fresh 24-minute
@@ -41,6 +60,20 @@ concurrency:
 permissions:
   contents: read
 
+# **Every job carries a `timeout-minutes`, and #364 is why.** `changes` was the only
+# one that did, so the other seven inherited GitHub's default of 360 minutes: one
+# stalled job would have held a required status check for six hours with nothing in
+# this repository ending it sooner. Nothing has been observed to stall - this is a
+# bound on the tail, not a fix for something seen.
+#
+# Each is set several times the job's observed worst case rather than just above it.
+# The bound exists to end a stall, and a timeout tight enough to trim a legitimately
+# cold cache is a red build for a reason unrelated to the diff. Measured on run
+# 31116003994, a full matrix on `main`: changes 7s, lint 22s, Security Scan 14s,
+# docker 2m30s, docs 4m34s, test-frontend 5m08s, test 7m43s, e2e 8m01s.
+#
+# `backend/tests/test_ci_workflow.py` asserts that a job added later carries one too,
+# and that none is so generous it bounds nothing.
 jobs:
   changes:
     # Which of the three expensive suites this pull request can provably not
@@ -105,6 +138,7 @@ jobs:
     # tracked file rather than a subtree, and a suite that skips is only defensible
     # while something still looks at the whole tree.
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -163,6 +197,7 @@ jobs:
     # check is a pass, so that is a green merge button over a suite that never ran.
     if: ${{ !cancelled() && needs.changes.outputs.backend != 'false' }}
     runs-on: ubuntu-latest
+    timeout-minutes: 25
     services:
       postgres:
         # pgvector, not stock postgres: the RAG store runs
@@ -265,6 +300,7 @@ jobs:
     needs: changes
     if: ${{ !cancelled() && needs.changes.outputs.frontend != 'false' }}
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -317,6 +353,7 @@ jobs:
     # backend, so it is only skippable for a change that touches neither.
     if: ${{ !cancelled() && needs.changes.outputs.e2e != 'false' }}
     runs-on: ubuntu-latest
+    timeout-minutes: 25
     services:
       postgres:
         # pgvector, not stock postgres: the RAG store runs
@@ -494,6 +531,7 @@ jobs:
   security:
     name: Security Scan
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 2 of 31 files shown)


DEENUU1:
Reviewed this as a maintainer and pushed two fixes rather than leaving them for a round trip. Both are things the branch's own argument implies but does not carry out.

**The route-level refusal existed on one write path in three.** `create_source` asked the connector; `update_source` and `clone_source` wrote the config straight to the row. So a `folder_id` refused at creation could be patched in afterwards:

```
POST  /rag/sync/sources        {"folder_id": "1AbC"}                 -> 201
PATCH /rag/sync/sources/{id}   {"folder_id": "x' in parents or ..."} -> 200
```

Nothing leaked — the sink check inside the connector still caught it. But the docstring here and two doc pages say the value is *"answered by the route that accepted it rather than by a sync log an hour later"*, and for two routes in three that was untrue. `update_source` now judges the **merged** config, because a caller patches one field and the connector runs the whole thing.

**The S3 connector still had the deployment-wide credential fallback this branch removed from Drive**, and the patch hole made it reachable. An empty `access_key_id` is stored verbatim by `_encrypt_config` (it skips falsy values), `or settings.S3_RAG_ACCESS_KEY` resolves to `None` because that setting defaults to empty, and boto3 then falls through to the container's own credential chain — so the caller's `bucket` chose what was read under whatever the task role could reach. Removed with the reasoning in the docstring. The endpoint and region still fall back; neither names a principal. `sources/s3.py` keeps reading the settings, exactly as the Drive CLI keeps `GOOGLE_DRIVE_CREDENTIALS_FILE`.

**Four escape assertions proved nothing.** They used two or three `..` segments, so under the old code the write landed *above* `tmp_path` and `assert not (tmp_path / "x").exists()` was true either way. One level now, so the negative assertion is the one doing the work. The tests still failed on revert before this — but on the equality above, not on the line meant to prove no escape happened.

Verified locally: `ruff` clean over the diff, **3448 passed / 10 skipped**, and each new test checked against the unfixed code (restoring the S3 fallback fails it).

Everything else I checked and could not fault: both Drive query sinks funnel through `checked_drive_folder_id` including the recursed sub-folder id; `_DRIVE_ID` anchors strictly and refuses a trailing newline and a Cyrillic homoglyph; `destination_within` refuses `..`, `.`, `/`, empty and NUL and lands everything else inside as one component; the TOCTOU window is a fresh 0700 temp directory nothing else can reach; and `test_no_shipped_connector_overrides_the_destination` genuinely enumerates the registry.

**CI has not run and will not** — Actions has been out since yesterday afternoon. Everything above is local.

Issue #434, opened 2026-08-06
The sync wizard's target collection picker cannot be reached by any call site


DEENUU1:
`ScheduleStep` in `components/rag/sync-source-wizard.tsx` renders a target
collection picker behind

```tsx
// sync-source-wizard.tsx:628
{!defaultCollection && collections.length > 0 && (
```

**No call site can satisfy both halves**, so the picker, its label, its helper
line and the `label` field of the `collections` prop are unreachable in the
product.

**Pre-existing; found reviewing the #341 branch**, which rewrites that
`SelectItem` and therefore had to establish whether anybody sees it.

## What breaks

Three call sites, and that is all of them:

| `file:line` | `collections` | `defaultCollection` | Picker renders |
|---|---|---|---|
| `app/[locale]/(dashboard)/rag/page.tsx:481` | `collections.map((c) => ({ name: c.name }))` | `selected` | no — `selected` is `chosen \|\| collections[0]?.name \|\| ""`, so it is only `""` when the list is empty, and then the second half fails |
| `app/[locale]/(dashboard)/kb/[id]/page.tsx:728` | `[{ name: kb.collection_name }]` | `kb.collection_name` | no — always set |
| `components/kb/reusable-integrations.tsx:137` | `[]` | — | no — empty list |

Nothing passes `label` either, so the mono `(collection_name)` qualifier beside
a display name has never been drawn.

The `kb/[id]` call site is the one that shows this is not merely dead but a
**lost feature**: a knowledge base has a human `name` and a physical
`collection_name`, and the row it builds discards the first
(`[{ name: kb.collection_name }]`). `label` is exactly the field for it.

## Sequence

1. `rg -n "<SyncSourceWizard" frontend/src` — three call sites.
2. Each passes a truthy `defaultCollection`, or an empty `collections`.
3. Add a sync source from any of the three: the wizard's schedule step has no
   target collection control, and never had one.

## How you would know it was fixed

One of two, and the choice is a product call rather than a cleanup:

- **Delete it** — the block, `targetCollection`, `selectCollectionOptional`,
  `leaveEmptySaveAs`, and `label` from the `collections` prop type. The repo's
  own rule: if a branch cannot be reached, delete it. This is the smaller change
  and loses nothing anybody has today.
- **Reach it** — decide the wizard may retarget, and stop passing
  `defaultCollection` from at least `rag/page.tsx`, where a person is looking at
  a list of collections and might reasonably want a different one. If so,
[...]

Issue #425, opened 2026-08-06
Keys extracted into en.json and never read, while the component keeps the literal


DEENUU1:
The extraction pass that produced `messages/en.json` sometimes lifted a
string into the catalog and **left the component reading the literal**. The
key is there, correct, and unused; the words on screen are the hardcoded copy
beside it. `scripts/check_i18n.py` sees a valid catalog and a line it cannot
parse, and reports both halves clean.

Worse, one sentence was split across the boundary: its head went into a key,
its tail went into a *second* key, and the middle stayed in the JSX. The tail
key's value begins with a full stop.

**Pre-existing; found working #341**, whose seven select call sites sat on
three of these lines.

## What breaks

**A sentence in three pieces** — `components/agents/workspace-section.tsx:372`:

```tsx
<p className="text-destructive text-xs">
  This connection no longer allows <span className="font-mono">{runtime}</span>
  {t("pickOneDoesAgent")}
</p>
```

```json
"agents.connectionNoLongerAllows": "This connection no longer allows",   // never read
"agents.pickOneDoesAgent": ". Pick one it does, or the agent fails on its first tool call."
```

In `pl` this renders as English, then an alias, then an English clause opening
with a stray full stop. **A catalog value beginning with punctuation is the
signature**: no sentence starts with `.`, so any such value is half of one.

**A key that exists, unused, beside its own hardcoded words:**

| `file:line` | On screen | The key already holding it |
|---|---|---|
| `components/rag/sync-source-wizard.tsx:327` | `Continue` (the wizard's Next button) | `rag.continue` = `"Continue"` |
| `components/agents/exposures-panel.tsx:242` | `" (inactive)"` | `agents.inactive` = `"inactive"` |

Both keys were reachable and neither was read. The bot picker's `(inactive)`
is the one a person actually hits: it is how somebody is told, at the moment
of binding, that the handle they are about to publish to answers nothing.

## Sequence

1. `python3 scripts/check_i18n.py` → "No hardcoded copy in frontend/src."
2. Sign in with `pl`. Builder → Channels on an agent with an inactive bot; the
   picker says `(inactive)` in English.
3. Builder → Workspace, on a spec naming a runtime the connection dropped: an
   English clause, the alias, then `. Pick one it does, …`.

## How you would know it was fixed

Two things, and the second is what stops this recurring:

1. No key in `en.json` is unread while a component hardcodes its value, and no
   value begins with punctuation.
[...]

Issue #407, opened 2026-08-06
main does not pass make lint-backend: two ruff errors merged while Actions was down


DEENUU1:
## What breaks

`main` does not pass its own lint. `make lint-backend` on `16a31b9` (the `0.0.59`
release commit) reports two ruff errors, so every branch cut from `main` starts red
on a gate it did not break:

```
RET501 backend/tests/integration/test_vector_store_reserved_names.py:42
       Do not explicitly `return None` in function if it is the only possible return value

RUF100 backend/tests/test_migrations.py:99
       Unused `noqa` directive (non-enabled: `BLE001`)
```

Both are deterministic under the locked ruff (`0.16.1` in `backend/uv.lock`, which is
what CI installs), so CI's `lint` job would report exactly the same two.

## How it got in

The GitHub Actions outage of 2026-08-06 (incident opened 15:22 UTC — runs failing or
delayed in starting; 150+ queued, none in progress). Two pull requests merged with
their whole check list **pending** rather than green:

- `5705503` — `test(rag): stub a resolver in the reserved-names integration test`
  (#397), which introduced the `RET501`.
- `8e7a08c` — `test(migrations): create the database the chain is tested on` (#302),
  which introduced the `RUF100`.

`gh pr checks 397` and `gh pr checks 302` still answer `pending` for `lint` on both.
Neither is a defect in the change itself — a lint error is exactly what the gate
exists to catch before a merge, and the gate did not run.

Worth recording separately, because it outlives the outage: the ruleset that requires
six green contexts was necessarily off for those two merges (`docs/branching.md`, "The
escape hatch"). That is the documented procedure and there is nothing to fix about it,
but the price of using it is that `main` can acquire a failure nothing reports until
somebody runs the target by hand — which is how this was found, three hours later.

## Sequence

1. `git switch main && git pull`
2. `make lint-backend`
3. Two errors, in files the branch never touched.

## How you would know it was fixed

`make lint-backend` passes on `main`. Restoring that means keeping both explanations,
not just deleting the lines ruff points at: the stub's docstring is *about* answering
`None`, and the `noqa` carries the reason the broad `except` is deliberate. Both
survive as prose.

## Where it sits

Cluster: **Tests, CI, dependencies** on #168. Nearest neighbours #143 and #165 — a
gate that reads green because the thing is not in it — though this is the milder
relative: the gate is correct and simply did not run. Closest live issue is #359,
which is the same sentence about a different cause (a checks list that is empty reads
[...]

Issue #365, opened 2026-08-06
Create knowledge base cannot say the embedding model list failed to load


DEENUU1:
The Embeddings section of Create knowledge base says "Loading models…" whether
`/rag/embedding-models` is still in flight or has failed, and there is no third
state. A request that answers 502 leaves the section saying it is loading for as
long as the dialog is open, with no model select at all — and the collection is
created on the deployment default with nobody told the choice was never offered.

`.claude/rules/frontend.md`: "An empty page is ambiguous … 'No skills yet' and
'the request answered 502' are the same pixels."

## What breaks

`frontend/src/components/kb/create-kb-dialog.tsx:77` reads the list and never
reads `isError`:

```tsx
const { data: embeddingModels } = useQuery({
  queryKey: ["rag", "embedding-models"],
  queryFn: () => apiClient.get<EmbeddingModels>("/rag/embedding-models"),
  staleTime: Infinity,
});
```

`create-kb-dialog.tsx:222` then branches on `embeddingModels === undefined`,
which is both "not yet" and "never". With `staleTime: Infinity` and the client's
single retry, "never" is where it stays until the dialog is closed and reopened —
and reopening re-mounts against the same cached rejection.

The embedding model is frozen at creation (the vector column is created at that
model's width), so this is the one choice in the dialog nobody can revisit. A
failure that silently removes it is worth more than a spinner.

## Sequence

1. Make `GET /rag/embedding-models` fail — stop the API mid-session, or a
   deployment where the route 500s.
2. Knowledge bases → Create → Embeddings.
3. "Loading models…", indefinitely. No select, no message, and Create still works.

## How you would know it was fixed

Three states rather than two: in flight, answered, and refused. The refused one
says the list could not be read and that the collection will be created on the
deployment's default — which is what happens either way, and is fine to be told.
An integration test with the query rejecting, alongside the two in
`create-kb-dialog.integration.test.tsx` that already cover the other two states.

## Where it sits

Cluster: **frontend surfaces** (#168). **Pre-existing; found reviewing PR #360**,
which fixed #328 in this exact block — #328 is the nearest neighbour and this is
not a duplicate of it: #328 was the select clobbering its own value to `""`, this
is the section having no way to say the list never arrived. Before #360 the same
hole existed behind a disabled trigger showing the same words; #360 made it an
explicit branch, which is what made it visible. Its siblings in the same section
are #305 and #306.

DEENUU1:
[...]

PR #435, opened 2026-08-06
Keep a comparative badge in the list, and say when the model list failed

### LINKED ISSUE #341: Seven select triggers repeat a badge that only means something in the list
Radix renders a `SelectItem`'s `ItemText` children into `SelectValue`, so **the
closed trigger draws whatever the selected option drew**. That is a feature for
anything describing the option itself — a brand mark, an avatar — and a defect
for anything that only means something *against the other options*: a tick, a
count, a warning, an explanatory second line. In a trigger there is nothing to
compare it with, and it reads as a claim about the selection.

#304 hit this with the "this provider already has a key" tick and gave
`components/ui/select.tsx` a `trailing` slot rendered outside `ItemText`
(PR #330). These are the other call sites, none of which use it yet.

## Where

| | | What the closed trigger draws |
|---|---|---|
| `components/sandboxes/runtime-field.tsx:97-118` | alias + `BUILDS` badge + amber `NOT ON THIS HOST` + a description line | all four, stacked |
| `components/teams/members-table.tsx:112-119` | role + the per-role explanation | two lines inside a `h-7 w-36` trigger |
| `components/agents/embeds-panel.tsx:180-195` | label + the sentence distinguishing it from the other option | both |
| `app/[locale]/(dashboard)/rag/page.tsx:517-520` | `{col.name} · {n} vectors` | the count — which the page **already prints beside the trigger** at `:525-528` |
| `components/agents/exposures-panel.tsx:240-243` | `… (inactive)` | the state badge |
| `components/rag/sync-source-wizard.tsx:639-649` | label + a mono `(name)` qualifier | the qualifier |
| `components/agents/workspace-section.tsx:347-352` | alias + ` · {mem_limit}` | milder — a property of the option itself |

`runtime-field.tsx` is the one worth reading first, because the workaround is
already in the file. Its own comment at `:87-90` records that the trigger had to
be given `min-w-0` and `truncate` because "the trigger grew to fit it and pushed
the dialog wider than the viewport". That is this defect, papered over at the
trigger instead of fixed at the item.

Checked and **not** in this list: `components/teams/org-switcher.tsx:84` (a
`DropdownMenu`, which has no `ItemText` and mirrors nothing),
`components/agents/version-history.tsx:73` (a menu whose `value` is pinned to
`""`, so it only ever shows its placeholder), and
`app/[locale]/(dashboard)/admin/conversations/page.tsx:316` (an avatar, which
describes the option — the same argument that puts the brand mark *in* the row).

## What to do

Move the comparative part into `trailing`, which is where #330 put the tick and
the "deployment default" badge:

```tsx
<SelectItem value={x} trailing={<Badge className="ml-auto shrink-0">…</Badge>}>
  <TheRow />
</SelectItem>
```

For `runtime-field.tsx` the second line is a description of the runtime, so it
belongs in `children`; the two badges do not.

Select the affected option, close the select, and read the trigger: it names the
option and nothing else. `provider-row.test.tsx` in #330 has the shape of the
test — assert the badge is absent from the trigger and present on the option.
Once these are done, `runtime-field.tsx`'s `truncate` workaround can be revisited.

## Where it sits

Cluster: frontend surfaces (#168). Found sweeping the siblings of #304 / PR #330,
which is where the `trailing` slot comes from — this issue is only worth doing
after that lands. Not a duplicate of #139 (four near-copies of one row); this is
one mechanism misused in seven places, and each fix is two lines.

DEENUU1:
Two select-behaviour defects in the same component family.

## #341 — a badge that only means something in the list

Radix renders a `SelectItem`'s `ItemText` into `SelectValue`, so **the closed
trigger draws whatever the selected option drew**. That is the mechanism that
gets a brand mark onto the trigger for free, and the same mechanism that puts a
tick, a count or a warning there — where there is nothing left to compare them
with, and they read as a claim about the selection.

v0.0.49 (#304 / PR #330) gave `SelectItem` a `trailing` slot rendered outside
`ItemText` for exactly this. These are the seven call sites that never used it.

| | What the closed trigger used to draw |
|---|---|
| `sandboxes/runtime-field.tsx` | `BUILDS` and an amber `NOT ON THIS HOST` — the field said "not on this host" beside the runtime it was about to save |
| `teams/members-table.tsx` | the per-role blurb, wrapped inside `h-7 w-36`, explaining a role nobody is choosing any more |
| `agents/embeds-panel.tsx` | the sentence telling the two auth modes apart, left describing the one that remained |
| `agents/exposures-panel.tsx` | `(inactive)`, which read as the state of the binding rather than of the bot |
| `agents/workspace-section.tsx` | a memory cap, as a bare number |
| `rag/sync-source-wizard.tsx` | the physical name qualifying a label, left disambiguating a set of one |
| `(dashboard)/rag/page.tsx` | the vector count — four inches from the same number the page already prints for the chosen collection |

`runtime-field.tsx`'s description stays in `children`: it describes the option
wherever it is drawn, so the trigger is right to inherit it. That also means
its `min-w-0` / `truncate` workaround **stays** — the issue suggests revisiting
it once these land, and it is still load-bearing, because a description is
still a sentence and the trigger can still outgrow the dialog.

## #365 — "Loading models…" for a list that was never coming

The Embeddings section of Create knowledge base said the same thing whether
`/rag/embedding-models` was in flight or had answered 502, and there was no
third state. With `staleTime: Infinity` and the client's one retry, refused is
where it stays for as long as the dialog is open — reopening re-mounts against
the same cached rejection.

The embedding model is frozen at creation (the vector column is made at that
model's width), so it is the one choice in this dialog nobody can revisit.
Create still worked, still produced a collection on the deployment's default,
and nobody was told the choice had never been offered. Reading `isError` splits
the two, and the refused branch says which model the collection gets anyway.

## What the maintainer pass found, and what it cost

The automated reviewer is off (#311 / #313), so the sweep was done by hand and
by subagent. It found one **real regression in this branch**, which `13e5d9f`
fixes:

**A refused runtime became invisible once chosen.** Moving the amber
`NOT ON THIS HOST` badge out of `ItemText` is right, and it was also the only
warning in `runtime-field.tsx` — `connection-dialog.tsx` saves
`default_runtime` without checking it against the allowlist. After the move a
person could probe the host, pick an alias it had just refused, close the list
and save, with nothing on screen dissenting. The field now says so **under**
the control, about the alias actually selected, where it is a sentence rather
than a badge with nothing to compare against.

That also repairs an accessibility loss the same mechanism causes everywhere:
Radix names an option by its `ItemText` alone, so anything in `trailing` is no
longer in the option's accessible name. A screen-reader user had lost the
warning outright. Judged case by case:

- **`runtime-field`** — real loss, fixed above, because nothing else said it.
- **`exposures-panel`** (`inactive`) — accepted. The exposure row that appears
  immediately after binding says `Paused`, so the fact is one beat away rather
  than gone, and #341 asks for exactly this badge to leave the trigger.

Two smaller things fixed in the same commit: `embeds-panel`'s popover is now
capped at `min(26rem,90vw)` (stacking two lines made a ~30-character column;
side by side it is one ~60-character row, and Radix sizes a popper to
`max-content`), and a comment in `create-kb-dialog.tsx` that claimed reopening
re-mounts against a cached rejection — `retryOnMount` defaults to true, so it
refetches; what matters is that after one retry the query is settled in error
for the life of *that* dialog.

**Seen and deliberately not fixed:** whether the description in a
`runtime-field` option can now be squeezed by a long `trailing` badge. It
cannot be reproduced in jsdom, the mechanism predates this branch, and the only
fix touches `components/ui/select.tsx` for all seven call sites — changing a
shared primitive on an unverified hypothesis is the worse trade.

## Also fixed here, and filed

Three untranslated strings were **inside the JSX being rewritten**, so they are
fixed rather than left: `" (inactive)"` in the bot picker, and two vector
counts built as English with an `s` glued on (now ICU `plural`). One more —
`sync-source-wizard.tsx`'s Continue button — was hardcoded while `rag.continue`
= `"Continue"` sat unread in the catalog.

**#434** came out of the review pass and is not fixed here: the sync wizard's
target collection picker needs `defaultCollection` absent and `collections`
non-empty, and **all three call sites** pass a truthy `defaultCollection` or an
empty list — so that one of the seven is a control nobody can reach, and this
branch's fix to it is correct and **latent**. The new test file mounts the
wizard the way no call site does and says so in its docstring rather than
implying it proves live behaviour. #434 has both ways out (delete it, or reach
it and pass `label: kb.name`, which is the display name `kb/[id]/page.tsx`
currently throws away).

The Continue button is a signature rather than a one-off, and is filed as **#425**:
keys extracted into `en.json` and never read while the component keeps the
literal, including one sentence split across two keys and a JSX literal
(`agents.connectionNoLongerAllows` held its head, `agents.pickOneDoesAgent`
held its tail — and a catalog value beginning with a full stop is how you find
the rest of them). The sweep and the two cheap guard rules are #425's; the
three instances this branch sat on are done here.

**Not** taken: `rag.flexItemsCenterGap2` and `rag.flexH9W2`, two Tailwind class
lists in the same file. They are **#348**'s, which a sibling branch is working
while editing `messages/en.json`. They were inlined and then deliberately put
back (`efd2d5d`) — noted on #348. This branch's `en.json` diff is four lines,
all its own.

## How the tests prove it

The trap this issue is: an assertion that a badge is *absent from the trigger*
also passes against a component that stopped drawing the badge at all. So every
regression test asserts **both** halves — the badge **is** inside the list row
and **is not** on the trigger — and the trigger half is `not.toHaveTextContent`
rather than `queryByText`, because the old markup sometimes glued the badge to
the label in a single text node, which `queryByText` would not have matched.

**Proved by reverting, not by reading.** Each of the eight changed source files
was reverted to `origin/main` on its own and its test file re-run:

| Reverted | Failing tests |
|---|---|
| `runtime-field.tsx` | 1 |
| `members-table.tsx` | 1 |
| `embeds-panel.tsx` | 1 |
| `exposures-panel.tsx` | 2 |
| `workspace-section.tsx` | 1 |
| `sync-source-wizard.tsx` | 2 |
| `(dashboard)/rag/page.tsx` | 2 |
| `create-kb-dialog.tsx` | 3 |

One thing that changed for existing tests: Radix sets `aria-labelledby` on an
option pointing at its `ItemText`, so a badge in `trailing` **leaves the
accessible name**. Three tests matched a badge through
`getByRole("option", { name: /…/ })` and now match it with
`within(option).getByText(…)` instead.

Two new test files: `components/rag/sync-source-wizard.integration.test.tsx`
(drives the wizard to its schedule step) and
`app/[locale]/(dashboard)/rag/page.integration.test.tsx`. The second is a page
test, which `vitest.config.ts` says are deliberately last for *coverage* — it
is here because it is the only place the seventh call site can be exercised,
not to move a number, and the page is not in the coverage `include` list.

## Verification — CI could not run

**GitHub Actions is in a major outage** (incident opened 15:22 UTC; 150+ runs
queued, none in progress). No check on this pull request has executed, and a
pending check here is not a signal. Run locally instead, from the repository
root:

| | |
|---|---|
| `make lint-frontend` | eslint, prettier, tsc — clean |
| `make test` | 3404 passed, 6 skipped; backend coverage 100.00% |
| `make test-frontend-cov` | 100% statements / lines / functions, **97.58%** branches (gate 97.5) — see the re-run note |
| `make build-frontend` | clean |
| `make db-check` | "No new upgrade operations detected" |
| `make docs-build` | clean (`--strict`) |
| `make audit` | no known vulnerabilities |
| `python3 scripts/check_i18n.py` | "No hardcoded copy in frontend/src" |
| `make check` | see below |

**The coverage run needed a second attempt, and not because of this branch.**
The first `make test-frontend-cov` reported 8 failures across
`chat-parts`, `file-preview`, `command-palette` and `create-skill-dialog` —
none of them files this branch touches — all of them
`Error: Test timed out in 5000ms`. The machine was at load 40 with eight
worktrees on one Postgres, which is the trap `.claude/rules/testing.md` names:
coverage instrumentation is slow enough to trip the 5s `testTimeout`.
`chat-parts` passes **26/26** run alone, and the whole gate passes clean at
load 5. Another agent hit the same two chat specs independently.

**`make check` cannot go green on any branch right now, and not because of
this one.** `make lint-backend` is red on `main` — two ruff findings
(`RET501`, `RUF100`) in `tests/test_migrations.py` and
`tests/integration/test_vector_store_reserved_names.py`, merged during the
Actions outage. Tracked as **#407** and fixed in **#426**; deliberately not
fixed and not re-filed here, because several branches carrying the same
two-line fix is several conflicts. The two fixes were applied to the working
tree *locally* to let the rest of `make check` run, and were **not**
committed — this branch contains no Python.

With those two applied, `make check` got through `lint`, `test` (3404 passed,
100% coverage) and `db-check` before being SIGTERMed during
`test-frontend-cov` — a killed process, not a failing one. The remaining
targets were then run individually, all clean, and are listed above. So every
job `check` runs has answered on this tree; it just did not answer in one
invocation.

Closes #341, #365


DEENUU1:
Coordination note for the merge, from #445 (#356/#434/#436), which landed after this.

**You collide in the sync wizard.** #445 rewrites the same `SelectItem` in `sync-source-wizard.tsx` and adds a test file at the same path. Its author's reading — which I pass on rather than endorse, since it is their diff: **your wizard fix has nothing left to fix once #445 lands**, because #445 makes the picker reachable in the first place and deletes the `label` qualifier your hunk edits.

You already flagged that one of your seven `trailing` fixes was latent and filed it as #434; #445 is what makes it live, and it takes the same file further.

So whichever merges second should drop the wizard hunk rather than resolve it line by line. Your `Continue` string is untouched by them, so only that one component is in question.

Issue #206, opened 2026-08-05
Wire the Logfire deep link end to end - logfire_trace_id is documented and never written


DEENUU1:
## What breaks

`AgentRunRead.logfire_trace_id` is documented as *"Deep-link into the full trace; spans are not duplicated here"* (`backend/app/schemas/agent_run.py:29`) and is **always `null`**. The public API promises a capability it has never delivered.

The column exists on `agent_runs`, and `AgentRunner.finish()` accepts a `logfire_trace_id` parameter (`backend/app/services/agent_runner.py`), which `finish_run` writes only `if logfire_trace_id is not None` (`backend/app/repositories/agent_run.py:78`). No caller passes it — not the playground, not web chat, not the HTTP API, not a channel, not a schedule. So the guard is always false and the column is always `NULL`.

## Sequence

1. Run an agent on any surface.
2. `GET /api/v1/runs` — the row comes back with `"logfire_trace_id": null`.
3. Read the field's own description, which says it is a deep link into the trace.

Any client written against the documented contract — including a customer reading the OpenAPI schema — builds a link that can never resolve.

## The decision, and what it now costs

This was originally filed to be **removed**: #202's design overturned the "Logfire is the drill-down" approach on four findings, and the run detail view it proposes works from `agent_runs` / `messages` / `tool_calls` with no Logfire in the path.

That decision has since been reversed — the Logfire deep-link is wanted, so this field has to be **wired rather than deleted**. Three pieces, and only the first is small:

1. **Write the id.** Every caller of `AgentRunner.finish()` passes the current trace id. Cheap, and worth doing regardless of the rest, because a stored trace id is useful to anyone with Logfire access even without a URL in the product.
2. **Store a project URL.** `LOGFIRE_TOKEN` is a *write* credential (`app/core/config.py:43`); a deep link needs an organization and project slug, which is a new setting.
3. **Per-agent redirection.** `ObservabilitySpec.token_secret_id` (`app/agents/spec.py:311`) exists precisely so a client's agent traces into the client's own Logfire project. Linking correctly therefore means a project slug **on the spec**, and a `SPEC_VERSION` bump.

[...]

PR #202, opened 2026-08-05
feat(activity): rebuild /runs on our own rows, not on Logfire

OchnikBartek:
## Summary

Stage 1 of #45: a written plan and a self-contained HTML mockup for the Activity
page, reviewed before any product code is written. Three files, no product code.

The central decision is recorded twice on purpose. §1 argued for dropping the
Logfire deep link on four findings; review kept it (#206). Both halves are in the
document, because a plan that deletes the case it lost is one nobody can audit.
What this page refuses either way is to *depend* on Logfire: the drill-down is a
run detail view built on `agent_runs`, `messages` and `tool_calls`, so no panel
goes blank on a deployment that never set `LOGFIRE_TOKEN`.

## Related Issue

Closes #45. Verified against the tree rather than from memory, entry by entry:
#45's own Done-when list is met, and the two plan items that went beyond it are
now their own issues rather than a paragraph that disappears with this merge.

**Closed since the last review of this branch:**

1. **A 502 per tab, asserting the error state rather than the empty one.**
   `tab-failures.integration.test.tsx` fails one of the three requests with the
   other two served, because they have to fail *independently* — one page-wide
   error state would hide a Spend tab that is answering perfectly well. Each tab
   now has its own `ErrorState` and a retry proven to reach the server rather
   than re-render a cached rejection.
2. **Approve/Reject and the Approvals tab absent without `approvals:decide`** —
   `80d84a7`. This was a live defect, not only a missing test: the page computed
   `canDecide` and gated the two buttons with it while the tab, its position as
   `defaultValue` and the figure above it rendered unconditionally, so a member
   without the permission read a 403 as *"Nothing waiting · Agents are running
   without needing you."* `useApprovals` also stops polling a 403 every thirty
   seconds. Four assertions on the refusal and one positive control; verified
   against the un-gated page, where the four fail.
3. **The three tabs as separate components** — `05338ad`. Five components under
   `components/runs/`, none over 100 lines, page down from 358 to 98.

**Two plan entries are out of #45's Done-when and are now filed:**

- **#489** — the version strip's `completed / total`, which has to equal the
  dashboard's Outcomes donut over the same rows (§8a.4). Blocked on that donut:
  it is on `feat/dashboard-page-demo`, not `main`, and writing the agreement test
  against a component nobody can see means guessing at the arithmetic the test
  exists to pin.
- **#490** — a route that reads a run's transcript, **authorized rather than
  owned**. `runs.py` has six routes and none answers "the steps of this run", so
[...]

Issue #16, opened 2026-08-01
Meter the web chat surface so knowledge searches are billed


DEENUU1:
**Severity: Medium** · AUD-014

## What's wrong

`agent_chat.py:247` opens `async with prepared.built.agent.iter(...)` with **no `metered_by(...)`**, unlike `agent_runner.py:722-732`.

## Where

- `backend/app/services/agent_chat.py:247`
- `backend/app/services/rag/embeddings.py:81-95`
- `backend/app/agents/capabilities/budget/_capability.py:216-229`

## How it fails

A user in the web chat asks a knowledge-enabled agent a question. `search_documents` embeds the query; `record_ambient_usage` reads `_active_ledger`, finds `None`, and drops the spend on the floor.

The run's `cost_usd` under-reports, `cost_is_partial` is **not** set (so the UI shows no `+`), and the organization's monthly total never sees it — nor does `ingestion_spend`, which only the worker writes.

`embeddings.py:83-85` says this "used to be dropped on the floor, which made every knowledge search invisible to the monthly budgets". It still is, on the product's primary surface.

`rg metered_by app` finds three worker call sites and `agent_runner._run`. `agent_chat.py` is absent.

## Fix

Wrap the block in `with metered_by(prepared.built.ledger):`. Better: move the meter into `prepare`/`PreparedRun` so a surface cannot forget it — the same reasoning that put budget resolution in the runner.

## Acceptance criteria

- [ ] A web-chat knowledge search's embedding cost lands on the run
- [ ] Regression test in `tests/test_agent_chat.py`: drive `ChatAgentRunner.run` with a stubbed agent whose iteration calls `record_ambient_usage`, assert the value reaches `finish`'s `cost_usd`

PR #458, opened 2026-08-06
feat(agents): record where a run's trace went

### LINKED ISSUE #407: main does not pass make lint-backend: two ruff errors merged while Actions was down

OchnikBartek:
`AgentRunRead.logfire_trace_id` has been documented as *"Deep-link into the full
trace"* since the baseline and has been `null` on every run ever recorded.
`finish_run` writes the column only `if logfire_trace_id is not None`, and no
caller ever passed one — the guard had nothing to guard. Any client written
against the OpenAPI schema built a link that could not resolve.

## Changed

Filling the id is only half a link: a Logfire URL names a project, and the only
thing this platform holds is a **write** token, which carries no project name.

- **`trace_location(spec)`** — one function, four outcomes, all tested. It answers
  both halves together because neither is useful alone.
- **`LOGFIRE_ORGANIZATION` / `LOGFIRE_PROJECT`** settings.
- **`ObservabilitySpec.project`**, `SPEC_VERSION` 7 → 8, YAML export/import.
- **`agent_environments.logfire_project`** — the third level of the chain. The
  issue names one level of redirection; there are two, and the environment's is
  the one that actually runs.
- **`agent_runs.logfire_project`**, written beside the trace id.

**The project follows the token, never the deployment.** An agent redirected to a
client's Logfire exports *there*, so pairing its traces with the operator's slug
links into a project those traces are not in — worse than the no-link a missing
slug gives. That is why the environment merge stopped falling through field by
field for this one field: whichever level supplies the token supplies the project
with it.

Both facts are **recorded on the run** rather than resolved at read time, for the
reason `provider` and `model_label` are on the row: they say where the run
actually went, and re-deriving them would relabel every historical run the day
somebody repoints an agent.

The id is taken in `finish` rather than passed by each surface — the same argument
as the spend meter in #16 (PR #454): a surface that has to remember is a surface
that will not, and the failure is silent. `finish` runs from every surface's
`finally`, so this reaches the failed run too. The unused `logfire_trace_id`
parameter `finish` used to accept is gone.

## One correction to the issue's premise

The issue says a deployment with no `LOGFIRE_TOKEN` produces an all-zero trace id,
so `None` should be stored there. **It does not.** `send_to_logfire="if-token-present"`
builds real spans with real ids and drops them, so an id read off the span there
is a syntactically valid identifier naming a trace that was never exported
anywhere — the same lie as the null column, written the other way round.

The conclusion holds, so the id is recorded only when something is actually
exporting. All-zeros does happen, for a different reason — no active span at all,
in a CLI command or a worker path nothing instruments — and is refused too. Both
are tested against a real OpenTelemetry span rather than a mocked one.

## Testing

`tests/test_trace_location.py` (7) and `TestRecordingWhereTheTraceWent` in
`tests/test_agent_runner.py` (4), plus three on the environment/spec project
pairing. Between them:

- the id is the 32 hex characters a Logfire URL wants, read off a real span;
- **a `failed` run records its trace** — explicitly, since that is the run
  somebody wants it for;
- a token with no slug still records the id (link absent, schema honest);
- a deployment with no Logfire records neither;
- a redirected agent never borrows the deployment's slug, and an environment that
  redirects without a project of its own links nowhere rather than wrongly.

`make test` — 3416 passed, 100.00% on the platform layer. Migration applied,
downgraded to base and re-applied against a throwaway database; `alembic check`
reports no drift.

**`test_an_unknown_key_is_refused` used `project` as its stand-in for an unknown
key**, which is now a real field. It now uses one that is not, and a new test
loads a verbatim `spec_version: 7` document to prove stored specs still validate.

## What is not here — `Refs`, not `Closes`

The issue's last checkbox is the link in **Activity's run detail view**. Those
files belong to `feat/activity-page` / PR #202, being written in another session;
editing them here would guarantee a conflict. This PR gives that view everything
it needs — `logfire_trace_id` and `logfire_project` on `AgentRunRead` — and stops
there. **#206 stays open with that box unticked.**

**Migration number collides with #457.** Both this branch and
`fix/expire-stale-approvals` add a `0010_*` revision on top of `0009`. Whichever
merges second needs its revision renumbered and its `down_revision` repointed;
there is no other interaction between them.

**This branch also carries `test: drop two lint errors main has been failing on`**
(#451, `Closes #407`) — `main` does not pass `make lint-backend`, and the
pre-commit hook made committing impossible without it. That commit drops out of
this diff once #451 merges.

Refs #206

### DIFF
--- a/backend/alembic/versions/0010_logfire_project.py
+++ b/backend/alembic/versions/0010_logfire_project.py
@@ -1,0 +1,51 @@
+"""Which Logfire project a trace went to
+
+Revision ID: 0010_logfire_project
+Revises: 0009_align_index_names
+Create Date: 2026-08-07
+
+`agent_runs.logfire_trace_id` has existed since the baseline and has never been
+written. Filling it is only half a deep link: a Logfire URL needs the project as
+well, and a *write* token - which is all the platform ever holds - carries no
+project name. So the slug has to be configured, and it has to be configured
+wherever the token is, or a link points into a project the trace never reached.
+
+Two columns for two levels of that:
+
+`agent_environments.logfire_project` sits beside the token that column already
+holds. An environment redirecting its runs to a client's project supplies both
+halves or neither; it never borrows the spec's slug, because that would pair one
+project's traces with another project's URL. (The third level, an agent's own
+spec, is a spec field rather than a column - `ObservabilitySpec.project`, spec
+version 8.)
+
+`agent_runs.logfire_project` records where the trace actually went, the way
+`provider` and `model_label` record what actually ran. Resolving it at read time
+instead would relabel every historical run the day somebody repoints an agent -
+links into a project those traces are not in.
+
+Both nullable, no backfill and no default. Every existing run has a null trace id
+and therefore no project to name; inventing the deployment's current slug for
+them would claim traces exist that were never exported.
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "0010_logfire_project"
+down_revision = "0009_align_index_names"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "agent_environments", sa.Column("logfire_project", sa.String(length=128), nullable=True)
+    )
+    op.add_column("agent_runs", sa.Column("logfire_project", sa.String(length=128), nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column("agent_runs", "logfire_project")
+    op.drop_column("agent_environments", "logfire_project")
--- a/backend/app/agents/observability.py
+++ b/backend/app/agents/observability.py
@@ -14,19 +14,92 @@ a long way from the code that caused it.
 
 The token is unsealed from the vault by the caller and passed in. Nothing here
 reads a secret, logs one, or puts one in a span attribute.
+
+The other half of this module is the *return* trip: recording where a run's
+trace went, so somebody reading run history can open it.
 """
 
 from __future__ import annotations
 
 import logging
-from typing import Any
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
 
 import logfire
+from opentelemetry import trace
 from pydantic_ai import Agent as PydanticAgent
+
+from app.core.config import settings
+
+if TYPE_CHECKING:
+    from app.agents.spec import AgentSpec
 
 logger = logging.getLogger(__name__)
 
 _instances: dict[tuple[str, str, str], logfire.Logfire] = {}
+
+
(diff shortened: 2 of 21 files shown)


OchnikBartek:
#206 landed in #202 (`dc1379b`) — closing this in favour of it.

PR #488, opened 2026-08-07
fix(chat): drop the duplicate usePermissions key in the picker's mock

### LINKED ISSUE #487: tsc fails on main: a duplicate usePermissions key in chat-model-picker.test.tsx
## What breaks

`bun run type-check` fails on `main`, so `make lint-frontend` and CI's `test-frontend`
job are red for every branch cut from it:

```
src/components/chat/chat-model-picker.test.tsx(37,3): error TS1117:
An object literal cannot have multiple properties with the same name.
```

The `vi.mock("@/hooks", ...)` factory declares `usePermissions` twice, at lines 31–33
and 37–39, with identical bodies:

```ts
vi.mock("@/hooks", () => ({
  usePermissions: () => ({
    can: (permission: Permission) => held.permissions.includes(permission),
  }),
  useProviderModels: () => ({ models: listedModels(), source: "curated", isLoading: false }),
  useSecretPurposes: () => ({ purposes: PURPOSES, isLoading: false }),
  useSecrets: () => ({ secrets: listedSecrets(), create: { mutate: vi.fn(), isPending: false } }),
  usePermissions: () => ({                                    // <- duplicate
    can: (permission: Permission) => held.permissions.includes(permission),
  }),
}));
```

The second wins at runtime and is identical to the first, so **the tests pass** —
`tsc` is the only thing that notices. That is what let it through: `bun run test:run`
and `bun run test:coverage` are both green on this file.

## Sequence that triggers it

```bash
git worktree add --detach /tmp/chk origin/main
cd /tmp/chk/frontend && bun install && bun run type-check
```

Reproduced on a throwaway worktree of pristine `origin/main` at `9401bf7` — one error,
that one.

## Where it came from

`frontend/src/components/chat/chat-model-picker.test.tsx:37`, last touched by **#360**
(`fix(kb): the embedding model select, two ungated writes, and two buttons called "Add
a key"`, `8a6ba58`). A mock factory gained a `usePermissions` entry while already
having one.

`cd frontend && bun run type-check` exits 0 on `main`. The fix is deleting one of the
two identical entries.

## Related

Third inherited red on `main` this morning, and each one a different gate:

- **#476** — `ruff format` red on `main` (`test_check_i18n_nodes.py`, from #403).
- **#485** — two integration tests need a `POSTGRES_PASSWORD` the app settings do not
  default, so they fail on any worktree without a `backend/.env`.
- this one — `tsc` red on `main`.

All three share a shape worth naming: **a gate that the change's own test run does not
exercise**. A test-only PR still has to pass `ruff format` and `tsc`, and neither is
part of "I ran the tests and they were green". That is the same argument #203 makes
about the pre-commit basics reading only changed files.

Found merging `main` into `feat/rag-one-knowledge-page` (PR #221). Checked the map in
#168 — belongs with the CI-hygiene cluster beside #203, #476 and #485.

OchnikBartek:
## What changed

Deleted one of two identical `usePermissions` entries in the `vi.mock("@/hooks", ...)`
factory of `frontend/src/components/chat/chat-model-picker.test.tsx`.

That was enough to make `bun run type-check` exit 0 on `main`. Until now it failed:

```
src/components/chat/chat-model-picker.test.tsx(37,3): error TS1117:
An object literal cannot have multiple properties with the same name.
```

`make lint-frontend` runs `tsc`, and so does CI's `test-frontend` job, so this was red
on every branch cut from `main` — including the four open ones.

## Why nothing caught it

The second entry wins at runtime and its body is identical to the first, so it changed
no behaviour: the suite is green with the duplicate and green without it, same 16
tests. `bun run test:run` and `bun run test:coverage` both pass on this file either
way. `tsc` was the only gate that could see it, and it is not part of "I ran the tests
and they were green" — which is the shape #487 names, and the same one behind #476.

## The issue's attribution is wrong, and the real story is worth recording

#487 blames #360 (`8a6ba58`). That commit touched this file, but only to rewrite two
assertion strings — `git show 8a6ba58 -- <file>` is four lines, none of them in the
mock factory.

It is two pull requests landing the same mock entry a day apart, neither seeing the
other:

- **#433** (`3254969`, *check secrets:edit inside the inline secret form*) appended
  `usePermissions` after `useSecrets`.
- **#442** (`63f250b`, *check the permission before offering the write*) added it again
  in alphabetical position, between `useModelProviders` and `useProviderModels`.

`chat-model-picker.test.tsx` is the only source file both touched. That is why this
happened here and nowhere else, and it is a plain merge collision rather than a
copy-paste inside one edit.

**#442's is the one kept**, because its position preserves the factory's alphabetical
order.

## How it was verified

- `cd frontend && bun run type-check` — exits 0, no output.
- `bunx vitest run src/components/chat/chat-model-picker.test.tsx
  src/components/chat/chat-model-picker.integration.test.tsx` — 2 files, 16 tests,
  passed. Unchanged from before, as expected: the deleted entry never ran.
- **Checked for the same shape elsewhere, two ways.** `tsc` reporting nothing else is
  itself the repo-wide answer for duplicate keys — TS1117 is raised per object
  literal, so a second occurrence anywhere would still be in that output. And for the
  shape `tsc` *cannot* see — two separate `vi.mock` calls naming the same module,
  where the second silently replaces the first — a scan over every test file in
  `frontend/src` found no module mocked twice.

Frontend-only and test-only, so no page under `docs/` describes anything this changes.

Closes #487

### DIFF
--- a/frontend/src/components/chat/chat-model-picker.test.tsx
+++ b/frontend/src/components/chat/chat-model-picker.test.tsx
@@ -34,9 +34,6 @@ vi.mock("@/hooks", () => ({
   useProviderModels: () => ({ models: listedModels(), source: "curated", isLoading: false }),
   useSecretPurposes: () => ({ purposes: PURPOSES, isLoading: false }),
   useSecrets: () => ({ secrets: listedSecrets(), create: { mutate: vi.fn(), isPending: false } }),
-  usePermissions: () => ({
-    can: (permission: Permission) => held.permissions.includes(permission),
-  }),
 }));
 
 const purpose = (


OchnikBartek:
Closing unmerged: the identical fix reached `main` independently while this PR was open.

#149 (`feat(dashboard): add the role-aware dashboard`, merged as `16aec9e`) deleted the same second `usePermissions` entry from `chat-model-picker.test.tsx` — `git show 16aec9e -- <file>` shows the `-  usePermissions: () => ({` line. Rebasing this branch onto current `main` reports `patch contents already upstream` and drops the commit, leaving an empty diff.

Verified `main` is actually fixed rather than merely different: `cd frontend && bun run type-check` exits 0 on `16aec9e`, and the mock factory now holds exactly one `usePermissions` (line 31).

#487 is fixed on `main`; closing it too. Nothing to merge here.

frontend/src/components/chat/chat-model-picker.test.tsx
```diff
@@ -34,9 +34,6 @@ vi.mock("@/hooks", () => ({
   useProviderModels: () => ({ models: listedModels(), source: "curated", isLoading: false }),
   useSecretPurposes: () => ({ purposes: PURPOSES, isLoading: false }),
   useSecrets: () => ({ secrets: listedSecrets(), create: { mutate: vi.fn(), isPending: false } }),
-  usePermissions: () => ({
```
DEENUU1, line 37: These are the right three lines. Worth saying for the record: the entry you deleted is the one that was actually running — last duplicate key wins — so the one that stayed was the dead copy. No behaviour change because the bodies are byte-identical, which you checked and I confirmed.

The catch is that `main` already has precisely this deletion, from #149. More in my summary.

DEENUU1 (commented):
Good detective work, and the write-up is better than the issue it closes. But this one has been overtaken, so I don't think we should merge it.

**`main` already has exactly this fix.** #149 (`16aec9e`, the role-aware dashboard PR) landed the byte-identical change at 12:20, fifteen minutes after this branch's CI ran at 12:05. `git show 16aec9e -- frontend/src/components/chat/chat-model-picker.test.tsx` is character-for-character your diff, blob for blob (`0485eea..0049697`). The file is identical on `main` and on this branch, so merging changes nothing:

```
$ git show origin/main:frontend/src/components/chat/chat-model-picker.test.tsx | grep -c 'usePermissions:'
1
$ cd frontend && bun run type-check && echo "exit $?"    # on main
exit 0
```

So: **close this without merging** — but don't lose the body. Two things in it are worth keeping, and I checked both:

1. **The attribution correction is right.** `git show 8a6ba58 -- <file>` is four lines, both assertion strings inside a `describe`, nothing in the mock factory. #487 blaming #360 is wrong, and the real story is #433 (`3254969`, appended after `useSecrets`) and #442 (`63f250b`, inserted alphabetically) landing an hour apart. Put that on #487 as a comment before closing it — the issue is where anyone will look for it.

2. **#442 called this exact collision and it still happened.** Its body says "They conflict textually… Whichever lands second resolves it; the resolution is mechanical." The bit worth recording is *why* nobody had to resolve it: the two insertion points are four lines apart, so git merged both cleanly and never raised a conflict. A predicted textual conflict that git cannot see is a conflict nobody resolves.

Three other things I found while checking:

- **The red `lint` check isn't yours.** It's `ruff format` on `backend/tests/test_check_i18n_nodes.py` — two missing blank lines from #403, i.e. #476. CI merged your branch into `main` at `d29d118`, which still carried it; #149 fixed that too. `origin/main` is clean now, so there's nothing to chase.

- **#221 is the actual remaining work from #487.** `feat/rag-one-knowledge-page` still has both entries:
  ```
  $ git show origin/feat/rag-one-knowledge-page:frontend/src/components/chat/chat-model-picker.test.tsx | grep -c 'usePermissions:'
  2
  ```
  #455 has one and #202 has none, so those two are fine. #221 needs a rebase on `main`, not a second copy of this fix.

- **The reason this got fixed twice is that #149 carried it.** A PR titled "add the role-aware dashboard" with 11,816 deletions quietly included an unrelated three-line fix to a chat test. Nobody reading #487 could have known it was already done. That's the scoped-diffs rule earning its keep — and worth a line in #487, so the next person checks `git log -- <file>` on `main` before cutting the branch.

Issue #167, opened 2026-08-04
A retried channel delivery is a second agent run: nothing dedupes on message_id


DEENUU1:
**Severity: Medium** · Found reviewing `mattermost_webhook.py` on `feat/subagents`, 2026-08-04. Pre-existing, not caused by that branch.

## What's wrong

Nothing in the inbound channel path recognises a message it has already answered. Every delivery — first or fifth — becomes a full agent run: a model call, a spend record against the organization's budget, and a reply posted into the thread.

The only defence in the code is the docstring at `mattermost_webhook.py:7-9`:

> Answers 200 and does the work afterwards: Mattermost retries a slow webhook, and a retried message is a second answer to a question that was only asked once.

That reasoning is right and the implementation matches it — the 200 is returned as soon as the task is scheduled, before any model call — so the *slow-handler* retry is genuinely prevented. It is the only retry cause that is. A platform redelivers for other reasons, and for those the fast 200 does nothing: it was never received.

## Where

All three inbound routes schedule the same coroutine, and none of them looks at delivery identity:

- `app/api/routes/v1/mattermost_webhook.py:75`
- `app/api/routes/v1/slack_webhook.py:71`
- `app/api/routes/v1/telegram_webhook.py:47`

and the three polling/stream paths converge on the same `process_channel_event` → `ChannelMessageRouter.route`. `app/worker/background/channel.py:16` is the single choke point; `app/services/channels/router.py` has no dedupe at any depth.

**The key material already exists.** `IncomingMessage.message_id` (`app/services/channels/base.py:92`) is populated by all three adapters — `mattermost.py:418` from `post_id`, `slack.py:270` from the message `ts`, `telegram.py:243` from `message_id`. Nothing reads it except `router.py:494`, to thread the reply. No new field is needed.

**Slack is the clearest case.** Slack retries an event up to 3 times when it does not receive a 2xx, and sends `x-slack-retry-num` / `x-slack-retry-reason` so a receiver can tell. `rg retry app/api/routes/v1/slack_webhook.py` returns nothing — the header is not read, and the retry is indistinguishable from a new message.

## How it fails

1. A user @-mentions the agent in Slack. The route verifies, schedules the run, returns 200.
2. The 200 is lost — a proxy drops it, an ingress restarts, the pod is rotated between `spawn` and the socket flush. Slack sees no acknowledgement.
[...]

PR #513, opened 2026-08-09
feat(channels): a working Mattermost integration, end to end

### LINKED ISSUE #2: Give both channel supervisors an unconditional backoff
**Severity: Critical** · AUD-002

## What's wrong

Both channel supervisors are `while True: await self._run_x(...)` with `asyncio.sleep(5)` only inside the `except Exception` branch, and both inner coroutines have early-return paths that never await.

## Where

- `backend/app/services/channels/slack.py:93-124`
- `backend/app/services/channels/mattermost.py:125-158`
- Reached from `backend/app/main.py:51-70` (`_start_channel_polling`)

```python
# slack.py:93 — current
while True:
    try:
        await self._run_socket_mode(bot_id, bot_token)   # can return without ever suspending
    except asyncio.CancelledError:
        break
    except Exception:
        logger.exception(...)
        await asyncio.sleep(5)                            # the only sleep

# slack.py:117 — the early return
app_token = self._app_tokens.get(bot_id)
if not app_token:
    logger.warning("Slack bot %s has no app-level token - Socket Mode not started. ...", bot_id)
    return
```

## How it fails

Awaiting a coroutine that never suspends does not yield to the event loop. The supervisor spins at 100% CPU and **no other task on the process ever runs again** — not a request, not a health check, not the WebSocket streams. The API is up and answers nothing.

Both trigger states are ordinary rows, not corruption:

- Slack: `slack_app_token_encrypted IS NULL`. The warning text instructs the operator to go add it, so it is an expected state, and `main.py:64-68` only calls `remember_app_token` when the unseal returns one.
- Mattermost: `api_base_url IS NULL`, which is `nullable=True` and — per the Mattermost half of the webhook-config issue — cannot currently be set at all.

`get_active_polling_bots` filters on `is_active` and `webhook_mode` only. Create a Slack bot with polling enabled, don't paste the `xapp-` token, restart the API: the process comes up, lifespan yields, and the deployment goes dark with one WARNING line as the only clue.

## Fix

```python
while True:
    try:
        await self._run_socket_mode(bot_id, bot_token)
    except asyncio.CancelledError:
        break
    except _NotConfigured:      # nothing a retry can fix
        return
    except Exception:
        logger.exception(...)
    await asyncio.sleep(5)      # unconditional, and cancellable
```

Better still, raise from the inner coroutine's configuration branches so "could not start" is distinguishable from "the session ended".

## Acceptance criteria

- [ ] Both supervisors sleep on every iteration
- [ ] A configuration failure stops retrying rather than looping
- [ ] Regression test, both adapters: with the inner call returning immediately, a concurrent `asyncio.sleep(0)` still completes
- [ ] Consider adding `app/services/channels/slack.py` and `mattermost.py` to the coverage/ty lists — this is exactly the test the 100% gate would have forced

### LINKED ISSUE #10: Give account linking something that mints a code
**Severity: High** · AUD-010

## What's wrong

`ChannelIdentity.link_code` is **never written with a value by any code in the repository.** Searching `app/`, `tests/`, `frontend/src` and `docs/` for `link_code` / `linkCode` / `link_to_user` finds only reads (`get_by_link_code`) and clears (`link_code = None`). No endpoint, service, CLI command or UI mints one, and `link_to_user()` has no callers at all.

## Where

- `backend/app/services/channels/router.py:266-302` (the `/link` redemption path)
- `backend/app/repositories/channel_identity.py:32-40` (`get_by_link_code`), `:82-95` (`link_to_user`, uncalled)
- Refusal at `backend/app/services/channels/mentions.py:277-291`

## How it fails

`get_by_link_code` filters on `link_code == arg`, which never matches a NULL column, so `/link <code>` always answers *"Invalid or expired link code. Please generate a new one from the web app"* — from a web app that has no such control.

Every `ChannelIdentity` therefore keeps `user_id = None`, and `ChannelAgentRouter._context` refuses on exactly that:

```python
# mentions.py:278 — current
if user_id is None:
    raise AuthorizationError(message=_LINK_FIRST, details={"agent": slug})
```

Both `answer` and `answer_default` go through it. So **every** channel message, on any platform, mention or not, is answered with "Link your account before talking to an agent" — permanently. `docs/channels.md:182-184` describes the flow as working.

`tests/test_channel_mentions.py` covers the *refusal* of an unlinked sender thoroughly, and the success path only by injecting a `user_id` directly. That asymmetry is why CI is green.

## Fix

Add `POST /api/v1/me/channel-link`: create or update the caller's `ChannelIdentity` with a short random `link_code` and an expiry, return it, surface it in the UI. Use the existing `link_to_user` on redemption instead of the two hand-rolled `update` calls at `router.py:281-297`.

## Acceptance criteria

- [ ] A signed-in user can obtain a link code
- [ ] End-to-end test: mint a code → `/link <code>` → the next mention runs as that user

### LINKED ISSUE #22: Seal channel_bots.webhook_secret
**Severity: Medium** · AUD-020

## What's wrong

CLAUDE.md: *"Every secret at rest goes through `app/core/vault.py`. **There is no second mechanism.**"*

`webhook_secret` is a 32-byte shared secret (`secrets.token_urlsafe(32)`) written straight to a `String(255)` column — no vault, no encryption — sitting in the same row as `token_encrypted`, `slack_signing_secret_encrypted` and `slack_app_token_encrypted`, all of which are sealed.

## Where

- `backend/app/db/models/channel_bot.py:44`
- `backend/app/services/channel_bot.py:110`
- Consumed at `backend/app/api/routes/v1/telegram_webhook.py:40` and `mattermost_webhook.py:64-65`

## How it fails

It is the only thing authenticating inbound Telegram and Mattermost webhooks. Anyone with read access to the database or a backup — a replica, a support dump, a restored snapshot — recovers it and can POST forged updates, driving the agent as any chat user, spending the organization's model budget and reaching whatever the agent's tools reach. The bot token immediately beside it is protected against exactly that attacker.

`docs/secrets.md` does not mention this column, and `app/core/crypto.py:1-19` carefully documents its one remaining accepted exception (`sync_source`) — this is not it.

## Fix

`webhook_secret_encrypted`, sealed via `_seal_at(..., key_version=bot.secret_key_version)`, unsealed in the two webhook routes. One migration re-sealing existing rows. It is never returned in a response (`ChannelBotRead` is clean), so the change is contained.

## Acceptance criteria

- [ ] The generated value does not appear in the persisted column
- [ ] Scope test mirroring `test_channel_bot_org_scope.py`: a ciphertext lifted into another organization's row fails to unwrap

### LINKED ISSUE #24: Fix the webhook URL, and let Mattermost bots be configured
**Severity: Medium** · AUD-022

Two halves of the same "the channel config surface is incomplete" problem.

## 22a. The registered webhook URL 404s

`services/channel_bot.py:270` and `commands/channel.py:112` both build:

```python
webhook_url = f"{base}/api/v1/channels/{bot.platform}/{bot_id}/webhook"
```

The receivers are mounted at `/telegram`, `/slack`, `/mattermost` (`api/routes/v1/__init__.py:99-103`); the `/channels` router holds only management endpoints. So `bot.set_webhook()` succeeds, the endpoint returns `{"success": true, "webhook_url": ...}`, and Telegram POSTs into a 404 forever — the bot silently never receives a message.

`docs/channels.md:170` documents the *correct* path for Mattermost, which is how the drift is visible.

Fix: `f"{base}/api/v1/{bot.platform}/{bot_id}/webhook"` at both call sites. Then assert the built URL resolves against `app.routes`, so the two cannot drift again.

## 22b. Mattermost bots cannot be configured at all

`ChannelBot.api_base_url` and `ChannelBot.webhook_secret` exist on the model (`db/models/channel_bot.py:43-44`) and are read by the adapter (`mattermost.py:73-75`, `main.py:59-60`) and the webhook route — but **neither is a field on `ChannelBotCreate` or `ChannelBotUpdate`**, neither is a parameter of `channel_bot_repo.create`, and neither appears anywhere in `frontend/src` or `app/commands`.

So the event-stream path logs *"Mattermost bot X has no server URL; cannot open a stream"* and returns (`mattermost.py:155-158`) — which is the trigger for the supervisor busy-loop issue — `send_message` raises `ValueError`, and the outgoing-webhook path can never match, because `verify_webhook_signature` compares Mattermost's token against a locally-generated random string the operator cannot overwrite.

`docs/channels.md:163-172` instructs the operator to *"register it here with the server URL"* and to *"copy the token Mattermost generates into the bot's webhook secret here"*. Neither field exists.

`tests/test_mattermost_channel.py:123` asserts the loud failure when the URL is missing — which is the state every bot is permanently in.

Fix: add `api_base_url: str | None` and `webhook_secret: str | None` to both schemas and to `channel_bot_repo.create`, validating `api_base_url` through `app.core.sanitize.validate_webhook_url`.

## Ordering note

Do the supervisor backoff fix **first**. Fixing 22b without it turns a dead feature into a live outage.

## Acceptance criteria

- [ ] The URL `register_webhook` hands the adapter resolves against `app.routes` — asserted in a test
- [ ] A Mattermost bot created with a server URL and a webhook secret answers a signed outgoing-webhook call
- [ ] `docs/channels.md` matches whichever path is chosen

### LINKED ISSUE #26: Observe the failures of background work
**Severity: Medium** · AUD-024

Two places where a background failure disappears.

## 24a. Webhook tasks bypass `spawn`

`api/routes/v1/slack_webhook.py:19,71-73`, `telegram_webhook.py:47-49`, `mattermost_webhook.py:75-77`.

All three do `asyncio.create_task(...)` into a private module set with `add_done_callback(_background_tasks.discard)` and nothing else. The strong reference is held, so nothing is GC'd mid-flight — but there is **no exception observation**. `process_channel_event` catches `Exception` internally, so the common case is covered; anything it cannot catch vanishes into asyncio's "Task exception was never retrieved" at GC time, outside the app's logger and without the bot or event context.

More materially, these three sets are invisible to `core/background.drain()` — see the shutdown issue.

The repo has exactly the right helper (`core/background.spawn`, whose module docstring is a lecture on this failure mode) and two other correct implementations (`worker/background/__init__.py:23-40`, `agent_session.py:102-111`). These three predate it.

Fix: `spawn(process_channel_event(incoming), name=f"channel-event-{bot_id}")` in all three; delete the local sets.

## 24b. The scheduled sync flow discards every exception

`worker/tasks/rag_tasks.py:155-161`:

```python
tasks = [asyncio.create_task(sync_single_source_flow(str(source.id))) for source in sources]
if tasks:
    await asyncio.gather(*tasks, return_exceptions=True)
logger.info("Scheduled sync check: dispatched %d source(s)", len(sources))
```

Every exception — including `CancelledError` — is collected and thrown away. A Google Drive source whose credentials expired raises on every scheduled run, and the flow logs `dispatched 3 source(s)` and reports `Completed()` to Prefect.

`_run_source_sync` writes `status="error"` on the source row for failures it catches internally, but anything raised in its setup block (`rag_tasks.py:400-437`, before the `try` at 444) escapes into the gather and leaves both the sync log and the source row untouched. Nobody is told, and Prefect is green.

Fix: inspect the results — count and log failures per source id, and fail the flow if any raised. `asyncio.TaskGroup` is the better shape if a failure should stop the batch.

## Acceptance criteria

- [ ] A raising `process_channel_event` produces an ERROR through the app logger, and the task is registered in `background._running`
- [ ] One source whose flow raises → the outcome is logged and recorded, not swallowed

### LINKED ISSUE #41: Mattermost setup
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680c5a744df6b0972c6a3) · Priority: **Mid** · Category: Development

Notion card is a title only. The code says what "setup" has to mean here, and it is more than configuration.

## Mattermost is currently impossible to configure

`ChannelBot.api_base_url` and `ChannelBot.webhook_secret` exist on the model (`app/db/models/channel_bot.py:43-44`) and are read by the adapter (`channels/mattermost.py:73-75`) and by `main.py:59-60` — but **neither is a field on `ChannelBotCreate` or `ChannelBotUpdate`**, neither is a parameter of `channel_bot_repo.create`, and neither appears anywhere in `frontend/src` or `app/commands`.

Consequences today:

- The event stream logs *"Mattermost bot X has no server URL; cannot open a stream"* and returns — **which is the trigger for the event-loop starvation in #2**
- `send_message` raises `ValueError`
- The outgoing-webhook path can never match: `verify_webhook_signature` compares Mattermost's token against a locally generated random string the operator cannot overwrite

`docs/channels.md:163-172` instructs the operator to "register it here with the server URL" and to "copy the token Mattermost generates into the bot's webhook secret here". Neither field exists. This is filed as **#24 (22b)**.

**Order matters: fix #2 before this.** Giving Mattermost a server URL turns a dead feature into a live outage path if the supervisor still busy-loops.

## What is already right

`channels/mattermost.py:219-241` verifies the token fail-closed with `secrets.compare_digest`, handles JSON and form encodings, and `_supervise` re-raises `CancelledError` and backs off exponentially. Covered by `tests/test_mattermost_channel.py:96-121`. Do not rewrite that part.

## Read this before starting any channel work

The 2026-08-01 audit found the channel layer is **not currently functional end to end**, and three of those findings block this task rather than sit beside it:

| | What it means here |
|---|---|
| **#10** | `ChannelIdentity.link_code` is never written by any code in the repo. `/link <code>` can therefore never succeed, `user_id` stays `NULL`, and `ChannelAgentRouter._context` refuses **every** message on **every** platform with "Link your account first". Until this is fixed, no channel answers anything. |
| **#2** | A Slack bot with no app token — or a Mattermost bot with no server URL — makes its supervisor spin without yielding, starving the event loop. The whole API stops answering, not just the bot. |
| **#24** | The three webhook routes use bare `asyncio.create_task`, so their failures are unobserved and they are invisible to `background.drain()` on shutdown. |

**Do #10 first, or this task cannot be tested at all** — you will wire a bot up perfectly and get "Link your account first" back, and spend a day looking in the wrong place.

## Done when

- [ ] #2 is fixed first
- [ ] `api_base_url` and `webhook_secret` are on both schemas and on `channel_bot_repo.create`, with `api_base_url` validated through `app.core.sanitize.validate_webhook_url` (which today has no callers at all — see #33)
- [ ] A Mattermost bot created with a server URL and a webhook secret answers a signed outgoing-webhook call, end to end
- [ ] `webhook_secret` is sealed, not stored in the clear (#22)
- [ ] `docs/channels.md` matches reality

### LINKED ISSUE #152: Tool approval in a channel: today it says "check the approvals queue" and the thread never hears back
An agent with a side-effecting tool works in the web chat and dead-ends in Slack. This is the sentence a channel user gets today, verbatim from `app/services/channels/router.py:169`:

```python
answer or "That needs approval before it can run - check the approvals queue."
```

No link, no queue they can reach, and — for anybody who talks to the bot but has never signed into the web app — no account to reach it with. The run is parked correctly, the notification fires correctly, and the person in the conversation is told to go somewhere they have never been.

## And even when somebody does approve, the channel never hears back

`AgentRunnerService.resume` (`agent_runner.py:774`) returns `(output, run)` to its HTTP caller. `POST /api/v1/runs/{id}/resume` returns it to the browser that clicked. **Nothing routes the answer to the Slack thread the question came from.** So the working path today is: ask in Slack → "check the approvals queue" → approve in the web UI → the answer appears in a browser belonging to whoever approved, and the thread stays silent forever.

The delivery path exists and needs no migration: `agent_runs.conversation_id` → `channel_sessions.conversation_id` → `bot_id` + `platform_chat_id` + `chat_type`. That join is the whole "where does the answer go" question.

## Two ways to ask, and the answer is probably both

**Interactive, in the thread.** Every platform has this and each spells it differently: Slack Block Kit buttons with an `interactivity` request URL, Telegram inline keyboards with `callback_query`, Mattermost interactive message actions. It is the right experience — the decision happens where the question was asked, in one tap — and it is three integrations rather than one, each needing its own inbound handler beside the existing webhook receivers.

**A signed link, as the fallback and the first thing to build.** One deep link into the approvals UI carrying the approval id, so the message says *what* is being approved rather than *that* something is. This is also the only option that works for the case the interactive path cannot cover: an approver who is not in the conversation. The link has to be treated as an auth surface — see below.

Ship the link first; it makes the feature usable and is a prerequisite for the notification content anyway. Then add buttons per platform, Slack first because #43 is where its setup lives.

## The part that decides whether this is safe

**Who may approve is not "whoever can press the button".** `Perm.APPROVALS_DECIDE` gates the resume route today and must gate this identically. Concretely:

- **A channel sender is not automatically an approver.** The sender is resolved to a `ChannelIdentity` and possibly a linked user; the person who *triggered* a side-effecting call is very often exactly the person who should not be the one waving it through. `ApprovalService.decide` is where the check belongs, not in the adapter.
- **An unlinked sender cannot approve anything**, and the message must say so usefully rather than offering a button that fails. Related: **#10** — nothing in the repository mints a `link_code`, so "link your account first" is currently advice nobody can follow.
- **A signed link is a credential.** Single use, short-lived, bound to one approval id and one organization, and it authorises *reaching the queue*, not *deciding* — the decision still runs through the permission check with a real session. A link that decides on click is a link forwarded into a channel that approves an `execute()` on somebody else's behalf. **#23** is the same mistake in the embed token: a claim nobody checked.
- **The decision is once.** `ApprovalService.decide` and `claim_parked_run` already carry the row-locking story (**#17** covers the races). A button pressed twice, or by two people, must not run the tool twice.

## Also worth getting right

- **Say what is being approved.** "That needs approval" names nothing. The tool, its arguments in a readable form, the agent and the conversation — an approver who cannot see what they are approving is a rubber stamp with a permission check.
- **The parked message should update, not accumulate.** Once decided, edit the original message to say who decided and when. Three platforms, three edit APIs, and Slack's `response_url` makes it easy for interactive messages specifically.
- **A refusal is an outcome too.** Rejected means the thread gets told, with the run marked and the reason if one was given.
- **Timeouts.** A parked run in a channel is a conversation that stopped mid-sentence. Decide what happens after some period and tell the thread when it happens.
- **`notifications.py:143`** builds `approvals_url` as `{frontend}/agents/{agent.id}` — the agent page, not the approval. Whatever deep link this issue defines should replace it there too, so the email and the channel message point at the same thing.

## Watch out

- **#3** — a failed run on any surface but web chat is rolled back and leaves no record. A parked run that is then abandoned has the same shape; verify what actually persists before building on top of it.
- The three webhook receivers (`slack_webhook.py`, `telegram_webhook.py`, `mattermost_webhook.py`) already differ in how they verify signatures — **#4** is Telegram failing open. An interactivity endpoint is a second inbound surface per platform and must not repeat that.
- Every one of these paths is org-scoped. An approval id from another organization resolves to nothing, including for an app admin acting in the wrong org.

## Done when

- [ ] A parked run in a channel produces a message naming the agent, the tool and its arguments, with a way to act on it
- [ ] Approving resumes the run and **the answer arrives in the thread the question came from**, on all three platforms
- [ ] Rejecting tells the thread, with the reason
- [ ] A sender without `approvals:decide` is refused, and told what to do rather than shown a button that fails
- [ ] An unlinked sender is refused clearly
- [ ] A signed link is single-use, expiring, bound to one approval and one organization, and does not itself decide
- [ ] The same approval cannot be decided twice, including by two people at once
- [ ] The decided message is updated in place, naming who decided
- [ ] A cross-organization approval id refuses
- [ ] `docs/governance.md` and `docs/channels.md` describe the channel approval path

### LINKED ISSUE #153: A per-channel prompt on the exposure, so one agent can write for Slack and for Telegram
The same agent answers in the web chat, in Slack and in Telegram, and the three render its markdown differently — so an agent tuned to look right in one looks broken in the other two. Binding it to a bot should let you say how to write *there*, without editing the spec every surface shares.

## What the platforms actually do with the reply

**Telegram** sends with `parse_mode=MARKDOWN` and, on `TelegramBadRequest`, silently resends the same text with `parse_mode=None` (`app/services/channels/telegram.py:83-92`). So a reply the model formatted with a table, a nested list or an unescaped `_` does not fail — it arrives as **raw markdown source**, asterisks and hashes visible, and nothing anywhere says it happened.

**Slack** sends `text` straight to `chat.postMessage` with no conversion and no blocks (`slack.py:57-60`). Slack's mrkdwn is not markdown: `**bold**` renders as literal asterisks, `#` headings are not headings, and tables do not exist. Every heading an agent writes is noise.

**Mattermost** takes real markdown and is mostly fine, which is why the difference is easy to miss until somebody adds Slack.

A per-platform renderer is a legitimate answer to some of this and a different, larger ticket. **This one is the cheap half that also covers the things a renderer cannot**: tone, length, whether to use threads, whether to @-mention, and "in this workspace, always answer in Polish".

## Where it goes: `agent_exposures`, and the precedent is already written down

The binding row already carries two per-surface overrides, and `session_scope`'s docstring is the argument for this field verbatim (`app/db/models/agent_exposure.py:96-104`):

> The spec carries the default and stays portable by not knowing what surfaces exist. This is where "on *this* bot, one workspace per channel" belongs — beside `environment_id`, which is already the statement "this bot serves dev".

A channel prompt is the same shape: `AgentSpec` stays portable and surface-agnostic, and the statement "on Telegram, keep it short and skip the headings" lives on the row that says "this agent is on this bot".

So: a nullable `channel_instructions` (or `reply_style`) text column on `agent_exposures`, appended to `spec.instructions` at build time. `factory.py:209` passes `system_prompt=spec.instructions or ""`, so the append happens where the exposure is already resolved — before the factory, not inside it, because the factory should not learn what a channel is.

**Null means the spec decides**, exactly as `session_scope` and `environment_id` already mean.

## The decisions worth making before writing it

**1. Appended or prepended, and can it override?** Appended is the safe default — the spec's instructions win a contradiction, and a binding cannot quietly retask an agent. Say so in the field's help text, because somebody will try to use it to change what the agent *does* rather than how it writes, and the honest answer is that it is a draft-and-publish change to the spec.

**2. A seeded default per platform.** A blank box is a box nobody fills. Ship a sensible starting text per surface — for Slack: no headings, no tables, `*bold*` not `**bold**`, keep it under a screen; for Telegram: short, plain, no tables. That default is also the documentation of what the platform cannot render, in the place somebody is looking.

**3. Does it belong in the exported YAML?** Almost certainly not: the spec exports to a client's git repository and a bot id means nothing there. But `agent_exposures` rows are then configuration that lives only in the database, which is worth stating in `docs/channels.md` rather than discovering during a migration.

**4. Length.** A prompt fragment that grows to a page is a spec somebody wrote in the wrong place. Cap it, and put the cap in the field description.

## Watch out

- **Every published version of an agent shares one exposure row.** Changing the channel prompt changes behaviour for the published agent with no version bump and no audit trail — unlike the spec, which is versioned on publish. Either it gets an audit entry, or the difference is stated where somebody editing it will read it. The second is probably enough; the first is cheap.
- The field is org-scoped through the exposure. A caller editing it needs the same access `resolve_access` grants for editing the agent, checked in the service, **not** as a `require(...)` route gate — the per-resource rule in CLAUDE.md.
- It goes into the model's context on every turn, so it costs tokens on every turn. Not a reason not to do it; a reason to cap it.
- Frontend: the field belongs next to the environment and session-scope controls in the exposure editor, through `next-intl` like everything else, and hidden from a caller who may not edit the agent (`.claude/rules/frontend.md:44`).

## Done when

- [ ] `agent_exposures` carries a nullable, length-capped channel prompt, with a migration
- [ ] It is appended to the agent's instructions for runs on that binding, and nowhere else
- [ ] Null keeps today's behaviour exactly
- [ ] A per-platform default text is offered when a binding is created
- [ ] The exposure editor exposes the field, gated on the same access editing the agent needs
- [ ] A test that two bindings of one agent produce different system prompts, and that the web chat's is unchanged
- [ ] `docs/channels.md` says where per-channel configuration lives and that it is not in the exported spec

### LINKED ISSUE #157: Charts on a channel: three docstrings promise a PNG that nothing renders
Found while auditing backend dependencies for #155, not caused by it.

Three places state that a chart an agent produces is rendered to a PNG for Slack and Telegram:

- `app/agents/capabilities/charts/_toolset.py` — *"The tool does not draw anything: it validates the request into a `ChartSpec`, which the web chat renders with Recharts and **the channel adapters render to a PNG**."*
- `app/agents/capabilities/charts/_spec.py` — *"the channel adapters render it server-side to a PNG for Slack and Telegram."*
- `app/agents/capabilities/charts/README.md` — the same claim again.

**Nothing renders anything.** `OutgoingMessage.image_png` exists on the dataclass (`app/services/channels/base.py`), all three adapters handle it correctly when set — Slack via `files_upload_v2`, Telegram via `send_photo` — and `rg image_png app/` finds no producer. The field is written by no code path in the repository.

Consistent with that: no image library is imported anywhere in `app/`. `pillow` is declared in `pyproject.toml` and has zero import sites, which is what sent me looking.

## What actually happens

An agent with the `charts` capability, exposed on Slack, asked to plot something: `create_chart` validates and returns a `ChartSpec` as a JSON string, the runner puts it in the reply, and the Slack user receives **the JSON of a chart specification** as a message. The docstring says otherwise in three files, and the capability's whole stated value on a channel does not exist.

(The web chat half is real — `ChartMessage` and `chart-message.impl.tsx` render it with Recharts. Though see #144: the frontend matches `create_chart_tool` and the backend registers `create_chart`, so the web chat is currently showing the JSON too, for a different reason.)

## Two ways out, and they are not close in cost

1. **Build it.** Server-side chart rendering in Python — matplotlib, or a headless browser against the same Recharts spec so the two surfaces cannot drift. The second is the only way "the same spec" stays true and it means a browser in the image. Neither is small.
2. **Stop claiming it.** Correct the three docstrings and the README, and have the channel adapter send something honest — a formatted table for a small series, or a line saying the chart is in the web conversation with a link to it. Small, and it makes the product truthful today.

Option 2 now, option 1 as its own scoped piece of work if charts-on-channels turns out to matter.

Either way **the docstrings are the deliverable**: `docs/reference/capabilities.md` is generated from them by mkdocstrings, so the published documentation currently carries this claim too. This is the same class as #29 — *"Correct three documentation claims that describe behaviour that does not exist"* — and the same rule from CLAUDE.md applies: fix the code or fix the page, they cannot both stand.

Ask an agent with the charts capability, bound to a Slack bot, to plot three numbers. Either an image arrives, or the reply says plainly what it can and cannot do — but not a JSON object, and not a docstring promising an image nobody sends.

## Done when

- [ ] The approach is chosen and stated
- [ ] A chart request on Slack and on Telegram produces something a person can use
- [ ] The three docstrings and the README say what the code does
- [ ] `docs/reference/capabilities.md` rebuilds without the false claim
- [ ] A test asserting the channel reply for a chart, whichever shape is chosen

### LINKED ISSUE #205: Embed and channel-mention runs record no messages, so a run has a cost and no content
## What happens

Every surface reaches `AgentRunnerService`, so every run gets its row — cost, status, tokens, and the budget enforced against it. Writing the **transcript** is the caller's job, and only one caller does it in full.

A run through the embedded widget creates a conversation and leaves it empty. A `@mention` on Slack, Telegram or Mattermost writes nothing. A web-chat run that fails keeps the user's prompt and loses the answer. For the widget that is a record nobody can reconstruct afterwards: the organization is billed for an answer given to a visitor on a client's site, and no row says what was asked or what was said back.

## Where

| Surface | What reaches `messages` / `tool_calls` | Where |
|---|---|---|
| Embedded widget | **Nothing.** A conversation row is created and left empty | `app/services/embed_session.py:142` |
| `@mention` on a channel | **Nothing**, though `conversation_id` is passed through | `app/services/channels/router.py:148` |
| A channel bot's default agent | `role` and `content` only — no tool calls, `model_name`, `thinking` or `agent_version_id` | `app/services/channels/router.py:108`, `:136` |
| Web chat, run failed | The user's prompt only. The exception skips `persist_assistant_turn` | `app/services/agent_session.py:222` |
| A run resumed after an approval | **Nothing.** `resume` replays through `_run`, which writes no messages | `app/services/agent_runner.py:600` |
| API | **Nothing**, although the caller may pass a `conversation_id` | `app/api/routes/v1/agents.py:368` |

Web chat on the success path is the one surface that records everything.

## Why only one surface has it

`tool_calls` rows are written in exactly one place in the repository — `app/services/agent.py:225`, inside `persist_assistant_turn` — and it is fed by `collected_tool_calls`, which only `_stream_tool_events` fills (`app/services/agent_session.py:391`). Tool calls are therefore observable **only** on the streaming path. `execute()` returns `tuple[str, AgentRun]` (`app/services/agent_runner.py:503`), so a non-streaming surface has no access to them at all.

That splits the work in two, and the halves are not the same size. The messages half is a call site each. The tool-calls half needs the runner to report what a run called, which is a change to how `execute` hands its results back rather than an addition at the caller.

This is only about the transcript. Which `RunSurface` each run is stamped with is a separate gap, already being closed on the dashboard branch (`e76af9d`).

## Not from #45, but it blocks half of it

Pre-existing on every path above; found while writing `docs/design/activity-plan.md` for #45.

#45 adds `messages.run_id` and a run detail view built on these rows, so on the surfaces above that drill-down has nothing to show — and the migration does not change it, because it links messages that were never written. #45 therefore ships the honest version: the detail view states that steps were not recorded for that surface, and fills in on its own once this lands, with no further frontend work.

Filed rather than folded in, because this is five write paths in the chat and channel subsystems rather than a change to the Activity page.

## Done when

- An embed run and a channel mention leave the same transcript a web-chat run does.
- A channel bot's default agent records its tool calls, model and agent version, not only text.
- A web-chat run that fails keeps whatever the assistant produced before it failed.
- A run resumed after an approval records its continuation.
- `execute()` can report the tool calls a run made, so a non-streaming surface can persist them.
- A regression test per surface — embed, mention, channel default, a failed web run, a resumed run — each asserting the rows exist, not that a mock was called.

### LINKED ISSUE #208: An embed run is recorded as web and a Mattermost run as api
## What breaks

`agent_runs.surface` is the column that answers "where did this run come from", and two of its values are never recorded, so the column quietly attributes those runs to something else:

- **An embedded-widget run is stamped `web`.** `backend/app/services/embed_session.py:164` passes `surface=RunSurface.WEB`. `RunSurface.EMBED` does not exist in the enum at all, so there is nothing to stamp it with — this needs the value adding as well as the call site changing.
- **A Mattermost mention falls through to `api`.** `_SURFACES` in `backend/app/services/channels/mentions.py:66` maps only `slack` and `telegram`; `mentions.py:253` and `:328` call `_SURFACES.get(platform, RunSurface.API)`, so a Mattermost run is recorded as an HTTP API call. `RunSurface.MATTERMOST` **is** defined on the enum (`app/db/models/agent_run.py:70`) and never assigned — the map is the only thing missing.

Both are silent. Nothing errors; the number just lands in the wrong bucket, and every consumer of the column inherits it.

## Why it matters now

This blocks the honesty of two designs currently in review, and both flagged it independently:

- **#37 / PR #149** — the dashboard's "Where runs come from" card. It is the single best answer to the issue's central question (*how do people actually use the system*), and today it would report every embedded widget as web chat and every Mattermost workspace as an API integration.
- **#45 / PR #202** — the Activity page's Surface multi-select. Its design offers `embed` and `mattermost` but badges them as not yet truthful, which is the right call and not a fix.

Both plans price the same two-service change and say "whichever branch lands it first ships it, the other inherits". That is how work with two owners ends up with none, which is why it is its own issue.

## Sequence

1. Talk to an agent through the embed widget. `GET /api/v1/runs` reports `surface: "web"`.
2. Mention an agent in Mattermost. The run reports `surface: "api"`.
3. Filter run history by surface. There is no honest way to count either, and no way to tell an embedded run from a web-chat one at all.

- `RunSurface.EMBED` exists, `embed_session.py` stamps it, and an embedded run's row carries `surface = "embed"`.
- `_SURFACES` maps `mattermost`, and a Mattermost mention's row carries `surface = "mattermost"` rather than `api`.
- A test per surface asserting the recorded value — the kind of assertion that would have caught both of these, since the code paths themselves are exercised today.
- Existing rows need no backfill and no caveat: nothing is in real use, so there is no body of mis-stamped rows worth reinterpreting or explaining on screen.

## What to build

- [ ] Add `EMBED = "embed"` to `RunSurface` — it does not exist on the enum yet, so there is nothing to stamp with.
- [ ] `backend/app/services/embed_session.py:164` passes `surface=RunSurface.EMBED` instead of `WEB`.
- [ ] Map `"mattermost": RunSurface.MATTERMOST` in `_SURFACES` (`backend/app/services/channels/mentions.py:66`). The enum value already exists; only the map is missing.
- [ ] A test per surface asserting the recorded value. Both code paths run today, so a test that only exercises them is not enough — it has to assert what lands in the column.
- [ ] Backfill is not needed and neither is a caveat about history: nothing is in real use, so there is no meaningful body of `web` rows to reinterpret. Do not add copy to either page saying the series starts on a deploy date — that is a concession to data that does not exist.

## Where it sits

Pre-existing; found in the reviews of PR #149 and PR #202, recorded in both designs and filed by neither.

Cluster: **Channels** (#24 #41 #42 #43 #152) for the Mattermost half, **Frontend surfaces** (#37 #45) for the consumers. Sibling issue: `RunSurface.SCHEDULE` is defined and never assigned — the same column wrong in the other direction (a value with no writer rather than a writer with the wrong value); pick them up together. **Blocks** the "Where runs come from" card in #37 and the Surface filter in #45 from being truthful, though neither is blocked from shipping. Nearest duplicate: none among the 137 open issues.

### LINKED ISSUE #514: A channel bot goes silent while it works: stream the reply instead
A channel bot posts one finished message and nothing before it. Ask an
agent something that takes twelve seconds and three tool calls, and the
chat is silent for twelve seconds — indistinguishable from a bot that
crashed. Found by using the Mattermost integration end to end (#41): the
first thing anybody says about it is that it feels dead.

The web chat has none of this problem, and not because it is special.
`AgentRunnerService.iterate()` already drives a metered stream and
`agent_session.py` already turns the graph's nodes into events. Channels
call the non-streaming half and wait.

## What a channel can actually render

**Mattermost** — a post can be edited after it is sent
(`PUT /api/v4/posts/{id}/patch`), and every connected client sees the
edit. Its own AI plugin streams this way. The WebSocket we already hold
for the event stream also accepts a `user_typing` action, which is how the
webapp shows "… is typing".

**Slack** — `chat.update` on the `ts` of the posted message.

**Telegram** — `editMessageText` on the returned `message_id`, plus
`sendChatAction: typing`.

So all three support the same shape: **post something immediately, then
edit it as the answer arrives.**

## Three layers, one mechanism

1. **A reply that exists from the moment the question was asked** — a
   placeholder post, so the thread shows the bot has the message.
2. **Text that grows in place**, throttled to about one edit a second or a
   sentence boundary. Editing per token is a denial-of-service attack on
   somebody else's server.
3. **What the agent is doing meanwhile** — "searching the knowledge
   base…", "drawing a chart…" — in that same post, replaced by the answer
   when it arrives. This is the one that matters most: the silence is
   longest exactly when tools are running.

## What it needs here

- `AgentRunnerService.execute(stream=...)`, using `prepared.iterate()`
  when a stream is given. The settle path — usage, budget, transcript,
  `finish()` — stays where it is, so a streamed channel run is metered
  identically to an API one. Not available on the resume path:
  `iterate()` takes no `deferred_tool_results`.
- Two optional methods on `ChannelAdapter`, defaulting to "not supported"
  so an adapter that cannot edit a message keeps working unchanged.
- The throttle belongs in the router, once, rather than in three adapters.

## Done when

- [ ] A Mattermost reply appears immediately and fills in as the agent writes
- [ ] Tool steps are visible while they run, and replaced by the answer
- [ ] The typing indicator shows on the event-stream path
- [ ] Slack and Telegram do the same through the same seam
- [ ] An adapter with no streaming support still posts one finished message
- [ ] Edits are throttled, and the throttle is tested rather than assumed

DEENUU1:
Makes Mattermost a channel somebody can actually register and talk to, then
closes the gaps that stop any channel from being a complete surface. Opened as a
draft and built commit by commit; the first commit is the plan
(`docs/plans/mattermost-integration.md`), which stays the running ledger.

## What this is about

**The adapter is finished. The configuration surface does not exist.** Nothing
about Mattermost is half-written — `app/services/channels/mattermost.py` sends
posts with attachments, folds threads, runs an authenticated WebSocket stream
with a keepalive and an exponential backoff, receives outgoing webhooks
fail-closed with `secrets.compare_digest`, and decodes a body by what it *is*
rather than by what its header claimed. Sixteen tests cover it.

What is missing is the ability to give a bot the one thing a self-hosted platform
needs: the address of its own server.

`ChannelBot.api_base_url` is on the model (`db/models/channel_bot.py:43`) and is
read by `main.py:61` and `services/channels/router.py:486`. It is **not** a field
on `ChannelBotCreate`, not on `ChannelBotUpdate`, **not a parameter of
`channel_bot_repo.create`**, and appears nowhere in `app/commands` or
`frontend/src`. There is no way to write it short of an `UPDATE` statement.

So every Mattermost bot has `api_base_url IS NULL`, permanently:

- `_run_stream` raises `ChannelNotConfigured`, the supervisor logs *"no server
  URL; cannot open a stream"* and returns — the bot is registered and deaf
- `send_message` raises `ValueError`
- `download_attachment` refuses
- `tests/test_mattermost_channel.py:159` asserts that loud failure. It is
  asserting the state every bot is permanently in.

Everything else follows from the same gap: `webhook_secret` is *minted* locally
rather than accepting the token Mattermost generates, so the webhook path
compares two unrelated strings and can never authenticate.

## Decisions, settled before any code

**1. `api_base_url` is validated on scheme and shape only.**

[#41](https://github.com/vstorm-co/agenticos/issues/41) asks for
`validate_webhook_url`, which blocks private, reserved, loopback and link-local
addresses and resolves DNS to prove a host is public (`core/sanitize.py:200-206`).
A Mattermost server is somebody's own — `docs/channels.md:174` gives
`https://mattermost.acme.internal` as the example and calls the event stream *"the
right choice behind a VPN"*. Validating as #41 asks would refuse exactly the
deployment the feature is for.

So: reject userinfo, non-`http(s)` schemes and malformed URLs; private addresses
pass. The validator's docstring carries the reason so nobody "fixes" it back. The
cost is stated rather than hidden — a member who may manage channel bots can
point one at an internal address and the bot token goes with it. That is a
consequence of the permission, not of this field.

**2. `webhook_secret` is sealed first** (#22), in its own commit, before any field
that writes one exists. Adding a plaintext write path we intend to remove is work
done twice.

**3. The "fix #2 first" ordering warning in #41 and #24 is stale.** Both
supervisors now sleep unconditionally and Mattermost's backs off exponentially
(`mattermost.py:166-194`), covered by `test_channel_supervisor_yields.py`. What
is left of [#2](https://github.com/vstorm-co/agenticos/issues/2) is Slack's flat
`asyncio.sleep(5)` (`slack.py:140`), which is not in the path of this work.

### Phase 1 — Mattermost becomes usable

| | Step |
|---|---|
| S1 | Seal `webhook_secret` (#22) — the only thing authenticating inbound Telegram and Mattermost webhooks, and a plaintext column beside three sealed ones |
| S2 | Fix the webhook URL at both call sites (#24) — `services/channel_bot.py:274` and `commands/channel.py:111` build `/api/v1/channels/{platform}/…`, the receivers are mounted at `/api/v1/{platform}/…`. Assert the built URL resolves against `app.routes` |
| S3 | `api_base_url` and `webhook_secret` on both schemas, on `channel_bot_repo.create` and on the update path |
| S4 | Accept an operator-supplied `webhook_secret` for Mattermost instead of minting one; keep minting for Telegram, where we hand it out |
| S5 | The panel: a server-URL and a webhook-secret field, shown for Mattermost the way the Slack credentials are shown for Slack |
| S6 | `app/commands/channel.py` — register a Mattermost bot from the CLI, because a deployment behind a VPN has no browser pointed at it |
| S7 | The webhook route hands its work to `spawn_after_commit` rather than a bare `asyncio.create_task` (#26) |
| S8 | Stamp a channel run with its channel surface (#208) |
| S9 | `docs/channels.md` matches what was built, down to which token comes from where |

### Phase 2 — the gaps every channel shares

None of these is Mattermost-only; Slack and Telegram get the same fix and the
same tests.

| | Step |
|---|---|
| S10 | `/link` mints a code (#10). Nothing writes `link_code` — it is only read and cleared — so `/link <code>` cannot succeed and `@slug` refuses every sender. **First in this phase**, or half of manual testing comes back as *"Link your account first"* and reads like a broken Mattermost |
| S11 | Dedupe on `message_id` (#167). Mattermost retries a slow webhook and our route answers 200 before doing the work, so a retry is a second run with a second bill |
| S12 | A channel run records its messages (#205) — today it has a cost and no content |
| S13 | Charts on a channel (#157). Three docstrings promise a PNG that nothing renders; the adapter already uploads `image_png` correctly |
| S14 | A tool approval answers in the thread (#152). Last, because it builds on S12 |

## Done when

Phase 1: a bot registered with a server URL and a webhook secret answers a signed
outgoing-webhook call end to end, and the same bot in event-stream mode answers a
`posted` event.

Phase 2: `/link` completes from a Mattermost DM, `@slug` runs as the person who
typed it, a retried webhook produces one run rather than two, that run has its
messages in Activity, a chart arrives as an image, and an approval is answered in
the thread that asked for it.

## How this is verified

Unit and API layers, plus a manual pass against a real Mattermost server the
author runs. **`tests/integration/` and the Playwright suite are deliberately not
extended here**, and nothing is added to `docker-compose.dev.yml`.

What that leaves unproven is what only a real server shows: whether Mattermost's
actual `posted` frame matches the shape `_from_socket` expects, and whether the
outgoing-webhook body arrives in the encoding we assume. That is what the manual
pass is for, and it is why `docs/channels.md` owes an exact account of which
token comes from where.

## State right now

Two commits, no code changed yet: the plan, and the decisions above recorded in
it. `make docs-build` (strict) and the codespell hook pass. Nothing else is
claimed.

Closes #41
Closes #153
Closes #24
Closes #22
Closes #10
Closes #205
Closes #157
Closes #152
Closes #208
Closes #26
Closes #514

Refs #167 — the fast 200 defuses the *slow-webhook* retry, but dropping a
redelivery whose 200 was lost in transit still needs a persisted
`(bot, message_id)` guard, which this branch deliberately does not add. Left
open on purpose; the receiver's docstring says the same.


backend/app/agents/capabilities/channel_tools/_directory.py
```diff
@@ -0,0 +1,137 @@
...
+    that could pass a channel id would be an agent that could read any channel
+    its bot happens to be in from a conversation in another one.
+
+    :meth:`search` is the one exception and it returns *summaries*, never
+    contents - finding that `#billing` exists is not reading it.
+    """
+
+    async def details(self) -> ChannelDetails:
+        """Name, purpose, topic and size of the channel this run is in."""
+        ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/app/agents/capabilities/channel_tools/_directory.py
```diff
@@ -0,0 +1,137 @@
...
+
+    async def members(self, *, limit: int) -> list[ChannelMember]:
+        """Who is in it, up to `limit`."""
+        ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/app/agents/capabilities/channel_tools/_directory.py
```diff
@@ -0,0 +1,137 @@
...
+
+    async def search(self, query: str, *, limit: int) -> list[ChannelSummary]:
+        """Channels whose name or purpose matches `query`, within the bot's reach."""
+        ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/app/agents/capabilities/channel_tools/_directory.py
```diff
@@ -0,0 +1,137 @@
...
+
+    async def history(self, *, limit: int) -> list[ChannelPost]:
+        """The last `limit` messages, oldest last."""
+        ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/tests/test_channel_tools.py
```diff
@@ -0,0 +1,437 @@
...
+        assert adapter.search_channels.await_args.kwargs["query"] == "billing"
+
+    async def test_an_adapter_that_does_not_implement_one_refuses_by_default(self):
+        """A new adapter is correct on the day it is written, and says what it
+        cannot do rather than answering an empty list."""
+
+        class Bare(ChannelAdapter):
+            platform = "carrier-pigeon"
+
+            async def send_message(self, bot_token: str, msg: Any) -> None: ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/tests/test_channel_tools.py
```diff
@@ -0,0 +1,437 @@
...
+            async def start_polling(self, bot_id: str, bot_token: str) -> None: ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/tests/test_channel_tools.py
```diff
@@ -0,0 +1,437 @@
...
+            async def stop_polling(self, bot_id: str) -> None: ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/tests/test_channel_tools.py
```diff
@@ -0,0 +1,437 @@
...
+            async def register_webhook(
+                self, bot_token: str, url: str, secret: str | None
+            ) -> bool: ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/tests/test_channel_tools.py
```diff
@@ -0,0 +1,437 @@
...
+            async def delete_webhook(self, bot_token: str) -> bool: ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/tests/test_channel_tools.py
```diff
@@ -0,0 +1,437 @@
...
+            def verify_webhook_signature(
+                self, headers: dict[str, str], secret: str, body: str | None = None
+            ) -> bool: ...
```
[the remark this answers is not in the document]
DEENUU1: Adjudicated in [`docs/code-review.md` → Five findings already adjudicated](https://github.com/vstorm-co/agenticos/blob/main/docs/code-review.md#five-findings-already-adjudicated): `...` is PEP 544's canonical `Protocol`/abstract-method body — `pass` has no more effect and reads worse. Not changing the code.

backend/app/services/channels/router.py
```diff
@@ -157,15 +221,59 @@ async def _route_inner(self, incoming: IncomingMessage, db: Any) -> None:
...
+                    OutgoingMessage(
+                        platform_chat_id=incoming.platform_chat_id,
+                        text=text,
+                        api_base_url=getattr(bot, "api_base_url", None),
+                    ),
+                    handle,
+                )
+                if answered.image_png is None and not answered.attachments:
+                    return
+            text = ""
```
DEENUU1: This one loses the answer on a failure path. `text = ""` sits *outside* the `with contextlib.suppress(...)`, so it runs whether the edit succeeded or raised. If the final `update_reply` throws - a rate-limit on the last edit, or the user deleted the placeholder so the PATCH 404s - the exception is swallowed, `text` is blanked, and `_send_reply` posts an empty message while the `…` placeholder stays on screen forever. That is the opposite of what `live_reply.py` promises ("the answer is sent whole at the end whatever happened").

Blank `text` only when the edit actually succeeded:

```python
if handle is not None:
    adapter = get_adapter(incoming.platform)
    try:
        await adapter.update_reply(
            unseal_bot_token(bot),
            OutgoingMessage(
                platform_chat_id=incoming.platform_chat_id,
                text=text,
                api_base_url=getattr(bot, "api_base_url", None),
            ),
            handle,
        )
    except Exception:
        logger.warning("live reply final edit failed; re-posting the answer whole", exc_info=True)
    else:
        if answered.image_png is None and not answered.attachments:
            return
        text = ""
```

Now a failed edit falls through to `_send_reply` with the full answer instead of an empty post.

backend/app/services/channels/chart_png.py
```diff
@@ -0,0 +1,382 @@
...
+    """
+    values = [
+        number
+        for row in spec.data
+        for key in keys
+        if (number := _numeric(row.get(key))) is not None
+    ]
+    if not values:
+        return 0.0, 1.0
+    low, high = min(min(values), 0.0), max(max(values), 0.0)
```
DEENUU1, line 132: Stacked bars overflow the plot. `_bounds` takes the min/max of *individual* values, but `_draw_bars` in the stacked branch plots the running cumulative total (`running += value`). For two positive series - which is the whole point of a stack - the stack reaches `a + b` while `high` is only `max(a, b)`, so `y_for(running, ...)` lands above `canvas.top` and the bar climbs into the title, with the axis labels still reading the smaller max. It happens on essentially every stacked bar with more than one positive series, not an edge case - and `test_a_stacked_bar_chart_draws` only asserts the PNG magic bytes, so it passes on a broken picture.

When the chart is stacked, the bounds need the per-row cumulative sums (positive and negative stacks separately), not the raw values. Worth a quick look at whether stacked *area* wants the same treatment - `_draw_lines` ignores `style.stacked` entirely, so a stacked area currently renders as overlapping unstacked bands.

DEENUU1 (commented):
Solid piece of work, and the commit-by-commit build with the plan as a running ledger makes it easy to follow. The parts I went in worried about are the parts that hold up best: the channel-tools directory is bound server-side from the bot row that received the message and no tool ever takes a channel id, so there is genuinely no path from one channel to another's, another bot's, or another org's. `@slug` resolving only against this bot's exposure, the run taking the *sender's* identity, and the migrations re-sealing rather than dropping secrets are all right. The `api_base_url` SSRF trade-off is documented where it belongs.

Headline: **CI's `test` job is red - platform coverage is 99.86%, the gate is 100%.** `make test` locally will name the uncovered lines; that is the blocker for merge.

Beyond that I found three real bugs worth fixing before this lands: a failed final live-reply edit blanks the whole answer and posts an empty message (`router.py`), stacked bars draw past the top of the plot because the axis bounds ignore the stack sum (`chart_png.py`), and a non-6-digit-hex colour crashes the area-chart renderer so the agent says "here is the chart" and sends nothing. The rest are suggestions - the biggest being that #167 (dedup) is not actually in this branch yet, which the webhook's own docstring half-promises. Details inline.

backend/app/services/channels/chart_png.py
```diff
@@ -0,0 +1,382 @@
...
+            if value is None:
+                continue
+            x = canvas.x_for(index, count)
+            y = canvas.y_for(value, low, high)
+            canvas.draw.ellipse([(x - 5, y - 5), (x + 5, y + 5)], fill=color)
+
+
+def _translucent(color: str) -> tuple[int, int, int]:
+    """The series colour, lightened, for the band under an area chart."""
+    red, green, blue = (int(color[index : index + 2], 16) for index in (1, 3, 5))
```
DEENUU1: This crashes on any colour that is not full 6-digit hex, which sends no chart at all - the #157 regression you wrote this module to kill. `color` and `palette` are free-form strings on the spec (only the *description* says hex, nothing validates it), and the model picks them. `_translucent("#f00")` slices `color[5:7]` -> `""` -> `int("", 16)` -> `ValueError`; a named colour like `"red"` blows up the same way. It fires on any area chart, gets caught at `mentions.py:104`, and the agent says "here is the chart" while nothing arrives.

Parse through PIL instead of slicing by hand, so `_translucent` accepts everything the rest of the renderer already does:

```python
from PIL import ImageColor

def _translucent(color: str) -> tuple[int, int, int]:
    """The series colour, lightened, for the band under an area chart."""
    red, green, blue = ImageColor.getrgb(color)[:3]
    return (
        red + (255 - red) * 3 // 4,
        green + (255 - green) * 3 // 4,
        blue + (255 - blue) * 3 // 4,
    )
```

Better still, validate `color`/`palette` at the spec with a `field_validator` so a bad colour is a clean refusal the model can retry, rather than a swallowed exception.

backend/app/services/embed_session.py
```diff
@@ -151,11 +188,22 @@ async def _answer(self, text: str) -> str:
...
+                    f"[Context for this placement: {self.embed.context}]"
+                    if self.embed.context
+                    else "",
+                    self._supplied_block(),
+                )
+                if part
+            )
+            if preamble:
+                prompt = f"{preamble}\n\n{text}"
+                self._context_sent = True
```
DEENUU1: The `_context_sent` latch defeats the per-frame `_supplied` update you added just above. You refresh `self._supplied` on every frame precisely so a single-page app that signs the visitor in on turn 2 gets picked up (the comment at line 133-137 says exactly this). But `_context_sent` latches on the *first* turn: if the placement `context` is set and the visitor is not yet signed in on turn 1, the preamble is non-empty, the flag flips, and turn 2's supplied block - including any `required` variables, and the `embed_context_missing` warning - never reaches the agent. It is frozen at whatever turn 1 had, which is worse than reading at connect time.

Track the supplied block separately from the placement context: send the placement context once, but re-send the supplied block when `self._supplied` actually changes (or at least do not latch until a non-empty supplied block has been sent when variables are declared).

backend/app/api/routes/v1/mattermost_webhook.py
```diff
@@ -47,16 +46,13 @@ async def mattermost_webhook(
...
         raise HTTPException(status_code=403, detail="Invalid webhook token")
 
     incoming = adapter.parse_incoming(decode_webhook_body(raw), str(bot_id))
     if incoming is None:
         return Response(status_code=200)
 
-    task = asyncio.create_task(process_channel_event(incoming))
-    _background_tasks.add(task)
-    task.add_done_callback(_background_tasks.discard)
+    spawn(process_channel_event(incoming), name=f"mattermost_event:{bot_id}")
```
DEENUU1, line 61: The docstring up top frames this as fixing "a second answer to a question that was only asked once", but there is no dedup on `message_id` anywhere in the branch - I grepped `services/channels/**`, `worker/**`, the repos, the models and the migrations, and `incoming.message_id` is only ever used as `reply_to_message_id`. The fast-200 does defuse the *slow-webhook* retry (the ack no longer waits on the agent), which is real. But if the 200 is lost in transit Mattermost redelivers the same `post_id` and the agent runs again on the sender's budget. The per-chat lock in `router.py` does not help - it *serializes* the two retries so both complete and both bill, rather than dropping the second.

That is #167 / plan step S11, so I read this as "not done yet" rather than broken - but the title says end to end and this docstring half-promises the fix, so either soften the wording here or land the guard: persist `(platform, bot_id, message_id)` with a unique constraint and insert-and-catch-`IntegrityError` (not read-then-write, so it survives concurrent delivery), guarding the `message_id is None` case.

backend/app/agents/capabilities/channel_tools/_directory.py
```diff
@@ -0,0 +1,137 @@
...
+    async def members(self, *, limit: int) -> list[ChannelMember]:
+        """Who is in it, up to `limit`."""
+        ...
+
+    async def search(self, query: str, *, limit: int) -> list[ChannelSummary]:
+        """Channels whose name or purpose matches `query`, within the bot's reach."""
+        ...
+
+    async def history(self, *, limit: int) -> list[ChannelPost]:
+        """The last `limit` messages, oldest last."""
```
DEENUU1: This says "oldest last", but the implementation and everything the model reads say the opposite. Mattermost's `channel_history` reverses Mattermost's newest-first `order`, so it returns `[oldest, ..., newest]` - i.e. newest *last* - and both model-facing descriptions (`_toolset.py:165`, `__init__.py:97`) correctly say "newest last". This Protocol docstring is the contract a future Slack/Telegram adapter author implements against, and the source the reference docs generate from, so "oldest last" here tells them to return the reverse of what the model expects. With `ChannelPost.posted_at` optional, a platform that omits it makes the mis-ordering invisible.

Same wrong phrase in `base.py:254` and `mattermost.py:425` (whose own next line describes newest-first input correctly). Change all three to "newest last".

backend/app/services/channels/chart_png.py
```diff
@@ -0,0 +1,382 @@
...
+    """The row value as a number, or None if it is not one.
+
+    A gap in a series is legitimate - a month with no figure yet - so a
+    non-number is skipped rather than treated as zero, which would draw a line
+    dropping to the axis and read as "sales were nil".
+    """
+    if isinstance(value, bool):
+        return None
+    if isinstance(value, int | float):
+        return float(value)
```
DEENUU1: `NaN`/`Infinity` slip through here and crash the renderer. `json.loads` accepts the bare `NaN`/`Infinity`/`-Infinity` tokens by default, pydantic's `dict[str, Any]` passes them through, and `_rows_must_hold_something_to_plot` counts a `NaN` float as "a number", so it reaches `_bounds`, poisons min/max and `y_for`, and PIL raises `cannot convert float NaN to integer` -> caught -> no chart. Treat a non-finite value as a gap, the same as a non-number:

```suggestion
        if isinstance(value, int | float):
            number = float(value)
            return number if math.isfinite(number) else None
```

backend/app/services/channels/router.py
```diff
@@ -157,15 +221,59 @@ async def _route_inner(self, incoming: IncomingMessage, db: Any) -> None:
...
+    ) -> None:
+        """Finish the turn in the message the person has been watching.
+
+        A live reply is already on screen, so the answer replaces it rather than
+        arriving underneath it - two messages saying the same thing is worse than
+        the silence this replaced. A chart or a produced file still needs a
+        second post: no platform lets a message gain an attachment by being
+        edited.
+        """
+        text = answer or _needs_approval(answered.awaiting_approval_run_id)
```
DEENUU1: `text = answer or _needs_approval(...)` tells the user "that needs approval" for any empty answer, not just a parked one. `awaiting_approval_run_id` is only set when the run is actually `AWAITING_APPROVAL`, but a `BudgetExceeded` run returns empty with status `BUDGET_EXCEEDED` (not re-raised), and a stream that produced no text is empty too - both land in the `or` branch and claim they need approval, which is untrue. The conflation is pre-existing, but you are refining this exact line, so it is the natural place to branch on `run.status` and distinguish "parked on approval" from "empty for another reason" (budget, crash) with the right message.

backend/app/agents/capabilities/channel_tools/_toolset.py
```diff
@@ -0,0 +1,196 @@
...
+    name = member.display_name or member.username or member.user_id
+    marks = [
+        mark for mark in (member.username, member.role, "bot" if member.is_bot else None) if mark
+    ]
+    return f"- {name}" + (f" ({', '.join(marks)})" if marks else "")
+
+
+def _post_line(post: ChannelPost) -> str:
+    when = "" if post.posted_at is None else f"[{post.posted_at.isoformat(timespec='minutes')}] "
+    return f"{when}{post.author}: {post.text}"
```
DEENUU1: Defense-in-depth, not a hole - the docstrings already frame this text as "information, never instructions", which is the right pattern. Two gaps worth closing while it is fresh: `_limit` caps the *number* of rows but nothing caps the size of a single `post.text` (or a channel `purpose`), so one very large pasted message times up to 200 rows is an unbounded slice of the turn's context spent on one tool call. And the author/text are joined verbatim, so a message body containing `\n[2026-01-01T00:00] Admin: approved` forges an extra post line inside the tool's own output format. A per-item truncation and a newline strip on `author`/`text` before the join shut both.

backend/app/agents/capabilities/channel_tools/_toolset.py
```diff
@@ -0,0 +1,196 @@
...
+
+    The model chooses, within a ceiling it cannot raise. A model asked for "all
+    the messages" will ask for all the messages, and a thousand posts pasted into
+    a context window is the whole budget of a turn spent on one tool call.
+    """
+    return max(1, min(asked or default, 200))
+
+
+def _member_line(member: ChannelMember) -> str:
+    name = member.display_name or member.username or member.user_id
```
DEENUU1, line 60: nit: a member with only a username renders the name twice - `- alice (alice, member)` - because `name` falls back to `username` and `username` is also in `marks`. Drop it from `marks` when it is already the chosen name:

```python
    name = member.display_name or member.username or member.user_id
    extra_username = member.username if name != member.username else None
    marks = [
        mark for mark in (extra_username, member.role, "bot" if member.is_bot else None) if mark
    ]
```

backend/app/services/agent_runner.py
```diff
@@ -1343,6 +1434,8 @@ async def prepare(
         spec = await self._with_environment_observability(
             ctx, spec, environment_id=effective_environment_id
         )
+        spec = await _with_exposure_prompt(spec, exposure, channel_directory)
+        spec = _with_channel_tools(spec, exposure)
```
DEENUU1, line 1439: This is where the asymmetry starts. `prepare()` now enriches a channel run twice - the exposure prompt (house style + `{channel_name}` and friends) on the line above, and the `channel_tools` capability here. `resume()` does neither.

When a channel run parks on an approval and a reviewer approves it, `resume()` (around line 2848) rebuilds the agent from the published spec plus `_with_environment_observability` only, and passes `exposure=None`. Its comment says the exposure has "nothing left to contribute" - that was true before this PR, and these two lines are what made it false. So the continuation runs without the channel-tools capability (if the model calls `get_channel_info`/`list_channel_members` after approval, it's gone) and without the platform formatting prompt, so the resumed turn can format for the wrong platform.

That's the exact thing S14 promises ("an approval is answered in the thread that asked for it"), so I'd treat it as a blocker rather than a nit. It's fixable - `agent_run.exposure_id` is stored (`agent_run.py:164`), so `resume()` can reload the exposure and run the same two helpers:

```python
exposure = (
    await agent_exposure_repo.get_by_id(self.db, run.exposure_id)
    if run.exposure_id is not None
    else None
)
spec = await _with_exposure_prompt(spec, exposure, channel_directory=None)
spec = _with_channel_tools(spec, exposure)
```

The directory can legitimately stay `None` here; `_with_channel_tools` still restores the binding so the toolset builds when a directory is present. Worth a regression test on the approve-then-resume path for a channel agent - nothing covers it today, which is why the drop is invisible.

backend/alembic/versions/0016_seed_exposure_prompts.py
```diff
@@ -0,0 +1,57 @@
...
+Revises: 0015_exposure_prompt
+Create Date: 2026-08-09
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+from app.services.channels.formatting import HOUSE_STYLE
```
DEENUU1: This imports the *mutable* `HOUSE_STYLE` constant into a migration. Migration 0019 makes the opposite call on purpose and writes the reason in-file: a migration must not import application code, because "a migration that ran once has to keep meaning what it meant."

`0013` importing `vault` is justified - the crypto has to stay in sync. But `HOUSE_STYLE` is just data, and `downgrade()` matches `prompt = :prompt` against the *current* text. Edit those strings later and this migration silently stops clearing the rows it seeded, because the text no longer matches. Inline the seed strings as literals the way 0019 does, so upgrade and downgrade keep meaning what they meant when they ran.

backend/app/services/channel_link.py
```diff
@@ -0,0 +1,243 @@
...
+
+        Any request this chat account already has is dropped first, so a URL that
+        scrolled out of view stops working rather than lingering.
+        """
+        await channel_link_request_repo.delete_for_identity(
+            self.db,
+            platform=incoming.platform,
+            platform_user_id=incoming.platform_user_id,
+        )
+        request = await channel_link_request_repo.create(
```
DEENUU1: nit: this is check-then-act across sessions - `delete_for_identity` then `create`, guarded only by the `channel_link_requests_identity_key` unique constraint. Two near-simultaneous first messages from the same chat account run in two sessions, both delete (neither sees the other's uncommitted row), both insert the same `(platform, platform_user_id)`, and the second commit raises `IntegrityError`.

On the `/link` command path that's caught. On the unlinked-first-message path (`router.py` around 151, `_invite_to_link`) it isn't, so it bubbles to the catch-all and the user just gets no reply - which reads like a broken bot. No data corruption (the constraint holds), so it's a nit, but cheap to close: catch `IntegrityError` in `request()` and re-read the surviving row rather than letting it propagate.

DEENUU1 (commented):
Second pass over this, building on the review already inline. Everything I flagged before still stands at this HEAD - the only commit since was the capability-drop fix, which touched none of those lines - so I'm not repeating them here. Three new things.

**The merge blocker is the refusal tests, not a cosmetic number.** `make test` against a real DB at this HEAD lands at 99.56%, and the gap is almost all the `@slug` refusal branches: `mentions.py` (unknown handle, paused/unexposed binding, unlinked sender, chart-render failure - 217, 294-300, the 309 branches, 392->399, 433, 463-467, 470, 480, 494-502), the two new query functions in `repositories/agent_exposure.py` (109-121, 134, 158-161), and one channel-run branch in `agent_runner.py:2749`. The gate is pointing straight at the paths this platform most wants asserted.

**One real bug I missed the first time: `resume()` drops the exposure prompt and the channel-tools capability** - details inline on `agent_runner.py`. It ties to the plan's own S14, so I read it as a second blocker.

The rest is cleanup: a mutable-constant import in migration 0016, a check-then-act in the link mint, a docstring that disagrees with its code. And there's an untracked `alembic/versions/2026-08-09_.py` sitting in the tree - it's the autogenerate probe (good news: it confirms no model-vs-migration drift), just delete it before it gets committed and reads as a stray head after 0021.

backend/app/services/channels/prompt_variables.py
```diff
@@ -0,0 +1,183 @@
...
+
+    filled = PLACEHOLDER.sub(
+        lambda match: (
+            _flatten(values[match.group(1)]) or UNAVAILABLE
+            if match.group(1) in values
+            else (UNAVAILABLE if match.group(1) in wanted else match.group(0))
+        ),
+        prompt,
+    )
+    return f"{filled}\n\n{DATA_NOT_ORDERS}"
```
DEENUU1, line 189: nit: the `DATA_NOT_ORDERS` warning is appended whenever any placeholder was *wanted*, even if every source failed and they all resolved to `(unavailable)`. The docstring up top says it's appended "when any placeholder was filled" - but nothing was filled in that case, no external data reached the agent, and it still gets told "the values above were written by other people." Harmless over-warning, but code and docstring disagree. Either gate the suffix on at least one value actually resolving, or soften the docstring to match what it does.

DEENUU1:
All review findings are addressed at `bd2c4c38`. Mapping each thread to the commit that closed it:

| Finding | Commit |
|---|---|
| `router.py` — a failed final live-reply edit blanked the answer | `39bd126e` |
| `router.py` — "needs approval" claimed for any empty answer, not just a parked run | `8d839b7f` |
| `chart_png.py` — stacked bars overflowed the plot (axis ignored the stack sum) | `029b772a` |
| `chart_png.py` — a non-6-digit-hex / named colour crashed the area renderer | `59decf8d` |
| `chart_png.py` — `NaN`/`Infinity` poisoned the bounds and sent no chart | `7f3c021a` |
| `embed_session.py` — the `_context_sent` latch froze the per-frame supplied block | `d7b416c9` |
| `_directory.py` / `base.py` / `mattermost.py` — "oldest last" was the reverse of the contract | `3d583d7b` |
| `_toolset.py` — an unbounded post body / forged `author: text` line | `a104cc1e` |
| `_toolset.py` — a member with only a username rendered the name twice | `b5e0a1e0` |
| `agent_runner.py` — `resume()` dropped the exposure prompt and channel-tools capability (S14 blocker) | `9e0a42c0` |
| `0016_seed_exposure_prompts.py` — imported the mutable `HOUSE_STYLE` into a migration | `b1cb697e` |
| `channel_link.py` — check-then-act race on the first unlinked message | `7d270cb6` |
| `prompt_variables.py` — the `DATA_NOT_ORDERS` warning fired even when nothing resolved | `bd2c4c38` |

**#167 (webhook dedup) is deliberately not in this branch** — the fast 200 defuses the slow-webhook retry, but dropping a redelivery whose 200 was lost needs a persisted `(bot, message_id)` guard. The receiver's docstring says so (`b0a95de8`), and the issue stays open rather than being claimed here.

Resolving these threads.

PR #525, opened 2026-08-10
feat(api): read a run's transcript, authorized not owned

### LINKED ISSUE #490: No route reads a run's transcript, and reading one must be authorized not owned
## What is missing

`messages.run_id` exists, the transcript is written for every non-streaming surface, and
**there is no route that reads it.** `app/api/routes/v1/runs.py` has six endpoints and
none of them answers "the steps of this run":

```
GET  /runs            GET  /runs/{id}       POST /runs/{id}/resume
GET  /approvals       POST /approvals/{id}  GET  /spend
```

So the work that made a run's transcript recordable is not yet readable, and the Activity
page has no run detail view to open.

## Why this is a route and not a filter on an existing one

The obvious shortcut is to let the caller ask the conversation endpoint for messages and
filter by `run_id` client-side. That is wrong on authorization, which is the whole point
of the issue:

**Reading a run must be authorized, not owned.** The two are different questions and this
platform's value is in the difference:

- an **owner** opening a *colleague's* run gets its steps - authority inside an
  organization is a membership row plus the permission catalog, not "did you start it";
- the same request for **another tenant's** run is refused, and refused the same way a
  run that does not exist is, so the response cannot be used to discover that a run
  exists;
- **the conversation endpoint one route over is left exactly as strict as it is today.**
  A run's transcript being readable by a colleague must not widen who can read the
  *thread* it sits in - a conversation is a person's, and a run is the organization's.

A route with `resolve_access` in the service behind it can hold all three. A filter on
the conversation endpoint inherits that endpoint's ownership check, which fails the first
condition, and widening *that* check to fix it fails the third.

## Scope

- The route, with the read in a service and `resolve_access` deciding it - never a
  `require(...)` gate, which cannot see the grants on a row (per-resource routes are in
  `CLAUDE.md`'s hard boundaries).
- Its refusal tests: the three cases above, plus a run whose conversation is `NULL`
  (an API call that passed no `conversation_id`), which has no transcript and must say
  so rather than answering an empty list that reads as "it did nothing".
- The frontend detail view that opens it, with the same empty-versus-failed distinction
  the three tabs now have.

An integration test where an owner reads a colleague's run and gets its steps; the same
call for another organization's run is refused; and a test asserting the conversation
endpoint's own authorization is unchanged - that last one is what stops this being fixed
by loosening the wrong thing.

## Why it is not in #202

Not in #45's Done-when list: #45 asks for the three tabs, and this is a fourth surface
plus a new route. #202 records it under "not `Closes`" rather than leaving it implied.
`docs/design/activity-plan.md` §7 priced it as "a new route, not a reuse", which is the
call this issue carries forward.

## Where it sits

Activity cluster on #168, under #45, and it is the *consumer* of the `messages.run_id`
work that landed in #202 - so it is the piece that makes that migration worth having.
Nearest neighbour: #139 for the table it will render steps in.

Found finishing #202. Refs #202.

OchnikBartek:
## Summary

Adds `GET /api/v1/runs/{run_id}/transcript`, which returns a run's messages to any colleague in the same organization who holds `runs:view` — a run is read by authorization, not by ownership. A caller from another tenant is refused exactly as a run that does not exist is, so existence never leaks.

## Related Issue

Closes #490.

## Added

- `GET /api/v1/runs/{run_id}/transcript` (paginated `skip`/`limit`), returning `RunTranscript` (`run_id`, `conversation_id | null`, `items: MessageRead[]`, `total`). A `null` `conversation_id` says the run has no transcript, distinct from an empty `items`.
- `AgentRunnerService.get_run_transcript`: resolves the run org-scoped (404 before the permission is read), then checks `runs:view` (403), then short-circuits a run with no conversation.
- `conversation` repository `get_messages_by_run` / `count_messages_by_run`.

## Changed

- Documented the endpoint and its "authorized, not owned" rule in `docs/governance.md` and `docs/concepts.md`.

## Checks

## Notes for Reviewers

- Cross-tenant and unknown-id return an identical `404` body — a test asserts the two responses match. The adjacent conversation-messages endpoint is left owner-strict (a test proves it still refuses a non-owner colleague).
- Verified locally: `make lint-backend` clean, `make test` 4050 passed / platform layer 100%, `make docs-build` (strict) clean.


backend/app/api/routes/v1/runs.py
```diff
@@ -156,6 +157,36 @@ async def get_run(run_id: UUID, service: AgentRunnerSvc, ctx: Auth) -> Any:
...
+
+    `conversation_id` is null when the run ran with no conversation, which is how
+    the response says "there is no transcript" rather than answering an empty list
+    that reads as "the run did nothing".
+    """
+    run, messages, total = await service.get_run_transcript(ctx, run_id, skip=skip, limit=limit)
+    return RunTranscript(
+        run_id=run.id,
+        conversation_id=run.conversation_id,
+        items=messages,  # ty: ignore[invalid-argument-type]
```
DEENUU1: Small one: this `ty: ignore` has no comment saying what makes it safe, which our hard boundary asks for. You can also just remove it — validate per item like `list_messages` does:

```suggestion
        items=[MessageRead.model_validate(m) for m in messages],
```

(add `from app.schemas.conversation import MessageRead` up top). Typed correctly, no suppression, and the two message-serializing paths match.

DEENUU1 (dismissed):
Nice one. This does exactly what #490 asked for, and the test coverage is the best part of it — the cross-tenant 404 parity, the 404-before-403 ordering, the null-conversation case, and especially the test proving `GET /conversations/{id}/messages` stays owner-scoped. That last one is the whole point of the issue and it's good to see it pinned down.

A few small things, none blocking:

- `backend/app/services/agent_runner.py` `get_run_transcript`: putting the `runs:view` check in the service reads a bit heavier than what it does — for a run there are no per-row grants, so `ctx.has(Perm.RUNS_VIEW)` is the right primitive (not `resolve_access`, despite the issue's wording). It works, but note the sibling `GET /runs/{run_id}` one route up guards with a plain `require(Perm.RUNS_VIEW)` gate and lets `service.get_run` do only the org-scoping. Both are tenant-safe. Worth a thought whether you want the two run-read routes to guard the same way — not a change I'd hold the PR for.
- Heads-up for the frontend detail view: this serializes raw messages, so `user_rating`/`rating_count` come back null (the conversation endpoint fills them only when a `user_id` is passed). Fine for a step view; just flagging it before someone wires up the thumbs and wonders why they're always empty.

I checked the things that could have bitten: `messages.run_id` is indexed so the new query is fine, the null-conversation short-circuit is safe because `record()` returns early when there's no conversation and `messages.conversation_id` is NOT NULL, and `le=500` matches the messages endpoint it mirrors. Good to merge once you've had a look at the inline note.

PR #526, opened 2026-08-10
feat(activity): add version strip sharing the donut's completed share

### LINKED ISSUE #489: Activity has no version strip, and its completed share must match the dashboard donut
## What is missing

A run's agent **version** is recorded on every row, and nothing on the Activity page
reads it. So "we published v7 on Tuesday and Wednesday's runs started failing" is a
question the data can answer and the product cannot.

The strip is one row per version of the agent being looked at, each with:

- how many runs that version served, and what share of them **completed**;
- its cost per run, so a version that got more expensive is visible as one;
- the window it covers, shared with every other figure on the page.

## The condition that makes it worth building

**Its completed share has to be the same number as the dashboard's Outcomes donut over
the same rows, and a test has to say so.** Both read one `by_status`, and the way they
drift apart is one word: a donut that treats `cancelled` as "not a failure" and a strip
that counts it as "not completed" are both defensible and disagree on the same rows. The
figure is `completed / total` with **nothing excluded** - a window holding one
`cancelled` and one `budget_exceeded` run produces the same percentage in both places.

That is §8a.4 of `docs/design/activity-plan.md`, and it is the reason this is not just
"add a table".

## Why it is not in #202

Two reasons, and the second is the blocking one.

1. It is not in #45's Done-when list. #202 closes #45 on that list; this is the plan
   going further than the issue, which is worth keeping and not worth holding a merged
   branch for.
2. **The donut it must agree with is not on `main`.** It is on
   `feat/dashboard-page-demo`, being merged with main by another session as #202 lands.
   Writing the agreement test against a component that is not there yet means either
   guessing at its arithmetic or waiting - and guessing at it is exactly the drift the
   test exists to prevent.

So this waits for that branch, then gets built against the real donut.

- The strip renders on the Activity page, one row per version, over the page's window.
- A test seeds a window with one `completed`, one `cancelled` and one `budget_exceeded`
  run and asserts the strip's percentage **equals** the dashboard donut's for the same
  rows - failing if either side starts excluding a status.
- An agent used *only* as somebody's delegate has a non-empty strip, and its cost per
  run does not count a delegation twice (§2a).

## Where it sits

Part of the Activity cluster on #168, under #45. Nearest neighbours: #139 (the shared
table primitive - this strip should use it once it exists rather than growing a second
bespoke table), and the dashboard work on #149 which owns the donut this has to agree
with.

Found finishing #202. Refs #202.

OchnikBartek:
## Summary

Adds the Activity version strip and makes its "completed share" structurally identical to the dashboard Outcomes donut's, by routing both through one shared computation. `cancelled` and `budget_exceeded` sit in the denominator on both sides, so the two figures cannot drift.

## Related Issue

Closes #489.

## Added

- `src/lib/run-outcomes.ts` — `completedShare`, `statusTally`, `versionTally`, `formatCompletedShare`; the single source both surfaces compute through.
- `src/components/runs/version-strip.tsx` — one card per version (runs, completed share, cost/run, p95, current-version marker), wired into `run-history-tab.tsx` above the table when narrowed to an agent.

## Changed

- The dashboard Outcomes donut (`dashboard/widgets/outcomes.tsx`) now derives its centre label from the shared helper instead of an inline computation (same displayed number).

## Checks

## Notes for Reviewers

- The invariant is enforced by a test that renders the real `OutcomesWidget` and the real `VersionStrip` from the same rows and asserts identical shares (1 completed of 3 → 33%, not 50%).
- Deferred to later Activity slices, not in scope here: the strip's shared date-range control (a 30-day default for now) and click-to-filter version chips (the `/runs` `agent_version_id` filter isn't built yet). This PR closes the version-strip + donut-parity contract of #489; those two remainders should carry into a follow-up.
- Verified locally: `make lint` clean, `make test-frontend-cov` green (branches 97.65%).


frontend/src/lib/run-outcomes.ts
```diff
@@ -0,0 +1,65 @@
...
+  total: number;
+}
+
+/**
+ * The completed share as a fraction in [0, 1], or null for an empty window.
+ *
+ * Null rather than zero: a share of no runs is unknown, not zero percent, and
+ * a caller renders the two differently ("—" versus "0%").
+ */
+export function completedShare({ completed, total }: RunTally): number | null {
```
DEENUU1, line 29: This is the right shape for the whole thing. Pulling the share into one function with `cancelled` and `budget_exceeded` in the denominator on both sides is exactly what the issue asked for, and returning `null` for an empty window instead of `0` — so the caller renders "—" rather than a misleading "0%" — is a nice call.

I checked the backend to be sure the donut's denominator swap is really a no-op: `count_runs` (the old `total_runs`) and `runs_by_dimension(status)` (the summed tally) both run through the same `_window_conditions` with delegations off, so they're equal, and Intl percent rounding matches the old `Math.round` for a fraction in [0, 1]. Same displayed number. Solid.

frontend/src/components/runs/version-strip.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * `by_status` must not drift apart under one word.
+ *
+ * Runs count what the agent did as somebody's delegate too, so a specialist
+ * used only as a delegate is not left with an empty strip; cost per run is an
+ * average of the agent's own rows, so it never counts a delegation twice the
+ * way a sum across a parent and its child would (activity-plan.md §2a, §6).
+ */
+export function VersionStrip({ agentId }: { agentId: string }) {
+  const t = useTranslations("pages.runs");
+  const period = useMemo(() => resolvePreset(DEFAULT_PRESET), []);
```
DEENUU1, line 39: One thing to watch: the strip is fixed to a 30-day window, but the table right below it (`useRuns` in `run-history-tab`) isn't windowed at all — it shows the recent runs the endpoint returns. So on the same screen a version can read "4 runs" in the strip while the table shows more, with nothing telling the user the two count different spans.

I saw from the PR notes that the shared date-range control is deferred, which is fine. Until it lands I'd just name the window in the caption (e.g. "over the last 30 days") so the smaller number doesn't get read as a bug. Cheap now, saves a "why don't these match" later.

frontend/src/components/runs/run-history-tab.tsx
```diff
@@ -36,52 +37,59 @@ export function RunHistoryTab({
...
-          Every run records the agent <em>{t("version")}</em>
-          {t("executedSoWhatHappened")}
-        </CardDescription>
-        {/* Said out loud, with the way out beside it. A filtered table that
+    <div className="space-y-4">
+      {/* Narrowed to an agent, a per-version summary sits above the table - the
+          builder's "did v4 behave better than v3" answered where the evidence
+          is. Its completed share is the shared `completedShare`, so it reads as
+          the same figure the dashboard's Outcomes donut shows (§8a.4). */}
+      {agentId !== null && <VersionStrip agentId={agentId} />}
```
DEENUU1: Small one: this renders whenever there's an `agentId`, including when we're also focused on a single run (`?agent=X&run=Y`). In that mode the card below switches to a single `FocusedRun`, so a per-version summary above it is out of place. I'd gate it on the focused case too:

```suggestion
      {agentId !== null && focusedRunId === null && <VersionStrip agentId={agentId} />}
```

The delegation handover uses `?run=` on its own (agentId null), so you won't usually hit it — but it's a one-word guard.

frontend/src/components/runs/version-strip.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { byVersion, isLoading, error, refetch } = useVersionUsage(agentId, period);
+  const { agent } = useAgent(agentId);
+  const currentVersionId = agent?.current_version_id ?? null;
+
+  if (isLoading) {
+    return <Skeleton className="h-28 w-full" />;
```
DEENUU1: Nit: the loading and error states return a bare `<Skeleton>` / `<ErrorState>`, but the loaded state is a full `Card` with a title and caption — so the block jumps shape as it resolves. Not a big deal, but wrapping the skeleton in the same `Card`/`CardHeader` shell (title + a skeleton body) would keep the layout from shifting under the user.

DEENUU1 (dismissed):
Nice work — this does what #489 asked and does it carefully. The shared `run-outcomes` helper is the right call, the parity test genuinely renders both the real donut and the real strip from one set of rows (not a stub asserting against itself), and returning `null` for an empty window instead of `0` so the caller can render "—" vs "0%" is a thoughtful touch. I traced the backend too: `total_runs` and the sum of `by_status` come from the same window conditions, so swapping the donut's denominator to the summed tally really is the same number it showed before — down to the rounding, since Intl's percent rounding matches the old `Math.round` over [0,1]. No regression there.

Nothing blocking. Two small things worth a look: the strip's 30-day window diverges from the unwindowed table right below it (worth naming the window in the caption until the shared date control lands), and the strip shows even when the page is focused on a single run, where it's out of place. Both minor — the rest I'd merge as is.

PR #527, opened 2026-08-10
feat(governance): show per-person spend on the Activity tab

### LINKED ISSUE #214: Per-person usage and cost breakdown on the dashboard and Spend tab
## What this is

**Show usage and cost broken down per person** — who ran how many agents, and what it cost — on the dashboard and on Activity's Spend tab.

Both stage-1 designs refused this, in the same words. #37's design note, *Decisions taken in stage 1*, decision 3: *"Adoption is a count, not a table of names. '14 of 23 members ran an agent' answers the steward's question without shipping a surveillance table."* #45's design repeated it, calling a per-person spend table *"the surveillance table with a dollar sign on it"*.

**That decision is reversed — the breakdown is in scope.** The reasoning above is not deleted, it is overruled: a steward asking "who is actually using this, and who is spending the budget" is a legitimate question, and the answer currently requires database access.

## It is cheaper than either design implies

`group_by=user` is **already in the frozen `/stats/usage` vocabulary** agreed on PR #149 — `day | surface | agent | version | user | status | model | exposure | environment`. So the endpoint was never what was refused; only the rendering was. Same on the runs side: #45's `list_runs` work already adds a `user_id` filter.

What is left is a table, and three decisions that belong on the page rather than in code review.

## The three decisions to make first

- [ ] **Whose names are visible to whom.** The obvious gate is `runs:view` — but that is held by builder and operator as well as the two stewards, which is a wider audience than the person on the list may assume. If a narrower gate is wanted (`members:manage`, say, which is already the gate on the members list), decide that here.
- [ ] **Usage, cost, or both.** "Who ran the most agents" and "who spent the most money" read very differently to the person being listed, and only the second is a budget question.
- [ ] **Whether a member can see the table at all.** They can already see their own numbers at `scope=own` (#37 decision 2, unchanged). A member seeing colleagues ranked by spend is a different thing.

## What to build

- [ ] `GET /api/v1/stats/usage?group_by=user` returning `user_id`, a display name, run count and cost — one row per person, with the caller's chosen gate enforced in the route.
- [ ] A per-person table on the dashboard, **under** the existing aggregate card rather than instead of it: "14 of 23 members ran an agent" stays as the headline, the table is the evidence beneath it. That is the same summary-and-its-evidence pattern the rest of both pages follow.
- [ ] The same breakdown on Activity's Spend tab, alongside By provider / By key / By agent, obeying that tab's date range.
- [ ] A person with zero runs is a row saying zero, not an absent row — "who has *not* adopted this" is the more useful half of the question and it disappears if the table only lists people with activity.
- [ ] A deleted user's rows are kept and muted, the way `CostByProvider` handles an unrecorded provider and `CostByKey` a deleted key. The spend happened.
- [ ] Partial cost marked as a floor, and a wholly unpriced run not rendered as `$0.00` — the same rule as every other cost figure on these pages.
- [ ] Every string through `next-intl`.

## Tests

- [ ] A caller without the chosen permission does not see the table, and the control that reveals it is **absent** rather than disabled or 403.
- [ ] Tenant isolation: no row from another organization, tested including the case where the caller owns it.
- [ ] The rows sum to the same total the aggregate card shows — a breakdown that disagrees with the number above it is worse than no breakdown.
- [ ] On a deployment using delegation, the totals do not double-count: a delegate's run row is already inside its parent's cost (#170).

## Where it sits

Cluster: **Frontend surfaces** (#37 #45) plus **Accounting and correctness** for the cost columns. Depends on **#37** stage 2 (the `/stats/usage` endpoint). Related: **#170** — three spend aggregates double-count a delegated run, and a per-person cost column inherits exactly that question; **#211** (CSV export), where a naive export of these rows reintroduces the same table as a download and needs the same gate decision. Reverses #37 decision 3 and #45's §6; sign-off comments: https://github.com/vstorm-co/agenticos/issues/37#issuecomment-5190861344

OchnikBartek:
## Summary

Adds a per-person spend breakdown to Activity's Spend tab, reading the existing `/stats/usage?group_by=user` endpoint and gated on `runs:view`. The dashboard per-person widgets already shipped in #149; this completes the Spend-tab side.

## Related Issue

Closes #214.

## Added

- `src/components/runs/spend-by-person.tsx` — the `SpendByPerson` card, gated on `runs:view` (renders nothing and issues no query without it), reading `/stats/usage?group_by=user` over the tab's date window, delegated runs excluded.
- A "+N others" line (mirroring the dashboard `TopPeopleWidget`) when `active_users` exceeds the rows rendered, so a hard top-10 list doesn't read as the whole org.

## Changed

- `src/components/runs/spend-tab.tsx` mounts `SpendByPerson` beneath By agent, passing the tab's resolved window.

## Checks

## Notes for Reviewers

- The `/stats/usage` `group_by=user` gate was already correct (scope decided in `StatsService`, `runs:view` for org scope) — no backend change was needed.
- Absence-without-permission is proven by an integration test (control not in the document, and no usage query is issued).
- **Known gap — this card can under-count against the By agent card above it, tracked in #539.** Two causes: the windows differ (By agent ends at `now` via `/spend`; this uses `/stats/usage` truncated to whole UTC dates), and `usage_by_user` inner-joins `User`, so a deleted user's runs (and account-less channel runs) drop off this card while By agent still counts them — which is #214's "deleted-user rows kept and muted" AC, structurally impossible on the current endpoint. The two cards should not be summed against each other until #539 extends the endpoint.
- Verified locally: `make lint-frontend` clean, `make test-frontend-cov` green; the "+N others" disclosure and its ICU-plural key covered by `spend-by-person.integration.test.tsx` (7 passed) and `messages/catalog.test.ts`.


DEENUU1 (commented):
Nice piece of work — this slots cleanly into the pattern the rest of the Spend tab and the #149 dashboard card already set. It reuses `usePeopleUsage` and `PersonUsageRow` rather than growing a parallel path, the docstrings carry the reasoning, and `docs/governance.md` moved in the same change. The two things I'd most want to see here are done right: the refusal is proven by the card being absent *and* the query never firing, and a failed request is kept distinct from a quiet month. A few non-blocking notes below, mostly about how honestly this card reconciles with the By agent card above it — worth a sentence in the PR rather than code changes.

frontend/src/components/runs/spend-tab.tsx
```diff
@@ -134,6 +135,19 @@ export function SpendTab() {
...
         </CardContent>
       </Card>
+
+      {/* Who spent it, over the same window the breakdowns above read. The
+          people rows come from `/stats/usage?group_by=user` rather than `/spend`,
+          so the window is handed over as the range this tab resolved to: the
+          `from`/`to` the caption already reflects, with an open-ended window read
+          as "up to now". Gated on runs:view inside the card, so it is absent for
+          a caller without it rather than a refused request. */}
+      {spend?.from_date != null && (
```
DEENUU1, line 145: One thing worth stating plainly in the PR: this card and the By agent card above it won't always sum to the same total, and #214 specifically called that out ("a breakdown that disagrees with the number above it is worse than no breakdown"). Two reasons.

The By agent rows come from `/spend` (a precise `now` end), this one from `/stats/usage` over `from`/`to` truncated to whole dates — so the window edges can differ by up to a day. And more concretely: `usage_by_user` inner-joins `User`, so any run whose user was deleted (or a channel run with no account behind it) is dropped from this card entirely, while By agent still counts it. That's the "deleted user's rows are kept and muted" AC — it's not just deferred on the frontend, it's structurally impossible on this endpoint.

All reasonable to punt, but a steward comparing the two cards will see this one under-count with nothing explaining the gap. I'd say so in the PR body, and ideally file a follow-up issue so it isn't lost.

frontend/src/components/runs/spend-by-person.tsx
```diff
@@ -0,0 +1,79 @@
...
+                <span className="min-w-0 flex-1 truncate font-medium">
+                  {person.full_name ?? person.email}
+                </span>
+                <span className="text-muted-foreground text-xs">
+                  {t("runCount", { count: person.runs })}
+                </span>
+                <span className="font-mono">${Number(person.cost_usd).toFixed(4)}</span>
+              </div>
+            ))}
+            <p className="text-muted-foreground text-xs">{t("perPersonDisclosure")}</p>
```
DEENUU1, line 82: The dashboard sibling (`TopPeopleWidget`) shows a "+N others" line when `active_users` exceeds the rows it renders, so a reader can tell top-6 from all. Here it's a hard top-10 with no such affordance — if an org has 40 people, this looks like the whole org.

You've already got the composed `active_users` count available cheaply via `useUsageStats` for the same window (that's exactly what the dashboard card does). Worth adding the same "+N others" line so the list doesn't read as complete when it isn't.

frontend/src/components/runs/spend-by-person.tsx
```diff
@@ -0,0 +1,79 @@
...
+                key={person.user_id}
+                className="flex items-center justify-between gap-3 rounded-md border p-3 text-sm"
+              >
+                <span className="min-w-0 flex-1 truncate font-medium">
+                  {person.full_name ?? person.email}
+                </span>
+                <span className="text-muted-foreground text-xs">
+                  {t("runCount", { count: person.runs })}
+                </span>
+                <span className="font-mono">${Number(person.cost_usd).toFixed(4)}</span>
```
DEENUU1, line 74: nit: a person whose runs were all on an unpriced model renders `$0.0000`, which reads as "spent nothing" on a page about money — the exact trap the tab warns about. The tab-level `someRunsCouldNotBePriced` note at the top does cover it, so this is fine as-is; just flagging that `PersonUsageRow` has no `partial_run_count`, so unlike the By agent rows this card can't mark its own floor. Endpoint limitation, not yours to fix here.

frontend/src/components/runs/spend-by-person.integration.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+    // The refusal that is the point: not disabled, not a 403 after the fact - the
+    // card is not rendered and the question is never put to the server.
+    auth.can = holds("agents:view", "agents:run");
+    serve(PEOPLE);
+
+    const { container } = render(<SpendByPerson from="2026-07-08" to="2026-08-07" />, { wrapper });
+
+    expect(container).toBeEmptyDOMElement();
+    expect(screen.queryByText("Katarzyna Nowak")).toBeNull();
+    expect(usageCalls()).toHaveLength(0);
```
DEENUU1, line 150: This is the right test, and the right assertion — `container` empty *and* `usageCalls()` length 0. Proving the question is never put to the server, not just that the control is hidden, is exactly what the frontend rule asks for and what most people skip.

frontend/src/components/runs/spend-tab.tsx
OchnikBartek: Called out in the PR body now, and filed as #539 to extend the endpoint so the two cards reconcile.

DEENUU1:
Landed on main via #578 — identical content, opened from a non-stacked branch. GitHub only offers the async merge endpoint for stacked PRs, and that endpoint does not honour the admin review-bypass this solo-maintainer repo relies on, so #527 could not be merged directly. Closing as superseded; the change shipped in 0.0.95.

PR #528, opened 2026-08-10
feat(runs): sort and filter run history by duration in the UI

### LINKED ISSUE #210: Sort and filter run history by duration, so p95 has rows behind it
## What this is for

The dashboard (#37 / PR #149) reports latency as `p50` and `p95` from `/stats/usage`. There is no way to get from "p95 is 14.8s" to **those runs**.

Activity (#45 / PR #202) renders a `Took` column and a per-version `p50` in its version strip, but neither is sortable and there is no duration filter — so the slowest runs on a deployment are reachable only by paging through history reading a column by eye.

That is the same gap the rest of the Activity design closes for every other dimension: a number on the dashboard, and the rows behind it one click away. Duration is the one dimension where the rows are unreachable.

## What to build

- [ ] **Sort by duration** on `GET /runs`, both directions. `ended_at - started_at`, computed in SQL rather than over a page — sorting one page of 25 is sorting the wrong set.
- [ ] **A minimum-duration filter** (`took_over_ms` or similar), so "everything slower than 30 seconds" is a query rather than a scan.
- [ ] The Activity table's `Took` header becomes a sort control, matching the `DataTable` pattern `admin/conversations` already uses.
- [ ] A "slow runs" entry in the canned-view row Activity ships, so the common case is one click.
- [ ] The dashboard's p95 figure links here, sorted by duration over the same window.

## Notes

- **`ended_at` is nullable.** A `running` or `awaiting_approval` run has no duration, and the sort has to place those deliberately rather than by accident — `NULLS LAST` on the descending sort, and they are not "fast". A still-running run's *age* is a separate and arguably more urgent question, handled in #45's stage 2.
- There is no index for this: `ix_agent_runs_org_started` covers `(organization_id, started_at)`. A sort on a computed interval over a large history is a sequential scan per page. Either accept it inside an already-narrowed window (the date range is mandatory on this page anyway) or add a generated column — measure before adding one.
- Latency belongs to the run row, so this needs no new aggregate. It is a repository filter plus an order-by, in the same change as #45's other `list_runs` parameters if that lands first.

- [ ] Sorting run history by duration descending puts the slowest completed run first, with unfinished runs not competing for that position, proven against a real database.
- [ ] A duration filter returns only runs over the threshold, and a run with no `ended_at` is excluded rather than treated as zero.
- [ ] The p95 figure on the dashboard reaches run history sorted by duration over the same window — the number and its evidence, which is the rule the rest of these two pages already follow.

## Where it sits

Cluster: **Frontend surfaces** (#37 #45). Depends on nothing, but is cheapest inside #45's `list_runs` work, which already adds `status`, `surface`, `user_id`, `started_from`/`started_to`, `environment_id`, `exposure_id` and `agent_version_id` — this is one more parameter on the same route and repository. Came out of reviewing PR #202 against PR #149's widget map. Nearest duplicate: none; #198 is the closest neighbour in that it is also a figure on this page that cannot be reconciled with what is under it.

OchnikBartek:
## Summary

Adds the run-history duration controls the issue asked for: a sortable `Took` column, a "slow runs" canned view, and a dashboard p95 deep-link into the sorted list. The backend query (`ended_at - started_at` in SQL, `order_by=duration`, `took_over_ms`, nulls-last) landed with #202 and is reused unchanged.

## Related Issue

Closes #210.

## Added

- A sortable `Took` column and Started/Took `SortButton` headers on the run table (server-side sort over the whole narrowed set, not one page).
- "Slow runs" / "All runs" canned-view toggles.
- A dashboard p95 deep-link (`/runs?sort=duration&started_from=…&started_to=…`) that seeds the sort and window; `formatRunDuration` in `src/lib/utils.ts`.

## Changed

- `use-runs.ts` and `qk.runs.list` take the new sort/filter/window options.

## Checks

## Notes for Reviewers

- Null-duration semantics (a run with no `ended_at`) mirror the backend: it sorts last in both directions, is excluded from the `took_over_ms` filter, and renders `-`.
- Verified locally: `make lint` clean, `make test-frontend-cov` green (branches 97.67%); the #202 backend contract re-checked (17 route + 5 integration tests).


frontend/src/components/runs/run-history-tab.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+import messages from "../../../messages/en.json";
+import { RunHistoryTab } from "./run-history-tab";
+import type { AgentRun } from "@/types/runs";
+
+/**
+ * The sort, the window and the "slow runs" preset - the controls #210 adds.
+ *
+ * All three are the server's over the whole narrowed set, so the assertion is on
+ * the options `useRuns` is called with rather than on which rows come back: a
```
DEENUU1, line 14: Nice — asserting on the options `useRuns` is called with, not on which rows render, is exactly right here. A test that read rows off the page would pass just as happily when someone sorted one page of 25 instead of the whole set, which is the bug #210 is about. Good instinct.

frontend/src/lib/utils.ts
```diff
@@ -112,6 +112,22 @@ export function formatDateTime(date: Date | string): string {
...
+ * different fact from a run that was fast. The Took column and the duration sort
+ * both rest on that distinction, so an unfinished run reads as unknown here and
+ * never as zero.
+ */
+export function formatRunDuration(startedAt: string | null, endedAt: string | null): string {
+  if (!startedAt || !endedAt) return "-";
+  const ms = new Date(endedAt).getTime() - new Date(startedAt).getTime();
+  if (Number.isNaN(ms) || ms < 0) return "-";
+  if (ms < 1000) return `${Math.round(ms)} ms`;
+  return `${(ms / 1000).toLocaleString(undefined, { maximumFractionDigits: 1 })} s`;
```
DEENUU1, line 128: These two lines are a verbatim copy of `formatMs` in `dashboard/format.ts` (same `< 1000 ? ms : s` split, same `toLocaleString`). Worth knowing they've now diverged in one spot — this returns `"-"` for the empty case, `formatMs` returns `"—"`. That difference is deliberate and correct (the table uses `-` like `formatDate` does), but the *number* formatting drifting apart later is the risk. Not blocking; if you wanted to dedupe, a shared `formatDurationMs(ms)` both call would do it, though `formatMs` living in the dashboard module makes that a slightly awkward import direction.

frontend/src/components/dashboard/widgets/latency.tsx
```diff
@@ -1,38 +1,72 @@
...
+/**
+ * Run history sorted by duration over this widget's window - where the p95
+ * figure points. The dashboard reports a period as inclusive whole days, so the
+ * window is widened to cover the last day rather than cut off at its midnight.
+ */
+function slowestRunsHref(period: Period): string {
+  const params = new URLSearchParams({
+    sort: "duration",
+    started_from: `${period.from}T00:00:00.000Z`,
+    started_to: `${period.to}T23:59:59.999Z`,
```
DEENUU1, line 22: Just flagging that I checked this: the stats window is half-open (`started_at >= from-00:00 AND < (to+1)-00:00`), and this link is inclusive (`<= to-23:59:59.999Z`) because the `/runs` filter is `<=`. So a run started in the final sub-millisecond of the last day is in the p95 distribution but not in the linked list. It's a microsecond edge nobody will hit, and given the inclusive `<=` filter your choice is the safest one available — anything at `(to+1)-00:00` would over-include instead. No change needed; the reasoning in the docstring is right.

DEENUU1 (approved):
This is a clean close on #210 — thanks. The part that mattered most is right: the p95 figure and the run list it links to are computed over the same top-level-only population (both filter `parent_run_id IS NULL`), so the number and its evidence actually agree instead of just looking like they do. The tests assert on the options `useRuns` gets rather than on rows off the page, which is the one thing that would have let a "sort the wrong 25" bug through green CI. Two minor, non-blocking notes below and one thing I liked.

PR #529, opened 2026-08-10
refactor(rag): split sync source wizard into per-step components

### LINKED ISSUE #461: Decompose sync-source-wizard.tsx — 761 lines in one component
## What

`frontend/src/components/rag/sync-source-wizard.tsx` is **761 lines** in one component. It is now the largest file in `components/rag/` by a factor of four, and the last one in that directory still shaped the way `rag/[id]/page.tsx` was before #61.

Found while decomposing that page on `feat/rag-one-knowledge-page` (PR #221). Deliberately **not** folded into that branch — it is a real problem and it is not #61's.

## Why it matters

The wizard is where a connector, an org integration and a target collection are chosen and a sync source is created or cloned. It holds the whole multi-step flow, its form state, the connector picker, the integration reuse path and the schedule, in one function. Two consequences:

- A reviewer cannot read one step without the other four in view — which is the same argument #61 made about the page it was extracted from.
- It already carries the failure flags `connectorsFailed` and `orgIntegrationsFailed` (added in #221 for #32). Any future step that reads a list has to thread its own flag through the same single component, and the honest distinction between "empty" and "failed" gets harder to keep with each one.

## Shape it should take

The same one PR #221 used next door, and the pattern is now established in that directory: one component per step under `components/rag/`, each taking the state only it reads, with whatever genuinely spans two steps staying in the wizard and going down as props. `.claude/rules/frontend.md` asks for components under ~100 lines as a direction, not a number to game.

- `sync-source-wizard.tsx` is composition plus the state that spans its steps; no step's markup lives in it.
- The failure flags reach the step that renders the list, still as required booleans — a step handed an already-empty array reintroduces #32.
- `bun run type-check && bun run lint && bun run test:coverage` clean from `frontend/`, and the sync-source specs pass unchanged.

## Related

Part of the same cluster as #61 (the RAG/knowledge pages) — see the map in #168. Closest sibling: #61 itself, whose last Done-when item was this exact problem one file over. Not a duplicate of #141 or #249, which are about `check_i18n.py`'s blind spots rather than file size.

### LINKED ISSUE #540: Hardcoded English in sync-source configure step bypasses the i18n guard
## What breaks

`frontend/src/components/rag/sync-source-configure-step.tsx:26` renders
"No additional configuration needed for {name}" as hardcoded English. Under `pl` it
shows English instead of a translation, and `check_i18n.py` does not catch it —
it is a plain text node broken across lines, which is the #141 guard blind spot.

The string was copied verbatim from the old `sync-source-wizard.tsx` during the
#461 decomposition, so it predates that refactor rather than being introduced by it
(flagged in review of PR #529 so it isn't lost).

## Where

- `frontend/src/components/rag/sync-source-configure-step.tsx:26`

- The string is routed through `next-intl` — a message key with a `{name}`
  parameter, read via `t` — and renders translated under `pl`.
- `check_i18n.py` / `messages/catalog.test.ts` account for it (and, ideally, the
  underlying #141 gap is narrowed so a plain node broken across lines is caught).

## Where it sits

Found reviewing PR #529 (#461). Rides on the guard blind spot tracked by #141;
same "sentence the guard can't see" family as #32 (empty-vs-error). Nearest related
open issue: #141.

OchnikBartek:
## Summary

Decomposes `sync-source-wizard.tsx` (761 lines) into one component per wizard step, following the pattern PR #221 established in `components/rag/`. A structural refactor, plus one review-flagged i18n fix folded in (#540, see Notes) because it lives in a file this branch created.

## Related Issue

Closes #461. Closes #540.

## Changed

- Extracted `sync-source-{clone,connector,configure,schedule}-step.tsx`. The wizard shell (`sync-source-wizard.tsx`, now ~320 lines) keeps cross-step flow, the shared form, the header/step-indicator chrome, and the `connectorsFailed`/`orgIntegrationsFailed` flags it hands to the steps.

## Checks

## Notes for Reviewers

- Behaviour-preserving: the two existing integration tests that exercise the wizard as a whole pass unchanged. No per-step tests were added — the dir is outside the coverage `include` list, and #221 tested its rag components the same way (as a composed whole).
- Carries the #540 fix (commit `76cdb45`): the review flagged `sync-source-configure-step.tsx`'s hardcoded "No additional configuration needed for {name}" — a plain node split across lines, the #141 `check_i18n.py` blind spot, so it rendered English under `pl`. It is now routed through `t.rich` with an `emphasis` tag, matching the sibling `configureRequiredFields`, with a `pl` translation added. Folded in here rather than a separate PR because it is this branch's own file.
- Verified locally: `make lint-frontend` clean, `make test-frontend-cov` green (branches 97.64%); the #540 fix separately checked — `check_i18n.py` clean, `catalog.test.ts` and the wizard integration test green, `tsc` clean.


frontend/src/components/rag/sync-source-connector-step.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { cn } from "@/lib/utils";
+
+export function ConnectorStep({
+  connectors,
+  connectorsFailed,
+  form,
+  setForm,
+}: {
+  connectors: ConnectorInfo[];
+  connectorsFailed?: boolean;
```
DEENUU1: nit: #461's "how you'd know it was fixed" asked for the failure flag to reach the step as a *required* boolean, the idea being a future caller physically can't drop it and quietly bring back #32 (empty list read as "no connectors" when the request actually failed). It was optional before too, so this isn't a regression — but since this is a fresh file it's a cheap moment to tighten it. The wizard already always passes it:
```suggestion
  connectorsFailed: boolean;
```

frontend/src/components/rag/sync-source-configure-step.tsx
```diff
@@ -0,0 +1,109 @@
...
+}) {
+  const t = useTranslations("rag");
+  const fields = Object.entries(connector.config_schema);
+
+  if (fields.length === 0) {
+    return (
+      <div className="border-foreground/10 bg-foreground/[0.03] rounded-xl border p-5 text-center">
+        <Cog className="text-foreground/45 mx-auto h-6 w-6" />
+        <p className="text-foreground/70 mt-3 text-sm">
+          No additional configuration needed for{" "}
```
DEENUU1: Not for this PR — you copied this over verbatim from the old file, so it's not yours to fix here. But flagging it so it doesn't get lost: this "No additional configuration needed for {name}" string is hardcoded English, and `check_i18n.py` misses it only because it's a plain node broken across lines (the #141 blind spot). Worth a quick follow-up issue to route it through `next-intl` like the rest of the step. Mentioning it here so we have a record.

DEENUU1 (dismissed):
Clean, honest refactor. I pulled the branch and diffed each extracted step against the version on `main` — all four bodies are byte-for-byte identical (the schedule step only picks up the `SYNC_MODES`/`SCHEDULE_PRESETS` constants that moved with it), the wizard shell's logic is untouched, and since these steps were already separate functions in the same file the risk here is about as low as a refactor gets. The three callers don't change and CI is green. The PR notes are accurate too — `src/components/rag/**` really is outside the coverage `include`, so testing the wizard as a composed whole (as #221 did) is the right call, not a coverage dodge.

Two optional things below, neither blocking. One matches a line the issue explicitly asked for; the other is a pre-existing i18n gap you inherited, not something this PR caused — worth a follow-up issue rather than folding in here.

frontend/src/components/rag/sync-source-configure-step.tsx
OchnikBartek: Filed as #540 to route it through next-intl (it's the #141 guard blind spot).

PR #530, opened 2026-08-10
test(runs): cover embed and channel runs recording a transcript

### LINKED ISSUE #205: Embed and channel-mention runs record no messages, so a run has a cost and no content

OchnikBartek:
## Summary

Adds regression coverage proving that embed/widget, channel-mention, and default-agent runs record their full transcript — the question, the answer, and the tool calls with the model and version that actually ran — closing out #205's requirement that the fix ship with a regression test.

## Related Issue

Closes #205.

## Added

- `backend/tests/test_surface_transcripts.py` — asserts at the repository boundary that a widget run, a mention, and the default agent each record their turns (role, content, run id, model/version, tool-call args and result), and that a broken widget run still records what the visitor asked.

## Checks

## Notes for Reviewers

- This branch is test-only: the recording behaviour it covers is already on `main`, and these tests are the regression net #205 was owed — they fail if the surface recording is reverted.
- Verified locally: `uv run pytest tests/test_surface_transcripts.py` — 4 passed.


backend/tests/test_surface_transcripts.py
```diff
@@ -0,0 +1,319 @@
...
+client's site, with no row saying what was asked or what was said back (#205).
+
+The fix moved the write into :meth:`AgentRunnerService._run`, the one place a
+non-streaming run executes, so a surface cannot forget it. These tests drive each
+surface's *real* entry point - the widget session, an `@slug` mention, a bot's
+default agent - and assert the turns land at the repository boundary: the
+question, the answer attributed to the model and version that produced it, and
+the tool calls with their arguments and their results. Asserting that
+`transcript.record` was called would prove the wiring and nothing about the
+rows; asserting the rows is the point, because the rows are what was missing.
```
DEENUU1, line 18: Asserting the rows instead of `transcript.record` being called is the whole point of #205: the bug was the missing row, not an unhit mock. Good that this is stated up front.

backend/tests/test_surface_transcripts.py
```diff
@@ -0,0 +1,319 @@
...
+    user, assistant = (call.kwargs for call in conversations.create_message.await_args_list)
+    assert (user["role"], user["content"], user["run_id"]) == ("user", question, run.id)
+    assert (assistant["role"], assistant["content"], assistant["run_id"]) == (
+        "assistant",
+        answer,
+        run.id,
+    )
+    # Attributed to the model and version that ran, not to whatever the agent
+    # says today - a bot that recorded "role and content only" was the #205 case.
+    assert assistant["model_name"] == "gpt-4.1"
```
DEENUU1, line 151: This is the assertion that actually pins the default-agent case — it used to record role and content only. Checking `model_name` and `agent_version_id` is what proves the tool call, model and version all land, not just text.

backend/tests/test_surface_transcripts.py
```diff
@@ -0,0 +1,319 @@
...
+
+        assert answer == "The refund window is 30 days."
+        _assert_full_turn(
+            conversations,
+            captured["prepared"].run,
+            question="What's your refund window?",
+            answer="The refund window is 30 days.",
+        )
+
+    async def test_a_broken_widget_run_still_records_what_the_visitor_asked(self) -> None:
```
DEENUU1, line 202: The broken-run test is the one I most wanted here: the failed run is the one someone opens, and asserting the question survives even when the provider throws is the billing-integrity half of the issue.

backend/tests/test_surface_transcripts.py
```diff
@@ -0,0 +1,319 @@
...
+        conversations.complete_tool_call = AsyncMock()
+        conversations.get_open_tool_call_in_run = AsyncMock(return_value=None)
+        yield captured, conversations
+
+
+def _assert_full_turn(
+    conversations: MagicMock, run: MagicMock, *, question: str, answer: str
+) -> None:
+    """The question, the answer, and the tool call all reached the transcript."""
+    user, assistant = (call.kwargs for call in conversations.create_message.await_args_list)
```
DEENUU1, line 142: This unpack quietly doubles as 'exactly two messages were written', which I like — but if a surface ever writes a third turn it fails with a bare `ValueError: too many values to unpack` that won't say what broke. Optional, but a length assert first reads clearer:

```suggestion
    assert len(conversations.create_message.await_args_list) == 2
    user, assistant = (call.kwargs for call in conversations.create_message.await_args_list)
```

DEENUU1 (dismissed):
This closes the regression net #205 was owed, and it does it the right way — each test drives the surface's real entry point (the widget session, an `@slug` mention, the bot's default agent) and asserts the actual rows landing at the repository boundary, not that `transcript.record` was called. I checked the existing surface tests: `test_agent_embed` and `test_channel_mentions` only assert the run surface and the usage footer, never the transcript rows, so this genuinely fills the gap instead of duplicating them, and it fails if the write in `_run` is reverted. One thing to confirm before merge — the `make check` box is unticked in the description; CI's lint/test/e2e are all green so I think it's fine, just worth ticking or calling out. Nice work.

PR #531, opened 2026-08-10
feat(api): export runs, approvals and spend as CSV

### LINKED ISSUE #211: Export runs, approvals and spend as CSV
## What this is for

Export run history, the approvals record and the spend breakdown as CSV, from the filters already applied on the Activity page.

#45's stage-1 design (PR #202 §6) considered this and deliberately left it out, with a reason worth keeping: *"real want, own decision — an export endpoint has its own tenancy and size questions, and folding it in here would bury them."* This issue is that decision, taken separately rather than dropped.

The demand is not speculative. Everything Activity shows is what somebody reconciles against an invoice, hands to a finance team, or attaches to an audit. A page that can answer the question on screen and not off it sends people to the database.

## The questions this has to answer, which is why it is not a button

- **Tenancy.** An export is a bulk read of `agent_runs` for one organization. It carries the same `runs:view` gate as the page, and — if the member/viewer scope decision in #45 §4 lands — the same `Scope.OWN` floor, enforced in the query rather than as a filter a caller can widen.
- **Size.** `/runs` pages at `limit ≤ 100` precisely so nothing streams the whole table. An export has no such ceiling by nature, so it needs one by design: a row cap with an honest refusal above it, a mandatory date range, or a streaming response — not a synchronous request that holds a connection while it serialises three years of history.
- **Partial cost.** `cost_is_partial` must survive the export. A CSV column of numbers where some are floors and nothing says which is worse than the same problem on screen, because a spreadsheet gets summed. Same for a run whose only model was unpriced and whose cost is `0`.
- **Delegated runs.** After #170 / #181, a run history export must state whether it is top-level only. A CSV that includes both parent and child rows and gets summed double-counts every delegation — which is exactly the bug #170 describes, exported to somewhere nobody can see the fix.
- **PII.** Rows carry `user_id`; a CSV of who ran what, with names resolved, is the per-person table #37's decision 3 refused, arriving via a download. Decide what identity column ships.
- **Audit.** An export is a privileged bulk read and should probably write an `audit_log` entry. Cheap now, impossible to reconstruct later.

## What to build

- [ ] Answer the six questions above **first**, and write the answers into `docs/` — they are the design, and the endpoint is the easy part.
- [ ] `GET /api/v1/runs/export` and the equivalents for approvals and spend, taking the same filter parameters as the list routes, so the export is *what is on screen* rather than a second query with its own semantics.
- [ ] A row ceiling with an honest refusal above it, or a streaming response — not a synchronous request holding a connection while it serialises three years of history.
- [ ] A mandatory date range, for the same reason.
- [ ] `cost_is_partial` as its own column, so a floor survives a spreadsheet sum; a wholly unpriced run must not export as `0`.
- [ ] An explicit stance on delegated rows, stated in the file itself (a header comment or a column), not only in the docs.
- [ ] A download control on Activity, gated identically to the tab it sits on and **absent** — not disabled — without the permission.
- [ ] An `audit_log` entry per export.

- [ ] An export from a filtered view contains exactly the rows the table showed, proven against a real database.
- [ ] A caller from another organization gets nothing, including for a row they own — tested.
- [ ] Every partial cost is marked in a way that survives a spreadsheet sum, and a wholly unpriced run does not export as `0`.
- [ ] Over the row ceiling, the request is refused with a message saying what to narrow, **not truncated silently**.
- [ ] Summing the cost column of a run export on a deployment that uses delegation gives the organization's actual bill, not double it.

## Where it sits

Cluster: **Frontend surfaces** (#45 #37) plus **Accounting and correctness** for the cost columns. Related: #170 (three spend aggregates double-count a delegated run — an export inherits that question), #45 §4 (the member/viewer scope decision, which decides this route's floor), #37 decision 3 (the per-person table, which a naive CSV reintroduces). Deliberately **not** part of PR #202: the design says so, and folding it in would bury the six questions above under a download icon. Nearest duplicate: none among the 137 open issues.

OchnikBartek:
## Summary

Adds three CSV export endpoints — `GET /runs/export`, `/approvals/export`, `/spend/export` — each serialising exactly the rows its list route would return, so the file matches what is on screen. An unbounded export gets the two rules it needs by design: a mandatory date range, and a row cap that refuses (never truncates) above it.

## Related Issue

Closes #211.

## Added

- `RunExportService` (`app/services/run_export.py`) and the three routes, each gated as its list sibling is (runs/spend on `runs:view`, approvals on `approvals:decide`) with the `Scope.OWN` floor enforced in-query.
- Columns that survive a spreadsheet sum: `cost_is_partial` on runs, `partial_run_count` on spend; a wholly unpriced run exports its real `0` beside `cost_is_partial=true`, never a bare `0`. CSV formula-injection is neutralised.
- An `audit_log` entry per export (window, applied filter names, row count — never the request body or a resolved row).
- An export menu on the Activity page, gated on `runs:view` and carrying the applied filters.

## Changed

- Registered the three gated routes in the `CALLS` permission table (`tests/api/test_platform_routes.py`) so the guard proves each gate is the intended one.

## Checks

## Notes for Reviewers

- Delegation stance is stated in the export file: top-level-only by default (so summing `cost_usd` gives the bill, not double it), with a `parent_run_id` column when `include_delegations` is opted into.
- Verified locally: `make test` 4084 passed / platform layer 100%, `make lint` clean, `make test-frontend-cov` green (branches 97.64%).


backend/app/services/run_export.py
```diff
@@ -0,0 +1,499 @@
...
+
+def _cell(value: object) -> str:
+    """One value as text, with `None` an empty cell rather than the word "None"."""
+    if value is None:
+        return ""
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, datetime):
+        return value.isoformat()
+    return _escape(str(value))
```
DEENUU1: One real thing here. `_cell` runs `_escape` over *every* value, numbers included, and `-` is in `_CSV_INJECTION_PREFIXES`. So the moment a `cost_usd` (or any number) is negative, `_escape("-1.50")` returns `'-1.50` — which a spreadsheet reads as text, not a number, and silently leaves out of a `SUM`. That's exactly the sum-safety this whole export is built to protect, defeated by the guard that's meant to protect it.

Today every cost and token count is non-negative so it never fires, but it's latent: the first credit, refund or correction row that lands negative exports as un-summable text, with nothing on screen to say so. A formula can only ever start a *string* cell (a tool arg, an agent name, a note) — a `Decimal`, an `int`, a `UUID` never can — so I'd scope the escape to strings and leave the rest alone:

```suggestion
def _cell(value: object) -> str:
    """One value as text, with `None` an empty cell rather than the word "None"."""
    if value is None:
        return ""
    if isinstance(value, bool):
        return "true" if value else "false"
    if isinstance(value, datetime):
        return value.isoformat()
    if isinstance(value, str):
        return _escape(value)
    return str(value)
```

`_json_args` still hands you a `str`, but a JSON object opens on `{`, so it stays escape-safe. Might be worth a test with a negative `cost_usd` so this can't quietly come back.

backend/app/services/run_export.py
```diff
@@ -0,0 +1,499 @@
...
+        """
+        since, until = self._require_range(since, until)
+        floor = self._own_floor(ctx)
+
+        rows = await agent_run_repo.spend_by_agent(
+            self.db,
+            organization_id=ctx.organization_id,
+            since=since,
+            until=until,
+            month_since=month_start(),
```
DEENUU1, line 382: A semantics call worth making deliberately: the spend export takes a mandatory window (`since`/`until`), but `month_to_date_usd` and `monthly_cap_usd` come from `month_start()` — the current calendar month, whatever the window is. Through the UI `to` is always "now", so it mostly lines up, but a "Last 90 days" export still puts a 90-day `cost_usd` next to a this-month MTD in the very next column.

In a downloaded CSV, two dollar columns with different time bases side by side is the same class of footgun the `cost_is_partial` column exists to stop — a reader sums across them by accident. Two clean options: drop the two current-month columns from a windowed export (they're a live-dashboard idea, not a windowed-report one), or keep them and say in the docs that those two are always current-month, never the window. I'd lean toward dropping them.

backend/app/services/run_export.py
```diff
@@ -0,0 +1,499 @@
...
+            user_id=floor,
+        )
+        self._guard_cap(len(rows))
```
DEENUU1: nit: the shared `_guard_cap` message says "Narrow the date range and try again", which is exactly right for runs and approvals but not for spend — here a row is an agent, and narrowing the window won't change how many agents exist. It's effectively unreachable (you'd need 10k+ agents in one org to hit it), so not worth much, but if you ever want it accurate, spend could carry its own message or the copy could be range-agnostic.

backend/app/repositories/agent_run.py
```diff
@@ -587,6 +594,7 @@ async def spend_by_agent(
         .outerjoin(AgentVersion, AgentVersion.id == Agent.current_version_id)
         .where(
             Agent.organization_id == organization_id,
+            *([] if user_id is None else [AgentRun.user_id == user_id]),
```
DEENUU1, line 597: Nice — pushing the `Scope.OWN` floor into the `WHERE` here rather than filtering after the aggregate is the right call, and it's why both windows and the counts describe the same person's rows instead of disagreeing. The comment right below about `and_(*window)` vs spreading into `or_` is the kind of thing that saves the next reader an afternoon.

DEENUU1 (commented):
Really solid piece of work. The six design questions from #211 are answered in the file itself rather than hand-waved, and the tests actually cover the refusals that matter here — tenancy including a row the caller owns in another org, the `Scope.OWN` floor as a `WHERE` and not a post-filter, over-cap refused not truncated, and a wholly-unpriced run marked instead of read as free. Path ordering (`/runs/export` before `/runs/{run_id}`) and the org header on `apiClient.raw` are both easy to get wrong and both right here.

Two things I'd change before calling it done, plus a nit:

- The CSV-injection guard runs over every cell, numbers included, and `-` is a neutralised prefix — so any numeric that ever goes negative exports as un-summable text. That defeats the sum-safety the whole export exists to protect. Latent today, but cheap to fix now.
- The spend export carries two current-month columns (`month_to_date_usd`, `monthly_cap_usd`) next to a windowed `cost_usd`. Worth a decision.

Details inline.

backend/app/api/routes/v1/runs.py
```diff
@@ -140,6 +141,72 @@ def _parse_statuses(raw: str | None) -> list[str] | None:
     return values or None
 
 
+def _csv_response(result: ExportResult) -> Response:
```
DEENUU1: One more, and this one I do want changed. This file should hold route handlers and nothing else. `_csv_response` isn't a route — it handles no request, it just turns an `ExportResult` into a `Response` — so it doesn't belong in the router module. Move it out into a small shared helper (e.g. `app/api/responses.py`) and import it here: that keeps the DRY win across the three export routes and leaves `runs.py` as only routes.

The rest of the repo actually builds the `Response` inline inside the handler instead — see `admin_ratings.py` and the YAML export in `agents.py:290` — so if you'd rather not add a module, inlining also satisfies "only routes here". With three call sites I'd lean toward the shared helper.

(`_parse_statuses` above is the same shape and is pre-existing, so the file has already drifted once — this is the moment to not add to it, not a licence.)

Issue #520, opened 2026-08-09
Research and cut the test suite's runtime — it is too slow to run locally


DEENUU1:
## What

The test suites take long enough that the loop actively discourages running
them locally — `make test`, `make test-integration` and `make test-frontend-cov`
all cost enough wall-clock that the documented habit is to lean on CI instead.
Research where that time goes and cut it.

Concrete starting points:

- **Profile before touching anything.** Where does the wall-clock actually go in
  `make test`, `make test-integration`, `make test-frontend-cov`? `pytest
  --durations=25`, vitest's reporter, and the coverage-instrumentation overhead
  (coverage is known to slow specs enough to trip a 5s `testTimeout`).
- **The integration suite rebuilds the whole schema before every test** — ~92s,
  nearly all of it there (#215). This is the single biggest known lever.
- Fixture setup, DB reuse across tests, transaction rollback instead of drop/create.
- Parallelism: `pytest-xdist` for the backend, vitest sharding for the frontend.
- Whether the 100% coverage gate and its instrumentation can run once at the end
  rather than on every scoped run.

## Why

A slow suite is a suite people don't run, which is exactly the failure the loop
warns about — the fast net becomes the only net, and a scoped answer that should
take one second takes ninety. Cutting the runtime makes local verification cheap
enough to be the default again, and shrinks the CI bill alongside it.

## Done when

- A written finding (in the issue or a short note) of where the time goes, per suite.
- A measurable, committed reduction — the integration suite's schema rebuild is the
  first target.

## Related

Part of the test-speed / CI-cost cluster: **#215** (schema rebuilt before every
integration test), **#317** (CI burns ~8,900 billed Actions minutes a week).
Adjacent doc drift: **#212**. See the issue map **#168**.

DEENUU1:
### First lever landed — the integration schema rebuild (#215)

PR #535 builds the integration schema **once per process** and resets the data
between tests with `TRUNCATE`, instead of `drop_all` + `create_all` before every test.
Measured on `pgvector/pgvector:pg16`, warm:

- `pytest tests/integration --no-cov`: **~125s → ~53s** (ordered and random-order alike).
- full `make test` (unit + integration + 100% gate): **~175s → ~103s**.
- Still passes in random order and twice concurrently — the per-process `_p<pid>`
  database from #189 is untouched.

Closes #215.

### The finding — where the time goes, per suite

**Backend `make test` — the CI long pole (~103s now).**
[...]

PR #535, opened 2026-08-10
perf(tests): build the integration schema once, not before every test

### LINKED ISSUE #215: The integration suite rebuilds the whole schema before every test, which is nearly all of its 92s
**Pre-existing, found while fixing #189.** Not caused by that branch — it is what made
#189 expensive as well as wrong, since the thing two runs fought over was 216 rounds of
DDL.

## What it costs

`backend/tests/integration/conftest.py` builds the schema from the models in a
function-scoped fixture: `drop_all` + `create_all` before **every** test. Measured on
this machine (pgvector/pgvector:pg16 in Docker, warm):

- `setup` for a single integration test: **0.91s**, essentially all of it that DDL;
- the whole suite: **216 tests in 92s** — about 0.43s per test, so the schema rebuild
  is very nearly the entire runtime. The assertions themselves are microseconds of
  Postgres work.

That is 90 seconds of every `make test`, and the reason `make test-integration` is the
slowest thing a person waits for locally.

## Why it was written that way

Complete isolation between tests, cheaply expressed: nothing has to clean up after
itself, and no test can see another's rows. That was the right trade when the fixture
was written, and the docstring says so.

What changed is #189: each pytest process now owns its database
(`<POSTGRES_DB>_p<pid>`, created and dropped by the session), so "the schema is fresh"
no longer has to be re-established against a database somebody else might be using.

## The shape of the fix

Build the schema **once per session** and reset the data between tests. Two candidates,
both need the numbers checked rather than assumed:

- `TRUNCATE … RESTART IDENTITY CASCADE` over `Base.metadata.sorted_tables` after each
  test — one statement, no DDL, and it handles the tests that commit through
  `get_db_session` (which the API-flow tests in `test_platform_flows.py` do, so a
  rollback-only strategy is not enough on its own).
- A per-test outer transaction the fixture rolls back, with the session joined to it.
  Cheaper still, but every path that commits has to be understood first — the
  `_managed_session` override in `test_platform_flows.py` commits by design.

Either way the session-scoped schema needs the same care the current fixture documents:
anyio's backend fixture is per-function, so the schema build cannot live in a
session-scoped *async* fixture. `database_url` (session-scoped, synchronous,
`asyncio.run`) is the pattern to follow.

- `uv run pytest tests/integration -q --no-cov` is a small multiple of the Postgres work
  rather than a multiple of the DDL — target under 20s for the 216 tests.
- The suite still passes in a random order (`pytest-randomly` is on by default here), and
  still passes twice concurrently — the property #189 bought.
- No test depends on another's rows: `-p no:randomly` and the default random order agree.

## Where it sits

Cluster: **Tests, CI, dependencies** (#109 #130 #132 #143 #154 #155 #156 #158 #162 #165
#188 #203). Nearest: **#189**, whose fix this builds on and which is the reason a
session-scoped schema is now safe at all. Not a duplicate — #189 is about two runs
corrupting each other, this is about what one run spends its time on. Nothing here
blocks anything; it is a straight speed-up of the slowest local command.

DEENUU1:
## What this fixes

The integration suite rebuilt the whole schema from the models before **every**
test — `drop_all` + `create_all` in a function-scoped fixture, ~0.4s of DDL each.
On a suite whose assertions are microseconds of Postgres work, that DDL was very
nearly the entire runtime. This builds the schema **once per process** and resets
the *data* between tests instead.

`Closes #215`. `Refs #520` (the first, and biggest, committed lever of the
test-speed work — the finding for all three suites is below).

## What changed

- `backend/tests/integration/conftest.py`
  - New session-scoped, synchronous `schema_url` fixture — `create_all` once, via
    `asyncio.run` (the pattern `database_url` already uses; a session-scoped *async*
    fixture cannot depend on anyio's per-function backend).
  - `engine` (function-scoped) now calls `_reset`: `TRUNCATE … RESTART IDENTITY
    CASCADE` over every model table, plus a `DROP` of any table a test created
    outside the models (a runtime `rag_<collection>`, an ordering probe). No DDL for
    the schema itself, ever, after the first build.
- `test_commit_before_response.py`, `test_flow_starts_after_commit.py` — docstrings
  on the two ordering-probe fixtures updated (they referenced the old `drop_all`).
- `backend/tests/integration/conftest.py` — the module docstring and
  `_refuse_a_real_database` docstring reworded off the old per-test `drop_all`
  onto the `TRUNCATE`/schema-once reset (review follow-up, 6db994b8).
- `docs/testing.md`, `.claude/rules/testing.md` — the fixture description updated to
  the build-once mechanism.

`TRUNCATE`, not a transaction rollback: the API-flow tests in `test_platform_flows.py`
and the dispatch-ordering tests commit through the real `get_db_session` by design, so
their rows outlive a rollback — rollback-only isolation would leak between tests.

## How it failed before

Not a failure, a cost. `pytest tests/integration --no-cov`: **400 tests in ~125s**,
almost all of it in the `setup` phase (~0.3–0.5s each), which is the schema DDL.

## Testing

Measured on `pgvector/pgvector:pg16`, warm, this machine:

| | Before | After |
|---|---|---|
| `pytest tests/integration --no-cov` (ordered) | ~125s | **~53s** |
| same, verified order-independent (two `--randomly-seed`s, #571) | — | ~57s |
| full `make test` (unit + integration + 100% gate) | ~175s | **~103s** |

- 400 integration tests pass ordered (`-p no:randomly`) **and** in random order.
- Two runs concurrently: both green (154/154 on an overlapping subset) — the
  per-process `_p<pid>` database from #189 is untouched, so the property it bought
  survives.
- `ruff`, `ruff format`, `ty` clean on the changed files; pre-commit green;
  `scripts/docs_drift.py` clean.

## The #520 finding — where the time goes, per suite

Numbers from this machine, same container.

**Backend `make test` — the CI long pole (~103s after this PR).**
- *Integration slice* — was the whole story: ~125s, essentially all schema DDL. **This
  PR: ~53s.** The remaining per-test cost is `create_async_engine` + one `pg_tables`
  scan + one `TRUNCATE` + dispose — bounded by anyio giving each test its own event
  loop, so the engine cannot be shared session-wide (the "Future attached to a
  different loop" trap the probe fixtures already document).
- *Unit slice* — ~32–44s `--no-cov`, no single hotspot; it is app-import + per-test
  fixtures across ~3,900 tests. Runs **single-process** — `pytest-xdist` is the next
  lever and unclaimed.
- *Coverage instrumentation* — real but modest, ~15–20% on top. Not the dominant cost;
  the dominant cost was the DDL, now gone.

**Frontend `make test-frontend-cov` — ~53s, not the bottleneck.** `test:run` ~45s,
`test:coverage` ~53s (coverage ~+18%). vitest already runs fully parallel across cores
(~850% CPU). Sharding would only help CI split across machines, not a local run.

## Noticed, not fixed

Filed as follow-ups under #520, not folded into this diff:

- **`pytest-xdist` for the backend unit slice** — the largest remaining local lever,
  and it interacts with the per-process `_p<pid>` DB naming (each xdist worker needs its
  own database). Real design work, its own PR.
- **Run the coverage gate once at the end**, not on every scoped run — the issue's
  "coverage instrumentation can run once" idea. Tooling change, separate.
- **`tests/test_approval_expiry.py::TestTheScheduledSweep::test_the_flow_answers_with_what_it_expired`
  fails locally without a live Prefect API at `localhost:4200`** — pre-existing (fails
  identically on `main`), environmental, unrelated to this change. Directly relevant to
  #520's "too hard to run locally" theme, so worth a guard/skip; filing separately.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -117,8 +117,8 @@ name meant two runs on one machine dropping each other's tables mid-test (#189).
 drops it at the end, even when the suite fails, so **two concurrent runs are safe and
 nothing has to be passed to make them so**. It still skips when no database is
 reachable (a laptop without Docker) and still refuses any database whose name contains
-neither `test` nor `ci`, or that is not a plain identifier — it calls `drop_all`
-unconditionally and drops the database itself afterwards.
+neither `test` nor `ci`, or that is not a plain identifier — it empties every table
+unconditionally between tests and drops the database itself afterwards.
 
 A run killed outright (`SIGKILL`) leaks its database; the next run with that pid drops
 it before creating its own. Anything else named `agenticos_*` on a shared Postgres was
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -6,12 +6,16 @@ unique index actually prevents a second default. Those are the guarantees the
 schema is supposed to provide, and the only way to know is to ask Postgres.
 
 The database belongs to this pytest process alone: it is created here and
-dropped again when the session ends. `drop_all` below is unconditional, so a
-shared name meant two runs on one machine demolishing each other's tables
-mid-test - deadlocks between one run's `DROP TABLE` and another's `CREATE
-INDEX`, a column that "does not exist" on a database that had it, and two runs
-of the same commit reporting different failures (#189). `tests/conftest.py`
-derives the per-process name; this module owns its lifecycle.
+dropped again when the session ends. The per-process name is what keeps two
+runs on one machine off each other's tables - without it a shared name once had
+them demolishing each other mid-test, back when each test rebuilt the schema
+with `drop_all` + `create_all` and the collision was a deadlock between one
+run's `DROP TABLE` and another's `CREATE INDEX`, a column that "does not exist"
+on a database that had it, and two runs of the same commit reporting different
+failures (#189). The schema is built once now and the reset between tests is
+`TRUNCATE` (#215), so that DDL deadlock can no longer happen - but the
+per-process name still earns its place. `tests/conftest.py` derives it; this
+module owns its lifecycle.
 
 The whole module is skipped when no database is reachable, so `make test` on a
 laptop without Docker still runs everything else.
@@ -99,12 +103,12 @@ _PLAIN_IDENTIFIER = re.compile(r"[A-Za-z0-9_]+")
 def _refuse_a_real_database(name: str) -> None:
     """Refuse to run against a database that is not obviously a test one.
 
-    `drop_all` here is unconditional and this module drops the database itself
-    afterwards, so pointing the suite at a development database destroys it.
-    Nothing in the fixture can tell the difference once it has happened; the only
-    moment it can be caught is before the first drop. The per-process suffix is
-    appended before this runs, so what is checked is the name that will actually
-    be dropped.
+    The reset between tests `TRUNCATE`s every table unconditionally and this
+    module drops the database itself afterwards, so pointing the suite at a
+    development database destroys it. Nothing in the fixture can tell the
+    difference once it has happened; the only moment it can be caught is before
+    the first of those runs. The per-process suffix is appended before this runs,
+    so what is checked is the name that will actually be dropped.
     """
     if not _PLAIN_IDENTIFIER.fullmatch(name):
         raise RuntimeError(
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -25,6 +25,7 @@ import re
 from collections.abc import AsyncGenerator, Iterator
 
 import pytest
+from sqlalchemy import text
 from sqlalchemy.ext.asyncio import (
     AsyncEngine,
     AsyncSession,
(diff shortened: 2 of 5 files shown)


DEENUU1:
Correction on the testing claim above: I wrote that this passes "in random order (pytest-randomly on)" — that was wrong. `pytest-randomly` turns out not to be installed here despite the docs saying it is (filed as #571), so both my ordered and "random" runs were the same collection order. The order-independence still holds — I've since confirmed the full suite green across two `--randomly-seed` values and under pytest-xdist's shuffled load distribution (#570) — but it was verified after the fact, not by the run I originally cited.

DEENUU1 (commented):
Nice change. I traced the two things this design rests on and both hold: no integration test mutates the *model* schema (the only `CREATE TABLE`s are the two ordering probes and the raw `rag_company_handbook`), so never rebuilding the schema can't leak drift between tests; and no model sets an explicit `schema=`, so `set(Base.metadata.tables)` keys are bare names that match `pg_tables.tablename` and the model-vs-runtime check in `_reset` is right.

It's also quietly a correctness win, not just a speed one. `test_vector_store_listing.py` creates `rag_company_handbook` and never drops it, and the old `drop_all` never touched non-model tables — so that table (and any leaked runtime `rag_<collection>`) used to survive into later tests. `_reset` dropping non-model tables at the start of every test makes that test's `present == modelled | {...}` assertion order-independent instead of quietly fragile. The injection-safety comment and the TRUNCATE-not-rollback rationale are both exactly right.

One thing to fix before merge, in the same spirit as the doc updates you already made: two `drop_all` references got left behind in `conftest.py`'s own docstrings, and the file no longer calls `drop_all`.

- `backend/tests/integration/conftest.py:9` (module docstring): "`drop_all` below is unconditional, so a shared name meant two runs … demolishing each other's tables mid-test — deadlocks between one run's `DROP TABLE` and another's `CREATE INDEX`…". The reset is `TRUNCATE` now, so the present-tense `drop_all` and the DDL-deadlock framing describe a mechanism this file no longer has. Since the #189 story it tells is now history rather than current behaviour, worth rewording to "emptied every table between tests" (and the deadlock was DDL-vs-DDL; two `TRUNCATE`s wouldn't reproduce it).
- `backend/tests/integration/conftest.py:101` (`_refuse_a_real_database` docstring): "`drop_all` here is unconditional and this module drops the database itself afterwards…". Still correctly destructive, but it's `TRUNCATE` that's unconditional now — `drop_all` reads as stale next to the code. The error message just below it (line 115, "drops every table before each test") is already generic, so only the docstring needs the touch.

Both are outside the diff so I couldn't hang them inline. Not blocking — CI's green and the mechanism is sound — but the whole point of this PR is that one description stays true to the code, and these two are the only spots still describing the old one.

DEENUU1:
Both stale docstrings fixed in 6db994b8. The module docstring now tells the #189 story as history — the per-process `_p<pid>` name stays the isolation between two runs, but the DDL deadlock it once prevented can't recur now that the schema is built once and the reset is `TRUNCATE` (#215). And `_refuse_a_real_database` now says it's the `TRUNCATE` that's unconditional, not `drop_all`. Docstrings only; the error message at line 115 was already generic, left as-is.

PR #538, opened 2026-08-10
feat(runs): filter run history by rating and flag it on the run

### LINKED ISSUE #209: Surface message ratings below the app admin, and on the run that earned them
## What this is for

`message_ratings` stores a 👍/👎 **and an optional free-text comment** per assistant message (`backend/app/db/models/message_rating.py`), and nothing below the app-admin level surfaces any of it. `admin_ratings.py` is the only consumer: a deployment-wide list, summary and export.

So the platform holds the highest-signal debugging queue it will ever have — the answers real people said were wrong, in their own words — and no operator, builder or steward can see it.

Two halves, and the second is the point:

1. **An organization-level summary.** `GET /api/v1/ratings/summary` with a `scope` parameter (`org` under `runs:view`, `own` for the caller's own conversations), returning the headline percentage **plus a per-day series** so the card shows how quality *moves* rather than where it stands. This is the card #37 / PR #149 proposes and flags as the one widget answering "is it any good"; its stage-1 note leaves it as an explicit review call, and that call is now yes.
2. **The rating on the run.** A 👎 rendered on the row in Activity's run history, with the comment in the run detail view, and a "rated down" filter over the list. This is the part that makes the percentage actionable: the dashboard says quality fell four points, and this is where you read the eleven conversations that did it.

Half 2 is the reason to build half 1 rather than a nicer number on a card.

## Depends on

**`messages.run_id`** (#45, PR #202 §2). A rating hangs off a message; a run has no link to its messages today, so "was this run rated down" is not answerable until that column exists and is populated. Everything else here is a join.

The repository layer is largely in place: `get_rating_counts_for_messages` and `get_ratings_with_users_for_messages` (`backend/app/repositories/message_rating.py`) already batch exactly the shape a run detail view needs.

## What to build

- [ ] `GET /api/v1/ratings/summary?scope=org|own&from=&to=` — same time parameters as `/stats/usage` (#37), so the two pages narrow by one vocabulary. `scope=org` requires `runs:view`; `scope=own` requires only a signed-in membership and reads the caller's own conversations.
- [ ] Headline percentage **and** a per-day series in the response, so the card can plot how quality moves rather than only where it stands.
- [ ] `rated=down` (or equivalent) as a filter parameter on `GET /runs`.
- [ ] A 👎 marker on the row in Activity's run history.
- [ ] The rating and its comment in the run detail view.
- [ ] The org answer-quality card on the dashboard (#37), and "Answer quality in your chats" for a member at `scope=own`.
- [ ] Every string through `next-intl`.

## Guardrails

- **A comment is user-written text about a specific conversation.** `scope=org` shows it to anyone with `runs:view`, which is a wider audience than the person who wrote it may expect. Decide deliberately whether the comment text is org-visible or whether the org card is counts-only with the text behind the run detail; do not let it happen by default.
- **Tenant isolation on both scopes**, tested — including the case where the caller owns the conversation.
- **Empty is not failed.** A collection with no ratings and a request that 502'd must not render the same pixels (`.claude/rules/frontend.md`, live defect #32).
- A rating is nullable in every direction: a run with no rated message, a message rated by nobody, a rating whose user was deleted.

- [ ] A run whose assistant message was rated 👎 shows it in run history, and the comment is readable in the detail view.
- [ ] Filtering run history to "rated down" returns exactly those runs, proven against a real database.
- [ ] The org quality card moves when a rating is added.
- [ ] A member at `scope=own` sees only their own conversations — proven by a refusal test, not by the happy path.
- [ ] A caller without `runs:view` gets no org card at all: absent, not disabled, not a 403.

## Where it sits

Cluster: **Frontend surfaces** (#37 #45) and **Agents, spec, capabilities** for the quality question. Blocked by #45 (`messages.run_id`). Enables the answer-quality card #37 flags as undecided in its stage-1 note — that decision is now yes, which is why this exists as work rather than a question. Came out of reviewing PR #202 alongside PR #149: the dashboard has the number and Activity has the rows, and nothing joined them. Nearest duplicate: none — nothing among the 137 open issues mentions ratings.

OchnikBartek:
## Summary

#209 has two halves, and the second is the point. Half 1 — the organization
answer-quality card, its `scope=own` sibling, and the `GET /ratings/summary`
endpoint behind them — already shipped with #37 / #149; the card is registered,
gated on `runs:view`, and tested. This branch delivers **half 2**, the run-side
that makes that percentage actionable: a rating filter on run history, a 👎 marker
on the row, and the rating with its comment in the run detail. The dashboard says
quality moved; this is where you read the conversations that moved it.

## Related Issue

Closes #209.

## Added

- `RunFilters.rated` (`RunRating.UP` / `RunRating.DOWN`) on `GET /runs`, and
  `down_rated_run_ids` in `agent_run` — one query per page for the marker, drawn
  from top-level runs, bounded to the caller's organization, `distinct` so several
  thumbs-down on one run are one id.
- `run-feedback` on the run detail (the rating and its free-text comment), and the
  👎 marker on the run-history row.
- Real-database coverage in `tests/integration/test_run_history_filters.py`:
  `rated=down` returns exactly the disliked runs, several dislikes collapse to one
  row, a divisive run appears under both filters, and a neighbouring organization's
  down-rated run is never marked even when asked with the wrong organization id.

## Changed

- `GET /runs` route + `RunFilters`; `AgentRunnerService.down_rated_run_ids` (adds
  only the tenant bound over the repository query); `use-runs` hook and
  `qk.runs` keys; run-history tab and run table (the marker); `types/runs`;
  `frontend/messages/en.json`; `docs/governance.md`.

## Checks

## Notes for Reviewers

- **Half 1 is deliberately not rebuilt here.** The org answer-quality card
  (`widgets/org-ratings.tsx`, gated on `runs:view` in the dashboard registry), its
  `scope=own` sibling (`widgets/my-quality.tsx`), and the `/ratings/summary`
  endpoint already exist on `main` from #37 / #149. Adding them again would be
  duplicate code; this branch completes the issue's remaining half.
- **Comment-visibility guardrail (issue).** The org card is counts-only — the
  headline percentage and the per-day trend — and the free-text comment lives
  behind the run detail, not on the card. That is the deliberate choice the issue
  asked us to make rather than let happen by default.
- **Rebased onto current `main` (`eb99799`).** One conflict, in
  `tests/test_agent_runner.py`: #513 (Mattermost) and this branch each append an
  independent test class at end-of-file (`TestWhatTheChannelLetsTheAgentLookUp`
  and `TestTheDownRatedMarker`). Resolved by keeping both.
- **Verified locally:** `uv run pytest tests/test_agent_runner.py` — 78 passed
  (the file the conflict touched). The run-side suites this branch adds were run
  green when committed; I did not re-run the full suite for this PR, so `make check`
  is left unticked above — CI runs it.

- **Self-contained — no merge-order dependency (updated).** Review flagged that
  the feedback panel calls `GET /runs/{run_id}/transcript` (from #525) and reads
  `rating_comment` / `user_rating` / `rating_count`, none of which the transcript
  route actually returned — #525 merged exposing a plain `MessageRead`. #525 is now
  on `main` and merged in here, and the missing backend has been folded into this
  branch: `RunTranscriptMessage` carries the three fields, populated by
  `AgentRunnerService.transcript_ratings` (batched, one query each) sourced from
  `message_ratings`. The comment shown is the most recent down rating's. So the
  feature is whole in one merge; there is no longer any "#525 must merge first"
  ordering to honour. `docs/governance.md` is updated to describe it.


frontend/src/components/runs/focused-run.tsx
```diff
@@ -67,6 +68,10 @@ export function FocusedRun({ runId }: { runId: string }) {
           {t("delegationsAlreadyCounted", { count: delegated.length })}
         </p>
       )}
+
+      {/* Where the 👎 on the row above becomes readable: the answers people
+          rated down in this run, and the words they left. */}
+      <RunFeedback runId={runId} />
```
DEENUU1, line 74: This is the one I'd block on. `RunFeedback` calls `GET /runs/{run_id}/transcript`, and that route doesn't exist on this branch or on `main` — it's added by PR #525 (`feat/run-transcript-route`), which is still open. So if #538 merges before #525, this panel renders on every run detail and always fails: the request 404s, `useRunTranscript` returns an `error`, and the reader gets "The feedback could not be read" for a run that's perfectly fine.

The tests don't catch it because they mock `apiClient.get` at the boundary, and e2e never opens the panel. That's the "green CI isn't correctness" trap — the code ran, against a contract nobody serves yet.

Two honest ways out: land #525 first and rebase this on top (and add a line to the PR checklist saying it's blocked on #525), or fold the transcript route into this branch so the feature is whole in one merge. Either is fine — but merging this alone ships a broken panel.

frontend/src/components/runs/run-feedback.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <div className="space-y-3">
+      <h3 className="flex items-center gap-1.5 text-sm font-medium">
+        <MessageSquareText className="h-4 w-4" />
+        {t("whatPeopleSaidWasWrong")}
+      </h3>
+      <ul className="space-y-3">
+        {ratedDown.map((message) => (
+          <li key={message.id} className="border-destructive/40 rounded-md border-l-2 pl-3">
+            {message.rating_comment ? (
```
DEENUU1, line 63: Second half of the same problem, and this one survives even after #525 lands. This renders `message.rating_comment`, but #525's `MessageRead` (what the transcript route returns) only exposes `user_rating` and `rating_count` — there's no `rating_comment` on it, and there isn't one on `main` either. The `message_ratings` table *does* have a `comment` column (`CheckConstraint("rating IN (1, -1)")`, `comment: Mapped[str | None]`), it just never gets mapped into the read schema.

So in practice every down-rated turn falls through to the `else` and shows "Rated down, with no comment.", forever. That's the whole point of the run-detail view per #209 ("the comment is readable in the detail view") not working — and `docs/governance.md` in this same PR asserts it *is* readable, so the docs would be wrong too.

The fix belongs on the backend side: `MessageRead` (or whatever the transcript route serializes) needs a `rating_comment` field sourced from `MessageRating.comment`, batched the way `get_ratings_with_users_for_messages` already does. Worth coordinating with #525 so the two branches agree on the shape rather than each guessing — right now the frontend type declares a field the API never sends.

backend/tests/integration/test_run_history_filters.py
```diff
@@ -477,6 +477,84 @@ async def test_a_rating_on_another_runs_message_does_not_pull_this_one_in(self,
...
+        agent = await _agent(db, org)
+        disliked = await _run(db, org, agent)
+        for _ in range(3):
+            await self._rated(db, disliked, by=await _user(db), rating=-1)
+
+        marked = await down_rated_run_ids(db, organization_id=org.id, run_ids=[disliked.id])
+
+        assert marked == {disliked.id}
+
+    async def test_a_neighbours_down_rated_run_is_never_marked(self, db) -> None:
```
DEENUU1, line 537: Nice — this is exactly the refusal test I'd want: a real run, really rated down, asked about with the wrong org id, and the marker still empty. Testing the tenant bound where it actually bites (a borrowed id at the one query that takes ids from the caller) rather than just the happy path. This is the kind of coverage that makes the marker safe to trust.

DEENUU1 (dismissed):
This completes the run-side of #209 and the backend marker is genuinely nice work: `down_rated_run_ids` is tenant-bound, `distinct`, and provably the same "rated down" definition (`rating < 0`) that the `rated=down` filter uses, so a marked row is exactly a row the filter returns. The integration tests prove it where it bites (the neighbour-tenant case), the empty-vs-failed handling is careful, the control is permission-gated, and the i18n is done right. Good stuff.

But the run-detail half, which the issue calls "the point", can't work as it stands, and both reasons are about the boundary with PR #525. I want to hold on merging until they're sorted.

1. `RunFeedback` reads `GET /runs/{run_id}/transcript`, which exists only on #525 (open, unmerged). If this branch merges first, every run detail renders the error state. So there's a hard merge-order dependency that isn't called out in the checklist.
2. Even after #525 lands, `rating_comment` is never in the response — #525's `MessageRead` carries `user_rating` and `rating_count` but not the comment (the model has the column, the schema just doesn't map it). So the panel always shows "Rated down, with no comment", and reading the comment is the acceptance criterion. The frontend suite is green because it mocks a transcript that includes `rating_comment` — a contract the backend doesn't actually provide.

Details inline. Neither is hard to fix, mostly a question of what lands where.

PR #541, opened 2026-08-10
feat(governance): reconcile per-person spend with an unattributed bucket

### LINKED ISSUE #539: Per-person spend card under-counts vs By agent, with nothing explaining the gap
## What breaks

On Activity → Spend, the **By person** card (`spend-by-person.tsx`, reading
`/stats/usage?group_by=user`) and the **By agent** card above it (reading `/spend`)
do not always sum to the same total — the exact failure #214 called out: "a
breakdown that disagrees with the number above it is worse than no breakdown."

Two independent causes:

1. **Different window bases.** By agent ends at a precise `now` (`/spend`); By
   person runs over `from`/`to` truncated to whole UTC dates (`/stats/usage`). The
   edges can differ by up to a day.
2. **Structural under-count.** `usage_by_user` inner-joins `User`, so any run whose
   user was deleted (the FK SET-NULLs) or a channel run with no account behind it is
   dropped from the By-person card entirely, while By agent still counts it. This is
   #214's "a deleted user's rows are kept and muted" acceptance criterion — not
   merely deferred on the frontend, but **structurally impossible on the current
   `group_by=user` endpoint**.

A steward comparing the two cards sees this one under-count with nothing on screen
explaining the gap.

## Where

- `frontend/src/components/runs/spend-by-person.tsx`
- `backend/app/services/stats.py` — `usage_by_user`
- `backend/app/repositories/agent_run.py`

- An org with a deleted user's priced runs shows those runs as a muted /
  "no longer in org" row (or an explicit unattributed bucket) rather than dropping
  them, so the By-person total reconciles with By agent — proven against a real
  database, including the deleted-user case.
- Or, if the gap is kept deliberately, the card carries an on-screen note naming it,
  so the two cards can't be silently summed against each other.

## Where it sits

Found reviewing PR #527 (which closes #214). Part of the per-person usage /
dashboard cluster — #214, and the #37 / #149 dashboard cards. The frontend "+N
others" disclosure added in #527 covers *truncation*, not this *under-count*; they
are different gaps. Nearest related open issue: #214 (its parent).

OchnikBartek:
## Summary

The Spend tab's **By person** card could under-count against the **By agent** card above it — the disagreement #214 warned against ("a breakdown that disagrees with the number above it is worse than no breakdown"). `usage_by_user` inner-joins `User`, so a priced run whose user was deleted (the FK SET-NULLs) or an account-less channel run was dropped from By person while By agent still counted it. This gathers those runs into an unattributed bucket so the two cards reconcile.

## Related Issue

Closes #539.

## Added

- `unattributed_usage` (`app/repositories/agent_run.py`) — the window's `user_id IS NULL` runs as one bucket (runs, cost, last run), top-level only. `UnattributedUsage` schema plus an `unattributed_usage` field on `UsageStats`.
- A muted bucket row ("Deleted users and anonymous runs") with its cost on `spend-by-person.tsx`, and a note naming the window basis.

## Changed

- `StatsService.usage_by_user` composes the bucket for `scope=org` only — a caller's own rows are never unattributed. `PersonUsageRow`'s docstring updated: null-user runs are gathered into the bucket rather than silently dropped, and `by_user`'s named rows are left untouched, so they stay consistent with `active_users`.
- `usePeopleUsage` exposes `unattributed`.

## Checks

## Notes for Reviewers

- **Stacked on #527** (`feat/per-person-usage`): this PR is based on that branch, so the diff shows only the reconciliation change and not #527's card. Once #527 merges, GitHub auto-retargets this to `main`, and `Closes #539` fires when it lands there.
- **The window is not aligned, by design.** By agent (`/spend`) ends at a precise `now`; By person (`/stats/usage`) truncates to whole UTC days — aligning would move every dashboard card. The bucket closes the structural gap; the card's window note names the residual edge difference so the two are never silently summed.
- **Verified against a real database:** a deleted user's priced runs and account-less runs land in the bucket, and the named rows plus the bucket sum to the by-agent totals (`test_usage_stats_sql.py`, 31 passed against pgvector). Plus service unit tests (org present/absent, own skips the query), frontend bucket present/absent, and `ty` / `ruff` / `check_i18n` / `catalog` / `tsc` all clean. `make check` left unticked above — CI is the wide net.

### DIFF
--- a/backend/tests/integration/test_usage_stats_sql.py
+++ b/backend/tests/integration/test_usage_stats_sql.py
@@ -801,6 +801,79 @@ class TestPersonRows:
         assert [(row.user_id, row.runs) for row in result.by_user] == [(owner.id, 1)]
         assert result.total_runs is None
 
+    async def test_the_unattributed_bucket_reconciles_by_person_with_by_agent(self, db) -> None:
+        """A deleted user's priced runs, and account-less ones, are gathered
+        into the bucket rather than dropped, so the per-person total lines up
+        with the by-agent total, which still counts them under their agent."""
+        organization, owner = await _org_with_owner(db, "Reconcile")
+        agent = await _agent(db, organization, owner)
+        named = await _user(db)
+        leaver = await _user(db)
+        await _run(
+            db,
+            organization=organization,
+            agent=agent,
+            started_at=START,
+            user_id=named.id,
+            cost=Decimal("2.00"),
+        )
+        await _run(
+            db,
+            organization=organization,
+            agent=agent,
+            started_at=START + timedelta(days=1),
+            user_id=leaver.id,
+            cost=Decimal("3.00"),
+        )
+        # A channel or widget run that never had an account behind it.
+        await _run(
+            db,
+            organization=organization,
+            agent=agent,
+            started_at=START,
+            user_id=None,
+            cost=Decimal("0.50"),
+        )
+        # The user leaves the organization: ondelete=SET NULL nulls their run.
+        await db.delete(leaver)
+        await db.flush()
+
+        named_rows = await agent_run_repo.usage_by_user(
+            db, organization_id=organization.id, start=START, end=END, limit=10
+        )
+        bucket = await agent_run_repo.unattributed_usage(
+            db, organization_id=organization.id, start=START, end=END
+        )
+        total_runs = await agent_run_repo.count_runs(
+            db, organization_id=organization.id, start=START, end=END
+        )
+        total_cost = await agent_run_repo.sum_cost_window(
+            db, organization_id=organization.id, start=START, end=END
+        )
+
+        # The leaver has dropped out of the named rows; the account-less run
+        # never was one. Both land in the bucket.
+        assert [row[0] for row in named_rows] == [named.id]
+        assert bucket is not None
+        bucket_runs, bucket_cost, _last = bucket
+        assert (bucket_runs, bucket_cost) == (2, Decimal("3.50"))
+        # Named rows plus the bucket reconcile with the by-agent totals.
+        assert sum(row[3] for row in named_rows) + bucket_runs == total_runs
+        assert sum(row[4] for row in named_rows) + bucket_cost == total_cost
+
+    async def test_a_window_with_every_run_named_has_no_bucket(self, db) -> None:
+        organization, owner = await _org_with_owner(db, "AllNamed")
+        agent = await _agent(db, organization, owner)
+        person = await _user(db)
+        await _run(db, organization=organization, agent=agent, started_at=START, user_id=person.id)
+
+        assert (
+            await agent_run_repo.unattributed_usage(
+                db, organization_id=organization.id, start=START, end=END
+            )
+            is None
+        )
+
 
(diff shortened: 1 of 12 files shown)


backend/tests/integration/test_usage_stats_sql.py
```diff
@@ -801,6 +801,79 @@ async def test_own_scope_answers_with_the_callers_single_row(self, db) -> None:
...
+
+        # The leaver has dropped out of the named rows; the account-less run
+        # never was one. Both land in the bucket.
+        assert [row[0] for row in named_rows] == [named.id]
+        assert bucket is not None
+        bucket_runs, bucket_cost, _last = bucket
+        assert (bucket_runs, bucket_cost) == (2, Decimal("3.50"))
+        # Named rows plus the bucket reconcile with the by-agent totals.
+        assert sum(row[3] for row in named_rows) + bucket_runs == total_runs
+        assert sum(row[4] for row in named_rows) + bucket_cost == total_cost
```
DEENUU1, line 862: This is the assertion that makes the whole PR trustworthy — checking named + bucket against `count_runs` and `sum_cost_window`, not against a hand-typed number. And it matches reality: `spend_by_agent` filters on `parent_run_id IS NULL` too, so the by-agent card really is summing the same top-level rows.

backend/tests/integration/test_usage_stats_sql.py
```diff
@@ -801,6 +801,79 @@ async def test_own_scope_answers_with_the_callers_single_row(self, db) -> None:
         assert [(row.user_id, row.runs) for row in result.by_user] == [(owner.id, 1)]
         assert result.total_runs is None
 
+    async def test_the_unattributed_bucket_reconciles_by_person_with_by_agent(self, db) -> None:
```
DEENUU1, line 804: One sibling test worth adding: `unattributed_usage` sums cost across the whole org's null-user runs, so a null-user run in another org leaking into this bucket would be a silent cross-tenant cost disclosure. It's provably safe today — `_window_conditions` pins `organization_id` — but there's already a tenant-isolation test in this file (the "other tenant's run owned by the same user" case, ~line 250), and a new aggregation path is exactly where this repo locks that down. A second org with a null-user priced run, asserting the bucket ignores it, is a cheap guard.

DEENUU1 (commented):
Nice piece of work — this closes the exact gap #539 named, and the honest way. The bucket and the named rows partition the same window (top-level only, delegations out), and that's the same window `spend_by_agent` sums with its own `parent_run_id IS NULL` filter — so the reconciliation isn't just asserted in a test, it holds by construction. The docstrings carry the reasoning where the next reader will actually find it.

Two non-blocking things before it lands:

1. **`docs/governance.md` is owed a sentence.** The "Who spent it" section (around line 312) describes this exact card, and #527 kept it current when the card was added. This PR changes that card's behaviour — the bucket, the window note, the whole reconciliation #539 is about — but the page still reads as if null-user runs just vanish. The docs-drift hook stayed quiet because the trigger map routes `stats.py` nowhere; that's a map gap, not your miss, but the page is still owed.
2. **A tenant-isolation one-liner for the new bucket query** — see the inline note.

Everything else looks right to me.

PR #544, opened 2026-08-10
perf(imports): defer aiogram and prefect out of app import

DEENUU1:
## What this fixes

Importing `app.main` took **~5.5s** warm, ~2.3s of which was two libraries the API
and the test suite never touch: **`aiogram`** (the Telegram SDK) and **`prefect`**.
Every scoped `pytest <file>` pays that fixed cost once before a single test runs, and
`tests/conftest.py` imports `app.main`, so all ~3,900 unit tests inherit it — the
exact "a few seconds, nearly all of it importing the app" cost the loop docs describe.

`Refs #520` — the second lever of the test-speed work, orthogonal to #535 (integration
schema-once): this one cuts the *per-invocation* import, #535 cut the *integration
per-test* DDL.

## What changed

- `app/main.py` — `TelegramAdapter` / `SlackAdapter` / `MattermostAdapter` were imported
  at module top but are only instantiated inside `lifespan` (registered at startup,
  stopped after the yield). Moved the three imports into `lifespan`. Production is
  identical (startup imports them anyway); the test suite drives the app over
  ASGITransport and **never runs `lifespan`**, so it stops importing aiogram + the Slack
  client entirely.
- `app/services/rag_sync.py` — imported `sync_collection_flow` at module top, the only
  top-level worker-task import in the services layer. Moved it into the dispatching
  method, matching `rag_document.py:279` and `sync_source.py:316`, which already do this.
  The flow only runs in the worker; nothing in the API path needs Prefect at import.
- `app/services/channels/telegram.py` — removed a dead `_telegram_router = Router()`
  module global that instantiated aiogram at import and is referenced nowhere.
- `tests/test_flow_dispatch.py` — the regression test patched
  `rag_sync.sync_collection_flow`; with the import moved it patches
  `rag_tasks.sync_collection_flow` (the source module), the same way its sibling
  assertions already patch the other two flows.

## How it failed before

Not a failure — a cost. `import app.main` eagerly loaded aiogram (~1.2s) and prefect
(~1.4s) plus their transitive deps, on every process start and every scoped test run.

## Testing

Measured warm, on this machine:

| | Before | After |
|---|---|---|
| `import app.main` | ~5.5s | **~2.3s** |

- `python -c "import app.main, sys; print('aiogram' in sys.modules, 'prefect' in sys.modules)"`
  → `False False` (both confirmed absent after the import).
- `tests/test_main.py::test_importing_the_app_does_not_pull_in_the_channel_sdks_or_prefect`
  guards it in a subprocess (cold import), so a re-added top-level adapter/flow
  import fails rather than silently restoring the cost (review follow-up).
- Full unit suite green in random order (3,899 passed).
- Integration suite green (400 passed).
- 100% platform-coverage gate holds — the moved lines run in `lifespan` and
  `start_local_sync`, both covered.
- `make lint-backend` clean; ty diagnostic count unchanged vs base (66 → 66, all
  pre-existing in template-inherited files).

## Noticed, not fixed

- The `import app.main` residual is still ~2.3s (sqlalchemy + pydantic-ai + the app
  graph). A `pytest-xdist` PR (next #520 lever) amortizes the full-suite cost; a warm
  test daemon would attack the single-invocation cost but carries real caveats here
  (our capability registry and Prefect `@flow` registration would need reload
  patch-points) — evaluated separately, not pursued yet.
- The `register_adapter` / `_slack_register` alias in `main.py` is removed
  (review follow-up, 3696fc59) — one name for one thing.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -33,11 +33,8 @@ from app.services.rag.vectorstore import BaseVectorStore
 from app.repositories.channel_bot import get_active_polling_bots
 from app.services.channel_bot import unseal_bot_token, unseal_slack_app_token
 from app.services.channels import register_adapter
-from app.services.channels.supervisor import open_inbound_stream
-from app.services.channels.telegram import TelegramAdapter
 from app.services.channels import register_adapter as _slack_register
-from app.services.channels.mattermost import MattermostAdapter
-from app.services.channels.slack import SlackAdapter
+from app.services.channels.supervisor import open_inbound_stream
 from app.core.rate_limit import limiter
 
 logger = logging.getLogger(__name__)
@@ -112,6 +109,15 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[LifespanState, None]:
             state["vector_store"] = vector_store
         except Exception as e:
             logger.error("pgvector connection failed: %s. Vector store will not be available.", e)
+
+    # Imported here rather than at module top so that importing `app.main` does
+    # not pull in the channel SDKs (aiogram, the Slack client) - ~1.2s of import
+    # that the API and the test suite pay for on every process start and never
+    # use. The adapters are only ever needed from this point on: registered at
+    # startup, and stopped after the yield below (#520).
+    from app.services.channels.mattermost import MattermostAdapter
+    from app.services.channels.slack import SlackAdapter
+    from app.services.channels.telegram import TelegramAdapter
 
     _telegram_adapter = TelegramAdapter()
     register_adapter(_telegram_adapter)
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -33,7 +33,6 @@ from app.services.rag.vectorstore import BaseVectorStore
 from app.repositories.channel_bot import get_active_polling_bots
 from app.services.channel_bot import unseal_bot_token, unseal_slack_app_token
 from app.services.channels import register_adapter
-from app.services.channels import register_adapter as _slack_register
 from app.services.channels.supervisor import open_inbound_stream
 from app.core.rate_limit import limiter
 
@@ -134,7 +133,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[LifespanState, None]:
         logger.error("Mattermost: failed to start the event stream: %s", _mm_exc)
 
     _slack_adapter = SlackAdapter()
-    _slack_register(_slack_adapter)
+    register_adapter(_slack_adapter)
     try:
         await _start_channel_polling("slack")
     except (OSError, ValueError, RuntimeError) as _slack_exc:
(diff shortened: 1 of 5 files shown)


DEENUU1 (commented):
Clean, and the reasoning in the comments is exactly right - production imports these anyway, and the suite never runs `lifespan`, so this is free. Nice catch on the dead `_telegram_router` too; it was pulling `Router` in for nothing. The test patch-point move is correct and now matches how its siblings patch the other two flows.

One thing I'd add before merging: a guard test. The whole win is "aiogram and prefect stay out of `sys.modules` after `import app.main`", and right now the first person to add a top-level `from ...telegram import TelegramAdapter` back silently restores the ~3s with zero signal - the exact regression this PR fixes, left undefended. We already have the pattern for this (`test_reload_supervisor.py:357` probes `sys.modules` in a subprocess). A subprocess check is worth it so the import is genuinely cold, not one another test already warmed.

Non-blocking: since you're already editing this import block, the `register_adapter` / `_slack_register` alias (`main.py:37`) could go now - you flagged it in the description, but it's one line and `_slack_register(...)` reads like a different function than the `register_adapter(...)` two lines up. Your call on scope.

backend/app/main.py
```diff
@@ -113,6 +110,15 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[LifespanState, None]:
...
 
+    # Imported here rather than at module top so that importing `app.main` does
+    # not pull in the channel SDKs (aiogram, the Slack client) - ~1.2s of import
+    # that the API and the test suite pay for on every process start and never
+    # use. The adapters are only ever needed from this point on: registered at
+    # startup, and stopped after the yield below (#520).
+    from app.services.channels.mattermost import MattermostAdapter
+    from app.services.channels.slack import SlackAdapter
+    from app.services.channels.telegram import TelegramAdapter
+
```
DEENUU1, line 120: This is the part that'll quietly regress. A small guard, in a subprocess so the import is actually cold:

```python
import subprocess
import sys


def test_importing_the_app_does_not_pull_in_the_channel_sdks_or_prefect():
    code = (
        "import app.main, sys; "
        "print('aiogram' in sys.modules, 'prefect' in sys.modules)"
    )
    out = subprocess.check_output([sys.executable, "-c", code], text=True).strip()
    assert out == "False False", out
```

Fails the moment someone re-adds a top-level adapter or flow import, which is exactly the signal we don't have today.

DEENUU1: Added as `tests/test_main.py::test_importing_the_app_does_not_pull_in_the_channel_sdks_or_prefect` (8106bcb2) — a subprocess probe mirroring the reloader probe in `test_reload_supervisor.py`, asserting `False False`, so the import is genuinely cold and the first re-added top-level adapter/flow import fails it. Also removed the `_slack_register` alias while here (3696fc59).

Issue #536, opened 2026-08-10
make test fails locally on any @flow test because backend/.env points Prefect at localhost:4200


DEENUU1:
## What breaks

`make test` / `make test-fast` fails locally on any test that invokes a Prefect
`@flow`, unless a Prefect server happens to be running on `localhost:4200`:

```
tests/test_approval_expiry.py::TestTheScheduledSweep::test_the_flow_answers_with_what_it_expired
    RuntimeError: Failed to reach API at http://localhost:4200/api/
    httpx.ConnectError: All connection attempts failed
```

The test itself is a pure unit test — it patches `get_db_context` and
`ApprovalService` and only asserts `await approval_expiry_sweep_flow() == 3`
(`tests/test_approval_expiry.py:321`). The connection attempt is not the test's
doing: `approval_expiry_sweep_flow` is `@flow`-decorated, so **calling it makes the
Prefect runtime try to reach `PREFECT_API_URL`.**

## The sequence that triggers it

- `backend/.env` sets `PREFECT_API_URL=http://localhost:4200/api` (line 90), for
  `make dev`, where a Prefect server is up.
- The backend test suites load `backend/.env` through the settings object, so that
  URL is in the environment for the run.
- Prefect 3 therefore talks to a *server* at that URL instead of falling back to its
  in-process ephemeral API. With no server running (i.e. any `make test` that isn't
  preceded by `make dev`), the `@flow` call fails to connect.

CI is green because it never loads `backend/.env` and sets no `PREFECT_API_URL`, so
Prefect there uses ephemeral mode. The failure is invisible to CI and hits exactly the
"run the suite locally" path.

## Why it matters (and where it sits)

This is the same class of problem as #520: a suite you cannot run locally without extra
ceremony is a suite people stop running. Found while working #215/#520 (integration
schema-rebuild fix, #535). Pre-existing — fails identically on `main`, not caused by
that branch.

Related: #520 (test-speed / local-runnability), #215. Part of the Tests/CI cluster
tracked in #168.

## How you would know it was fixed

- `make test` and `make test-fast` pass on a clean checkout with **no** Prefect server
  running and `backend/.env` present with `PREFECT_API_URL` set.
- The fix pins Prefect to ephemeral/test mode for the suite regardless of `.env` — e.g.
  `tests/conftest.py` clearing `PREFECT_API_URL` (and/or setting
  `PREFECT_SERVER_ALLOW_EPHEMERAL_MODE=true`) before the settings object is built, the
[...]

PR #570, opened 2026-08-10
perf(tests): run the whole-suite targets across worker processes

DEENUU1:
## What this fixes

`make test` was serial (~175s), and the integration slice inside it spends almost all
its time waiting on one Postgres — embarrassingly parallel work run one test at a time.
This runs the whole-suite targets across worker processes with `pytest-xdist`.

`Refs #520` — the third lever (parallelism). Orthogonal to and stacking with #535
(schema-once) and #544 (lazy imports).

## What changed

- `backend/pyproject.toml` + `uv.lock` — add `pytest-xdist>=3.8.0` to the `dev` group.
- `Makefile` — `test`, `test-fast`, `test-integration`, `test-cov` gain
  `-n auto --maxprocesses 4`.
- `docs/testing.md`, `docs/commands.md`, `.claude/rules/testing.md` — describe the
  parallel execution and why scoped runs stay serial.

## Why `-n auto --maxprocesses 4`, not bare `-n auto`

The two halves of the suite scale differently, and I measured both:

- **Integration is I/O-bound** on one database. Each worker already gets its own
  per-process `_p<pid>` database — the isolation #189 built extends to xdist workers for
  free — so they parallelise cleanly and saturate around four workers.
- **The unit suite is import-bound**: every worker imports the app once (~2–5s), and
  past four workers that startup grows faster than the parallelism pays back. On a
  ten-core laptop an uncapped `-n auto` (10 workers) ran the suite **slower than
  serial**, all of it worker startup. The cap makes `auto` mean "as many as help": four
  on CI's four-core runner, four on a bigger laptop, fewer on a smaller one.

Scoped runs (`pytest <file>`) stay **serial** — spawning workers to run one file costs
more than the file does, and the write-loop is scoped runs.

## Testing

Measured on `pgvector/pgvector:pg16`, warm, this machine:

| | serial | `-n auto --maxprocesses 4` |
|---|---|---|
| `make test` (unit + integration + gate) | ~175s | **~91s** |
| `make test-integration` | ~124s | **~52s** |
| integration with #535's TRUNCATE + xdist | ~62s | **~31s** |

- **100% platform-coverage gate holds** under xdist — `pytest-cov` combines the
  per-worker data automatically; verified `Required test coverage of 100.0% reached`
  under `-n auto`.
- Full suite green under xdist's load distribution (4299 passed, 6 skipped).
- `tests/test_ci_parity.py` passes — CI calls `make test`, so the internal flag change
  is transparent.
- `make lint-backend` clean. The lock adds `pytest-xdist` + its `execnet` dep, and
  also re-resolves `whenever`'s `tzdata`/`tzlocal` markers to gate them on
  `python_full_version >= '3.13'` — **not** a deliberate change: restoring main's
  lock and running `uv lock` on the pinned toolchain (uv 0.11.2) reproduces exactly
  that marker update with no dependency added, so it is uv catching main up, not
  this PR. Inert for us — pinned to 3.12 on Linux/macOS, where both packages are
  already excluded by their platform markers before and after.

## Noticed, not fixed

- **`pytest-randomly` is documented as on-by-default (CLAUDE.md, `.claude/rules/testing.md`)
  but is not installed and not in the lockfile** — so the suite runs in collection
  order, `-p no:randomly` in the docs is a no-op, and order-independence has never been
  verified by that mechanism. I confirmed the suite *is* order-independent anyway (full
  unit + integration green across two random seeds each with `pytest-randomly`
  temporarily installed, and green under xdist's shuffled distribution here). Filing as
  its own issue; enabling randomly for real is then a low-risk one-liner. Not folded in
  to keep this diff scoped and avoid introducing run-to-run CI order changes unannounced.
- The `#536` Prefect-local test failure is unrelated and pre-existing (CI passes it;
  locally it needs no Prefect server because `backend/.env` points at `localhost:4200`).

### DIFF
--- a/backend/uv.lock
+++ b/backend/uv.lock
@@ -68,6 +68,7 @@ dev = [
     { name = "pre-commit" },
     { name = "pytest" },
     { name = "pytest-cov" },
+    { name = "pytest-xdist" },
     { name = "ruff" },
     { name = "ty" },
 ]
@@ -137,6 +138,7 @@ dev = [
     { name = "pre-commit", specifier = ">=4.0.0" },
     { name = "pytest", specifier = ">=9.0.0" },
     { name = "pytest-cov", specifier = ">=6.1.0" },
+    { name = "pytest-xdist", specifier = ">=3.8.0" },
     { name = "ruff", specifier = ">=0.16.1" },
     { name = "ty", specifier = ">=0.0.69" },
 ]
@@ -1253,6 +1255,15 @@ dependencies = [
 sdist = { url = "https://files.pythonhosted.org/packages/50/79/66800aadf48771f6b62f7eb014e352e5d06856655206165d775e675a02c9/exceptiongroup-1.3.1.tar.gz", hash = "sha256:8b412432c6055b0b7d14c310000ae93352ed6754f70fa8f7c34141f91c4e3219", size = 30371, upload-time = "2025-11-21T23:01:54.787Z" }
 wheels = [
     { url = "https://files.pythonhosted.org/packages/8a/0e/97c33bf5009bdbac74fd2beace167cab3f978feb69cc36f1ef79360d6c4e/exceptiongroup-1.3.1-py3-none-any.whl", hash = "sha256:a7a39a3bd276781e98394987d3a5701d0c4edffb633bb7a5144577f82c773598", size = 16740, upload-time = "2025-11-21T23:01:53.443Z" },
+]
+
+[[package]]
+name = "execnet"
+version = "2.1.2"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/bf/89/780e11f9588d9e7128a3f87788354c7946a9cbb1401ad38a48c4db9a4f07/execnet-2.1.2.tar.gz", hash = "sha256:63d83bfdd9a23e35b9c6a3261412324f964c2ec8dcd8d3c6916ee9373e0befcd", size = 166622, upload-time = "2025-11-12T09:56:37.75Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/ab/84/02fc1827e8cdded4aa65baef11296a9bbe595c474f0d6d758af082d849fd/execnet-2.1.2-py3-none-any.whl", hash = "sha256:67fba928dd5a544b783f6056f449e5e3931a5c378b128bc18501f7ea79e296ec", size = 40708, upload-time = "2025-11-12T09:56:36.333Z" },
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


backend/uv.lock
```diff
@@ -5598,8 +5622,8 @@ name = "whenever"
 version = "0.10.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "tzdata", marker = "sys_platform == 'win32'" },
-    { name = "tzlocal", marker = "sys_platform != 'darwin' and sys_platform != 'linux'" },
+    { name = "tzdata", marker = "python_full_version >= '3.13' and sys_platform == 'win32'" },
+    { name = "tzlocal", marker = "python_full_version >= '3.13' and sys_platform != 'darwin' and sys_platform != 'linux'" },
```
DEENUU1, line 5626: nit: the lock also re-resolved `whenever`'s markers here (gating `tzdata`/`tzlocal` on `python_full_version >= '3.13'`), which the PR body doesn't mention — it says the lock "adds only `pytest-xdist` and its `execnet` dep."

Harmless in practice: we're pinned to 3.12 on Linux/macOS, so neither package was being installed either way. But it's unscoped churn from a newer `uv` re-resolving. Either drop it to keep the lock diff to the two packages you meant, or add a line to the PR body so the next person reading `git blame` on this marker isn't puzzled by why a test-parallelism PR touched a timezone dep.

DEENUU1 (commented):
This is a tidy win — halving `make test` by parallelising the part that's just waiting on Postgres, with the `--maxprocesses 4` cap justified by actual measurement rather than a guess. What I like most is that it needed almost no test-infra change: the `_p<pid>` isolation from #189 already separates xdist workers for free, and the Makefile comment plus the docs update explain *why scoped runs stay serial*, which is exactly the thing a future reader would otherwise undo. The "Noticed, not fixed" note on `pytest-randomly` is the right call too — flag it, don't fold it in.

One optional thing on `test-cov` (`Makefile:348`): it still runs the full suite serially. I get that it's not the write-loop and not CI-gated, so it's defensible to leave it — but it's the same ~175s full run, so if anyone uses it to inspect coverage locally it'd benefit from the same `-n auto --maxprocesses 4`. Your call whether it's in scope here.

Nothing blocks — one small note on the lockfile inline.

backend/uv.lock
DEENUU1: Went with option 2 (a note in the PR body) rather than dropping it, because it can't be dropped cleanly. Restoring main's lock and running `uv lock` on the pinned toolchain (uv 0.11.2) reproduces *exactly* this `whenever` marker change on its own, with no dependency added — main's lock is simply behind what current uv resolves, so this is uv catching up, not xdist dragging in a timezone dep. Hand-reverting the two lines would leave a lock that the next `uv lock`/`make deps` immediately wants to rewrite: the same puzzle, moved one person downstream. It's inert for us either way — pinned to 3.12 on Linux/macOS, where both `tzdata` and `tzlocal` are already excluded by their platform markers before and after. Documented in the body. Also took the `test-cov` note — it now runs `-n auto --maxprocesses 4` like its siblings (b44a1fdc).

Issue #11, opened 2026-08-01
Drain background work before tearing the process down


DEENUU1:
**Severity: High** · AUD-011

## What's wrong

`drain()`'s own docstring says: *"Called from the application lifespan. Without it, shutting down mid-flight cancels ingestion and sync work that was nearly done, which shows up later as a document stuck in `processing` forever."* It is not called from the lifespan — `rg drain app/ tests/` returns only `tests/test_background.py` and the definition.

And the shutdown order is close to inverted:

```python
# main.py:131 — current
yield state
if "vector_store" in state:
    await state["vector_store"].engine.dispose()   # engine disposed FIRST
for _bid in list(_telegram_adapter._polling_tasks.keys()):
    await _telegram_adapter.stop_polling(_bid)     # intake stopped SECOND
...
await close_db()
```

## Where

- `backend/app/core/background.py:66-85` (zero non-test callers)
- `backend/app/main.py:131-144`

## How it fails

Upload a document, `spawn(ingest_document_flow(...))`, deploy. The task is a plain `asyncio.Task` in the API process; uvicorn drains HTTP, runs lifespan shutdown, tears the loop down. `_run_ingestion` never reaches `complete_ingestion`, its `except` never fires, and the `rag_documents` row stays `processing` forever with nothing in the logs — exactly the symptom the docstring predicts.

Meanwhile a Telegram message arriving between the engine dispose and the poller stop runs a knowledge search against a disposed engine.

Compounding it, the three webhook routes bypass `spawn` entirely (`slack_webhook.py:71`, `telegram_webhook.py:47`, `mattermost_webhook.py:75`) with bare `asyncio.create_task` into private module sets that `drain()` cannot see — so a Slack message being answered during a deploy is dropped, and if `close_db()` wins the race it dies on a disposed engine.

## Fix

Stop intake → drain → close resources:

```python
yield state
for _bid in list(_telegram_adapter._polling_tasks): await _telegram_adapter.stop_polling(_bid)
for _sbid in list(_slack_adapter._socket_tasks):     await _slack_adapter.stop_polling(_sbid)
for _mbid in list(_mattermost_adapter._socket_tasks): await _mattermost_adapter.stop_polling(_mbid)
await background.drain()
if "vector_store" in state:
    with suppress(Exception):
        await state["vector_store"].engine.dispose()
if "redis" in state:
    await state["redis"].close()
await close_db()
```

and replace the three `create_task` calls with `spawn(...)`, deleting the local sets.

[...]

PR #579, opened 2026-08-10
refactor: strip dead weight across backend and frontend

### LINKED ISSUE #521: Audit the codebase for dead weight: strip needless comments and unused functions
## What

Audit the codebase for dead weight and strip it. The quality bar in `CLAUDE.md`
is explicit — "No dead weight. No speculative abstraction, unused parameter,
commented-out code or 'just in case' branch" — but nothing has swept the tree
against it as a whole.

Two passes:

- **Needless comments.** Comments that restate *what* the code already says,
  commented-out code, and stale comments that describe behaviour the code no
  longer has. Keep the comments that explain *why*; delete the ones that narrate
  *what*.
- **Unused / near-dead functions.** Functions, parameters and branches nothing
  reaches — unreferenced helpers, "just in case" branches, speculative
  abstractions with a single caller. A branch that cannot be reached is deleted;
  one that can is tested.

Method: prefer a mechanical first pass (`ruff`/`vulture`-style dead-code
detection, grep for orphaned symbols) to find candidates, then a human read
before removing — a symbol with zero callers can still be a public entry point
or a registry-driven capability.

## Why

Dead weight is a standing invitation to read code that lies. A comment that
describes the old behaviour is worse than no comment; a function nobody calls is
a maintenance cost with no return. This is the same standard the audit of
2026-08-01 applied piecemeal (#33) — this issue is the tree-wide sweep behind it.

## Scope note

Scoped, mechanical diffs — deletions with tests still green, not opportunistic
refactors folded in. Anything that turns out to be a real bug gets its own issue
per the filing rule, not a fix smuggled into this branch.

## Done when

- The comment and dead-function passes are done across `backend/app/**` and
  `frontend/src/**`, with green suites.
- A short written note of what was removed and what was deliberately kept.

## Related

Code-health cluster: **#33** (nine small items from the 2026-08-01 audit).
Adjacent dead-weight issues: **#155** / **#156** (unused backend/frontend
dependencies), **#158** (drop the langchain tree). See the issue map **#168**.

DEENUU1:
## What this fixes

The tree-wide dead-weight sweep behind #521. Removes functions, methods, files
and exports that nothing reaches — across `backend/app/**` and
`frontend/src/**` — plus the one item the issue asked to *wire* rather than
delete. Deletions only; no behaviour change, and the diff is scoped to what the
verified inventory named (with the corrections below).

`Closes #521`

## What changed

Three commits, by area:

- **`refactor: remove unused backend helpers, methods and dead abstractions`** —
  11 repository query helpers, ~11 service methods, the dead
  `RetrievalService.retrieve_by_document` abstraction, 5 never-wired sanitizers
  in `core/sanitize.py`, and the stray helpers `serializable_dict` /
  `is_model_provider`. Tests that only exercised a removed symbol go with it.
- **`refactor(rag): close the vector store through its public aclose()`** — the
  "wire it, don't delete it" item. See below.
- **`refactor(frontend): remove dead files, exports and unused deps`** — two
  dead files, six dead exports, four `export` keywords narrowed to internal, two
  redundant barrel entries, and the two now-unused radix dependencies.

## The comment pass

Confirmed near-empty, as the issue's own inventory predicted: backend carries no
TODO/FIXME/HACK and no commented-out code; frontend has no dead comment blocks.
The one stale comment found — a `middleware.ts` line naming the deleted
`LanguageSwitcher` — is updated. The rest of the `#`/`//` lines are "why"
comments and stay.

## `aclose()` — wired, not deleted

`PgVectorStore.aclose()` read as dead because the lifespan teardown reached past
it into `state["vector_store"].engine.dispose()` with a
`# type: ignore[attr-defined]`. `aclose()` is now abstract on `BaseVectorStore`,
the lifespan calls the public method (ignore gone), and `PgVectorStore` disposes
its pool as before. Same behaviour — the pool is still released once on shutdown —
now through the contract. Regression test added.

## Corrections to the inventory (the "human read" the issue asked for)

Four candidates were **kept** because deleting them would have removed real
behaviour, not dead weight:

- **`ingestion_config.overlap_fits_inside_a_chunk`**, **`rag.config.set_dim_from_model`**,
  **`rag.models.connect_pages`** — all `@model_validator(mode="after")`. Pydantic
  invokes them by registration, which is exactly why a static pass sees zero
  callers. They enforce chunk-overlap bounds, embedding-dimension derivation and
  page↔doc linkage respectively.
- **`channels.directory.supported_tools`** — has no production caller, but it is
  the sole mechanism behind the drift test that keeps `PLATFORM_TOOLS` honest
  against the adapter classes (both directions). Deleting it and its test removes
  a real invariant guard for no production gain; it stayed.

Two were **reclassified** from "dead" to "over-exported": `OAuthButtons` and
`OAuthDivider` are not dead — `OAuthBlock` renders both — so they were narrowed
to module-private rather than deleted. `refusesOrganization` keeps its `export`
(a co-located unit test imports it directly) and lost only its unused barrel
entry.

Two **cascade orphans** the sweep itself created were removed in the same pass:
deleting `list_tool_calls` / `get_message_ratings` / `create_default_for_org`
left `get_tool_calls_by_message`, `get_ratings_for_message` and
`get_default_for_org` with no callers.

## Testing

- `make test` — backend green, platform-layer coverage still **100%**
  (4309 passed; 5 tests removed with their dead symbols, 1 added for `aclose`).
- `make test-frontend-cov` — green (100% lines/functions, 97.6% branches).
- `make lint` / `tsc` / `eslint` / `prettier` — clean.
- **Pre-existing, not mine:** `tests/test_approval_expiry.py::…::test_the_flow_answers_with_what_it_expired`
  fails locally because it needs a Prefect API at `localhost:4200` that isn't
  running here. It fails on the base ref too; CI has the Prefect setup.

## Noticed, not fixed

- `email.EmailKey.NEWSLETTER_WELCOME` and its template are now unreferenced after
  removing `send_newsletter_welcome`. Left in place — an enum member + template
  is data/registry, and pruning it is a wider change than this scoped sweep; worth
  its own issue if the newsletter flow is truly gone.
- `PgVectorStore` still creates its own engine per instance (`deps.get_vectorstore`
  builds a fresh one per request). Wiring `aclose()` fixes the shutdown path, not
  the per-request pool churn — that is the design smell the class docstring already
  flags, and belongs to the #11 teardown cluster rather than here.


backend/app/main.py
```diff
@@ -149,7 +149,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[LifespanState, None]:
     yield state
     if "vector_store" in state:
         with suppress(Exception):
-            await state["vector_store"].engine.dispose()  # type: ignore[attr-defined]
+            await state["vector_store"].aclose()
```
DEENUU1, line 152: Nice — this is the right shape. The old line reached past the interface into `.engine` with a `# type: ignore[attr-defined]`; routing shutdown through the public `aclose()` means the teardown depends on the contract, not on `PgVectorStore`'s internals. Good call wiring it instead of deleting it.

backend/tests/test_vectorstore_aclose.py
```diff
@@ -0,0 +1,24 @@
+"""Shutdown teardown for the vector store.
+
+The pgvector store creates its own SQLAlchemy engine, so the application
+lifespan must release that pool on shutdown. It does so through the public
+`aclose`, not by reaching into `.engine`; these tests pin both halves of that
```
DEENUU1: Small accuracy thing: the docstring says "these tests pin **both halves** of that contract", but there's a single test here and it only pins the first half — that `aclose()` disposes the pool. The second half, that the lifespan calls `aclose()` rather than `.engine.dispose()`, isn't asserted anywhere. Either soften the wording, or (if you want the second half genuinely pinned) assert the lifespan path. The wording fix is the cheap one:

```suggestion
lifespan must release that pool on shutdown. It does so through the public
`aclose`, not by reaching into `.engine`; this test pins the store side of that
contract — `aclose` disposes the pool.
```

DEENUU1 (commented):
Clean sweep, and a careful one. I traced the blast radius on the backend removals independently — no external caller survives for any of the deleted repo helpers, service methods or sanitizers, and dropping the abstract `retrieve_by_document` is safe because `RetrievalService` is the only subclass. The frontend narrowings check out too: the four `export`→internal changes still have in-module callers, and `refusesOrganization`/`providerInfo` correctly keep their exports for the co-located tests.

The part I like most is that this isn't just deletion — turning `aclose()` from a dead method into an abstract contract, so the lifespan stops reaching past it into `.engine` with a `# type: ignore`, is a real improvement. The "kept" and "cascade orphan" notes in the description show the human-read the issue asked for actually happened.

One small thing on the new test's docstring (inline). Otherwise this is good to merge.

backend/tests/test_vectorstore_aclose.py
DEENUU1: Softened the wording (e8928bb8) — the docstring now says the test pins the store side (`aclose` disposes the pool), which is what the single test actually asserts. Left the lifespan half unasserted rather than force a full-lifespan test for it; the docstring no longer overclaims.

PR #587, opened 2026-08-10
feat(dashboard): let a person arrange their own dashboard layout

### LINKED ISSUE #213: Let a person arrange their own dashboard: pick widgets, order and width
## What this is

**Let a person arrange their own dashboard: choose which widgets appear, in what order, and how wide.** Today the dashboard has four fixed layouts, one per audience (deployment admin / org steward / operations / everyday user), and a person gets whichever one their permissions resolve to. This adds a layer on top: they keep the audience layout as the starting point, and can then reorder it, hide what they do not use, add a widget from a catalog, resize, and reset back to the default.

This was out of v1 in the stage-1 design (PR #149, `docs/design/dashboard-demo.md` → *Deliberately out of v1*). **That decision is reversed — it is in scope now.** The design note already settled the whole shape, which is why this is smaller than it sounds: implement what is written there, do not redesign it.

Depends on #37 stage 2 landing first. This is a layer over the widget registry, so there has to be a registry.

## The rule that must not bend

```
effective layout = stored preference ?? audience default
visible          = effective ∩ gate()
```

**A preference may only ever reorder or hide. It may never reveal.** The gate runs last, on every render, over whatever the preference says. A stored layout naming a widget the caller cannot see is dropped at render time rather than trusted — the realistic case is somebody demoted from admin to member six months after they arranged their page, not an attack, and it has to be correct for both.

Two consequences worth stating because they are easy to get wrong:

- **The "add a widget" catalog is itself gated.** The list of widgets somebody can add passes through `gate()` before it is rendered, or the catalog leaks the existence of what the page hides.
- **Nothing about a stored layout is trusted on read.** An unknown widget id (renamed, removed) is dropped, not rendered as a blank card. An unknown span is replaced with the widget's default.

## What to build

**Storage**

- [ ] A new table. It cannot be a column on `users`: the layout is scoped **per user and per organization** — the same person is a steward in one org and a member in another, and one saved layout across both is wrong in one of them. So `(user_id, organization_id)` with a unique constraint, and the layout as JSONB.
- [ ] `ondelete="CASCADE"` from both sides: a removed membership leaves no orphan layout.
- [ ] **Validate on write, not on read.** A rule narrowed later on a field already holding JSONB rows is the trap `.claude/skills/alembic-migration` exists to warn about — decide the stored shape now and reject anything else at the API boundary.

**API**

- [ ] `GET` / `PUT` / `DELETE` on the caller's layout for the active organization (`DELETE` is reset-to-default). Any signed-in member may read and write **their own**; there is no route that reads somebody else's, and no permission gates this — it is a personal preference, not org data.
- [ ] Tenant isolation tested, including the case where the caller owns the row.
- [ ] Widget ids validated against the registry on write, so a typo fails loudly instead of producing a card that never renders.

**Frontend**

- [ ] Widget definitions carry catalog metadata from day one: id (a stable registry key, because it is being persisted), title, description, the permission that reveals it, default span, default audiences, and a **"see all" destination pointing at a page that already exists**.
- [ ] An edit mode on the dashboard: reorder, hide, resize, add from the catalog, reset to default. Drag-and-drop is not required — move up / move down / hide / a span picker is a complete answer and is keyboard-reachable, which drag-and-drop on its own is not.
- [ ] Spans come from the closed set the grid already uses: `s4` / `s6` / `s8` / `s12`. No free resizing, no new spans.
- [ ] **Span and title ride the layout entry, not a widget-id lookup table.** The demo keys both off the widget id, and PR #149's note flags this as a thing that must not be ported: the moment a person can place the same widget twice ("runs over time, once org-wide and once for me") an id-keyed lookup collides.
- [ ] `gate()` takes what it needs — `(can, isAppAdmin) => boolean` — rather than closing over module state, so a widget is testable alone.
- [ ] Every string through `next-intl`, including the catalog descriptions.

**Tests**

- [ ] A stored layout naming a widget the caller may not see does **not** render it. This is the test the whole feature rests on.
- [ ] A stored layout naming a widget id that no longer exists renders the rest of the page rather than breaking.
- [ ] Reset-to-default returns the audience layout, not an empty page.
- [ ] The same user in two organizations gets two different layouts.
- [ ] An integration test for the catalog: a widget the caller cannot see is absent from the "add" list.
- [ ] Empty is not failed — a person who has hidden everything sees a state that offers reset, not a blank page that looks broken.

## Explicitly out

- **Sharing a layout with somebody else, or an org-wide default set by an admin.** Both are reasonable and both are a different feature, with their own permission questions.
- **New widgets.** This arranges what #37 stage 2 builds; anything new is its own issue.
- **Free-form grid positioning.** Order plus a span from the closed set, nothing more.

## Where it sits

Cluster: **Frontend surfaces** (#37 #45 #129 #139). Blocked by **#37** (the widget registry this layers over). Related: **#45** / PR #202 — Activity's filter state lives in the URL rather than in a stored preference, and that stays true; this is about widget placement, not filter state, and the two must not be conflated. The shape implemented here is specified in `docs/design/dashboard-demo.md` on PR #149; sign-off comment: https://github.com/vstorm-co/agenticos/pull/149#issuecomment-5190852925

OchnikBartek:
## Summary

The dashboard becomes arrangeable. A person can reorder, resize, hide, colour and group its cards, add more from a gated catalog, and keep the result — either as their single active arrangement or as any number of named presets they switch between. The arrangement is a third layer over the two the page already has (`effective = preference ?? audience default`), with the permission gate always running last, so a preference can reorder or hide but never reveal a widget the caller may not see. It is persisted per user per organization.

## Related Issue

Closes #213.

## Added

- Two tables with their repositories — `dashboard_layouts` (one active arrangement per user per organization) and `dashboard_presets` (named arrangements), placements stored as JSONB (migrations `0022`, `0023`, on top of `0021`).
- `schemas/dashboard_layout.py` with a deliberately asymmetric contract: writes validate every placement against the widget registry and the closed span/row sets (a typo is a 422 at the boundary); reads hand back what was stored (a retired widget id drops at render time, never 500s the page). An entry is a widget placement or a section divider, discriminated on `kind`.
- Services and routes under `/api/v1/me/dashboard-layout` (the active arrangement) with a `/presets` shelf beneath it; applying a preset is a `PUT` of its entries, so there is one write path. Same-origin frontend proxy routes for both.
- Frontend data layer: `lib/dashboard/registry.ts` (widgets, spans, rows, accents), `preference.ts` (group/flatten, sanitize, resolve), `layouts.ts` (grid vocabulary), typed API clients and their hooks.
- The arrange editor UI in `components/dashboard/` — a pointer-drag engine that reorders and resizes cards and animates on drop, section dividers a person can name/colour/collapse, an add-widget catalog with a live gated preview, and the named-preset menu with its save dialogs (including a warning before a from-blank layout is saved only to the throwaway active slot).
- Polish translations for every string the feature adds.

## Changed

- `dashboard/page.tsx` renders a saved arrangement — grouped into its own dividered sections — when one exists, still gated last.
- The period filter and section headings were restyled to match the Agents section (segmented pill control, `tracking-tight` titles); `globals.css` gains the accent and arrange-grid utilities.

## Checks

- [x] `make lint` passes (ruff, ruff-format, ty, eslint, prettier, tsc, the i18n and backtick guards, codespell)
- [x] Platform-layer coverage is 100% (`make test` — total 100.00%, gate reached)

## Notes for Reviewers

- Rebased onto current `origin/main`: the migrations were renumbered `0013/0014 → 0022/0023` and re-parented onto `0021_drop_channel_tools_bindings`, because a sibling migration merged to `main` since this branch was cut shared the same parent and the merge produced two alembic heads. `alembic heads` is now a single head; a full `downgrade base → upgrade head` cycle and `alembic check` are clean on a throwaway database.
- Verified locally, each green: `make lint`; the backend suite at **100.00% platform coverage** (`make test` — "Required test coverage of 100.0% reached"), all backend dashboard tests passing (unit + API + integration against a real pgvector database); `make build-frontend`, `make docs-build`.
- Two known non-blockers left for CI, neither introduced by this branch:
  - `make test` fails locally on one pre-existing, unrelated test — `tests/test_approval_expiry.py::TestTheScheduledSweep::test_the_flow_answers_with_what_it_expired` — which raises an outbound-HTTP `ConnectError` in this sandbox (no network egress). This branch touches nothing under approvals.
  - The frontend coverage gate meets every threshold (100% lines/statements/functions, 97.5% branches), but a handful of heavy pre-existing integration specs (kb, teams, skills, chat, runs) intermittently trip the 5s `testTimeout` under coverage instrumentation on a loaded machine — a documented flake (`.claude/rules/testing.md`). The set is non-deterministic and every flaking file passes in isolation; none are dashboard files.
- The FLIP animation measures positions relative to its own container rather than the viewport, because the dashboard scrolls an inner `overflow-auto` element and `document.scrollY` stays 0 — see `use-flip.ts` if reviewing the drag/animation.
- Commits are partitioned by layer (persistence → API → backend tests → frontend data → UI → i18n → docs); squash-merge collapses them.

---

## Rebased on `main` (0.0.151), and a visual pass over the whole dashboard

`main` moved a long way while this branch was open, and 0.0.151 in
particular standardised every list in the product onto `DataTable`,
`ListCard` and the `Card` primitive. The dashboard was the one page that
had not moved with it — and it is the page that shows every kind of
surface at once, so it was where the divergence read loudest.

**Merge.** Two conflicts, both frontend: `filter-row.tsx` took `main`'s
extracted `PeriodControl`, and `registry.ts` kept this branch's expanded
widget rows while dropping `ROUTES.ADMIN_RATINGS`, which left with the
admin overhaul. `main` also brought `0022_hosted_embeds..0027_message_ordinal`
off the same parent as this branch's two migrations, so those were
renumbered `0028`/`0029` and re-parented onto `0027`; `alembic heads`
answers with one head again.

**A widget is now the object every other page draws.** `WidgetFrame` is
built on `Card` — same corner, same elevation, the divider under the
heading that `ListCard` carries — instead of a hand-rolled `rounded-xl`
with no shadow. `Metric` and `DeltaChip` are the one way a figure and its
change are drawn, matching Activity's mono numerals under a small
upper-case label; thirteen widgets had each rolled their own.

**Every card explains itself, once.** The header carries an info icon
holding the widget's own `description` — the same sentence the add-widget
catalog lists it under. Five cards had been carrying that explanation as
grey prose under the data, two of them a full paragraph. Two explanations
deliberately stayed on screen: the sandbox card's missing host figure, and
the per-person table's audience note.

**The page opens on numbers.** A steward's dashboard led with "Needs
attention", every card of which is an empty state in an organization where
nothing is wrong — five "nothing here" boxes before the first number. The
new `summary` widget is a full-width strip of runs, completed share, spend
and active people, leading the steward, operator and builder layouts. It
costs no request: every figure is a slice of the composed `/stats/usage`
response the cards below already share, and the completed share reads
`run-outcomes`, so it and the Outcomes donut cannot disagree.

**And the rest of the pass:** section bands set as mono kickers so the page
has three typographic levels rather than two; the primary action moved to
the header where every other page puts its primary, leaving the control
strip to filters and three ghost shortcuts that wrap as groups; bar-list
labels widened (every row in two cards ended in an ellipsis); each info
icon named for its card rather than twenty-seven identical stops.

**One defect found and fixed along the way.** The sections filter offered a
section only if it carried a `titleKey`, so it disappeared the moment
somebody saved an arrangement — including one that kept every heading it
started with, because flattening the default turns each heading into a
divider named by `title` instead. `isFilterable` reads either name, and
`sectionLabel` is now the single copy of "the caption a person typed, else
the curated key".

### Verified

- `make check` — every CI job except e2e.
- Frontend: 284 files / 4344 tests, and the coverage gate (100% lines,
  statements, functions; 97.56% branches against a 97.5% floor).
- In a browser, against the dev stack: the summary strip renders, the
  period control writes `?period=`, the sections filter hides a band and
  never the strip, an arrangement saves and keeps its filter, "Default
  layout" resets it, and each info icon opens its own explanation. Light
  and dark, at 1600px and 1280px.
- Request fan-out measured: 25 distinct API calls for the app-admin page,
  `/stats/usage` asked three ways (org, per-user, own) and shared by every
  card that reads it. The summary strip added none.

`docs/design/dashboard-demo.md` records the three stage-1 calls this
superseded — the 8px card, the figure typography, and quick actions on the
filter row — rather than editing them away, because each was right when it
was written.

---

## A visual system for the dashboard, not another coat of paint

The first pass matched the dashboard's chrome to what 0.0.151 standardised.
The verdict on it was that the widgets still read weak — margins, colours —
and measuring rather than arguing showed why: **there was no system to be
consistent with.** Across `components/dashboard` the spacing classes in use
were eight `p-*` values, six `gap-*` and eight `py-*` — forty-one distinct
numbers — and the rest of the product is no better, so "match the app" was
never available. Five different tokens drew the same quiet surface. The
charts broke four named anti-patterns.

So this establishes the system, applies it, and fills what the page could
not answer.

**Rhythm.** Band-to-band was 24px against card-to-card's 16px: three levels
of structure inside eight pixels of each other, which is why five bands read
as one mass. Bands sit at 40px now — four times the card gap — while the
header and control strip keep the 24px every other page uses.
`lib/dashboard/system.ts` holds the four values and the reasoning; its test
asserts the *relationships*, since a constant equal to its own literal tests
nothing.

**Data ink, measured with the dataviz validator rather than argued.**
`--color-chart` is `brand-500` in both themes: 3.74:1 on the light card,
5.07:1 on the dark, both clear of the 3:1 floor a mark owes its surface and
both quieter than what shipped. New `--color-track` draws the unfilled part
of a bar in the fill's own hue — the meter rule — at 1.19:1 light and 1.11:1
dark. `brand-900` was tried in dark and is why the pair is measured: at
1.41:1 it read as a filled bar, so a provider that spent nothing showed a
full-width mark beside $0.00.

**One figure component.** There were three — `StatCard` on Admin (with a
sparkline and delta the dashboard never used), `Metric` on the dashboard, and
a private `Figure` inside `ActivityFigures` — so the same number changed
typeface between a card and the page its "see all" points at.
`components/ui/figure.tsx` is the one, and the number is **sans, semibold,
with the font's own figures**: `tabular-nums` and a mono face on a large
standalone value are both named anti-patterns, so equal-width digits stay in
table rows and axis ticks where columns align.

**Charts to spec.** Dashed gridlines gone (ink that is not data), the area
wash flat at 10% instead of a 0.25 gradient inventing a second encoding down
the y-axis, bars square where they start and rounded where they end over a
track in the fill's own hue, and a real hover where a truncated model id used
to hide in a native `title`.

**No skeleton flash.** A period change was a new query key, so ten cards
dropped to skeletons at once and the page emptied and reflowed.
`keepPreviousData` holds the last answer; `UsageBody` dims it and sets
`aria-busy` so it is visibly the answer being replaced.

### The defect this turned up

**The Spend card put two definitions of cost side by side.** The headline was
`/stats/usage → cost.period_usd` — `sum_cost_window`, model spend alone. The
line under it was `/spend → month_to_date_usd` — `organization_spend_since`,
runs **plus ingestion**, which is the arithmetic a monthly cap is actually
measured with. Nothing said they were different questions, and on any
deployment that indexes documents they disagreed; `ingestion_spend` is real
money already counting against the cap and appeared nowhere on the page.

`CostBlock.period_usd` is the whole bill now, with `model_usd` and
`ingestion_usd` beside it and `previous_period_usd` following so the change
compares like with like. At `scope=own` the ingestion half is zero rather
than a share — a document is indexed by a worker and the ledger records no
user.

### What the page can answer now

- **Channels** — `surfaces` says where work *came from*, so a channel nobody
  used is absent from it. This lists what is registered, and whether each bot
  listens on a webhook or polls: the difference between "silent because
  nobody asked" and "silent because nothing is listening".
- **Knowledge** — `knowledge-freshness` says whether documents are still
  arriving; this says whether the ones that did ever finished indexing. A
  collection can be perfectly fresh and hold two hundred documents nothing
  can retrieve.
- **When the work happens** — a week by an hour, shaded by runs, on a new
  `group_by=hour`. One hue in five steps, zero not a step, every cell
  hoverable.
- **Sparklines** on three of the four KPI tiles, free from `runs_by_day` now
  answering runs, completed and cost from the one scan it was already making.

Both list cards read hooks that already existed; only the heatmap and the
daily measures needed backend.

### And the layouts are rebuilt on one rule

**A row holds cards of comparable natural height.** That is what left a
four-line list beside a chart two-thirds empty. Charts pair with charts,
lists with lists, figures with figures; the heatmap takes a row to itself,
because anything beside seven-by-twenty-four is either dwarfed or stretched
to a height it has no content for.

### Verified

`make check` green — 4,959 backend tests at 100% platform coverage, 4,355
frontend tests and both gates. The colour decisions were run through
`validate_palette.js`, not eyeballed. In a browser against the dev stack, both
themes at 1600px and 1280px: every new card with real data, a period change
that dims rather than blanks, and the arrange editor and preset switcher still
round-tripping.

Scope note, plainly: this touches `components/runs/**` and `admin/` — files a
dashboard branch had no business in — because unifying the figure component is
what makes the typography rule true rather than aspirational. The commits are
partitioned by workstream so they can be read one at a time.


frontend/src/components/dashboard/dashboard-editor.tsx
```diff
@@ -0,0 +1,577 @@
...
+  // resize grip or a control (`data-resize`/`data-no-drag`) is not a drag; a
+  // press on a section grip drags the band, and anywhere else on a card drags
+  // the card. It arms on pointerdown and activates only past a small threshold,
+  // so a plain click never nudges the order. Nothing reorders while the pointer
+  // moves — the target is only tracked (and ringed) — and the move is applied
+  // once on drop, so the cards glide to their new places then rather than
+  // shuffling live under the cursor. Releasing over another card swaps the two;
+  // releasing in a gap or an empty section inserts, which is what carries a card
+  // across a divider.
+  const onGridPointerDown = (event: ReactPointerEvent) => {
```
bll-kacper-wlodarczyk, line 309: **Blocking:** reorder and resize are pointer-only, and #213 makes the keyboard path an explicit acceptance criterion: *"move up / move down / hide / a span picker is a complete answer and is keyboard-reachable, which drag-and-drop on its own is not."*

As shipped there is no `onKeyDown` anywhere under `components/dashboard/`, the resize grips are `aria-hidden` with `role="presentation"`, and `stepSpan`/`stepRows` (`layouts.ts:171-178`) are exported and tested but nothing consumes them. A keyboard user can hide, add, rename, recolour and collapse — but can never reorder or resize a card. The design doc update in this PR even says "a corner handle **(and discrete steppers, for the keyboard)**", so the doc currently describes controls that don't exist.

A focusable handle on `WidgetEditCard` with ArrowUp/ArrowDown wired to `moveWidget`, plus a small span/rows stepper built on the two functions that already exist for exactly this, would close it without touching the drag engine.

frontend/src/components/dashboard/dashboard-editor.tsx
```diff
@@ -0,0 +1,577 @@
...
+        setSections((current) => moveWidget(current, armed.uid, drop.sectionUid, drop.index));
+      } else if (active && armed.kind === "section" && sectionDrop !== null) {
+        const index = sectionDrop;
+        setSections((current) => moveSection(current, armed.uid, index));
+      }
+      setDrag(null);
+      setOverUid(null);
+    };
+    window.addEventListener("pointermove", move);
+    window.addEventListener("pointerup", up);
```
bll-kacper-wlodarczyk: **Blocking:** `move`/`up` are registered for `pointermove`/`pointerup` only — there is no `pointercancel` handler anywhere in the package, and no unmount cleanup.

When the browser cancels the gesture (touch scroll takeover is the realistic case — the section grip has no `touch-action: none`, unlike the widget card's `touch-none`), `up` never runs: both listeners stay on `window`, `document.body.style.userSelect` stays `"none"`, the drag state sticks, and `widgetDrop`/`sectionDrop` keep their last value — so the **next click anywhere fires the leaked `pointerup` and applies the stale move**. Unmounting mid-drag leaks the same pair.

```ts
const finish = (endEvent: PointerEvent) => {
  window.removeEventListener("pointermove", move);
  window.removeEventListener("pointerup", finish);
  window.removeEventListener("pointercancel", finish);
  document.body.style.userSelect = "";
  const cancelled = endEvent.type === "pointercancel";
  if (!cancelled && active && /* …apply the drop as today… */) { /* … */ }
  setDrag(null);
  setOverUid(null);
};
window.addEventListener("pointermove", move);
window.addEventListener("pointerup", finish);
window.addEventListener("pointercancel", finish);
```

(Or `setPointerCapture` on the pressed element and let `lostpointercapture` centralize the teardown.) An unmount cleanup for an in-flight drag is worth adding in the same pass.

frontend/src/components/dashboard/widget-edit-card.tsx
```diff
@@ -0,0 +1,227 @@
...
+      setResizingTo({ span, rows: nextRows });
+      onResize(span, nextRows);
+    };
+    const up = () => {
+      setResizingTo(null);
+      window.removeEventListener("pointermove", move);
+      window.removeEventListener("pointerup", up);
+    };
+    window.addEventListener("pointermove", move);
+    window.addEventListener("pointerup", up);
```
bll-kacper-wlodarczyk: **Blocking (same class as the reorder engine):** no `pointercancel`, no unmount cleanup — and here the resize applies live on every `move`, so after a cancelled gesture the leaked `pointermove` keeps resizing the card on bare mouse movement until the next click. Whatever shape the fix takes over in `dashboard-editor.tsx`, this engine wants the same one.

frontend/src/components/dashboard/dashboard-editor.tsx
```diff
@@ -0,0 +1,577 @@
...
+                      entry={section.divider}
+                      dragging={drag?.kind === "section" && drag.uid === section.uid}
+                      onLabelChange={(label) => setLabel(section.uid, label)}
+                      onAccentChange={(accent) => setAccent(section.uid, accent)}
+                      onToggleCollapse={() => toggleCollapse(section.uid)}
+                      onRemove={() => setSections((current) => removeSection(current, section.uid))}
+                    />
+                  </div>
+                ) : null}
+                {section.divider?.collapsed ? null : (
```
bll-kacper-wlodarczyk, line 543: A collapsed section still renders its `[data-sec-uid]` block but no `[data-sec-grid]`, so `hitTestWidget` resolves a drop over it (or a gap beside it, via nearest-block) to `{sectionUid, index: 0}` — and the card lands inside the folded section. From the person's side it just vanishes: the drag looks like it deleted the card.

`handlePick` already guards the equivalent add path with `patchDivider(..., { collapsed: false })`; the drop path wants the same expand-on-drop, or the hit test should skip collapsed sections.

frontend/src/app/[locale]/(dashboard)/dashboard/page.tsx
```diff
@@ -74,44 +119,219 @@ export default function DashboardPage() {
...
   };
 
+  const handleSave = async (entries: StoredEntry[]) => {
+    try {
+      await save(entries);
+      setEditing(false);
+    } catch {
+      toast.error(t("edit.saveFailed"));
+    }
+  };
```
Comment (bll-kacper-wlodarczyk) [inline frontend/src/app/[locale]/(dashboard)/dashboard/page.tsx]: `handleSave` swallows its own failure, so the blank-start "save as template" flow (`dashboard-editor.tsx:558`, which awaits this after the preset create) can't tell the apply half failed. The person then gets "Could not save your dashboard" *and* "Saved «name»" for one action, the dialog closes, and nothing hints which half went wrong. Rethrowing here (or returning a boolean) lets the composite flow report one truthful outcome.

Related, same catch: a 422 from the backend's 60-entry cap also lands in this generic toast — duplicates make 61+ placements reachable in the editor, and the save is refused with no hint why.

frontend/src/components/dashboard/dashboard-preset-menu.tsx
```diff
@@ -0,0 +1,139 @@
...
+    setBusy(true);
+    try {
+      await action();
+      if (close) setOpen(false);
+    } finally {
+      setBusy(false);
+    }
+  };
+
+  const remove = async (preset: DashboardPreset) => {
```
bll-kacper-wlodarczyk: `remove` bypasses `run`, so `busy` never sets and a double-click on the trash sends two DELETEs — the second 404s and toasts "Could not delete" for a delete that succeeded. Wrap it like apply/default:

```ts
const remove = (preset: DashboardPreset) =>
  run(async () => {
    try {
      await onDelete(preset.id);
      toast.success(t("presets.deleted", { name: preset.name }));
    } catch {
      toast.error(t("presets.deleteFailed"));
    }
  }, false);
```

frontend/src/hooks/use-dashboard-layout.ts
```diff
@@ -0,0 +1,61 @@
...
+export function useDashboardLayout(): UseDashboardLayoutResult {
+  const queryClient = useQueryClient();
+  const activeOrgId = useOrgStore((state) => state.activeOrgId);
+  const queryKey = qk.dashboard.layout(activeOrgId ?? "current");
+
+  const { data, isLoading } = useQuery({ queryKey, queryFn: getLayout });
+
+  const saveMutation = useMutation({
+    mutationFn: (entries: StoredEntry[]) => putLayout(entries),
+    onSuccess: (layout) => queryClient.setQueryData<StoredLayout | null>(queryKey, layout),
```
bll-kacper-wlodarczyk: `onSuccess` reads `queryKey` from the latest render, not from when the mutation started. Save, then switch organization before the PUT resolves: the tenant cache reset clears everything, then the late `onSuccess` writes org A's layout under org B's key, and it renders as fresh data until something refetches. Snapshot the key at mutate time:

```ts
const saveMutation = useMutation({
  mutationFn: async (entries: StoredEntry[]) => ({ key: queryKey, layout: await putLayout(entries) }),
  onSuccess: ({ key, layout }) => queryClient.setQueryData<StoredLayout | null>(key, layout),
});
```

Narrow window and it's the person's own data either way, so not blocking — but cheap to close. `use-dashboard-presets` has the same shape; there the wrong-key invalidate is harmless.

backend/app/repositories/dashboard_layout.py
```diff
@@ -0,0 +1,58 @@
...
+    organization_id: UUID,
+    entries: list[dict[str, Any]],
+) -> DashboardLayout:
+    """Store the arrangement, creating the row or replacing its entries.
+
+    One row per `(user_id, organization_id)` — enforced by the unique
+    constraint — so this reads the existing row and overwrites its entries
+    rather than inserting a second.
+    """
+    existing = await get(db, user_id=user_id, organization_id=organization_id)
```
bll-kacper-wlodarczyk: Read-then-insert races with itself: two first saves in flight together (the same person in two tabs) both see no row, and the second flush hits `uq_dashboard_layout_user_org` and surfaces as a 500 — there's no global `IntegrityError` handler. You handled exactly this race for preset names with the 409 translation; the active layout deserves the same care, and Postgres can make it atomic instead of translated:

```python
from sqlalchemy.dialects.postgresql import insert

stmt = (
    insert(DashboardLayout)
    .values(user_id=user_id, organization_id=organization_id, entries=entries)
    .on_conflict_do_update(
        constraint="uq_dashboard_layout_user_org",
        set_={"entries": entries},
    )
    .returning(DashboardLayout)
)
return (await db.execute(stmt)).scalar_one()
```

backend/tests/test_dashboard_layout.py
```diff
@@ -0,0 +1,213 @@
...
+
+    await service.reset(user_id=uuid4(), organization_id=uuid4())
+
+    delete.assert_not_called()
+
+
+# --- write schema ---------------------------------------------------------
+
+
+def test_a_valid_arrangement_is_accepted() -> None:
```
bll-kacper-wlodarczyk, line 151: The write schema's section-divider half has no test: nothing sends `kind: "section"` through a write anywhere in the backend suite. The accent validator — an invalid accent refused, `#AABBCC` lowercased, a named preset accepted — and the divider's acceptance alongside widgets are all unexercised, and since `app/schemas/` isn't under the coverage gate, nothing else will notice. Three small cases in this section would pin the schema docstring's own claims.

frontend/src/hooks/use-dashboard-layout.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+});
+
+describe("useDashboardLayout", () => {
+  it("reports no arrangement when none is saved", async () => {
+    vi.mocked(api.getLayout).mockResolvedValue(null);
+    const result = await hook();
+    expect(result.current.storedEntries).toBeNull();
+  });
+
+  it("exposes the saved arrangement's entries", async () => {
```
bll-kacper-wlodarczyk, line 41: nit: the matrix here misses the one cell the design leans on — `{ entries: [] }` must come back as `[]`, not `null` (the truthy-object check at `use-dashboard-layout.ts:56` is what keeps "hid everything" distinct from "no preference"). One three-line test pins it against a future `data?.entries?.length ? … : null` "simplification".

frontend/src/components/dashboard/save-preset-dialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const submit = async (event: FormEvent) => {
+    event.preventDefault();
+    const trimmed = name.trim();
+    if (!trimmed) return;
+    setBusy(true);
+    try {
+      await onSave(trimmed);
+      toast.success(t("presets.saved", { name: trimmed }));
+      setName("");
```
bll-kacper-wlodarczyk: nit: `name` only resets on a successful save, so cancelling and reopening shows the stale name. `add-widget-dialog.tsx` resets its state on close — same move here.

frontend/src/app/[locale]/(dashboard)/dashboard/page.tsx
```diff
@@ -1,44 +1,87 @@
...
     searchParams.get("sections"),
   );
+  const [editing, setEditing] = useState(false);
+  // Whether edit mode opened on a blank grid ("New blank layout") rather than on
+  // the current arrangement. Composing from nothing and saving under a name is
+  // how a person builds a preset that shares no cards with what they see today.
+  const [startBlank, setStartBlank] = useState(false);
+  // Ephemeral collapse of a section on the page — the divider's saved `collapsed`
+  // is the initial, and a click here folds or unfolds it for this visit only.
+  const [collapseOverrides, setCollapseOverrides] = useState<Record<string, boolean>>({});
```
Comment (bll-kacper-wlodarczyk) [inline frontend/src/app/[locale]/(dashboard)/dashboard/page.tsx:94]: nit: these overrides are keyed by positional section ids (`custom-1`, `custom-2`…), so saving an edit or applying a preset re-numbers the sections and a leftover override folds a *different* section for the rest of the visit. Clearing the map when `storedEntries` changes would do.

frontend/src/components/dashboard/use-flip.ts
```diff
@@ -0,0 +1,86 @@
...
+      const dx = previous.left - left;
+      const dy = previous.top - top;
+      if (dx === 0 && dy === 0) continue;
+
+      child.style.transition = "none";
+      child.style.transform = `translate(${dx}px, ${dy}px)`;
+      // Commit the inverted offset before scheduling the release, so the browser
+      // has a "from" frame to animate out of rather than collapsing both writes.
+      void child.offsetWidth;
+      requestAnimationFrame(() => {
```
bll-kacper-wlodarczyk: nit: the pending rAF is never cancelled, so two signature changes inside one frame let the first run's callback clear the second run's inverted transform — that move snaps instead of gliding. Store the handle and cancel it on re-run. Cosmetic, but a one-liner.

frontend/src/lib/dashboard/registry.test.ts
```diff
@@ -80,6 +88,12 @@ describe("the widget catalog", () => {
     }
   });
 
+  it("every widget's default height is a class the grid knows", () => {
+    for (const id of WIDGET_IDS) {
+      expect(ROW_CLASS[WIDGETS[id].defaultRows], id).toBeTruthy();
+    }
+  });
+
   it("every see-all destination is a route that exists", () => {
```
bll-kacper-wlodarczyk, line 103: Turning #213's "a see-all destination pointing at a page that already exists" checkbox into an assertion against `ROUTES` means a renamed route fails here naming the widget, instead of shipping a dead link in the catalog. Nice.

frontend/src/app/[locale]/(dashboard)/dashboard/dashboard.integration.test.tsx
```diff
@@ -473,3 +496,241 @@ describe("the member's dashboard", () => {
     expect(named).toHaveLength(0);
   });
 });
+
+describe("the arrangeable dashboard", () => {
+  it("drops a widget a stored layout names but the caller may no longer see", async () => {
```
Comment (bll-kacper-wlodarczyk) [inline frontend/src/app/[locale]/(dashboard)/dashboard/dashboard.integration.test.tsx:515]: This is the test the whole feature rests on, and it asserts the strong version: not just dropped from the DOM, but *never fetched*. Exactly right.

backend/tests/test_dashboard_registry.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def _union_members(source: str, type_name: str) -> frozenset[str]:
+    """The string-literal members of an exported `type X = "a" | "b" | ...` union."""
+    match = re.search(rf"export type {type_name} =(.*?);", source, flags=re.DOTALL)
+    assert match, f"could not find `export type {type_name}` in {REGISTRY_PATH.name}"
+    return frozenset(re.findall(r'"([^"]+)"', match.group(1)))
+
+
+def test_the_widget_ids_match_the_frontend_registry() -> None:
```
bll-kacper-wlodarczyk, line 34: Reading the union types out of the registry the frontend actually ships and asserting the mirror in both directions is the same move as the tool-catalog parity test, and it's what makes a backend copy of `WIDGET_IDS` safe to have at all. Good.

bll-kacper-wlodarczyk (changes requested):
This is a lot of feature landed very close to how the design note settled it, and the hard part is genuinely right: the gate discipline is airtight (resolve → sanitize → gate last, on the page, the add catalog and the preset apply, with integration tests asserting the refusal at the network layer — the forbidden widget is never even fetched), the write-strict/read-forgiving split is carried end to end, and the backend layering, tenant predicates and their tests are exactly the house pattern.

Two things block it. First, #213 makes keyboard reachability an explicit acceptance criterion, and as shipped reorder and resize are pointer-only — `stepSpan`/`stepRows` exist and are tested but nothing consumes them, and the design doc's "(and discrete steppers, for the keyboard)" describes something that didn't ship. Second, both pointer engines register `pointermove`/`pointerup` but never `pointercancel`, so a cancelled gesture leaks the listeners and the next unrelated click applies a stale move. Details and fixes inline; the rest is suggestions and nits.

One housekeeping note: the `github-code-quality` threads about unused `revision`/`down_revision` globals in the migrations are false positives — Alembic reads those module attributes — safe to resolve with that one-liner.

frontend/src/hooks/use-dashboard-layout.ts
```diff
@@ -0,0 +1,67 @@
...
+      key: queryKey,
+      layout: await putLayout(entries),
+    }),
+    onSuccess: ({ key, layout }) => queryClient.setQueryData<StoredLayout | null>(key, layout),
+  });
+
+  const resetMutation = useMutation({
+    mutationFn: deleteLayout,
+    onSuccess: () => queryClient.setQueryData<StoredLayout | null>(queryKey, null),
+  });
```
DEENUU1, line 55: Same race as the save mutation you just fixed, one mutation down: reset, then switch organization before the DELETE resolves, and this `onSuccess` reads the *new* render's `queryKey` — writing `null` under org B's key, so org B's saved arrangement renders as "no preference" until something refetches. Snapshot the key at mutate time, same as `saveMutation` above:

```suggestion
  const resetMutation = useMutation({
    mutationFn: async () => {
      const key = queryKey;
      await deleteLayout();
      return key;
    },
    onSuccess: (key) => queryClient.setQueryData<StoredLayout | null>(key, null),
  });
```

frontend/src/components/dashboard/editor-model.ts
```diff
@@ -0,0 +1,227 @@
...
+  widgetUid: string,
+  direction: -1 | 1,
+): EditorSection[] {
+  const order = sections.flatMap((section) => section.widgets.map((widget) => widget.uid));
+  const at = order.indexOf(widgetUid);
+  if (at < 0) return sections;
+  const neighbour = order[at + direction];
+  if (neighbour === undefined) return sections;
+  return swapWidgets(sections, widgetUid, neighbour);
+}
```
DEENUU1, line 150: The keyboard move can land a card in a *collapsed* section — the same vanishing-card case the drop hit-test now skips and `handlePick` guards with `collapsed: false`. Concretely: `s0 [a, b]`, `s1` collapsed with `[c]`; move-down on `b` swaps it with `c`, so `b` disappears into the folded band and a card the person didn't touch pops out where it was.

Since the add path answers this by unfolding the target, I'd do the same here — the card stays visible wherever it goes:

```suggestion
export function moveWidgetBy(
  sections: EditorSection[],
  widgetUid: string,
  direction: -1 | 1,
): EditorSection[] {
  const order = sections.flatMap((section) => section.widgets.map((widget) => widget.uid));
  const at = order.indexOf(widgetUid);
  if (at < 0) return sections;
  const neighbour = order[at + direction];
  if (neighbour === undefined) return sections;
  const swapped = swapWidgets(sections, widgetUid, neighbour);
  const landing = swapped.find((section) =>
    section.widgets.some((widget) => widget.uid === widgetUid),
  );
  return landing?.divider?.collapsed
    ? patchDivider(swapped, landing.uid, { collapsed: false })
    : swapped;
}
```

This module is inside the coverage gate, so the new branch needs a test anyway — "a step into a collapsed section unfolds it" next to the carry-across-a-divider case, and a clause in the docstring.

frontend/src/components/dashboard/widget-edit-card.tsx
```diff
@@ -0,0 +1,320 @@
...
+      setResizingTo({ span, rows: nextRows });
+      onResize(span, nextRows);
+    };
+    const finish = () => {
+      setResizingTo(null);
+      window.removeEventListener("pointermove", move);
+      window.removeEventListener("pointerup", finish);
+      window.removeEventListener("pointercancel", finish);
+      resizeTeardown.current = null;
+    };
```
DEENUU1: nit: the cancel path here keeps whatever size the moves already applied, while the drag engine discards its move on cancel — a touch scroll taking over mid-grip leaves the card at a half-resized size the person never chose. `entry.span` and `rows` in this closure are the press-time values, so reverting is one line:

```suggestion
    const finish = (endEvent?: PointerEvent) => {
      if (endEvent?.type === "pointercancel") onResize(entry.span, rows);
      setResizingTo(null);
      window.removeEventListener("pointermove", move);
      window.removeEventListener("pointerup", finish);
      window.removeEventListener("pointercancel", finish);
      resizeTeardown.current = null;
    };
```

(The optional parameter keeps the unmount teardown call — `finish()` with no event — from reverting, which is right: the component is gone.)

frontend/src/components/dashboard/section-divider-card.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  return (
+    <div
+      className={cn(
+        "dash-editable border-border bg-muted/30 group flex h-full flex-wrap items-center gap-2 rounded-xl border border-dashed px-2 py-1.5 transition",
+        dragging && "dash-dragging",
+      )}
+    >
+      <span
+        data-section-grip
```
DEENUU1, line 66: Non-blocking, your call whether here or as a follow-up: sections themselves still reorder only by pointer — this grip is `aria-hidden`, and there's no keyboard path to move a whole band. #213's keyboard criterion names widgets, and sections are an extra this PR added, so I'm not holding the branch on it — but the same reasoning applies, and the pieces exist (`moveSection` is tested). A move up/down pair next to the collapse toggle would close it; if you'd rather not grow this branch further, file the follow-up issue and link it here.

DEENUU1 (commented):
Re-review after the fixes. Both blockers are properly closed: the keyboard toolbar is exactly the shape #213 asked for — real buttons on the existing `stepSpan`/`stepRows`, translated labels in both languages, `moveWidgetBy` tested including the carry across a divider — and all three pointer engines now handle `pointercancel` and tear down on unmount. The atomic upsert is also the right fix, and asserting the statement shape in the unit test plus the replace path against a real database is exactly how I'd want it pinned. CI is green across the board, e2e included.

What's left are two small leftovers of the same classes you just fixed, found by sweeping the siblings: the reset mutation has the same stale-query-key race the save mutation was cured of, and the keyboard move can land a card in a collapsed section — the same vanishing-card case the drop hit-test and the add path both guard against. Both are ~5-line fixes, suggestions inline. Plus two nits. Fix those two and this is done from my side.

backend/tests/test_dashboard_layout.py
```diff
@@ -0,0 +1,218 @@
...
+    session = _RecordingSession(result)
+
+    returned = await dashboard_layout_repo.upsert(
+        session, user_id=user_id, organization_id=org_id, entries=[{"widget": "runs", "span": "s8"}]
+    )
+
+    assert returned is stored
+    assert session.added == []  # no ORM read-then-add; the database resolves the conflict
+    compiled = str(session.statements[-1].compile(dialect=postgresql.dialect()))
+    assert "ON CONFLICT ON CONSTRAINT uq_dashboard_layout_user_org DO UPDATE" in compiled
```
DEENUU1, line 93: Asserting the compiled statement shape — not just the returned row — is what makes this test worth having: a future refactor back to read-then-insert fails here by name, instead of passing until two tabs race. Nice.

frontend/src/components/dashboard/editor-model.ts
```diff
@@ -0,0 +1,257 @@
...
+  const section = index < 0 ? undefined : sections[index];
+  if (!section) return sections;
+  const next = sections.slice();
+  next.splice(index, 1);
+  if (section.widgets.length === 0) return next;
+  const previous = next[index - 1];
+  if (previous) {
+    next[index - 1] = { ...previous, widgets: [...previous.widgets, ...section.widgets] };
+    return next;
+  }
```
DEENUU1, line 271: Same vanishing-card class as the two you just fixed, third site: remove a heading whose *previous* section is collapsed, and the merged cards land in the folded band — a collapsed section renders no `[data-sec-grid]` (`dashboard-editor.tsx:525`), so on screen the person removed one heading and lost every card under it. The docstring's "removing a heading never removes cards" is true in the model and false on the screen.

Same answer as `moveWidgetBy` and the add path — unfold the receiver:

```suggestion
  if (previous) {
    next[index - 1] = {
      ...previous,
      divider: previous.divider?.collapsed
        ? { ...previous.divider, collapsed: false }
        : previous.divider,
      widgets: [...previous.widgets, ...section.widgets],
    };
    return next;
  }
```

This module is coverage-gated, so the branch needs a test either way — "removing a heading above a collapsed section unfolds it" next to the existing merge cases. I flagged this class twice and missed this site both times, so no blocking on it — fix here or file the follow-up and link it.

frontend/src/components/dashboard/editor-model.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      { divider: "Two", widgets: ["agents"] },
+      { divider: "One", widgets: ["spend"] },
+    ]);
+  });
+
+  it("is a no-op past the ends, against the pinned lead, and for an unknown section", () => {
+    const before = build();
+    expect(moveSectionBy(before, "s1", -1)).toBe(before);
+    expect(moveSectionBy(before, "s2", 1)).toBe(before);
+    expect(moveSectionBy(before, "lead", 1)).toBe(before);
```
DEENUU1, line 248: The pinned-lead no-op case is the one I'd have forgotten — `moveSectionBy(before, "lead", 1)` returning the same reference, not just an equal shape, is what lets the editor skip a re-render on a dead keypress. Good instinct testing identity here.

DEENUU1 (dismissed):
Third round, and everything from the last one is closed properly — I verified each fix in context at head, not just the patch. The reset mutation snapshots its key exactly like the save above it; `moveWidgetBy` unfolds a collapsed landing section with the test and docstring clause to pin it; the resize cancel reverts to the press-time size (and correctly *not* on pointerup or the unmount teardown); and the section reorder went beyond what I asked — `moveSectionBy` handles the pinned headingless lead, returns identity on every no-op, and the tests cover all of it in both directions. CI is green across the board.

Approving. One thing left, and it's on me as much as anyone — sweeping the vanishing-card class one more time turned up a third sibling I missed in both earlier rounds: `removeSection` merging cards into a collapsed previous section. Suggestion inline; it's the same five-line shape as the two you just fixed, so land it here or as a tiny follow-up, your call.

backend/alembic/versions/0028_dashboard_layout.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from alembic import op
+
+revision: str = "0028_dashboard_layout"
```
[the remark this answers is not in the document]
DEENUU1: Already adjudicated — `docs/code-review.md`, "Five findings already adjudicated" (`py/unused-global-variable`). Alembic reads `revision`, `down_revision`, `branch_labels` and `depends_on` off the module by name; nothing in the file uses them, which is what the query sees, and deleting one breaks the chain. Every revision in `backend/alembic/versions/` carries all four, so this recurs once per migration. The other seven threads on this PR are the same finding.

PR #537, opened 2026-08-10
feat(agents): heartbeat & agent triggers

OchnikBartek:
## Summary

Heartbeat & agent triggers for #44 — an agent can now run itself with nobody at the
keyboard. A trigger is a row beside the agent (modelled on `AgentExposure`, deliberately
outside the portable spec), a once-a-minute heartbeat fires the due ones through the
ordinary run path, and a fired run runs **as the trigger's creator** with membership
re-resolved each fire.

The branch was cut plan-first (`docs/design/heartbeat-triggers-plan.md`) and now carries
the full implementation: both **schedule** triggers (interval **and** cron) and **event**
triggers (GitHub / email / LinkedIn / webhook), an in-place cadence editor, optional human
titles, and five UI surfaces.

## What it introduces

**Two kinds of trigger, one table** (`agent_triggers`, `AgentTrigger`)

- **Schedule** (`ScheduleKind` = `interval` | `cron`) — an interval floor of 60s, or a
  crontab expression validated with `croniter`. `next_fire_at` is computed on write and
  advanced under the heartbeat's lock.
- **Event** (`EventSource` = `github` | `email` | `linkedin` | `webhook`) — a per-source
  `event_config` filter and a signing secret sealed through the one vault
  (`event_secret_encrypted` + `secret_key_version`), exactly as a channel bot stores its
  secret. An event row has a null `next_fire_at`; nothing is due until an event lands.
- A `trigger_type` discriminator and a single shape CHECK (`ck_trigger_shape`) keep "what
  makes this due" to exactly one answer per row.

**The cadence engine** — a trigger's cadence can be **changed in place** (interval↔cron, a
new interval, a new expression) without delete-and-recreate. `_resolve_cadence` infers the
kind from the fields sent, clears the opposite field, and refuses an unschedulable edit as
a 422 naming the field rather than a 500 from the shape CHECK.

**A human title** (`name`, migration `0024`) — a trigger is listed by an optional title,
falling back to the agent's name, so two schedules on one agent no longer read identically.

**`RunSurface.SCHEDULE`** — a fired run is stamped `schedule`, a first-class surface in run
history (`?surface=schedule`) beside web/api/slack/embed. Triggered runs stay in Runs as
well as the new Scheduled view.

**The heartbeat** — `check_agent_triggers_flow` (`IntervalSchedule(60)`) claims due
triggers `FOR UPDATE SKIP LOCKED`, advances `next_fire_at` under the lock, and submits each
[...]

Issue #44, opened 2026-08-01
Agent Schedules & Triggers: set up and manage unattended runs from the UI


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576806789a8f484571dbb35) · Priority: **High** · Category: Development

> **The plan is done.** The Notion card asked to plan the feature and contact @DEENUU1 first; that plan is written and approved, and this issue now describes the **full feature end to end** rather than only the plan. The engine has shipped (`AgentTrigger`, the heartbeat, `fire()` — PR #537); what remains is turning it into a first-class, UI-driven feature and filling the two gaps below. All of it lands under this one issue, on `feat/agent-triggers` / PR #537.

## The brief

Verbatim in substance: today an agent only runs when somebody writes to it. Add other triggers — **every X minutes**, **when an email arrives**, and so on.

## What already exists

More than the card assumes, and building without reading it means reinventing it:

- **`RunSurface.SCHEDULE` is already a value** on the run row. The concept of a run nobody typed into is modelled; the engine (PR #537) now produces one.
- **Prefect is already the background runtime** — `app/worker/`, with `rag_tasks.py:155` running a scheduled sync flow today. A periodic trigger is a Prefect deployment, not a new scheduler; the heartbeat `check_agent_triggers_flow` already claims due triggers.
- **`app/core/background.py`** is the in-process spawn/drain helper, correct and tested. Note that **nothing calls `drain()`** (#11) — a scheduled run started in the API process would be dropped on every deploy; the trigger path runs in the worker for this reason.
- **`AgentRunnerService.prepare/finish`** is the run path a trigger has to go through, so budgets, approvals and accounting apply identically.

## The design questions — and how they are answered

The five questions the plan had to answer, and the decisions this implementation ships:

**1. Who does a triggered run run *as*?** The member who created the trigger. `AuthContext` requires a subject for anything touching a resource grant, an audit entry or a budget — `subject_id` raises loudly when there is none, deliberately. No invented fallback user; the creator is the subject and every mutation gates on **`agents:run` on that agent**, resolved per-resource in the service.

[...]

PR #589, opened 2026-08-11
fix(triggers): record a model-errored fire instead of failing the flow

### LINKED ISSUE #588: Scheduled trigger fire() drops last_run_id and fails the flow on a model error
## What breaks

When a scheduled trigger fires and the agent's run fails on a **model/provider
error** (bad key, provider 5xx, timeout — anything `AgentRunnerService._run`
re-raises), `AgentTriggerService.fire` lets that exception propagate. The run row
itself *is* recorded (committed as `failed` in `_run`'s `finally`, before the
re-raise), but `fire` never reaches `trigger.last_run_id = run.id`.

Two consequences, both on a reachable path:

1. **`trigger.last_run_id` is left stale/NULL** even though a failed run for that
   trigger is sitting in Activity — the trigger's "last run" pointer lies.
2. **`run_scheduled_trigger_flow` is marked Failed** in Prefect for a run that was
   actually recorded and handled. `retries=0`, so there is no retry storm, but the
   flow-run history shows a failure the system already accounted for.

Budget and authz are **not** affected: `BudgetExceeded` is caught inside `_run`
(recorded, not re-raised) so `execute` returns normally and `last_run_id` is set;
`NotFoundError`/`AuthorizationError` correctly disable the trigger. Only the
model/provider-error path diverges.

1. A trigger's agent resolves to a model profile whose provider call fails
   (invalid key → 401, provider outage → 5xx, timeout).
2. The heartbeat (`check_agent_triggers_flow`) claims the trigger and dispatches
   `run_scheduled_trigger_flow`.
3. `fire` → `runner.execute(...)` → `_run` commits the `failed` run, then re-raises
   the provider error.
4. `fire` only catches `(NotFoundError, AuthorizationError)`, so the error escapes:
   `trigger.last_run_id` is never updated and the flow-run fails.

## Where

`backend/app/services/agent_trigger.py:249-263` — the `try/except` around
`runner.execute(...)` catches only `(NotFoundError, AuthorizationError)`; the
`trigger.last_run_id = run.id` on line 263 is unreachable when `execute` raises.

## How I found it

Verified live on an isolated stack: the heartbeat fired two triggers whose agent
had an invalid provider key. Two runs were recorded (`surface=schedule`,
`status=failed`, visible in Activity), but both triggers' `last_run_id` stayed
`NULL` and the two `run-scheduled-trigger` flow-runs were marked Failed.

**Pre-existing in the #44 feature branch (`feat/agent-triggers`, PR #537), found
while verifying #44 end-to-end.** Not caused by any change outside that branch.

## How you'd know it's fixed

A trigger whose fire hits a provider error:

- records the run (unchanged), **and** updates `trigger.last_run_id` to point at it;
- its `run_scheduled_trigger_flow` completes rather than failing;
- the trigger stays active for its next interval (a provider blip is not a reason to
  disable — only a withdrawn grant is).

Regression test covering the model-error path (the platform layer is at 100%, so the
new branch in `fire` must be covered).

Refs #44.

OchnikBartek:
## What & why

A scheduled trigger whose agent run fails on the **model/provider** (a 5xx, a
revoked key, a timeout) let that error propagate out of `AgentTriggerService.fire`,
because the `try/except` around `runner.execute` only caught
`NotFoundError`/`AuthorizationError`. `_run` commits the run row as `failed` before
re-raising, so the run *was* recorded in Activity — but:

- `trigger.last_run_id` never advanced past its previous value, and
- `run_scheduled_trigger_flow` was marked **Failed** for a run the system had
  already accounted for (`retries=0`, so no retry storm — but a false failure in the
  flow history).

`last_run_id` is not cosmetic: it is what the **no-overlap guard** in
`agent_trigger_repo.claim_due` joins against to skip a trigger whose previous run has
not reached a terminal status. A stale pointer means the trigger's own history lies
about its last run.

## The fix

`fire` now treats a re-raised run failure the way it already treats a budget stop —
the run is recorded, so it belongs to the schedule's history, not to a flow crash:

- recover the run from the trigger's one run-log conversation
  (`agent_run_repo.latest_run_for_conversation`), stamp `last_run_id`, and return
  **without raising** — a provider blip is neither a refusal nor a reason to disable
  or to retry the same outage against the same money;
- an error that strikes *before* a run row exists (a spec that no longer builds, a
  model profile deleted since publish) stamps nothing and still does not raise.

`execute` is untouched, so the interactive `POST /agents/{id}/run` still surfaces the
error as a 500 — only the unattended path changes.

## How it was verified

- **Live, before the fix**, on an isolated stack: the heartbeat fired two triggers
  whose agent had an invalid key. Both runs landed `surface=schedule` `failed` in
  Activity, both triggers' `last_run_id` stayed `NULL`, and both
  `run-scheduled-trigger` flow-runs were Failed — the exact symptom.
- **Unit** (`tests/test_agent_triggers.py`): the recover-and-stamp path (run
  recovered, `last_run_id` set, not disabled, not raised) and the before-any-run path
  (nothing to stamp, still not raised / not disabled).
- **Integration** (`tests/integration/test_run_history_filters.py`): the new
  `latest_run_for_conversation` lookup — newest-first, tenant-scoped, empty case.
- `app/services/agent_trigger.py` stays at **100%**; `make test` green locally.

## Base branch

Stacked on `feat/agent-triggers` (**#537**, still draft): `fire()` and the trigger
service do not exist on `main`, so this targets the feature branch and rides in with
it. `docs/governance.md` updated to note the model-error outcome alongside the budget
and authz ones.

Closes #588
Refs #44


backend/app/services/agent_trigger.py
```diff
@@ -259,6 +267,22 @@ async def fire(self, trigger_id: UUID) -> None:
...
             return
+        except Exception:
+            # The run failed on something the runner re-raises instead of recording
+            # as a terminal status - a provider 5xx, a revoked key, a timeout. `_run`
+            # commits the row as `failed` before re-raising, so the failure is in
+            # Activity; letting it propagate would only fail the Prefect flow for a
+            # run already accounted for and retry the same outage. Recover the row
+            # from the trigger's own conversation, stamp it, and return.
+            logger.exception("trigger_fire_run_errored", extra={"trigger_id": str(trigger.id)})
+            run = await agent_run_repo.latest_run_for_conversation(
```
DEENUU1, line 1122: **Blocking.** The comment's premise — "`_run` commits the row as `failed` before re-raising" — holds for errors raised inside the runner's execute loop, but `except Exception` catches more than that. `create_run` flushes the row as `running` partway through `_assemble` (`agent_runner.py:1458`), and `workspaces.open`, `_delegation_runtime` and `build_agent` all run *after* it. An error from any of those (a workspace backend hiccup, a delegate pin that no longer resolves) escapes before `_run`'s `finally` ever runs, so nothing marks the row terminal and nothing commits it.

This recovery query then runs in the same transaction, sees the flushed `running` row, and stamps `last_run_id` at it. `fire` returns cleanly, `get_worker_db_context` commits on the clean exit — and now `last_run_id` points at a run nothing will ever finish. `claim_due`'s no-overlap guard treats `running` as non-terminal, so it skips this trigger on every tick from then on: the trigger is still active, no audit entry, and it never fires again. Before this change that path rolled back and the next interval retried, which was the better outcome.

A non-terminal row recovered here can only be this fire's own orphan — `claim_due` proved the previous run terminal at claim time, and an in-flight run's row isn't visible to other transactions until its own commit lands — so it's safe to settle it before stamping:

```python
run = await agent_run_repo.latest_run_for_conversation(
    self.db, conversation_id, organization_id=ctx.organization_id
)
if run is None:
    return
if run.status == RunStatus.RUNNING.value:
    # The error struck between the row's creation and the runner's own
    # failure recording, so nothing else will ever finish it - and a
    # non-terminal last_run_id blocks claim_due forever.
    run.status = RunStatus.FAILED.value
    run.ended_at = datetime.now(UTC)
    run.error = "fire failed before the runner could record the run"
```

plus a test for the assemble-time-failure path (the platform layer is at 100%, so the new branch needs one anyway).

backend/app/repositories/agent_run.py
```diff
@@ -177,6 +177,28 @@ async def get_run(db: AsyncSession, run_id: UUID, *, organization_id: UUID) -> A
...
+    A scheduled trigger appends every fire to one conversation, so this is how
+    :meth:`app.services.agent_trigger.AgentTriggerService.fire` recovers the run a
+    provider error recorded but never handed back: `_run` commits the failed row
+    and re-raises, so `execute` returns nothing to stamp `last_run_id` against.
+    """
+    result = await db.execute(
+        select(AgentRun)
+        .where(
+            AgentRun.conversation_id == conversation_id,
+            AgentRun.organization_id == organization_id,
```
DEENUU1, line 202: **Blocking.** A delegated child run is recorded with the parent's `conversation_id` (`record_delegated_run(..., conversation_id=parent.conversation_id)` in `finish`), and it lands in the same transaction as the parent's terminal write — so both rows carry the exact same `created_at`, because `func.now()` is the transaction timestamp in Postgres. Between tied rows `ORDER BY created_at DESC` is unspecified, so for an agent that delegated before hitting the provider error this can hand back a child row, and `last_run_id` ends up pointing at a delegation instead of the fire — the pointer-lies problem this PR exists to fix, and the trigger's "last run" in any UI would open a child run.

The fire is always a top-level run, so filter to those:

```suggestion
            AgentRun.organization_id == organization_id,
            AgentRun.parent_run_id.is_(None),
```

backend/tests/integration/test_run_history_filters.py
```diff
@@ -477,6 +484,44 @@ async def test_a_rating_on_another_runs_message_does_not_pull_this_one_in(self,
...
+        convo = await _conversation(db, org)
+        await _run(db, org, agent, conversation_id=convo.id, created_at=_NOW - timedelta(minutes=1))
+        newest = await _run(db, org, agent, conversation_id=convo.id, created_at=_NOW)
+
+        found = await latest_run_for_conversation(db, convo.id, organization_id=org.id)
+
+        assert found is not None
+        assert found.id == newest.id
+
+    async def test_it_will_not_reach_another_organizations_run(self, db) -> None:
```
DEENUU1, line 805: Nice — covering the tenant boundary on a lookup that's only ever called by the worker is easy to skip ("the conversation id came off our own trigger row, who'd pass a foreign one"), and this is exactly the refusal-first testing this repo asks for. The docstring saying *why* a guessed UUID matters is a good touch.

DEENUU1 (changes requested):
Right diagnosis, and I like the shape: a provider blip is treated the way a budget stop already is, the run stays in the schedule's history, and the live verification in the description is exactly the evidence this needed. The docs update keeps governance.md honest too.

Two problems in the recovery path though, both in what `latest_run_for_conversation` can hand back. First, the premise "`_run` has already committed the row as failed" doesn't hold for everything `except Exception` catches — an error between `create_run` and the runner's own try block leaves a flushed `running` row that this stamps and commits, and `claim_due`'s no-overlap guard then skips the trigger forever, silently. Second, a delegated child run shares the parent's `conversation_id` and its exact `created_at`, so the query can recover a child instead of the fire. Details and fixes inline.

backend/app/worker/tasks/trigger_tasks.py
```diff
@@ -63,9 +63,21 @@ async def check_agent_triggers_flow() -> None:
...
+        try:
+            await _dispatch(str(trigger.id))
+        except Exception:
+            # The claims already committed, so a trigger whose dispatch fails has its
+            # `next_fire_at` advanced with no run - a missed fire for that one trigger.
+            # Isolate each dispatch so a single failed `run_deployment` (a transient
+            # Prefect API error) does not abort the loop and cost the rest of the batch
+            # their fire too; the failure is logged, and the next tick fires this
+            # trigger again on its new schedule.
+            logger.exception("agent_trigger_dispatch_failed", extra={"trigger_id": str(trigger.id)})
```
DEENUU1: **Blocking.** This comment was true when the isolation landed, but the in-flight marker made it false: the claim already committed `fire_in_flight_since = now` for this trigger, and the flow that would clear it never starts when `_dispatch` raises. `claim_due` now skips the trigger until the lease lapses — so one transient Prefect 503 costs up to an hour of fires (on a one-minute interval, ~59 missed fires), not "the next tick fires this trigger again on its new schedule".

Nothing is in flight for a trigger whose dispatch failed, so release the marker here:

```python
except Exception:
    logger.exception("agent_trigger_dispatch_failed", extra={"trigger_id": str(trigger.id)})
    async with get_worker_db_context() as db:
        await AgentTriggerService(db).release_fire_marker(trigger.id)
```

with a small service method that loads the row and nulls `fire_in_flight_since` (a fresh context per failure is fine — this is the rare path). If the release itself fails, the lease still bounds the damage, same as today. Plus a test: the middle trigger's dispatch fails, its marker is released, the other two are untouched.

backend/app/services/agent_trigger.py
```diff
@@ -536,12 +541,36 @@ async def fire(self, trigger_id: UUID, *, event_context: str | None = None) -> N
...
-
+        try:
+            await self._fire_loaded(trigger, event_context=event_context)
+        finally:
+            # Release the in-flight marker the claim set, however this fire ended - a
+            # completed run, a recorded failure, a disable, or a skip. The next tick's
+            # no-overlap and last_run_id guards still apply; leaving it set would park
+            # the trigger until the lease lapses.
+            trigger.fire_in_flight_since = None
+            await self.db.flush()
```
DEENUU1: **Blocking.** `fire` has three callers, and this `finally` releases a marker only one of them sets. `run_now` on a trigger whose scheduled run is still in flight — a click that gets *more* likely the slower the run, because the schedule looks stuck — runs its one extra fire and then clears the marker the *claim* set. After that, `last_run_id` names run-now's terminal run, so the next tick sees no marker and no non-terminal last run, and fires the schedule on top of the still-executing scheduled run — the exact self-overlap window `0025_trigger_fire_in_flight` exists to close, reopened through the run-now door. (The event-fire caller in `trigger_event.py` is harmless: an event trigger is never claimed, so its marker is always null.)

The release belongs to the caller that owns the claim, so make it opt-in:

```python
async def fire(
    self, trigger_id: UUID, *, event_context: str | None = None, release_claim: bool = False
) -> None:
```

clear in the `finally` only under `release_claim`, and pass `release_claim=True` from `run_scheduled_trigger_flow` alone. `test_a_fire_clears_the_in_flight_marker_when_the_run_completes` then fires with the flag, plus one test that a run-now fire leaves a set marker alone.

backend/tests/integration/test_agent_trigger_schema.py
```diff
@@ -220,6 +220,26 @@ async def test_a_trigger_whose_last_run_is_unfinished_is_skipped(self, db):
         claimed = await agent_trigger_repo.claim_due(db, now=datetime.now(UTC))
         assert [t.id for t in claimed] == [blocked.id]
 
+    async def test_a_trigger_with_a_fire_in_flight_marker_is_skipped_until_the_lease(self, db):
```
DEENUU1, line 273: Good test — it proves both halves of the lease in one place: the marker holds the trigger out of the claim, and a marker older than the lease frees it. The un-wedge assertion is the one that matters most here, and it would have been easy to skip.

DEENUU1 (changes requested):
Both blockers from my last review are fixed exactly right — the orphan settle with its own unit test, and `parent_run_id IS NULL` with an integration test that constructs the `created_at` tie instead of hoping for it. The in-flight marker is the right design too: set under the claim's own UPDATE, lease-bounded so a dead worker can't park a schedule forever.

But the two newest commits interact in two places: the marker made the dispatch-isolation comment false (a failed dispatch now parks the trigger for the lease hour, not one tick), and `fire`'s unconditional release lets a run-now clear a claim's marker and reopen the self-overlap window. Both inline, both with a concrete fix. Once those land I think this is done.

backend/app/worker/tasks/trigger_tasks.py
```diff
@@ -63,9 +63,25 @@ async def check_agent_triggers_flow() -> None:
...
+            # Prefect API error) does not abort the loop and cost the rest of the batch
+            # their fire too. The claim already committed this trigger's advanced
+            # `next_fire_at` and its `fire_in_flight_since` marker, and the run that
+            # would clear the marker never starts - so release it here, in its own
+            # transaction: nothing is in flight for a trigger whose dispatch failed,
+            # and a marker left set would hold it out of `claim_due` until the lease
+            # lapsed rather than the next tick firing it again on its new schedule.
+            logger.exception("agent_trigger_dispatch_failed", extra={"trigger_id": str(trigger.id)})
+            async with get_worker_db_context() as db:
+                await AgentTriggerService(db).release_fire_marker(trigger.id)
```
DEENUU1: This is the shape I suggested, so the gap here is mine as much as anyone's: an exception raised inside an `except` block propagates, so if this release round-trip fails — the session open, the `get_by_id`, or the commit on context exit — the loop aborts and every trigger after this one in the batch neither dispatches nor releases. Each of those is then parked for up to the one-hour lease with `next_fire_at` already advanced. And the failure is correlated, not independent: a flaky tick is precisely when many dispatches fail, so this path runs many times exactly when the infrastructure is most likely to fail it once.

My "the lease still bounds the damage" from last round was true for the failed trigger, not for the batch — the batch-isolation invariant `2dba7bb` established needs the release contained too:

```suggestion
            try:
                async with get_worker_db_context() as db:
                    await AgentTriggerService(db).release_fire_marker(trigger.id)
            except Exception:
                # The lease bounds this one trigger's damage; the rest of the
                # batch must still dispatch - the reason this loop isolates at all.
                logger.exception(
                    "agent_trigger_marker_release_failed",
                    extra={"trigger_id": str(trigger.id)},
                )
```

Plus one case in `test_trigger_heartbeat.py`: the middle trigger's dispatch fails *and* its release raises — the third trigger is still submitted.

backend/tests/test_agent_triggers.py
```diff
@@ -722,6 +733,170 @@ async def test_access_withdrawn_mid_run_disables_rather_than_retries(self):
...
+            patch("app.services.agent_trigger.record_audit", new=AsyncMock()),
+            patch("app.services.agent_runner.AgentRunnerService") as runner_cls,
+        ):
+            repo.get_by_id = AsyncMock(return_value=trigger)
+            members.get = AsyncMock(return_value=MagicMock(role=OrgRoleName.OWNER))
+            runner = runner_cls.return_value
+            runner.execute = AsyncMock(return_value=("done", run))
+            await service.fire(trigger.id)
+
+        assert trigger.fire_in_flight_since == in_flight
```
DEENUU1, line 1241: This assertion is the one that actually guards the reopened-window regression — it would have been easy to write this test as "run_now still stamps `last_run_id`" and let the marker go unasserted. Checking it against the exact timestamp the claim set, not just non-null, is right too: a test that only asserted truthiness would pass if `fire` overwrote the marker with its own `now`.

DEENUU1 (commented):
Both blockers from the last round are fixed exactly as they needed to be, and I verified them against every caller: `run_now` (agent_trigger.py:442) and the event fire (trigger_event.py:30) reach `fire` with the default and leave a concurrent claim's marker standing, and only `run_scheduled_trigger_flow` passes `release_marker=True`. The failed-dispatch release runs in its own transaction, and the four new tests cover the matrix that matters — including the one asserting the marker *stands*, which is the actual regression guard.

One thing left, and it's partly on me: the release call I suggested for the dispatch-failure path is unguarded inside the `except`, so a DB blip during release aborts the loop and costs the rest of the batch — the invariant the isolation commit exists to protect. Inline, with the fix. It's a compound-failure path, so I'm not blocking on it, but I'd land it before this merges.

DEENUU1:
@codex review

DEENUU1:
@codex review

DEENUU1:
Closed itself as merged: this branch was the base, and these commits are now in `feat/agent-triggers` (#537), which `main` has been merged into. The work reviewed here ships in #537 rather than separately — one pull request instead of a stack.

Two notes for anyone following the history here. The migrations are renumbered: `0042_trigger_fire_in_flight` is now `0047_trigger_fire_in_flight`, because both branches had numbered 0037–0041 independently and the merged tree had two heads. And a local database with the old chain applied names a revision that no longer exists, so recreate it rather than trying to move it.

Issue #7, opened 2026-08-01
Actually apply the rate limiter


DEENUU1:
**Severity: High** · AUD-007

## What's wrong

`app.state.limiter` is set and a `RateLimitExceeded` handler registered, but **`SlowAPIMiddleware` is never added** — `main.py` adds only `RequestIDMiddleware`, `CORSMiddleware` and `SessionMiddleware`. In slowapi, `default_limits` are evaluated exclusively inside the middleware; the alternative path, `@limiter.limit(...)` on a handler, has zero call sites. `rate_limit_low` / `_medium` / `_high` are defined and imported nowhere.

Confirmed: `limiter.limit` has three occurrences, all inside `rate_limit.py` itself; `add_middleware` has three, all in `main.py`.

## Where

- `backend/app/main.py:249-254`
- `backend/app/core/rate_limit.py:32-50`

## How it fails

`RATE_LIMIT_REQUESTS=100` enforces nothing. `POST /api/v1/auth/login`, register, password reset, `GET /api/v1/embed/{key}/config`, `widget.js` and all three webhook receivers are unmetered — brute-force and cost-amplification are free.

`docs/governance.md:156` states *"There is deployment-level rate limiting on the API"*; the bundled `nginx/nginx.conf` has no `limit_req` either, so the claim is false at both layers.

The only test is `tests/test_core.py:104` `test_limiter_exists`, which asserts `limiter is not None` — a test that passes whether or not the limiter does anything, which is what happened.

## Fix

- `app.add_middleware(SlowAPIMiddleware)` in `create_app`
- `Limiter(..., storage_uri=settings.REDIS_URL)` — the default `memory://` is per-process, so with N uvicorn workers the effective limit is N× the configured one and resets on every deploy
- Run uvicorn with `--proxy-headers --forwarded-allow-ips=<nginx>`: `key_func=get_remote_address` reads `request.client.host`, so behind the bundled nginx every request currently shares one bucket
- Decide which routes need the tighter tiers

## Acceptance criteria

- [ ] The limiter is on the middleware stack and backed by Redis
- [ ] `/auth/*` and `/embed/*` carry explicit tighter limits
- [ ] Regression test: the N+1st request in a window returns 429
- [ ] Delete `test_limiter_exists`

DEENUU1:
Worth reading #155 before starting this. Two things it turned up:

[...]

PR #599, opened 2026-08-11
fix(channels): answer a redelivered message once

### LINKED ISSUE #167: A retried channel delivery is a second agent run: nothing dedupes on message_id

DEENUU1:
## What this fixes

A channel message redelivered by the platform became a second full agent run — another model call, another spend record, and a second answer in the thread. The fast 200 the webhook routes return only prevents the slow-handler retry; a 200 lost on the wire (proxy drop, pod rotation) was never received, and the redelivery that follows is valid, signed and brand-new. Closes #167.

## What changed

- `backend/app/services/channels/dedupe.py` — new. The first delivery claims the message with one atomic `SET NX` against the shared Redis, so the claim holds across API workers. Claims live 15 minutes, which outlives every platform's retry window.
- `backend/app/services/channels/router.py` — the claim is taken at the top of `route()`, before the per-chat lock: the lock would otherwise queue the retry behind the run it duplicates and then answer again. A run that does not finish gives the claim back, under `BaseException` so a task cancelled while the pod drains counts as one.
- `backend/app/api/routes/v1/slack_webhook.py` — a request carrying `x-slack-retry-num` is logged and then processed like any other. The header is **not** a short-circuit: see below.
- `backend/app/clients/redis.py` — `set()` gains `nx` and reports whether this call wrote the key.
- `backend/app/main.py` — the lifespan hands the shared Redis client to the dedupe module (the router runs outside any request, so `request.state` cannot reach it) and withdraws it on shutdown.
- `backend/pyproject.toml` — `dedupe.py` added to the coverage and ty gates, both lists.
- `docs/channels.md` — the delivery guarantee is now stated, including how it degrades.
- `.gitignore` — `.claude/worktrees/`, so session worktrees stop showing up as untracked in the main checkout.

## Two deviations from the issue, both deliberate

- **The claim lives in `ChannelMessageRouter.route`, not `process_channel_event`.** The issue assumed all six inbound paths converge on the worker shim; in fact the three polling/stream paths (`slack.py:502`, `telegram.py:438`, `mattermost.py:576`) call the router directly and never pass through it. A claim in `process_channel_event` would have covered three paths of six. The acceptance test still asserts at the `process_channel_event` boundary, as the issue asked.
- **The key includes the chat id.** The issue's sketch keyed on `bot:platform:message_id`, but Telegram numbers messages per chat and Slack's `ts` is per channel — a bot-wide key would let one chat's message silently swallow another's.

## Policies chosen (the issue asked for them to be explicit)

The guarantee degrades open, never shut, and always with a log line: a message with no `message_id`, a module nothing configured, and a Redis that cannot be reached are all processed rather than dropped. A duplicated answer is the rarer, cheaper failure than a dropped question.

**Nothing is refused on a retry header alone**, which is a reversal of what this branch first did. The issue suggested short-circuiting a Slack request carrying `x-slack-retry-num` and the first commit here did exactly that — and it was wrong. The header says Slack is redelivering; it does not say the first attempt did any work, and `reason=http_error` means it explicitly did not: the route raised before `spawn`, so nothing was scheduled and no claim was ever taken. A transient database error in `find_active` was enough to lose a message that way — 500, redelivery, 200, and a log line reading like a success. The header is logged; the claim decides, because it is the only thing that knows whether the first delivery got far enough to matter.

**A claim is given back when the run under it does not finish.** It is taken on receipt, not on completion, so without this a run that died mid-flight swallowed every redelivery of that message for fifteen minutes. Harmless on the webhook paths — the 200 has gone out, nothing is coming — but the pollers re-read what the process died on: aiogram re-fetches an unconfirmed `getUpdates` batch and Socket Mode redelivers an unacknowledged envelope.

## Testing

- `tests/test_channel_dedupe.py` — the same `IncomingMessage` delivered twice through `process_channel_event` reaches the router once; two concurrent claims resolve to exactly one winner; chats sharing a message id do not collide; the no-`message_id`, unconfigured and Redis-down policies each have a test; a failed run and a cancelled run each give the claim back, and a Redis that cannot be reached during a release is logged rather than raised over the failure already being handled.
- `tests/test_slack_webhook.py` — a retried event reaches `spawn` rather than being acknowledged and dropped, a retried `url_verification` still echoes its challenge, and the retry header does not bypass signature verification.
- `tests/test_clients.py` — the `SET NX` contract on `RedisClient.set`.
- `make check` green locally.

Not verifiable end to end yet: no channel answers anything until #10 lands (`link_code` is never minted), as the issue itself notes.

## Noticed, not fixed

- `app/services/rate_limit/storage.py` cannot import — `from app.core.cache import get_redis` names a function `app/core/cache.py` no longer defines. Recorded on #7, which already owns that package's fate: https://github.com/vstorm-co/agenticos/issues/7#issuecomment-5253190202


backend/app/api/routes/v1/slack_webhook.py
```diff
@@ -55,6 +55,21 @@ async def slack_events(
...
+    # request cannot use the header to probe which events were processed.
+    retry_num = headers.get("x-slack-retry-num")
+    if retry_num is not None:
+        logger.info(
+            "Slack redelivery acknowledged without processing: bot=%s retry=%s reason=%s",
+            bot_id,
+            retry_num,
+            headers.get("x-slack-retry-reason"),
+        )
+        return Response(status_code=200)
```
DEENUU1: This drops messages that were never answered, and does it silently.

Slack sends `x-slack-retry-num` for two reasons. `http_timeout` means it got no response in 3s; `http_error` means it got a **non-2xx**. In the second case the route raised before reaching `spawn` on line 84 - so nothing was scheduled, no claim was taken in Redis, and the message has not been processed at all. This retry is the only chance it gets, and we answer 200 and bin it. The log line reads "acknowledged without processing", which looks like a success in the logs rather than a lost question.

Concretely: a transient DB error in `find_active` on line 42 gives Slack a 500. Slack redelivers with `retry_num: 1, reason: http_error`. The database is fine by then, but we never get past line 64 - the user's message is gone and nothing says so.

The claim in `dedupe.py` does not have this problem, because it knows whether the first delivery actually got as far as claiming. That is what makes it the authority and this a shortcut that can only be wrong. The header cannot distinguish the two cases either, so there is no version of the condition that fixes it. I would keep the log and drop the branch:

```suggestion
    retry_num = headers.get("x-slack-retry-num")
    if retry_num is not None:
        logger.info(
            "Slack redelivery: bot=%s retry=%s reason=%s - the Redis claim decides "
            "whether it runs, because a retry after an http_error was never processed",
            bot_id,
            retry_num,
            headers.get("x-slack-retry-reason"),
        )
```

Worth noting the branch also sits above the `url_verification` handling on line 73. I am not certain Slack sets the retry header on a challenge request, but if it ever does we would answer an empty 200 instead of the challenge and URL verification would stay broken with nothing to point at. Removing the short-circuit removes the question too.

DEENUU1 (commented):
The Redis claim itself is the right shape, and the reasoning behind both deviations from the issue holds up - putting it in `route()` rather than `process_channel_event` is the only placement that covers the three polling paths, and the chat id belongs in the key. The `SET NX` wrapper is careful in the way that matters: the nil-reply-to-`False` conversion is the bit that gets this wrong everywhere else, and there is a test for both sides of it.

One thing needs to change before this lands: the Slack `x-slack-retry-num` short-circuit drops messages that were never processed. `retry_reason: http_error` means Slack got a non-2xx from us - the route failed before `spawn`, nothing was scheduled, no claim was taken - and line 64 answers 200 and throws the message away. It is the one place this design degrades shut, which is the opposite of what `dedupe.py`'s docstring promises. The claim already handles that case correctly on its own, so the fix is to delete the short-circuit rather than repair it.

One suggestion and one nit inline as well.

backend/app/services/channels/router.py
```diff
@@ -119,12 +120,24 @@ class ChannelMessageRouter:
...
+            logger.info(
+                "Duplicate channel delivery ignored: bot=%s platform=%s message=%s",
+                incoming.bot_id,
+                incoming.platform,
+                incoming.message_id,
+            )
+            return
         lock = _get_chat_lock(incoming.bot_id, incoming.platform_chat_id)
         async with lock:
             await self._route_inner(incoming, db)
```
DEENUU1: The claim is taken on receipt and never released, so anything that dies between here and the reply loses the message for the full 15 minutes.

For the webhook paths that is mostly academic - the 200 already went out, so no platform is going to retry anyway. For the two polling paths it is a new loss. aiogram re-fetches an unconfirmed `getUpdates` batch after a crash, and Socket Mode redelivers an envelope it saw no ack for; both of those used to be processed on the second pass and now hit a live claim in a Redis that survived the restart. That is the pod-rotation scenario `dedupe.py`'s own docstring names, landing on the shut side of "degrades open, never shut".

Cheap to close for the exception case:

```suggestion
        lock = _get_chat_lock(incoming.bot_id, incoming.platform_chat_id)
        try:
            async with lock:
                await self._route_inner(incoming, db)
        except BaseException:
            await release_delivery(incoming)
            raise
```

with a `release_delivery` in `dedupe.py` doing the `delete` and swallowing a Redis error the same way `claim_delivery` does. `BaseException` rather than `Exception` on purpose - cancellation on shutdown is exactly the case worth releasing for. A hard `SIGKILL` still strands the claim, and that is fine to leave; it is the difference between "usually recoverable" and "never".

backend/tests/test_channel_dedupe.py
```diff
@@ -0,0 +1,164 @@
...
+
+    assert await dedupe.claim_delivery(incoming) is True
+    assert await dedupe.claim_delivery(incoming) is False
+
+    # The claim expires - the keyspace tracks recent traffic, not history.
+    assert list(fake_redis.ttls.values()) == [dedupe.SEEN_TTL_SECONDS]
+
+
+async def test_two_concurrent_deliveries_exactly_one_proceeds(fake_redis: _FakeRedis) -> None:
+    """The claim is one atomic SET NX, never a get-then-set race."""
```
DEENUU1: nit: the docstring claims more than the test can show. `_FakeRedis.set` never awaits anything, so `gather` runs the two claims one after the other - there is no interleaving for a get-then-set to lose. What the test does prove is that `claim_delivery` goes through `SET NX` (a get-then-set would blow up on the fake, which has no `get`), and that is worth keeping. I would just say that instead:

```suggestion
    """The claim goes through SET NX - the fake has no `get` to race on."""
```

The real atomicity contract is the one in `test_clients.py`.

backend/app/api/routes/v1/slack_webhook.py
DEENUU1: Fixed in c7a89950. The branch no longer short-circuits: the header is logged and the request goes down the same path as any other, so the Redis claim decides. `test_a_redelivery_is_scheduled_because_the_first_attempt_may_have_failed` pins it (a retried event reaches `spawn`), and `test_a_redelivered_challenge_still_echoes_the_challenge` covers the `url_verification` ordering you flagged — cheap to keep even if Slack never sets the header there.

backend/app/services/channels/router.py
DEENUU1: Fixed in c7a89950, in the shape suggested — `release_delivery` in `dedupe.py`, best effort, swallowing a Redis error the same way `claim_delivery` does, and `BaseException` so a task cancelled while the pod drains releases too. Two tests: a failed run and a cancelled run each leave the keyspace empty.

backend/tests/test_channel_dedupe.py
DEENUU1: Fixed in c7a89950 — the docstring now says what the test actually shows and points at `test_clients.py` for the wire contract.

Issue #505, opened 2026-08-09
Approving a gated call shows no work, asks again, and records nothing in between


DEENUU1:
Approving a gated tool call shows nothing running, then asks for another approval
for a step that was never drawn — and the work in between is recorded nowhere.

## Sequence

1. Upload an `.xlsx` to an agent with a sandbox and ask what it is.
2. The agent lists `/uploads` and asks to run `python - <<'PY' … openpyxl …`. The step
   is drawn, the panel opens, the turn ends with *"This run needs approval before it
   can go further"*.
3. **Approve.** The step flips to done. Nothing else appears.
4. A second approval panel opens, carrying a *different* command
   (`zipfile`, `xml.etree`). Nothing on screen says the first one ran, and no step
   exists for the second.
5. Approve again — same thing, a third command.
6. Eventually an answer arrives, describing six worksheets nothing on screen ever read.

Three commands ran in a sandbox and spent money. The transcript accounts for one.

## Two defects, one behaviour

**The continuation's steps reach nobody.** `POST /runs/{id}/resume` executes the agent
inside the request (`agent_runner.py:2752`) rather than on the socket the conversation
streams, so its `tool_call`/`tool_result` frames go to that HTTP response and not to the
chat. `AgentRunResult` carried `output`, `status`, the cost and `parked` — never what was
called. `use-chat.ts` therefore had nothing to draw: it marked the decided call finished,
re-opened the panel from `parked`, and appended the answer.

**A continuation with no answer records nothing at all.** `TranscriptService.record`
(`transcript.py:157`) wrote the assistant turn only `if answer:`. A continuation that runs
a command and then parks on a second one has `answer == ""`, so the command it ran, its
arguments and its result were never written — not to `messages`, not to `tool_calls`.
`_run` computes them (`agent_runner.py:2930`) and dropped them on the floor. The same
applies to any non-streaming run that fails or is stopped after calling something, which
is precisely the run somebody opens afterwards.

The docstring on `_run` already claimed the opposite: *"so a run that failed, parked or
was stopped still says what was asked and what it managed to do"*.

## How you would know it was fixed

- Approving a gated call draws the steps the continuation made, in order, before its
  answer — including the call it parked on again, as *waiting for a person*.
- Deciding that second call writes the outcome back onto **its own** step, not onto the
  turn that parked first.
- After a reload, a run that parked twice shows both commands and the result of the one
  that ran, rather than two approvals with nothing between them.

[...]

PR #600, opened 2026-08-11
fix(chat): stop storing the approval notice as the agent's words

### LINKED ISSUE #509: A parked run stores "waiting in the approvals queue" as the agent's words, and it outlives the wait
A run that parks stores *"This run needs approval before it can go further - it is
waiting in the approvals queue."* as the agent's words. Once somebody approves, that
sentence is false — and it stays in the transcript, in the middle of the turn, for good.

## Where it goes

`agent_chat.py:171` (`display_output`) answers with `_AWAITING_APPROVAL` whenever the
run ends on `DeferredToolRequests`, because silence would read as the agent ignoring
the person. That value becomes `ChatTurn.output`, and `agent_session.py` persists it as
the assistant message's `content`.

It is UI state, not model text. The panel below it and the step in *waiting for a
person* already say the same thing, and they stop saying it when the decision is made.
The sentence does not.

## Why it shows now

Since #505 the segments of one run are drawn as one turn (`continuesTurn` in
`MessageList`), so the reader sees:

> ▸ Ran python - openpyxl …
> **This run needs approval before it can go further - it is waiting in the approvals queue.**
> ▸ Ran python - zipfile …
> Widzę plik Excel „…" — wygląda jak roczny plan nauki…

The middle line is a sentence about a state that ended two steps earlier, attributed to
the agent, inside an answer that plainly did go further.

## What fixing it probably means

Persist what the model actually wrote for a parked turn — `TurnTimeline.text`, which is
usually empty — and leave the notice as something the socket says rather than something
the transcript keeps. `_persist_partial_turn` already writes from the timeline for the
turn that does not finish, so the shape exists. Worth checking what an empty parked turn
renders as before committing to it: the step and the panel have to carry the whole
meaning on their own.

Found while fixing #505 and #507. Same surface; independent of both.

DEENUU1:
## What this fixes

A chat turn that parked on an approval stored *"This run needs approval before it can
go further - it is waiting in the approvals queue."* as the assistant message's
`content`. The moment somebody approved, that sentence was false — and it stayed in
the transcript for good, attributed to the agent, in the middle of a turn that plainly
did go further. Since #505 draws the segments of one run as one turn, the reader sees
it between two steps that both ran.

It was never the model's text. It was UI state written into the one field that keeps
things forever.

Closes #509.

## What changed

- `backend/app/services/agent_chat.py` — `display_output` answers `""` for a run that
  ended on `DeferredToolRequests`, and `_AWAITING_APPROVAL` is gone. This is what every
  surface that does not stream already records (`AgentRunnerService._run` sets
  `output = ""` and hands that to `TranscriptService.record`), so web chat was the one
  place inventing a sentence.
- `backend/app/services/agent_session.py` — a parked turn persists what the model wrote
  before it stopped (`timeline.text`), the same route a turn that failed, was stopped or
  lost its socket already takes. Usually empty; not always, and a model that explained
  itself before asking for a gated call was previously overwritten by the notice.
- `backend/app/services/agent_session.py` — the in-memory history no longer gets an
  assistant entry when the turn said nothing. `build_message_history` turns each one into
  a `ModelResponse` holding a single `TextPart`, and an empty one is a part with nothing
  in it on the next request. The notice used to keep that field non-empty by accident.

That a run is parked is still said, by the two things that stop saying it when the
decision is made: the step it stopped on renders `awaiting_approval` ("waiting for
approval", amber, paused icon), and the `tool_approval_required` frame puts the panel in
front of whoever is looking. No frontend change — the client already appends nothing for
`final_result` with an empty `output`, and a message with tool-call parts and no text
renders its steps without an empty bubble.

## Testing

- `tests/test_agent_chat.py::TestWhatTheClientIsShown::test_a_parked_run_produces_no_text_of_its_own`
- `tests/test_agent_chat.py::TestPausingMidRun::test_a_parked_tool_call_leaves_the_run_resumable` — asserted the notice, now asserts no answer
- `tests/test_agent_session.py::TestATurnThatFinished` — three new: the parked turn stores
  no notice; it keeps what the model wrote (driven over a real `Agent.iter`, so the text
  travels the route the client's `text_delta` frames do); and a turn that answered nothing
  leaves no empty reply in the history
- `make lint` and `make test` clean, including the 100% gate — both files are in the
  platform layer

## Noticed, not fixed

**A still-parked run says nothing about waiting once the page is reloaded**, and this
change removes the sentence that was accidentally covering for it. The stored tool-call
row for a parked call keeps `status="running"`, which `replayedStatus` maps to
`unfinished` and `tool-call-card` draws as a finished step; and `pendingApproval` is only
ever set from a live socket frame, so the panel does not come back either. Every
non-streaming surface has always looked like this — the fix aligns web chat with them
rather than causing it — but it is worth its own issue, filed as #601.

### DIFF
--- a/backend/app/services/agent_chat.py
+++ b/backend/app/services/agent_chat.py
@@ -371,8 +371,8 @@ class ChatAgentRunner:
                 )
                 status = RunStatus.AWAITING_APPROVAL
             else:
+                output = result.output
                 status = RunStatus.COMPLETED
-            output = display_output(result.output)
         except asyncio.CancelledError:
             # The user pressed stop, or the socket went away mid-run. Cancelled
             # is not failed, and the tokens spent up to here were still spent.
--- a/backend/app/services/agent_chat.py
+++ b/backend/app/services/agent_chat.py
@@ -69,14 +69,6 @@ type ChatStream = Callable[[AgentRun[AgentDeps, str | DeferredToolRequests]], Aw
 # role implies.
 _NOT_A_MEMBER = "You are no longer a member of this organization."
 
-# What the person is told when the run stopped on an approval instead of an
-# answer. Silence would read as the agent ignoring them, and the run does not
-# continue in this conversation - it continues when somebody decides, from the
-# approvals queue.
-_AWAITING_APPROVAL = (
-    "This run needs approval before it can go further - it is waiting in the approvals queue."
-)
-
 
 def requested_agent_id(frame: Mapping[str, Any]) -> UUID | None:
     """Which published agent a chat frame is addressed to, if any.
@@ -152,8 +144,18 @@ def display_output(output: str | DeferredToolRequests) -> str:
     side-effecting call, Pydantic AI ends the run with the calls waiting on a
     human. There is no model text to show then, and the object itself is not
     something a client can render.
+
+    Empty, then - and deliberately not a sentence about the queue. This value
+    becomes the assistant message's `content`, so a notice put here is stored as
+    the agent's words: false the moment somebody approves, and false for good, in
+    the middle of a turn that plainly did go further (#509). That a run is parked
+    is state, and it is already carried by things that stop carrying it when the
+    decision is made - the step the run stopped on, and the approval panel put in
+    front of whoever is looking. Every surface that does not stream records an
+    empty answer here too (:meth:`AgentRunnerService._run`), so this is the whole
+    platform's answer rather than this one's.
     """
-    return output if isinstance(output, str) else _AWAITING_APPROVAL
+    return output if isinstance(output, str) else ""
 
 
 @dataclass(frozen=True)
@@ -182,6 +184,10 @@ class ChatTurn:
     """What one published-agent turn produced, for the surface to persist."""
 
     output: str
+    """The answer, or empty for a turn that parked on an approval - see
+    :func:`display_output`. A surface that streamed the turn holds what the model
+    said before it stopped; this carries only what the run *ended* with."""
+
     model_label: str
     agent_id: UUID
     agent_version_id: UUID | None
@@ -289,8 +295,8 @@ class ChatAgentRunner:
 
         Returns:
             The answer to show and persist, and the model that produced it. A
-            run parked on an approval returns the queue's explanation instead of
-            an answer - it did not fail, and it has not finished either.
+            run parked on an approval returns no answer at all - it did not fail,
+            and it has not finished either - and `parked` is what says so.
 
         Raises:
             AuthorizationError: If the user is not a member of this organization.
--- a/backend/app/services/agent_session.py
+++ b/backend/app/services/agent_session.py
@@ -200,11 +200,12 @@ class AgentSession:
         # stored, so a reloaded conversation is the one somebody watched rather
         # than a reconstruction of it; see `chat_timeline.TurnTimeline`.
         #
-        # It also holds what the model has said so far, for the turn that does not
-        # finish. On the success path `turn.output` is the answer and `.text` is
-        # never read; nothing can tell in advance which path a turn is on, and the
-        # alternative is throwing away a half-written answer on exactly the runs
-        # somebody opens afterwards.
+        # It also holds what the model has said so far, for every turn that does
+        # not end with an answer: one that failed, was stopped or lost its socket,
+        # and one that parked on an approval. `turn.output` is the answer where
+        # there is one and empty where there is not; nothing can tell in advance
+        # which path a turn is on, and the alternative is throwing away a
+        # half-written answer on exactly the runs somebody opens afterwards.
         timeline = TurnTimeline()
         # The run row, as soon as `prepare` opens one. A list because the
         # callback is `list.append` and a turn opens at most one run - it is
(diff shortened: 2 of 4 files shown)


backend/app/services/agent_session.py
```diff
@@ -258,12 +259,23 @@ async def stream(agent_run: Any) -> None:
...
                     environment_id=requested_environment_id(data),
                 )
-            output = turn.output
+            # What the model actually wrote. `turn.output` is what the run ended
+            # with, and a turn that parked on an approval ended with nothing - so
+            # what it said before it stopped is on the timeline, exactly as it is
+            # for a turn that failed. The notice that used to stand in for the
+            # answer was stored as the agent's words and stayed there after the
+            # decision, saying a turn was waiting inside a turn that went on
+            # (#509); the step and the approval panel say it instead, and stop.
```
DEENUU1: Three copies of one reason now. The block above `timeline = TurnTimeline()` (203-208) carries the invariant, `display_output`'s docstring carries the #509 story, and this retells both. `code-style.md` says the reasoning lives in the docstring, and the middle copy is the one that goes stale first.

I'd keep only the clause that stops a reader "fixing" the `or`:

```suggestion
            # `turn.output` is what the run *ended* with; a turn that parked ended
            # with nothing, so its words are on the timeline (#509).
```

DEENUU1 (commented):
This is the right fix and it's argued the right way round: the notice was UI state written into the one field that keeps things forever, and pushing it back onto the step and the panel means the two things that say "waiting" both stop saying it when the decision lands. Reusing the `_persist_partial_turn` route for the parked turn rather than inventing a second one is what makes it small.

Nothing blocking. Two things I'd change - the reason for this change is now written out three times across two files, and `agent_chat.py:375` re-asks a question it answered ten lines earlier - plus one verification I think belongs in the code rather than only in my head.

One more, outside the diff: `backend/tests/test_agent_session.py:468`'s docstring says the history "is appended only after a complete run", which is now half true - the user half is appended after a parked run too.

backend/app/services/agent_session.py
```diff
@@ -258,12 +259,23 @@ async def stream(agent_run: Any) -> None:
...
+            output = turn.output or timeline.text
             model_label = turn.model_label
             agent_version_id = turn.agent_version_id
 
             self.conversation_history.append({"role": "user", "content": user_message})
-            self.conversation_history.append({"role": "assistant", "content": output})
+            # Only when there was something to say. A parked turn is the ordinary
+            # way to reach this with nothing, and an empty assistant turn in the
+            # history is a `TextPart` with no content on the next request.
+            if output:
```
DEENUU1: Worth writing down what I checked here, because the obvious worry about this guard is the shape it leaves behind: skip the assistant entry and keep the user one, and `conversation_history` can hold two `user` entries in a row (park with no text, decide the approval, type again). `build_message_history` turns those into two consecutive `ModelRequest`s.

It's fine - `_agent_graph._clean_message_history` runs `_merge_consecutive_messages` over the whole history before every model request, so the two arrive as one `ModelRequest` with two `UserPromptPart`s. And the empty `TextPart` this avoids was already harmless: the Anthropic adapter skips an empty `TextPart` and drops an assistant message with no content blocks, and the OpenAI chat adapter sends `content: ""`, which the API accepts.

So the guard is tidiness rather than a broken request being fixed. I'd say that here, otherwise the next reader takes "a `TextPart` with no content on the next request" as a live failure mode and builds on it.

backend/app/services/agent_chat.py
```diff
@@ -152,8 +144,18 @@ def display_output(output: str | DeferredToolRequests) -> str:
...
+    the agent's words: false the moment somebody approves, and false for good, in
+    the middle of a turn that plainly did go further (#509). That a run is parked
+    is state, and it is already carried by things that stop carrying it when the
+    decision is made - the step the run stopped on, and the approval panel put in
+    front of whoever is looking. Every surface that does not stream records an
+    empty answer here too (:meth:`AgentRunnerService._run`), so this is the whole
+    platform's answer rather than this one's.
     """
-    return output if isinstance(output, str) else _AWAITING_APPROVAL
+    return output if isinstance(output, str) else ""
```
DEENUU1, line 158: Now that this answers `""`, the call site at line 375 re-asks a question it answered at 365:

```python
if isinstance(result.output, DeferredToolRequests):
    paused = PausedRunState(...)
    status = RunStatus.AWAITING_APPROVAL
else:
    output = result.output
    status = RunStatus.COMPLETED
```

`output` is already `""` from line 356, so the `else` is all it takes - and that is byte-for-byte the shape `AgentRunnerService._run` uses at 3158-3169, which is the alignment this PR is arguing for. It also leaves `display_output` with the one caller that genuinely cannot discriminate: the `final_result` frame.

backend/tests/test_agent_session.py
```diff
@@ -519,6 +531,52 @@ async def test_a_parked_call_is_put_in_front_of_whoever_is_looking(self):
...
+        nothing - so the words are read off the timeline, the route a turn that
+        failed already takes. Without it a model that explained itself and then
+        asked for a gated call is written down as having said nothing at all.
+
+        Driven over a real `Agent.iter` so the text travels the way the client's
+        `text_delta` frames do: what is stored is what its reader saw."""
+        session = _session()
+
+        async def stopped_on_an_approval(**kwargs: Any) -> ChatTurn:
+            async with _answering_agent().iter("how many are open?") as agent_run:
```
DEENUU1, line 560: This is the part I'd have got wrong. Stubbing `timeline.text` would have tested the `or`; driving it over a real `Agent.iter` tests that the words which reach the client as `text_delta` are the words that reach the column - which is the actual claim. Good test.

backend/app/services/agent_session.py
DEENUU1: Fixed in 256b7bf3 — cut to the two-line clause. The `TurnTimeline` block and `display_output`'s docstring carry the rest.

backend/app/services/agent_session.py
DEENUU1: Written down in 256b7bf3, in the shape you set out: the two consecutive `user` entries, `_clean_message_history` merging them before every model call, and both adapters' handling of the empty `TextPart` — so the comment says the guard is tidiness rather than a broken request.

backend/app/services/agent_chat.py
DEENUU1: Fixed in 256b7bf3 — the `else` sets `output = result.output`, matching `AgentRunnerService._run` at 3169-3171. `display_output` keeps the one caller that cannot discriminate, the `final_result` frame.

backend/tests/test_agent_session.py
DEENUU1: Agreed — keeping it driven over a real `Agent.iter`.

PR #604, opened 2026-08-11
fix(i18n): keep the chosen locale across navigations

### LINKED ISSUE #285: Changing the language does not persist: it reverts to the previous locale on navigation
## What breaks

Switching the language works for the current page, then reverts on the next navigation. Pick Polish → the page redraws in Polish; click through to another page → it is English again. The choice does not persist.

## Why (root cause, not just the symptom)

`frontend/src/middleware.ts` runs next-intl with `localePrefix: "as-needed"` (English, the default, has **no** URL prefix; Polish lives under `/pl/...`) and `localeDetection: false`, and **nothing writes a `NEXT_LOCALE` cookie**. So the *only* thing keeping the UI in Polish is the `/pl` prefix in the URL.

`frontend/src/components/language-switcher.tsx` switches by `router.push(buildLocalizedPath(pathname, newLocale))` (lines ~44/99/186) — it changes the URL prefix and nothing else. The moment you reach a path without the `/pl` prefix — an internal `<Link>`/`router.push` that was not built locale-aware, or any absolute path like `/agents` — the middleware resolves it to the default locale, because with detection off and no cookie there is nothing else to resolve from. The language "reverts" because its entire persistence is a prefix that ordinary navigation drops.

## Where

- `frontend/src/middleware.ts` — `localePrefix: "as-needed"`, `localeDetection: false`, no cookie.
- `frontend/src/components/language-switcher.tsx:~44,~99,~186` — `router.push(buildLocalizedPath(...))`, URL-prefix only, no persisted locale.
- Any internal navigation that uses plain `next/navigation` / `next/link` instead of next-intl's locale-aware navigation drops the prefix.

## How you'd know it's fixed

Pick Polish, then navigate across several pages (including a link to an unprefixed path) and reload — every page stays Polish. Concretely: the switch persists the locale (write the `NEXT_LOCALE` cookie the middleware reads, and/or route all internal navigation through next-intl's locale-aware `Link`/`useRouter` so the prefix is never dropped). A test covers that a locale chosen on one page is still in force after navigating to another.

## Where it sits

Frontend surfaces cluster (#168). This is the i18n *runtime/persistence* layer — distinct from the i18n *tooling* issues #141 and #249 (which are about `check_i18n.py` catching untranslated strings). Found by a maintainer using the app.

DEENUU1:
## What this fixes

Picking Polish redrew the current page and nothing more: the next link took the UI
back to English. Closes #285.

The locale's entire persistence was the `/pl` URL prefix. `localePrefix: "as-needed"`
means a path without a prefix *is* the default locale, and the switcher pushed a
hand-built path with `next/navigation` — so every ordinary `<Link href="/agents">`
and `router.push("/orgs")` in the app dropped the prefix and, with it, the language.
next-intl reads a `NEXT_LOCALE` cookie itself, but only under `localeDetection`,
which also turns on `accept-language` sniffing — and this deployment deliberately
serves English at the root whatever the browser asks for. So nothing wrote the
cookie and nothing read it.

## What changed

- `frontend/src/lib/locale-routing.ts` — new. One routing config for the middleware
  and the navigation APIs, naming the cookie explicitly (a drift between the write
  and the read is invisible: the switch still appears to work for one page) and
  giving it a year's `maxAge`, so the choice is a preference rather than a session.
- `frontend/src/lib/locale-navigation.ts` — new. `createNavigation(routing)`, so
  `router.push(pathname, { locale })` both prefixes the path and writes the cookie.
- `frontend/src/middleware.ts:23` — reads that cookie by hand and redirects an
  unprefixed path to the picked locale, leaving `localeDetection: false` in place so
  `accept-language` is still ignored. A path that names a locale always wins, so a
  shared `/pl/...` URL still means what it says.
- `frontend/src/components/language-switcher.tsx:30` — `router.push(pathname, { locale })`
  in place of `router.push(buildLocalizedPath(...))`; the hand-rolled path builder is gone.
- `frontend/vitest.config.ts` — inline `next-intl` so `next-intl/middleware`'s bare
  `next/server` import resolves under Vite; without it the middleware cannot be
  imported by a test at all.
- `frontend/vitest.setup.ts` — `redirect` / `permanentRedirect` added to the
  `next/navigation` mock. `createNavigation` wraps them at module scope, so a
  component reaching for locale-aware navigation otherwise fails to import.
- Three layout test files shadow that shared mock with a one-line override and
  render a tree that now imports the switcher, so they needed the same two
  exports. Found by the coverage gate, second commit.

## How it failed before

Switch to Polish on `/pl/dashboard`, then click anything that navigates with an
absolute path — `router.push("/orgs")` in `components/teams/org-switcher.tsx:43` is
one of dozens — and the middleware resolves `/orgs` to `en` because there is no
prefix, no cookie and no detection. English again, and a reload never brought Polish
back either.

## Testing

- `src/middleware.test.ts` — 8 tests. The three that matter fail on `main`
  (verified by stubbing the new call out): an unprefixed path with `NEXT_LOCALE=pl`
  redirects to the prefixed one, at the root too, query string intact. The rest are
  the refusals: a path that names a locale is left alone, an unknown or empty cookie
  changes nothing, and a Polish `accept-language` with no cookie still gets English —
  that last one guards the `localeDetection: false` decision against exactly the
  shortcut this change could have taken.
- `src/components/language-switcher.test.tsx` — the round trip, because either half
  alone still reverts: click Polski, then feed the cookie the component actually
  wrote to the real middleware and assert it redirects `/orgs` → `/pl/orgs`.
- `src/lib/locale-routing.test.ts` — `localePrefixOf`, plus the config assertions
  that keep the cookie name and lifetime honest.
- Locally: `make test-frontend-cov` green — 256 files, 3875 tests, 100%
  lines/statements/functions, 97.65% branches. `make build-frontend` compiles.
  `scripts/check_i18n.py` clean, and pre-commit ran eslint, prettier and tsc on
  both commits.
- Not written: an E2E spec. The switcher test already drives the real middleware,
  and `frontend/e2e` needs a seeded backend for a journey these three cover.

## Noticed, not fixed

- **The switch drops the query string.** `usePathname` carries no search params, so
  switching language on `/orgs?create=1` lands on `/orgs`. True before this change
  too, and unchanged by it.
- **The redirect costs a hop.** An unprefixed navigation under a non-default locale
  now round-trips through a 307. Routing every internal `<Link>` and `useRouter`
  through next-intl's locale-aware equivalents would avoid it — 49 files import
  `next/link` and 27 use `useRouter`, so that is its own change, and the cookie is
  needed regardless for a reload or a pasted URL.

## Review round (f3a57b3e)

Three findings from reviewing the branch, two of them defects:

- **`localePrefixOf` disagreed with next-intl about case.** next-intl's
  `getPathnameMatch` lowercases both sides, so `/PL/agents` is a Polish path to it and
  was an unprefixed one here — with a `pl` cookie the middleware prefixed it again and
  redirected to `/pl/PL/agents`, a 404 on a URL that worked before the cookie existed.
  Fixed, and asserted from both sides.
- **A shared `/pl/...` link persisted nothing for a Polish browser.** next-intl's
  `syncCookie` writes `NEXT_LOCALE` only when the request carries no cookie *and*
  `accept-language` disagrees with the resolved locale — so exactly the visitors most
  likely to want Polish got no cookie, and their next ordinary `<Link>` was English
  again. That is #285 arriving by URL rather than by the switcher, and it is why the
  middleware now records the locale a prefixed path names. Document requests only: the
  router revalidates routes of the locale just switched away from, and writing from one
  of those would undo the switch.
- **The cookie value is validated by `pickedLocale`**, beside the other locale parsing,
  so the middleware no longer reaches around the routing module back to `./i18n`.
  `.claude/rules/frontend.md` gains the invariant, because nothing recorded that a
  switch must go through `@/lib/locale-navigation`.

The `<html lang="en">` bug the review turned up is pre-existing and out of scope here —
filed as #619.

### DIFF
--- a/frontend/src/components/layout/app-sidebar.test.tsx
+++ b/frontend/src/components/layout/app-sidebar.test.tsx
@@ -7,7 +7,13 @@ const can = vi.fn<(permission: string) => boolean>();
 const currentPath = vi.fn<() => string>(() => "/dashboard");
 const currentUser = vi.fn<() => { is_app_admin?: boolean } | null>(() => ({}));
 
-vi.mock("next/navigation", () => ({ usePathname: () => currentPath() }));
+vi.mock("next/navigation", () => ({
+  usePathname: () => currentPath(),
+  // Wrapped at module scope by the language switcher's locale-aware navigation,
+  // which this tree imports; without them the file fails to import at all.
+  redirect: vi.fn(),
+  permanentRedirect: vi.fn(),
+}));
 vi.mock("next-intl", () => ({ useTranslations: () => (key: string) => key }));
 vi.mock("@/hooks/use-permissions", () => ({ usePermissions: () => ({ can }) }));
 vi.mock("@/stores", () => ({ useAuthStore: () => ({ user: currentUser() }) }));
--- a/frontend/src/components/layout/app-sidebar.test.tsx
+++ b/frontend/src/components/layout/app-sidebar.test.tsx
@@ -9,8 +9,7 @@ const currentUser = vi.fn<() => { is_app_admin?: boolean } | null>(() => ({}));
 
 vi.mock("next/navigation", () => ({
   usePathname: () => currentPath(),
-  // Wrapped at module scope by the language switcher's locale-aware navigation,
-  // which this tree imports; without them the file fails to import at all.
+  // next-intl's createNavigation wraps these at module scope; see `vitest.setup.ts`.
   redirect: vi.fn(),
   permanentRedirect: vi.fn(),
 }));
--- a/frontend/src/lib/locale-routing.ts
+++ b/frontend/src/lib/locale-routing.ts
@@ -1,0 +1,43 @@
+import { defineRouting } from "next-intl/routing";
+
+import { locales, defaultLocale, type Locale } from "@/i18n";
+
+/**
+ * The cookie holding the locale the visitor picked.
+ *
+ * `NEXT_LOCALE` is next-intl's own default name, written client-side by the
+ * navigation APIs below and read back by `src/middleware.ts`. It is named here
+ * rather than left to the default so that the write and the read are the same
+ * constant: a divergence between them is invisible - the switch appears to work,
+ * because the URL prefix carries the locale for exactly as long as the current
+ * page lasts.
+ */
+export const LOCALE_COOKIE_NAME = "NEXT_LOCALE";
+
+/** A year. The choice is a preference, not a session. */
+const LOCALE_COOKIE_MAX_AGE = 60 * 60 * 24 * 365;
+
+/**
+ * One routing configuration, shared by the middleware and the navigation APIs.
+ *
+ * `localePrefix: "as-needed"` - English, the default, has no prefix; Polish lives
+ * under `/pl/...`.
+ *
+ * `localeDetection: false` because it gates the `accept-language` header as well
+ * as the cookie, and a visitor with a Polish browser must still be served English
+ * at the root until they ask for Polish. The cookie half is therefore read by hand
+ * in `src/middleware.ts`; see the note there.
+ */
+export const routing = defineRouting({
+  locales,
+  defaultLocale,
+  localePrefix: "as-needed",
+  localeDetection: false,
+  localeCookie: { name: LOCALE_COOKIE_NAME, maxAge: LOCALE_COOKIE_MAX_AGE },
+});
+
+/** The locale a path names in its first segment, or `null` when it names none. */
+export function localePrefixOf(pathname: string): Locale | null {
+  const segment = pathname.split("/")[1];
+  return segment && (locales as readonly string[]).includes(segment) ? (segment as Locale) : null;
+}
--- a/frontend/src/lib/locale-routing.ts
+++ b/frontend/src/lib/locale-routing.ts
@@ -6,16 +6,17 @@ import { locales, defaultLocale, type Locale } from "@/i18n";
  * The cookie holding the locale the visitor picked.
  *
  * `NEXT_LOCALE` is next-intl's own default name, written client-side by the
- * navigation APIs below and read back by `src/middleware.ts`. It is named here
- * rather than left to the default so that the write and the read are the same
- * constant: a divergence between them is invisible - the switch appears to work,
+ * navigation APIs below, and read back - and written, for a path that names a
+ * locale - by `src/middleware.ts`. It is named here rather than left to the
+ * default so that every write and the read are the same constant: a divergence
+ * between them is invisible - the switch appears to work,
  * because the URL prefix carries the locale for exactly as long as the current
  * page lasts.
  */
 export const LOCALE_COOKIE_NAME = "NEXT_LOCALE";
 
 /** A year. The choice is a preference, not a session. */
-const LOCALE_COOKIE_MAX_AGE = 60 * 60 * 24 * 365;
+export const LOCALE_COOKIE_MAX_AGE = 60 * 60 * 24 * 365;
 
 /**
  * One routing configuration, shared by the middleware and the navigation APIs.
@@ -25,8 +26,8 @@ const LOCALE_COOKIE_MAX_AGE = 60 * 60 * 24 * 365;
  *
  * `localeDetection: false` because it gates the `accept-language` header as well
  * as the cookie, and a visitor with a Polish browser must still be served English
- * at the root until they ask for Polish. The cookie half is therefore read by hand
- * in `src/middleware.ts`; see the note there.
+ * at the root until they ask for Polish. The cookie half is therefore kept by hand
+ * in `src/middleware.ts`; see the notes there.
  */
 export const routing = defineRouting({
   locales,
@@ -36,8 +37,22 @@ export const routing = defineRouting({
   localeCookie: { name: LOCALE_COOKIE_NAME, maxAge: LOCALE_COOKIE_MAX_AGE },
 });
 
-/** The locale a path names in its first segment, or `null` when it names none. */
+/**
+ * The locale a path names in its first segment, or `null` when it names none.
+ *
+ * Matched case-insensitively, because next-intl's own prefix matching is:
+ * `/PL/agents` is a Polish path to it, redirected to the canonical `/pl/agents`.
+ * Reading it as unprefixed here would put a second prefix on the front of a path
+ * that already has one - `/pl/PL/agents`, which no route serves.
+ */
 export function localePrefixOf(pathname: string): Locale | null {
-  const segment = pathname.split("/")[1];
+  const segment = pathname.split("/")[1]?.toLowerCase();
   return segment && (locales as readonly string[]).includes(segment) ? (segment as Locale) : null;
+}
+
+/** The locale the visitor picked, when this deployment serves it and it is not the default. */
+export function pickedLocale(value: string | undefined): Locale | null {
+  return value && value !== defaultLocale && (locales as readonly string[]).includes(value)
+    ? (value as Locale)
+    : null;
 }
--- a/frontend/src/middleware.ts
+++ b/frontend/src/middleware.ts
@@ -1,8 +1,13 @@
 import createMiddleware from "next-intl/middleware";
 import { NextResponse, type NextRequest } from "next/server";
 
-import { locales, defaultLocale } from "./i18n";
-import { LOCALE_COOKIE_NAME, localePrefixOf, routing } from "./lib/locale-routing";
+import {
+  LOCALE_COOKIE_MAX_AGE,
+  LOCALE_COOKIE_NAME,
+  localePrefixOf,
+  pickedLocale,
+  routing,
+} from "./lib/locale-routing";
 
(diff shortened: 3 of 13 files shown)


frontend/src/lib/locale-routing.ts
```diff
@@ -0,0 +1,43 @@
...
+  locales,
+  defaultLocale,
+  localePrefix: "as-needed",
+  localeDetection: false,
+  localeCookie: { name: LOCALE_COOKIE_NAME, maxAge: LOCALE_COOKIE_MAX_AGE },
+});
+
+/** The locale a path names in its first segment, or `null` when it names none. */
+export function localePrefixOf(pathname: string): Locale | null {
+  const segment = pathname.split("/")[1];
```
DEENUU1: This is case-sensitive and next-intl's is not, which turns a URL that used to work into a 404.

`getPathnameMatch` in `next-intl/middleware/utils.js` lowercases both sides before comparing, so on `main` a request for `/PL/agents` matches the `pl` prefix and gets canonicalised to `/pl/agents`. Here, `localePrefixOf("/PL/agents")` answers `null`, so with a `pl` cookie we treat it as unprefixed and prepend — `/pl/PL/agents`, which no route serves.

```suggestion
  const segment = pathname.split("/")[1]?.toLowerCase();
```

I ran it: with that line, `/PL/agents` + a `pl` cookie lands back on `/pl/agents` and the three new test files stay green. Worth a row in `locale-routing.test.ts` beside the `/plans` one — a first segment that names a locale in the wrong case still names that locale.

frontend/src/middleware.ts
```diff
@@ -1,16 +1,42 @@
...
+  url.pathname = `/${picked}${pathname === "/" ? "" : pathname}`;
+  return NextResponse.redirect(url);
+}
 
-  // Always serve `defaultLocale` (en) at root, regardless of the visitor's
-  // Accept-Language header. The user opts into Polish via the language switcher.
-  localeDetection: false,
-});
+export default function middleware(request: NextRequest): NextResponse {
+  return restorePickedLocale(request) ?? handleI18nRouting(request);
```
DEENUU1: The cookie is the persistence, but only one of the two ways into Polish writes it.

next-intl's `syncCookie` writes `NEXT_LOCALE` on a prefixed request only when there is no cookie **and** `accept-language` disagrees with the locale it resolved. So a visitor whose browser sends `accept-language: pl-PL` and who opens a shared `/pl/agents` gets no `Set-Cookie` at all — I checked the response — and their next plain `<Link href="/orgs">` is English again. That is #285's symptom arriving by URL instead of by the switcher, and it hits precisely the people most likely to want Polish.

The switcher path is fine because the client writes the cookie itself. What is missing is the path that names a locale telling us so. This works, and the existing tests stay green with it (needs `LOCALE_COOKIE_MAX_AGE` exported):

```ts
function rememberPrefixedLocale(request: NextRequest, response: NextResponse): void {
  // A background request is how the router revalidates a locale you have just left;
  // writing the cookie from one would undo the switch. Same guard next-intl uses.
  if ((request.headers.get("sec-fetch-dest") ?? "document") !== "document") return;
  const named = localePrefixOf(request.nextUrl.pathname);
  if (!named || request.cookies.get(LOCALE_COOKIE_NAME)?.value === named) return;
  response.cookies.set(LOCALE_COOKIE_NAME, named, {
    path: "/",
    maxAge: LOCALE_COOKIE_MAX_AGE,
    sameSite: "lax",
  });
}

export default function middleware(request: NextRequest): NextResponse {
  const restored = restorePickedLocale(request);
  if (restored) return restored;
  const response = handleI18nRouting(request);
  rememberPrefixedLocale(request, response);
  return response;
}
```

If you would rather keep this branch to the switcher, a follow-up issue is fine by me — but it should not go unrecorded, because "the cookie is the persistence" is only half true until it is closed.

frontend/src/middleware.ts
```diff
@@ -1,16 +1,42 @@
...
+ * is not.
+ */
+function restorePickedLocale(request: NextRequest): NextResponse | null {
+  const { pathname } = request.nextUrl;
+  if (localePrefixOf(pathname)) return null;
+
+  const picked = request.cookies.get(LOCALE_COOKIE_NAME)?.value;
+  if (!picked || picked === defaultLocale || !(locales as readonly string[]).includes(picked)) {
+    return null;
+  }
```
DEENUU1: nit: the whole point of `locale-routing.ts` is one config, and this reaches around it back to `./i18n` for the same two values. I would move the check next to `localePrefixOf`, where the other locale-string parsing already lives:

```ts
/** The locale the visitor picked, when we serve it and it is not the default. */
export function pickedLocale(value: string | undefined): Locale | null {
  return value && value !== defaultLocale && (locales as readonly string[]).includes(value)
    ? (value as Locale)
    : null;
}
```

then here:

```suggestion
  const picked = pickedLocale(request.cookies.get(LOCALE_COOKIE_NAME)?.value);
  if (!picked) return null;
```

which also lets the `./i18n` import go from this file, so the middleware reads as routing decisions and nothing else.

DEENUU1 (commented):
This is the right shape for #285. The cookie is the persistence, the middleware is what restores it, and both names come from one constant so they cannot drift. The test I like most is the switcher one: it clicks Polski and then feeds the cookie the component actually wrote into the real middleware, so it covers the round trip rather than either half of it — plus the `accept-language` case, which locks the `localeDetection: false` decision against the shortcut this change could easily have taken. The "noticed, not fixed" section is honest about the dropped query string and the extra hop.

Two things I found by driving the real middleware locally, both verified rather than guessed:

1. **`/PL/agents` with a `pl` cookie now redirects to `/pl/PL/agents` → 404.** next-intl's prefix matching is case-insensitive and `localePrefixOf` is not, so a path it used to canonicalise is now read as "no prefix" and gets a second prefix stuck on the front. One line, inline.
2. **A shared `/pl/...` link opened by a Polish browser persists nothing** — next-intl's `syncCookie` skips the write when `accept-language` already agrees with the resolved locale, so no cookie is set and the next plain `<Link>` is English again. That is the issue's symptom coming in through the other door, for the people most likely to want Polish. Mechanism and a tested sketch inline.

One pre-existing thing worth knowing, out of scope for this branch: `src/app/layout.tsx:90` hard-codes `lang={defaultLocale}` and the `[locale]` layout does not render `<html>`, so every Polish page still declares itself English to screen readers and search engines. Not caused by this change, but more visible now that Polish actually sticks — worth its own issue.

frontend/src/components/layout/app-sidebar.test.tsx
```diff
@@ -7,7 +7,13 @@ const can = vi.fn<(permission: string) => boolean>();
 const currentPath = vi.fn<() => string>(() => "/dashboard");
 const currentUser = vi.fn<() => { is_app_admin?: boolean } | null>(() => ({}));
 
-vi.mock("next/navigation", () => ({ usePathname: () => currentPath() }));
+vi.mock("next/navigation", () => ({
+  usePathname: () => currentPath(),
+  // Wrapped at module scope by the language switcher's locale-aware navigation,
+  // which this tree imports; without them the file fails to import at all.
```
DEENUU1: nit: this comment is in three files verbatim, and the repo's own rule is that comments are scarce. One line would carry it — `// next-intl's createNavigation wraps these at module scope.` — or drop it in two of the three and leave the fullest version once.

The deeper fix is a shared factory so the fourth test that renders a tree containing the switcher does not hit the same import error again; if you go that way it has to be reached from *inside* the `vi.mock` factory (`await import(...)`), because `vi.mock` is hoisted above the imports. Not worth it for three files, so I would just trim the comment.

frontend/src/lib/locale-routing.ts
DEENUU1: Fixed in f3a57b3e. `localePrefixOf` lowercases the segment before comparing, with the why in its docstring, and it is covered from both sides: `locale-routing.test.ts` asserts `/PL/agents` names `pl`, and `middleware.test.ts` asserts that with a `pl` cookie it now lands on `/pl/orgs` rather than `/pl/PL/orgs`.

frontend/src/middleware.ts
DEENUU1: Taken as written — `rememberPrefixedLocale` in f3a57b3e, document requests only, with the `sec-fetch-dest` reasoning in the docstring. Covered four ways: the Polish browser that used to persist nothing, a browser asking for nothing in particular, a cookie that already agrees (writes nothing), and a background request (writes nothing).

frontend/src/middleware.ts
DEENUU1: Done. `pickedLocale` sits beside `localePrefixOf` in `locale-routing.ts`, both helpers are covered there, and the `./i18n` import is gone from the middleware.

frontend/src/components/layout/app-sidebar.test.tsx
DEENUU1: Trimmed to one line in all three files, pointing at `vitest.setup.ts` where the full explanation already lives. Left the shared factory alone for the hoisting reason.

PR #605, opened 2026-08-11
fix(frontend): self-host the three web fonts

### LINKED ISSUE #572: ci: frontend build depends on Google Fonts at build time and flakes on gstatic 404s
## What breaks

`src/app/layout.tsx` loads two families through `next/font/google` — **Inter** and
**Bricolage Grotesque**. `next build` (Turbopack) fetches their `.woff2` files from
`https://fonts.gstatic.com` **at build time**. When gstatic answers `404` for those
URLs, Turbopack turns the failed fetch into:

```
Module not found: Can't resolve '@vercel/turbopack-next/internal/font/google/font'
  [next]/internal/font/google/inter_3d693864.module.css
  ./src/app/layout.tsx
Error: Turbopack build failed with N errors
```

and the build exits non-zero — taking down **every CI job that builds the frontend**:
`test-frontend`'s `Build` step and `e2e`'s `Build the frontend` step.

## How it triggers

Any PR whose changed paths trigger a frontend build, whenever gstatic 404s the font
files. On 2026-08-10 it hit two PRs in the same push window:

- **#570** — `test-frontend` build failed on the Bricolage Grotesque woff2 (6 errors).
- **#544** — `e2e` "Build the frontend" failed on the Inter woff2 (28 errors).

**#535**, built in the same window, succeeded — so this is intermittent CDN behaviour,
not a code defect on any branch. Every green frontend build so far has been luck of the
CDN; nothing in the repo guarantees the fonts are reachable.

## file:line

- `frontend/src/app/layout.tsx` — the `next/font/google` imports for Inter and
  Bricolage Grotesque.

## Pre-existing, found — not caused

Found while getting #535 / #544 / #570 green (their `test`/`lint` jobs pass; only the
transient font-404 build failed). Not introduced by any of those branches — they touch
the backend, docs and the Makefile, never the frontend build.

## How you'd know it's fixed

`next build` makes **no request to `fonts.gstatic.com`** — the two families are
self-hosted with `next/font/local` and the `.woff2` files are vendored under the app —
so the build succeeds offline and regardless of what gstatic returns.

## Where it sits

Same CI-reliability theme as **#162** (an e2e spec that flakes on `main`). Surfaced by
the #520 test-speed work but orthogonal to it — a build-input dependency, not a test
timing one. Recorded on the map (#168).

DEENUU1:
## What this fixes

`next build` no longer touches the network. The three web families are vendored under
`frontend/src/app/fonts/` and read by `next/font/local`, so a 404 from
`fonts.gstatic.com` can no longer fail a frontend build on a branch that never touched
the frontend.

Closes #572

## What changed

- `frontend/src/app/layout.tsx:2` — one `localFont` import replaces the
  `next/font/google` imports of Bricolage Grotesque, Inter and Geist Mono. Each
  declaration names a vendored file and the weight *range* the variable font covers
  (`700 800`, `400 700`, `400 500`) instead of a list of static weights.
- `frontend/src/app/fonts/*.woff2` — the latin subset of each family, range-limited to
  those weights. 113 KB across the three.
- `frontend/src/app/fonts/OFL.txt` — SIL OFL 1.1 with the copyright notice of all three
  projects. Vendoring the files means shipping the licence with them.
- `frontend/src/app/fonts/README.md` — where each file came from, its Google Fonts
  version, and the two commands to refresh one.
- `frontend/src/app/fonts/fonts.test.ts` — the regression test. Two assertions: no module
  imports the Google helper, and the set of `.woff2` on disk is exactly the set
  `layout.tsx` declares.

The issue named two families. Geist Mono loads the same way and had the same exposure,
so all three moved.

## How it failed before

`next/font/google` resolves a family against `fonts.gstatic.com` at build time. When
gstatic 404s the `.woff2`, Turbopack surfaces the failed fetch as
`Module not found: Can't resolve '@vercel/turbopack-next/internal/font/google/font'` and
exits non-zero, which is `test-frontend`'s `Build` step and `e2e`'s
`Build the frontend` step. On 2026-08-10 it hit #570 (Bricolage, 6 errors) and #544
(Inter, 28 errors) inside one push window while #535 built fine — nothing in the
repository made the fonts reachable, so every green build so far was luck of the CDN.

## Testing

- `frontend/src/app/fonts/fonts.test.ts` — fails on the pre-fix `layout.tsx` (checked by
  restoring it) on both assertions. A path typo is the one remaining way to fail a build
  for want of a font, and the comparison runs in both directions, so a `.woff2` nobody
  declares fails too — that is what a swapped-out family leaves behind, with its licence
  obligation still sitting in `OFL.txt`.
- `bun run build` — succeeds, emits the three `.woff2` into `.next/static/media`, and the
  only occurrence of `gstatic` in the output is the source comment explaining why it is
  absent.
- The generated `@font-face` rules were read back to confirm the weight ranges survived
  and that Next still emits its metric-adjusted `local(Arial)` fallback, as it did for
  the Google-hosted versions. The emitted CSS differs only in where the file comes from.
- `make lint` (exit 0), `bun run test:coverage` — 254 files, 3863 tests, gate green at
  100/100/100 lines-statements-functions and 97.64% branches.
- Not verified: an actually air-gapped build. The build was run with a working network;
  what is checked is that no request to gstatic is made or referenced.

## Noticed, not fixed

The vendored subsets are `latin` only, which is exactly what `subsets: ["latin"]` asked
for before — so Polish diacritics (`ą ć ę ł ń ś ź ż`) on the shipped `pl` locale still
fall through to the system font, as they did on `main`. Pre-existing, unchanged here, and
now a one-file change to fix: filed separately rather than widened into this diff.


frontend/src/app/fonts/README.md
```diff
@@ -0,0 +1,34 @@
+# Vendored web fonts
+
+`next/font/local` reads these; nothing here is fetched at build time. That is the whole
+point — `next/font/google` resolved the same families against `fonts.gstatic.com` while
+`next build` ran, so a 404 from the CDN failed the build on branches that had not touched
+the frontend at all (#572). A vendored file cannot 404.
+
+Each file is the **latin** subset of the family's variable font, range-limited to the
+weights `layout.tsx` asks for. Weights outside the declared range are synthesized by the
+browser, so widening a range means re-fetching, not editing the declaration.
```
DEENUU1: This paragraph is wrong on all three claims, and I only noticed because I went to verify the files.

The files are not range-limited. `fvar` says Inter 100–900, Geist Mono 100–900, Bricolage 200–800 — every one of them the family's full axis. Google does not axis-subset: `css2?family=Inter:wght@400..700` and `...@100..900` return the *identical* URL, `.../v20/UcC73FwrK3iLTeHuS_nVMrMxCp50SjIa1ZL7W0Q5nw.woff2`. The range in the query only changes the `font-weight:` line it prints.

Weights outside the declared range are clamped, not synthesized — a variable face declared `400 700` asked for 800 renders at 700.

And so widening a range *is* editing the declaration. Re-fetching gets you the same bytes back. Telling a future maintainer to re-download is telling them to do a no-op and then wonder why nothing changed.

```suggestion
Each file is the **latin** subset of the family's variable font and carries that
family's full `wght` axis: Google does not axis-subset, so `wght@400..700` and
`wght@100..900` return the same bytes. The range in `layout.tsx` is a declaration
about which weights the face may serve, not a property of the file — a weight
outside it is clamped, not synthesized. Widening one is a one-line edit to
`layout.tsx`, with nothing to re-download.
```

The Weights column in the table below is fine as-is once this reads as "what we declare". The filenames are the leftover: `inter-latin-400-700.woff2` names a range the file does not have. Either `inter-latin.woff2` and friends (a `git mv` plus the three `src:` lines — the new test's both-ways comparison will catch it if you miss one), or keep them and let the README say the name records the declaration. I would lean to renaming, since the name is what people trust before they read anything.

frontend/src/app/fonts/fonts.test.ts
```diff
@@ -0,0 +1,59 @@
+import { readdirSync, readFileSync, statSync } from "fs";
+import { join } from "path";
```
DEENUU1: nit: the three other tests that sweep the source tree (`platform-proxy.test.ts:20`, `loading-state.test.tsx:1`, `settings-tabs.test.ts:1`) all use the `node:` prefix. Worth matching so it is uniform.

```suggestion
import { readdirSync, readFileSync } from "node:fs";
import { join } from "node:path";
```

This drops `statSync` too — see the comment on the walker below; apply both or neither.

frontend/src/app/fonts/fonts.test.ts
```diff
@@ -0,0 +1,59 @@
...
+const SRC = join(__dirname, "..", "..");
+const FONTS = __dirname;
+
+function sourceFiles(dir: string): string[] {
+  return readdirSync(dir).flatMap((entry) => {
+    const path = join(dir, entry);
+    if (statSync(path).isDirectory()) return sourceFiles(path);
+    return /\.tsx?$/.test(entry) ? [path] : [];
+  });
+}
```
DEENUU1, line 31: `readdirSync` already knows what is a directory, so the per-entry `statSync` is one extra syscall for each of the 804 files under `src/`. `loading-state.test.tsx:120` has the same function under the same name doing it the cheaper way:

```suggestion
function sourceFiles(dir: string): string[] {
  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
    const path = join(dir, entry.name);
    if (entry.isDirectory()) return sourceFiles(path);
    return /\.tsx?$/.test(entry.name) ? [path] : [];
  });
}
```

Which is the bigger point: this is now the third hand-rolled tree walker in the suite. Not something to widen this diff with — worth a follow-up to lift one into `src/test-utils/` (outside the coverage `include` list, so it does not drag the gate along).

DEENUU1 (commented):
Good fix, and the right one — vendoring is the only thing that actually removes the CDN from the build's critical path, and doing all three families rather than the two the issue named is correct.

I checked the vendored files against gstatic rather than taking them on trust: all three are byte-identical (sha256) to what Google serves today for exactly the declared ranges, and `Inter:wght@400;500;600;700` — what `next/font/google` requested on `main` — resolves to that same variable file. So this is a zero-pixel change, not just a green build. I also swept for the same class of build-time fetch elsewhere: the three `next/og` generators use `sans-serif`/`monospace` with no remote font, mkdocs pulls nothing at build time, and there is no leftover `preconnect`/CSP entry pointing at gstatic. Nothing else in the tree fetches a build input. Nice that #606 got filed rather than folded in.

One thing to change, and it is in the README rather than the code: the paragraph explaining how to refresh a file is wrong in three ways, and it is the one artifact whose whole job is to tell the next person what to do. Details inline, plus two nits on the test.

frontend/src/app/fonts/README.md
DEENUU1: Fixed in 1f04f90d. Rewrote the paragraph, and went with the rename — `bricolage-grotesque-latin.woff2`, `inter-latin.woff2`, `geist-mono-latin.woff2`, with the three `src:` lines in `layout.tsx` following. The table now carries both columns: what we declare, and the axis the file actually holds (200–800 / 100–900 / 100–900). The refresh recipe asks for the full axis, since any range returns the same bytes.

Checked the claim on the other two families before writing it down, not just Inter: `Bricolage+Grotesque:wght@200..800` and `Geist+Mono:wght@100..900` both return files sha256-identical to the vendored ones.

`bun run build` succeeds and emits all three under the new names; the only `gstatic` string left in `.next` is the source comment in `layout.tsx` explaining its absence.

frontend/src/app/fonts/fonts.test.ts
DEENUU1: Fixed in 418440f3 — `node:fs` / `node:path`, and `statSync` dropped along with the walker change below.

frontend/src/app/fonts/fonts.test.ts
DEENUU1: Fixed in 418440f3, `withFileTypes` as suggested.

Mutation-tested the sweep afterwards rather than trusting a green run: a path typo in `layout.tsx` fails the declarations-match-the-directory assertion, a module importing `next/font/google` fails the other, and both go green again on revert.

Shared helper filed as #618 rather than widened into this branch.

PR #607, opened 2026-08-11
test: run the backend suite in a random order

### LINKED ISSUE #571: pytest-randomly is documented as on-by-default but is not installed
## What breaks

`CLAUDE.md` and `.claude/rules/testing.md` both state that **pytest-randomly is on by
default** and that the suite "still passes in a random order":

- `.claude/rules/testing.md`: *"The suite still passes in a random order
  (`pytest-randomly` is on by default here)"* and a `-p no:randomly` example command.
- `CLAUDE.md` testing section repeats the claim.

It is not installed, and not in the lockfile:

```
$ uv run pip list | rg -i randomly    # (nothing)
$ rg -i pytest-randomly backend/uv.lock   # (nothing)
$ uv run pytest ... -p randomly
ImportError: Error importing plugin "randomly": No module named 'randomly'
```

So the suite runs in **collection order**, `-p no:randomly` in the documented commands
is a silent no-op, and the order-independence the docs claim is verified has in fact
never been exercised by that mechanism.

## Why it matters

Order-dependent tests are a classic source of "passes locally, flakes in CI" (and vice
versa). The docs promise a guard against them that does not exist, so a latent
order-dependent test would sit undetected — and a reader trusting the docs would
believe `-p no:randomly` means something.

## Evidence the suite *is* order-independent (so the fix is low-risk)

Found while adding `pytest-xdist` (#570). With `pytest-randomly` temporarily installed:

- full unit suite green across two seeds (`--randomly-seed=12345`, `=99999`);
- full integration suite green across two seeds (`=12345`, `=88888`);
- full suite green under xdist's load distribution (a shuffled order) at `-n 4`.

So enabling it for real is very likely a clean one-liner.

- `pytest-randomly` is in `backend/pyproject.toml` `[dependency-groups] dev` and the
  lockfile; `uv run pytest ... -p randomly` no longer errors.
- The suite passes with it enabled (CI proves it every run, in a fresh order).
- The `-p no:randomly` examples in the docs are then real, and the "on by default"
  claim is true.

## Where it sits

Same class as **#536** (docs/local-runnability drift found while working #520): a claim
in the docs that the environment does not back up. Part of the Tests/CI cluster in
**#168**. Found while working **#520 / #570**. Pre-existing — not caused by that branch.

DEENUU1:
## What this fixes

`pytest-randomly` was described as on by default in `docs/testing.md` and
`.claude/rules/testing.md` while being in neither `backend/pyproject.toml` nor the
lockfile — so the suite ran in collection order, the documented `-p no:randomly` was a
silent no-op, and the order-independence those pages called verified had never been
exercised by that mechanism. #579 removed the false claim; this installs the plugin and
makes the claim true instead.

It then immediately earned its keep: the first shuffled run went red on a pre-existing
connection-pool defect in the integration harness, fixed in the second commit here.

Closes #571.

## What changed

**Enabling the shuffle**

- `backend/pyproject.toml` — `pytest-randomly>=4.1.0` in `[dependency-groups] dev`, with
  a note on why the plugin's own xdist support is load-bearing here
- `backend/uv.lock` — the resolved package and nothing else
- `backend/tests/test_random_order.py` — a guard: the dev group must declare it, so
  removing it fails a test instead of silently un-shuffling the suite. It asserts the
  *declaration*, not that the shuffle is active in this process — `-p no:randomly` is the
  documented move while bisecting, and a guard that turned it red would be telling people
  to stop using it
- `docs/testing.md` — the shuffle, the seed in the header, reproducing a red run, pinning
  the order while bisecting
- `.claude/rules/testing.md` — the same as a trap: a test that passed yesterday and fails
  today may have been depending on what ran before it

**What the shuffle found** (`fix(tests): dispose the app engine before each integration test`)

- `backend/tests/integration/conftest.py:255` — the `engine` fixture disposes
  `app.db.session.engine` on the way *in*
- `test_commit_before_response.py`, `test_flow_starts_after_commit.py` — their per-file
  disposes removed, leaving one explanation instead of three

## How it failed before

```
$ uv run pytest tests/ -p randomly
ImportError: Error importing plugin "randomly": No module named 'randomly'
```

And once it was installed:

```
sqlalchemy.exc.InterfaceError: (asyncpg.exceptions._base.InterfaceError)
cannot perform operation: another operation is in progress
[SQL: INSERT INTO dispatch_ordering_probe (id) VALUES ($1)]
```

`app.db.session.engine` is a module-level object, so its pool outlives the test that
filled it, while anyio gives every test its own event loop. A connection created on a loop
that has since closed answers the next statement issued through it that way, in whichever
test checked it out. The two files that drive the real `get_db_session` each disposed the
engine on the way *out*, which only covers the pair of them — anything else sharing the
xdist worker can leave a connection in that pool.

**Found by this branch, not caused by it:** which tests share a worker was already decided
at run time by xdist's `--dist load`. The shuffle changes the adjacencies and made it
surface.

## Testing

Locally against a live pgvector Postgres, so `tests/integration/` ran rather than skipping:

| What | Result |
|---|---|
| baseline on `main`, collection order | 4379 passed, 6 skipped |
| `--randomly-seed=12345 -n auto --maxprocesses 4` | 4380 passed, 6 skipped |
| full suite, fresh shuffle each run, **before** the pool fix | **red on run 6 of 8** |
| the two engine-driving integration files alone, 25 shuffled runs | 25 green — not those two on their own |
| full suite, fresh shuffle each run, **after** the pool fix | green ×12 |
| CI `test` job **before** the pool fix | red, same `InterfaceError`, first attempt |
| CI `test` job **after** | pass, 100% coverage gate held |

The header now prints `Using --randomly-seed=<n>`, and the controller's seed reaches every
xdist worker through `workerinput`, so `-n auto` collects one order rather than four —
without that the workers would abort on a collection mismatch.

`tests/test_migrations.py` was the other place order-dependence looked likely — an
`upgrade head` test beside a `downgrade base` one — and each test upgrades for itself
first, so the shuffle is safe there.

## Noticed, not fixed

`docs/testing.md` says the suite "takes a minute and a half". With a database up it is one
to three minutes here depending on load; the minute and a half is the unit layer with
`tests/integration/` skipping itself. Separate claim, worth its own measurement rather than
a guess in this diff.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -58,9 +58,12 @@ Traps, each of which has cost a red job here:
 
 - **The order is shuffled every run** (`pytest-randomly`, on by default since #571), so
   a test that passed yesterday and fails today may have been depending on what ran
-  before it. The header prints `Using --randomly-seed=<n>`; pass that seed back to
-  reproduce the failure, and `-p no:randomly` to pin collection order while bisecting.
-  The controller's seed reaches every xdist worker, so `-n auto` collects one order.
+  before it. The header prints `Using --randomly-seed=<n>`; replay that seed to get the
+  same order back, and `-p no:randomly` to pin collection order while bisecting. The
+  seed fixes the order, not which worker runs what - `--dist load` decides that on
+  timing - so a failure that depended on what shared a worker only reproduces without
+  `-n`. The plugin also reseeds `random` identically before every test: anything using
+  it for uniqueness is unique within a test and repeats across them.
 - **`bun run test:run` measures no coverage.** The frontend gate is a separate command
   and CI runs it (`bun run test:coverage`); 168 green files still failed the job.
 - **Frontend commands run from `frontend/`.** At the repository root vitest finds no
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -56,6 +56,11 @@ job always has one.
 
 Traps, each of which has cost a red job here:
 
+- **The order is shuffled every run** (`pytest-randomly`, on by default since #571), so
+  a test that passed yesterday and fails today may have been depending on what ran
+  before it. The header prints `Using --randomly-seed=<n>`; pass that seed back to
+  reproduce the failure, and `-p no:randomly` to pin collection order while bisecting.
+  The controller's seed reaches every xdist worker, so `-n auto` collects one order.
 - **`bun run test:run` measures no coverage.** The frontend gate is a separate command
   and CI runs it (`bun run test:coverage`); 168 green files still failed the job.
 - **Frontend commands run from `frontend/`.** At the repository root vitest finds no
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -39,6 +39,7 @@ from sqlalchemy.ext.asyncio import (
 
 from app.core.config import settings
 from app.db.base import Base
+from app.db.session import engine as app_engine
 
 # Where to connect to issue `CREATE DATABASE` / `DROP DATABASE`, neither of
 # which can run from inside the database it names. Every Postgres image this
@@ -240,7 +241,18 @@ async def engine(schema_url: str) -> AsyncGenerator[AsyncEngine, None]:
     the same per-function event loop the test does. The schema itself is built
     once by `schema_url`; here each test is handed a clean slate cheaply, by
     emptying the data rather than rebuilding the schema (#215).
+
+    The application's own engine is disposed first. It is a module-level object,
+    so its pool outlives the test that filled it, while anyio gives each test an
+    event loop of its own - and a connection created on a loop that has since
+    closed answers the next statement issued through it with `InterfaceError:
+    another operation is in progress`, from whichever unlucky test checked it
+    out. The tests that drive the real `get_db_session` used to each dispose it
+    on the way out, which only covers the pair of them; disposing on the way *in*
+    covers every test that could be handed one, whatever ran before it in this
+    worker. Disposing an empty pool costs nothing, which is the common case.
     """
+    await app_engine.dispose()
     engine = create_async_engine(schema_url)
     await _reset(engine)
     yield engine
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -251,8 +251,13 @@ async def engine(schema_url: str) -> AsyncGenerator[AsyncEngine, None]:
     on the way out, which only covers the pair of them; disposing on the way *in*
     covers every test that could be handed one, whatever ran before it in this
     worker. Disposing an empty pool costs nothing, which is the common case.
+
+    `close=False` because those connections belong to loops that are already
+    closed: closing one from the loop running now is an operation that cannot
+    succeed, and SQLAlchemy answers a failing close by logging it at ERROR and
+    carrying on. De-referencing the pool is what is actually meant.
     """
-    await app_engine.dispose()
+    await app_engine.dispose(close=False)
     engine = create_async_engine(schema_url)
     await _reset(engine)
     yield engine
--- a/backend/tests/test_random_order.py
+++ b/backend/tests/test_random_order.py
@@ -1,0 +1,33 @@
+"""A guard on the plugin that shuffles this suite.
+
+`pytest-randomly` is the only thing standing between an order-dependent test and
+a green build, and its absence is silent: without it the suite runs in collection
+order, `-p no:randomly` becomes a no-op, and nothing anywhere says so. That is
+not hypothetical. Both `docs/testing.md` and `.claude/rules/testing.md` described
+it as on by default while it was in neither the dependency group nor the
+lockfile, so the order-independence the docs called verified had never once been
+exercised (#571).
+
+The declaration is what makes the shuffle true for everybody who runs `uv sync`,
+so that is what is asserted - deliberately not that the shuffle is active in this
+process. Pinning collection order with `-p no:randomly` is the documented move
+while bisecting an order-dependent failure, and a guard that turned it red would
+be telling people to stop using it.
+"""
+
+from __future__ import annotations
+
+import tomllib
+from pathlib import Path
+
+PYPROJECT = Path(__file__).resolve().parents[1] / "pyproject.toml"
+
+
+def test_the_suite_is_shuffled_because_the_docs_say_it_is() -> None:
+    config = tomllib.loads(PYPROJECT.read_text())
+    dev_group: list[str] = config["dependency-groups"]["dev"]
+    assert any(spec.startswith("pytest-randomly") for spec in dev_group), (
+        "pytest-randomly is gone from the dev dependency group, so the suite now runs "
+        "in collection order. Either put it back or correct docs/testing.md and "
+        ".claude/rules/testing.md, which both promise a shuffled run and a live seed."
+    )
--- a/backend/tests/test_random_order.py
+++ b/backend/tests/test_random_order.py
@@ -8,26 +8,41 @@ it as on by default while it was in neither the dependency group nor the
 lockfile, so the order-independence the docs called verified had never once been
 exercised (#571).
 
-The declaration is what makes the shuffle true for everybody who runs `uv sync`,
-so that is what is asserted - deliberately not that the shuffle is active in this
-process. Pinning collection order with `-p no:randomly` is the documented move
-while bisecting an order-dependent failure, and a guard that turned it red would
-be telling people to stop using it.
+Two halves, because #571 was both: declared nowhere *and* installed nowhere. The
+declaration is what makes the shuffle true for everybody who runs `uv sync`; the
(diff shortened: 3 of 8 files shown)


backend/tests/integration/conftest.py
```diff
@@ -240,7 +241,18 @@ async def engine(schema_url: str) -> AsyncGenerator[AsyncEngine, None]:
...
+    so its pool outlives the test that filled it, while anyio gives each test an
+    event loop of its own - and a connection created on a loop that has since
+    closed answers the next statement issued through it with `InterfaceError:
+    another operation is in progress`, from whichever unlucky test checked it
+    out. The tests that drive the real `get_db_session` used to each dispose it
+    on the way out, which only covers the pair of them; disposing on the way *in*
+    covers every test that could be handed one, whatever ran before it in this
+    worker. Disposing an empty pool costs nothing, which is the common case.
     """
+    await app_engine.dispose()
```
DEENUU1: Small one. The connections still sitting in that pool belong to event loops anyio has already closed, and `dispose()` at its default `close=True` tries to close each of them from the *current* loop. SQLAlchemy swallows a failing close in `Pool._close_connection` and logs it at ERROR, so the worst case is log noise rather than breakage - but `close=False` is the documented option for a pool that is to be abandoned rather than closed ("the previous connection pool is de-referenced, and otherwise not touched in any way"), and it says what we actually mean here.

```suggestion
    await app_engine.dispose(close=False)
```

I haven't confirmed the close attempt actually raises on a dead asyncpg transport, so this may be closing nothing either way - but there's no case where trying is the better of the two.

DEENUU1 (commented):
Two commits, and the second is the interesting half - turning the shuffle on and then immediately fixing what it found is the whole argument for doing it. Moving the dispose out of two per-file teardowns and into the `engine` fixture, on the way *in*, is the right shape: it covers every test that could be handed a stale connection rather than the pair that happened to notice, and it leaves one explanation instead of three. I checked the third file that looked like a candidate, `test_platform_flows.py`, and it overrides `get_db_session` with the test session, so "those two files" in the body is accurate. The guard test earns its place too - the failure mode here was silence, so something has to be loud about it.

Three things I'd change, none of them blocking: `dispose(close=False)` in the fixture, the docs promising more than the seed can deliver under `-n auto` (which is exactly the failure this branch found), and the new test reading `pyproject.toml` differently from `test_coverage_gate.py`, which already does the same thing three files over.

One question rather than a change: #571 named `CLAUDE.md` as one of the two places carrying the false claim, and #579 stripped it. Do we want the shuffle back in its testing section, or is the pointer to `.claude/rules/testing.md` enough now?

docs/testing.md
```diff
@@ -23,6 +23,22 @@ suite is import-bound — every worker imports the app once — and gains nothin
...
+
+```bash
+uv run pytest tests/ -q --randomly-seed=1697040112   # reproduce a red run exactly
+uv run pytest tests/ -q -p no:randomly               # collection order, while bisecting
+```
+
+The seed is chosen once by the controller and handed to each xdist worker, so `-n auto`
+collects one order rather than four. Until #571 the plugin was documented but not
+installed, `-p no:randomly` was a silent no-op, and nothing had ever exercised the
+claim.
```
DEENUU1: Right about the order, and I'd add the bit it leaves out: the seed fixes the *order*, not the *split*. `make test` runs `-n auto --maxprocesses 4` with xdist on its default `--dist load`, which hands each test to whichever worker is free - timing, not the seed. So a failure that depended on what shared a worker (the `InterfaceError` this branch fixed is exactly that) does not come back by re-running `make test` with the seed. The snippet above already gets this right by running serially, so it's only the words that need to catch up - "reproduce a red run exactly" on line 33 promises more than the seed can give.

```suggestion
The seed is chosen once by the controller and handed to each xdist worker, so `-n auto`
collects one order rather than four. It does not fix which worker runs what: `make test`
leaves xdist on its default `--dist load`, which hands each test to whichever worker is
free. So a failure that depended on what shared a worker - the `InterfaceError` fixed in
this change is one - comes back by replaying the seed *serially*, as above, and not by
re-running `make test`. Until #571 the plugin was documented but not installed,
`-p no:randomly` was a silent no-op, and nothing had ever exercised the claim.
```

.claude/rules/testing.md
```diff
@@ -56,6 +56,11 @@ job always has one.
 
 Traps, each of which has cost a red job here:
 
+- **The order is shuffled every run** (`pytest-randomly`, on by default since #571), so
+  a test that passed yesterday and fails today may have been depending on what ran
+  before it. The header prints `Using --randomly-seed=<n>`; pass that seed back to
+  reproduce the failure, and `-p no:randomly` to pin collection order while bisecting.
+  The controller's seed reaches every xdist worker, so `-n auto` collects one order.
```
DEENUU1: Same point as the docs one, plus one trap the plugin brings that isn't written down anywhere: `pytest-randomly` also reseeds `random` to the same value before every test. So anything leaning on `random` for uniqueness is unique *within* a test and identical *across* tests. `app/commands/seed.py::random_email` is the only place I found that does, and nothing drives it against a real database today - so a note for the next person rather than a bug now.

```suggestion
- **The order is shuffled every run** (`pytest-randomly`, on by default since #571), so
  a test that passed yesterday and fails today may have been depending on what ran
  before it. The header prints `Using --randomly-seed=<n>`; replay that seed to get the
  same order back, and `-p no:randomly` to pin collection order while bisecting. The
  seed fixes the order, not which worker runs what - `--dist load` decides that on
  timing - so a failure that depended on what shared a worker only reproduces without
  `-n`. The plugin also reseeds `random` identically before every test: anything using
  it for uniqueness is unique within a test and repeats across them.
```

backend/tests/test_random_order.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import annotations
+
+import tomllib
+from pathlib import Path
+
+PYPROJECT = Path(__file__).resolve().parents[1] / "pyproject.toml"
+
+
+def test_the_suite_is_shuffled_because_the_docs_say_it_is() -> None:
+    config = tomllib.loads(PYPROJECT.read_text())
```
DEENUU1: `test_coverage_gate.py` reads the same file three tests over, in binary through `tomllib.load` - worth matching, since `read_text()` uses the locale encoding and TOML is defined as UTF-8:

```suggestion
    with PYPROJECT.open("rb") as handle:
        config = tomllib.load(handle)
```

Separately, and genuinely optional: the declaration is what everybody's `uv sync` follows, but it isn't quite what was broken in #571 - that was declared nowhere *and* installed nowhere. `assert find_spec("pytest_randomly") is not None` stays green under `-p no:randomly` (the package is installed, just not registered), so it doesn't contradict the docstring's reasoning, and it would also catch a stale venv or a `pytest` run from outside `uv run`.

backend/tests/integration/conftest.py
DEENUU1: Done in 7d270e19 - `dispose(close=False)`, with the reason in the docstring rather than a comment. I did go and check the thing I said I was unsure about: the close is attempted every time and cannot land, and SQLAlchemy logs it rather than raising, so it was noise. Three shuffled runs of `tests/integration` against a live pgvector Postgres after the change, 419 passed each time.

docs/testing.md
DEENUU1: Done in 88f857d5, roughly as suggested, plus the `random` reseed sentence folded in here as well - it belongs on the page people read before writing a test, not only in the rules file. The `# reproduce a red run exactly` comment on the snippet went with it.

.claude/rules/testing.md
DEENUU1: Done in 88f857d5.

backend/tests/test_random_order.py
DEENUU1: Done in 81b84e0d - both halves. The `tomllib.load` idiom now matches `test_coverage_gate.py`, and the installed check is in as a second test rather than a second assert, so a failure names which half broke. Both stay green under `-p no:randomly`, which I checked rather than assumed.

DEENUU1:
All four points from the review are in, as three commits on top:

- `7d270e19` `fix(tests)` - `dispose(close=False)` in the `engine` fixture
- `81b84e0d` `test` - the guard now checks declared *and* installed, in the `tomllib.load` idiom `test_coverage_gate.py` already uses
- `88f857d5` `docs(testing)` - what the seed reproduces and what it does not, plus the `random` reseed trap

And the open question, answered: no third copy in `CLAUDE.md`. It keeps its testing section to what changes what you do before writing a line, #579 stripped the false claim from it, and the shuffle is fully carried by `docs/testing.md` and `.claude/rules/testing.md`. Adding it back there would be the second explanation the docs rule exists to prevent.

Verified locally against a live pgvector Postgres: three shuffled runs of `tests/integration` (419 each), the full backend suite once (4381 passed, 6 skipped - one more than before, since the guard is two tests now), `make lint-backend` and `make lint-spelling` clean.

PR #608, opened 2026-08-11
test(prefect): run the suite against no Prefect server at all

### LINKED ISSUE #536: make test fails locally on any @flow test because backend/.env points Prefect at localhost:4200

DEENUU1:
## What this fixes

`make test` and `make test-fast` failed on any test that calls a Prefect `@flow` unless
a Prefect server happened to be listening on `localhost:4200`. The suite reached for one
because Prefect resolves its own settings from `backend/.env` — its settings model
carries `env_file=".env"` — so `PREFECT_API_URL=http://localhost:4200/api`, the line
`make dev` needs, was also the address a test's flow call tried to reach. CI never saw
it: with no `.env` there is no URL, so what a laptop ran was never what CI ran.

Closes #536

## What changed

- `backend/tests/conftest.py:90` — assign `PREFECT_API_URL` **empty** before Prefect is
  imported, beside the database name and password already seeded there. Deleting the
  variable would not do it: an unset variable leaves the dotenv source to answer, and
  the dotenv source is what holds the URL. An empty assignment outranks it, and Prefect
  reads an empty URL as no URL — it then starts a temporary server for the call, which
  is what CI has always done. Unconditional, so a developer with `make dev` up gets the
  same run CI gets rather than a different code path.
- Same block — name ephemeral mode (`PREFECT_SERVER_EPHEMERAL_ENABLED`, the canonical
  alias of the `PREFECT_SERVER_ALLOW_EPHEMERAL_MODE` in the issue) rather than inherit
  it, and raise the allowance for starting that server past Prefect's own 20 seconds.
  See below — 20 is a first run that fails.
- `backend/tests/test_prefect_test_environment.py` — new, pins all three properties
  against the resolved Prefect settings rather than against the environment, because the
  environment is only the mechanism and it has two ways of not working.
- `docs/testing.md` — a "Prefect, and why no test reaches a server" section next to the
  test-database ones, which document the neighbouring seeds in that same conftest block.

## How it failed before

Reproduced through the real mechanism rather than by exporting a variable: with
`backend/.env` pointing at a port nothing listens on, on `origin/main`,

```
tests/test_approval_expiry.py::TestTheScheduledSweep::test_the_flow_answers_with_what_it_expired
    httpx.ConnectError: All connection attempts failed
    RuntimeError: Failed to reach API at http://localhost:4299/api/
```

out of a test that patches `get_db_context` and `ApprovalService` and asserts a return
value. Same file, same `.env`, with the fix: 18 passed.

**The obvious fix has a second failure in it, which is why the timeout is raised.** With
the URL emptied and Prefect's default 20-second allowance, a machine whose `PREFECT_HOME`
has never been written — a fresh container, or a developer whose Prefect runs in Docker
and whose `~/.prefect` is therefore empty — answers `RuntimeError: Timed out while
attempting to connect to ephemeral Prefect API server` on the **first** run and passes
on every one after. The server migrates a SQLite database before it answers: measured
here at ~75s cold against a fresh `PREFECT_HOME` and ~7s warm. Trading a deterministic
failure for a first-run one is not a fix, so the allowance is 90s.

And with ephemeral mode *off* instead, a flow call does not fail fast — it retries for
75 seconds and then fails. Hence naming the mode rather than relying on the default.

## Testing

- `tests/test_prefect_test_environment.py` — 3 of its 4 assertions fail on `main` with a
  `.env` present; all pass here.
- `tests/test_approval_expiry.py` — the reported failure. Verified in all four
  combinations: `.env` pointing at a dead port × with and without the fix, and `.env`
  pointing at a live server × with and without it. Only the pre-fix dead-port case
  fails, which is the issue, and the post-fix live-server case passes — so the fix does
  not depend on the server being absent.
- Cold-start path verified directly with a throwaway `PREFECT_HOME`: fails at 20s
  without the raised allowance, 74.7s and green with it.
- `make lint` — the backend half passes. The frontend half needed a `bun install` in
  this worktree; no frontend file is touched.
- `make test`, `make docs-build` — see the checks on this PR.

## Noticed, not fixed

**The unit suite now starts a temporary Prefect server it has no use for.** ~7s warm,
once per pytest process that touches a flow. The one test that invokes a flow patches
every collaborator it has and asserts a return value, so
`approval_expiry_sweep_flow.fn()` would assert exactly the same thing with no server at
all — cutting that time from every local run *and* from CI's `test` job, which pays it
today. Left alone deliberately: #536's acceptance criteria ask that a test invoking a
`@flow` still pass with a server also running, and this is the only test that exercises
that path. Worth its own issue against #520 if you want it — say the word and I will
file it.

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -86,6 +86,37 @@ def _a_password_is_already_configured() -> bool:
 if not _a_password_is_already_configured():
     os.environ["POSTGRES_PASSWORD"] = "postgres"
 
+
+# Prefect reads `.env` itself - its own settings model carries `env_file=".env"` - so
+# `PREFECT_API_URL=http://localhost:4200/api`, which `backend/.env` sets for `make dev`,
+# is also the address a `@flow` call in this suite tries to reach. With no server up
+# that is `RuntimeError: Failed to reach API at http://localhost:4200/api/` out of a
+# test that patched every collaborator it has, which is #536: the failure belongs to the
+# environment and reads as the test's. CI never saw it, having no `.env` and therefore no
+# URL at all, so what a laptop ran was never what CI ran.
+#
+# Deleting the variable would not do it: an unset variable leaves the dotenv source to
+# answer, and the dotenv source is what holds the URL. An empty *assignment* is what
+# outranks it, and Prefect reads an empty URL as no URL - it then runs the flow against
+# a temporary server of its own, which is what CI has always done. Unconditionally,
+# because the point is that the two agree: a developer with `make dev` up gets the same
+# run CI gets rather than a different code path.
+#
+# Ephemeral mode is named rather than assumed. It is the default today, but with no URL
+# and no ephemeral server a `@flow` call does not fail - it retries for 75 seconds and
+# then fails, which is a worse version of the bug this removes.
+#
+# The allowance for starting it is raised past Prefect's own 20 seconds, which is not
+# enough for the first run on a given machine: the server migrates a SQLite database
+# under `PREFECT_HOME` before it answers, taking ~75 seconds where nothing has written
+# one yet - a fresh container, or a developer whose Prefect runs in Docker and whose
+# `~/.prefect` is therefore empty - and about seven on every run after. Twenty is
+# therefore `RuntimeError: Timed out while attempting to connect to ephemeral Prefect
+# API server` once, then a pass, which is the same defect wearing a different message.
+os.environ["PREFECT_API_URL"] = ""
+os.environ["PREFECT_SERVER_EPHEMERAL_ENABLED"] = "true"
+os.environ["PREFECT_SERVER_EPHEMERAL_STARTUP_TIMEOUT_SECONDS"] = "90"
+
 from collections.abc import AsyncGenerator
 from unittest.mock import AsyncMock, MagicMock
 
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -8,6 +8,7 @@ See: https://anyio.readthedocs.io/en/stable/testing.html
 
 import os
 import re
+import tempfile
 from pathlib import Path
 
 # Before anything imports `app.core.config`, and therefore before anything can
@@ -97,23 +98,34 @@ if not _a_password_is_already_configured():
 #
 # Deleting the variable would not do it: an unset variable leaves the dotenv source to
 # answer, and the dotenv source is what holds the URL. An empty *assignment* is what
-# outranks it, and Prefect reads an empty URL as no URL - it then runs the flow against
-# a temporary server of its own, which is what CI has always done. Unconditionally,
-# because the point is that the two agree: a developer with `make dev` up gets the same
-# run CI gets rather than a different code path.
+# outranks it, because Prefect's settings model carries `env_ignore_empty=False` and an
+# empty string is therefore an answer rather than a silence. That is Prefect's rule and
+# not ours: `app/core/config.py` sets it the other way, so the same line against one of
+# *our* settings is discarded and the `.env` answers anyway - which is why the password
+# above is checked for rather than emptied. Prefect reads an empty URL as no URL and
+# runs the flow against a temporary server of its own, which is what CI has always done.
+# Unconditionally, because the point is that the two agree: a developer with `make dev`
+# up gets the same run CI gets rather than a different code path.
+#
+# That server keeps its state in a SQLite database under `PREFECT_HOME`, which is
+# `~/.prefect` unless something says otherwise - a developer's own Prefect data, and the
+# file a locally run `prefect server` has open. Pointed at a directory of the tests' own
+# for the same reason as the database name above: a unit run has no business writing
+# there. One directory rather than one per process, because what costs is creating it.
 #
 # Ephemeral mode is named rather than assumed. It is the default today, but with no URL
 # and no ephemeral server a `@flow` call does not fail - it retries for 75 seconds and
 # then fails, which is a worse version of the bug this removes.
 #
-# The allowance for starting it is raised past Prefect's own 20 seconds, which is not
-# enough for the first run on a given machine: the server migrates a SQLite database
-# under `PREFECT_HOME` before it answers, taking ~75 seconds where nothing has written
-# one yet - a fresh container, or a developer whose Prefect runs in Docker and whose
-# `~/.prefect` is therefore empty - and about seven on every run after. Twenty is
-# therefore `RuntimeError: Timed out while attempting to connect to ephemeral Prefect
-# API server` once, then a pass, which is the same defect wearing a different message.
+# Creating that database is a migration, and the allowance for it is raised past
+# Prefect's own 20 seconds as headroom rather than because 20 has been seen to fail:
+# against a `PREFECT_HOME` nothing had written yet the whole start takes about six
+# seconds here and about nine on a CI runner, which is cold on every run. What 90 buys
+# is that the one step whose cost nothing here bounds - a migration on a contended
+# machine, or a temporary directory that has been swept - waits instead of failing a
+# suite that a second run would pass.
 os.environ["PREFECT_API_URL"] = ""
+os.environ["PREFECT_HOME"] = str(Path(tempfile.gettempdir()) / "agenticos-prefect-test")
 os.environ["PREFECT_SERVER_EPHEMERAL_ENABLED"] = "true"
 os.environ["PREFECT_SERVER_EPHEMERAL_STARTUP_TIMEOUT_SECONDS"] = "90"
 
--- a/backend/tests/test_prefect_test_environment.py
+++ b/backend/tests/test_prefect_test_environment.py
@@ -1,0 +1,54 @@
+"""Where a `@flow` call in this suite sends its API requests, which is nowhere.
+
+`tests/conftest.py` seeds `PREFECT_API_URL` empty before anything imports Prefect,
+because Prefect resolves its own settings from `backend/.env` - the file `make dev`
+needs, pointing at `localhost:4200`. A suite that reads it makes every `@flow` call
+depend on a server nobody started (#536).
+
+Three of these assert the consequence - what Prefect resolved - because a variable set
+too late, or set in a shape the dotenv source outranks, leaves the URL in place while
+looking correct. The fourth pins the shape itself, which is the half nothing else can
+see.
+"""
+
+import os
+
+from prefect.settings import get_current_settings
+
+
+def test_the_suite_points_prefect_at_no_server() -> None:
+    """Otherwise a `@flow` call reaches for whatever `backend/.env` names.
+
+    Asserting the resolved setting is what makes the check mean anything: the
+    variable is set before `app.core.config` is imported, and Prefect's own dotenv
+    source holds the URL that would answer if it were merely unset.
+    """
+    assert not get_current_settings().api.url
+
+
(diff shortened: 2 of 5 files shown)


backend/tests/conftest.py
```diff
@@ -86,6 +86,37 @@ def _a_password_is_already_configured() -> bool:
...
+# The allowance for starting it is raised past Prefect's own 20 seconds, which is not
+# enough for the first run on a given machine: the server migrates a SQLite database
+# under `PREFECT_HOME` before it answers, taking ~75 seconds where nothing has written
+# one yet - a fresh container, or a developer whose Prefect runs in Docker and whose
+# `~/.prefect` is therefore empty - and about seven on every run after. Twenty is
+# therefore `RuntimeError: Timed out while attempting to connect to ephemeral Prefect
+# API server` once, then a pass, which is the same defect wearing a different message.
+os.environ["PREFECT_API_URL"] = ""
+os.environ["PREFECT_SERVER_EPHEMERAL_ENABLED"] = "true"
+os.environ["PREFECT_SERVER_EPHEMERAL_STARTUP_TIMEOUT_SECONDS"] = "90"
```
DEENUU1, line 130: The temporary server is not sandboxed — it writes into whatever `PREFECT_HOME` says, which on a laptop is `~/.prefect/prefect.db`. I checked rather than assumed: with these three variables set, two concurrent flow calls left two rows in my own Prefect database, alongside runs from `make dev`.

```
$ sqlite3 ~/.prefect/prefect.db 'select name, created from flow_run order by created desc limit 2'
terrestrial-pony|2026-08-11 17:38:54
flying-wasp|2026-08-11 17:38:52
```

That is the same class of thing the block twenty lines above exists to stop: `POSTGRES_DB` is made per-pid so that no unit test can reach a developer's data, after #189 had two runs dropping each other's tables. Here the suite writes test flow runs into a developer's real Prefect state, and if their Prefect runs on the host rather than in Docker it is the same SQLite file a running `prefect server start` has open.

I'd seed a home for it in the same breath:

```python
os.environ["PREFECT_HOME"] = str(Path(tempfile.gettempdir()) / "agenticos-prefect-test")
```

Stable rather than per-pid on purpose — a fresh home per process would repay the cold migration on every run instead of once, which is the cost this file already raises the timeout for. And if the temp directory is swept, the 90 seconds is exactly the allowance that covers rebuilding it.

backend/tests/conftest.py
```diff
@@ -86,6 +86,37 @@ def _a_password_is_already_configured() -> bool:
...
+# and no ephemeral server a `@flow` call does not fail - it retries for 75 seconds and
+# then fails, which is a worse version of the bug this removes.
+#
+# The allowance for starting it is raised past Prefect's own 20 seconds, which is not
+# enough for the first run on a given machine: the server migrates a SQLite database
+# under `PREFECT_HOME` before it answers, taking ~75 seconds where nothing has written
+# one yet - a fresh container, or a developer whose Prefect runs in Docker and whose
+# `~/.prefect` is therefore empty - and about seven on every run after. Twenty is
+# therefore `RuntimeError: Timed out while attempting to connect to ephemeral Prefect
+# API server` once, then a pass, which is the same defect wearing a different message.
```
DEENUU1: The allowance is worth raising, but "a fresh container" is the one machine we can check, and it disagrees. A GitHub runner has no `~/.prefect` on any run, so CI cold-starts this server every time — and in this PR's own `test` job the flow test ran in 9.2 seconds end to end (13:10:05.99 → 13:10:15.14), while `main` has been doing the same cold start under Prefect's own 20 seconds for as long as that test has existed, without ever going red.

So ~75s is a macOS measurement, not a property of a cold `PREFECT_HOME`, and as written this says CI was failing a first run it has never failed. The next person to read it will go looking for a CI problem that isn't there, or stop trusting the paragraph. I'd say where it was measured and keep the 90 as headroom rather than as a fix for an observed timeout — the same edit is owed on `docs/testing.md:323`, which carries the sentence in the same shape.

backend/tests/conftest.py
```diff
@@ -86,6 +86,37 @@ def _a_password_is_already_configured() -> bool:
...
+# `PREFECT_API_URL=http://localhost:4200/api`, which `backend/.env` sets for `make dev`,
+# is also the address a `@flow` call in this suite tries to reach. With no server up
+# that is `RuntimeError: Failed to reach API at http://localhost:4200/api/` out of a
+# test that patched every collaborator it has, which is #536: the failure belongs to the
+# environment and reads as the test's. CI never saw it, having no `.env` and therefore no
+# URL at all, so what a laptop ran was never what CI ran.
+#
+# Deleting the variable would not do it: an unset variable leaves the dotenv source to
+# answer, and the dotenv source is what holds the URL. An empty *assignment* is what
+# outranks it, and Prefect reads an empty URL as no URL - it then runs the flow against
```
DEENUU1: nit: worth one clause saying this is Prefect's rule and not ours. `prefect.settings.Settings` sets `env_ignore_empty=False`, which is what makes the empty assignment beat the dotenv source; `app/core/config.py:26` sets it **True**, so the identical line against one of our own settings is discarded and the `.env` answers anyway. That asymmetry is why the password block below checks instead of assigning empty, and this comment is exactly where somebody will pick the idiom up to reuse it.

backend/tests/test_prefect_test_environment.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_that_server_is_given_longer_than_prefect_would_allow_it() -> None:
+    """Prefect's own 20 seconds is not enough for the first run on a fresh machine.
+
+    The ephemeral server migrates a SQLite database before it answers, which took
+    75 seconds on a `PREFECT_HOME` nobody had written yet and 7 on every run after.
+    Twenty is therefore a first run that fails and a second that passes.
+    """
+    assert get_current_settings().server.ephemeral.startup_timeout_seconds > 20
```
DEENUU1: This asserts against Prefect's current default rather than against what the conftest sets, so it holds for a range of values that would not fix anything — 21 passes, and so does a Prefect release that raises its own default to 30 while our seed stays where it is. Pin the number the file actually chose:

```suggestion
    assert get_current_settings().server.ephemeral.startup_timeout_seconds >= 90
```

backend/tests/test_prefect_test_environment.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_the_url_is_emptied_rather_than_removed() -> None:
+    """Removing it hands the question back to `backend/.env`, which is the defect.
+
+    An unset variable is not an answer to a dotenv source; an empty assignment
+    outranks it.
+    """
+    assert os.environ["PREFECT_API_URL"] == ""
```
DEENUU1, line 72: This one is the reason the file earns its place. The other three assert what Prefect resolved, which on CI is true whether or not the fix is present — no `.env`, no URL, nothing to outrank. This is the only assertion that fails if somebody "tidies" the seed into a `del os.environ[...]`, which is the exact regression that would then be invisible until it hit a laptop.

DEENUU1 (commented):
The diagnosis is right and the fix is the smallest one that works — I re-ran it against Prefect 3.8.1 here and confirmed both halves: with `backend/.env` present, `get_current_settings().api.url` is `http://localhost:4200/api`, and an empty assignment does outrank the dotenv source. The four-combination verification in the description (dead port × live server × with and without the fix) is the part that makes this trustworthy, and `test_the_url_is_emptied_rather_than_removed` is the right fourth test — it pins the one property CI structurally cannot see.

Two things I'd change before merging, neither blocking.

First, the ephemeral server writes into the developer's **real** `~/.prefect/prefect.db` — I ran two concurrent flow calls with exactly this conftest's env and found both flow runs sitting in my own Prefect database afterwards, next to rows from `make dev`. The block directly above this one sets `POSTGRES_DB` per-pid precisely so the suite cannot touch a developer's data; this reintroduces that class one library over. Comment on `conftest.py:118`.

Second, the reason given for the raised timeout overreaches, and this repo's own CI is the counter-example: a GitHub runner has no `~/.prefect` on any run, and the flow test in **this PR's** `test` job took 9.2s end to end (13:10:05.99 → 13:10:15.14) — and `main` has been doing that same cold start under Prefect's 20-second default for as long as that test has existed, green every time. The 90s allowance is still worth having; the sentence that says a fresh container times out is not true of the fresh container we actually run. Comment on `conftest.py:112`, and the same claim is on `docs/testing.md:323`.

One thing outside the diff: `.claude/skills/backend-tests/SKILL.md:56` and `.claude/rules/testing.md:112` both describe what this conftest block seeds and why, and after this change they cover two of the three. A sentence in each keeps them from being confidently incomplete.

On the "noticed, not fixed" — yes, file it. Worth saying in the issue what would actually be lost: with the seeds in place, `test_prefect_test_environment.py` is what holds the three properties, so switching `test_the_flow_answers_with_what_it_expired` to `.fn()` costs the end-to-end proof that a real `@flow` call completes with no server, not the guarantee itself. That's a fair trade for ~7s off every local run and off CI's `test` job, but it should be the trade the issue states rather than one somebody discovers.

backend/tests/conftest.py
DEENUU1: Done in 20eab883 — `PREFECT_HOME` is seeded to a stable directory under the system temp dir, and the comment says why it is one directory rather than one per process.

Issue #603, opened 2026-08-11
BFF route payloads and the workspace step detail reach the toast in English


DEENUU1:
Copy reaches the screen from two places that have no translator in scope, so
`check_i18n.py` cannot ask them for a key and #446 had to exempt both.

## The BFF route handlers

`src/app/api/**/route.ts` sits outside the `[locale]` segment. No locale is in
scope, `next-intl` has no message to resolve, and what the handler writes is a
wire payload:

```ts
if (!accessToken) return NextResponse.json({ detail: "Not authenticated" }, { status: 401 });
// …
return NextResponse.json({ detail: "Internal server error" }, { status: 500 });
```
`src/app/api/orgs/[id]/route.ts:11,20`

That payload is not internal. `parseErrorMessage` reads `body.detail` verbatim
when the backend envelope is absent — `src/lib/api-error.ts:94` — so the string
reaches a toast exactly as written, in English, under every locale.

**Measured: 132 offences across the BFF routes, 110 of them two strings** —
`Internal server error` ×65 and `Not authenticated` ×45, repeated across fifty
route files.

## The workspace step detail

`subjectOf` joins a grep pattern to the file it ran against with an English
preposition, in a module function that takes no translator:

```ts
return where === null ? pattern : `${pattern} in ${basename(where)}`;
```
`src/lib/tool-steps.ts:230`

Same shape, one word wide.

## Why #446 did not fix it

Neither is fixable where the string is written. The client is the only side that
knows the locale, so the remedy is a change of contract rather than a change of
wording — the handler returns a code the client maps to a message, or
`subjectOf` starts taking the translator its caller already holds. Both are
bigger than widening a guard's glob, and folding either into #446 would have put
an error-contract change inside a copy migration.

So #446 skips `src/app/api/` in the offence sweep (`UNTRANSLATABLE` in
`scripts/check_i18n.py`, documented there) and marks the join `i18n-exempt`.
Both point here.

## How you would know it was fixed

Under `pl`, killing the backend and triggering any request shows a Polish toast
rather than `Internal server error`; a grep step against a pattern and a path
reads with a Polish preposition. `UNTRANSLATABLE` is empty and the sweep reads
`src/app/api/` with nothing to report, and the `i18n-exempt` on
`tool-steps.ts:230` is gone.

## Where it sits

The same chain as #199 → #246 → #249 → #314 → #141 → #395 → #446: the reading
model, and what it does not read. This one is the part #446 could see and
deliberately did not take.

[...]

Issue #569, opened 2026-08-10
Frontend duplication & quality cleanup (audit 2026-08-10)


DEENUU1:
Batched small frontend items from the 2026-08-10 duplication audit. Each is Low / effort:s.

- [ ] Delete the dead registered query keys `conversations.count` (`query-keys.ts:170`) and, once #557 lands, `kb.detail`/`kb.documents`.
- [ ] `patchKB` (`use-knowledge-bases.ts:90`) swallows the refusal into a hardcoded toast while its siblings (`createKB`, `updateIngestion`, `deleteCollection`) rethrow — the edit dialog can't show why a rename failed. Rethrow like the siblings.
- [ ] Sandbox `usage` discriminator built inline (`use-sandbox-connections.ts:226`) instead of in the `sessions(id, usage)` factory.
- [ ] `files/[id]/route.ts:24` interpolates a bare `${id}` while its binary siblings `encodeURIComponent` (intra-family drift on the AUD-030 encoding).
- [ ] `members/page.tsx:87` gates on `can()` without `permissionsLoading`, where `runs/page.tsx:45` guards the same trap — controls flicker in (safe direction, not a leak).
- [ ] `message-item.tsx:404-509` — a ~105-line render IIFE inside a 315-line component; extract a `MessageBody`.
- [ ] `capability-settings.tsx` (1003) and `mcp-server-list.tsx` (978) concentrate all builder/MCP churn; extract the two MCP dialogs first. (Related #461.)
- [ ] The untranslated string clusters in `mcp-server-list.tsx` and `capability-settings.tsx` (scope labels, auth choices, toasts, a hand-rolled English plural) — part of the check_i18n.py blind-spot family; see #446.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

PR #609, opened 2026-08-11
fix(i18n): read a .ts file in the offence sweep, and migrate what it found

### LINKED ISSUE #446: check_i18n.py never reads a .ts file, so every hook toast is invisible to it
`scripts/check_i18n.py` walks `frontend/src/**/*.tsx` and nothing else:

```python
for path in sorted(SRC.rglob("*.ts*")):
    ...
    if path.suffix != ".tsx" or any(part in relative for part in SKIPPED_DIRS):
        continue
```
`scripts/check_i18n.py:590`

So every `.ts` file under `src/` — the hooks, the API clients, the
module-level tables, the route handlers — has never been read by the
offence sweep since the guard was written. A `.ts` file holds no JSX, but
it holds toasts, error messages and tables of labels, and all three are
copy somebody reads.

**Measured, not estimated: 406 offences across 91 files**, on `main` at
0.0.59 plus #427, #435 and #425's branch. Reproduce with:

```python
for p in sorted(SRC.rglob("*.ts")):
    if p.name.endswith(SKIPPED_NAMES) or "/dev/" in str(p): continue
    print(p, len(offences(p)))
```

## What breaks

Under `pl`, in English:

- `src/hooks/**` — `toast.success("Ingestion settings saved")`,
  `toast.error("Failed to update organization")`,
  `` toast.success(`Published v${version.version}`) ``. #425 fixed the
  nineteen of these whose words the catalog already held; the rest have
  no message at all.
- module tables of labels — `MCP_AUTH_LABEL` (`No credentials`,
  `API token`), `SCOPE_LABEL` (`Organization`, `You`) in
  `mcp-server-list.tsx`, `src/lib/agent-step-captions.ts`.
- `src/lib/seo.ts`, `src/lib/mcp-catalog.ts` — page descriptions and
  catalog blurbs.

`make lint` reads `.ts` as well as `.tsx` and reports nothing, having
reported the 406 before. The rules that only make sense against JSX
(`JSX_TEXT`, `MIXED`, `COUNT`, `LEAD`) must not fire on a `.ts` file —
`; return` and `a > b` are not text nodes — so this is not just widening
the glob.

## Where it sits

**Tests, CI, dependencies** on #168, and the same chain as
#199 → #246 → #249 → #314 → #141 → #395: the reading model, and what it
does not read. Closest sibling is **#395**, which argues the whole script
should become a `ts.createSourceFile` pass — a parser reads a `.ts` file
by construction, so **doing #395 first makes most of this free** and
whoever takes #395 should take this. Filed separately because the
backlog is a copy migration of its own size, the way #141 is.

Found working #425, which added two rules that *do* read `.ts` — both
anchored on the catalog rather than on the source — and which is how the
nineteen hook toasts were found at all. Pre-existing; #425's branch did
not cause it.

DEENUU1:
## What this fixes

`scripts/check_i18n.py` walked `frontend/src/**/*.tsx` and nothing else, so no `.ts` file under `src/` had ever been read by its offence sweep — the hooks, the API clients, the module tables of labels. **381 offences across 90 files**, measured on `main` at 0.0.101 rather than estimated, all of them English under `pl`. Two commits: the migration, then the guard that keeps it migrated, so each is green on its own.

Closes #446.

## What changed

**The reading model** — `scripts/check_i18n.py`

- `path.suffix` gates the four rules that anchor on a JSX bracket (`JSX_TEXT`, `MIXED`, `COUNT`, `LEAD`) to `.tsx`. Widening the glob alone would have reported the *syntax*: `; return` is a text node and `a > b` is a count. What is left reads a string literal wherever it sits.
- `UNTRANSLATABLE = ("app/api/",)` skips the BFF route handlers. They sit outside the `[locale]` segment, so no locale is in scope and there is nothing to resolve a message against. 110 of the 132 offences there are two strings repeated across fifty files.
- The import/export line-skip now keys on the **module specifier** rather than the keyword. In a `.tsx` file that is a no-op; in a `.ts` file the keyword form hid every `export const LABEL = "…"` and every default parameter on an `export function`. Two more real offences, zero false positives.
- `backend/tests/test_check_i18n_ts_files.py:1` — 16 tests, half of them the near-miss, because what a `.ts` file must *not* be reported for is as much of the specification as what it must.

**The migration** — 381 → 0, 233 messages added to `messages/en.json`

- 25 hooks read their toasts and error fallbacks from the catalog.
- The module tables hold **keys** and the component resolves them (`TOOL_CATALOG`, `MCP_AUTH_LABEL`, `BUILTIN_COMMANDS`, `ingestion-config.ts`'s `Choice` rows). The pure functions over them take the caller's `t` — `toolStep`, `toolCaption`, `ingestionProblems`, `labelOf`, `mergeWithUserCommands` — typed by `Translate` in `agent-step-captions.ts`.
- A workspace step's verb is **not** a parameter: `{verb} {subject}` is the `{noun}` defect of #362 under another name, so each tense is one message selecting on whether the call named a subject — `{named, select, no {Writing…} other {Writing {subject}}}`.
- `${count} file${count === 1 ? "" : "s"} uploaded` became an ICU `plural`.
- What genuinely cannot translate takes an exemption with its reason, never a key: the BFF helpers in `lib/`, `getErrorMessage`'s fallback, `seo.ts`'s tagline (read above the locale segment), and the two default field values posted to the API. #603 covers the ones that should stop needing one.

## How it failed before

Under `pl`, in English: `toast.success("Ingestion settings saved")` and eighteen like it; `MCP_AUTH_LABEL`'s `No credentials`; every `hint` in the ingestion dialog; `Rolled back - now running v${version}`. #425 could only see the nineteen whose words the catalog already held — the rest had no message at all.

## Three defects found on the way, fixed here

- **`mcp-server-list.tsx:715` rendered a catalog key on screen.** `AUTH_CHOICES` held `hint: "authTokenHint"` and the paragraph under the radio group printed it verbatim, in every locale. Neither a hardcoded string nor a missing key, so no guard could see it — the duplication rule pointed at the table once the catalog held the sentence.
- **`optional-setting.tsx` had a stranded `useTranslations` import mid-file**, immediately above `export const PROVIDER_DEFAULT = "Provider default"` — a migration somebody started and abandoned. Now `ui.providerDefault`.
- **The test translator was rebuilt on every call.** The real `useTranslations` is a `useMemo` over stable inputs; the mock's was not, so a hook putting `t` in a `useCallback`'s dependencies handed a new function to every render, and an effect keyed on that callback re-fired forever. The admin conversations screen and the RAG detail page both loaded in a loop and never left their spinners. `vitest.setup.ts` and the one spec with its own mock cache per namespace, and the callbacks an effect keys on resolve their message to a *string* above the callback so the dependency is stable by value rather than by memoisation holding.

## Removed, because copy nothing renders is deleted rather than translated

- `lib/mcp-catalog.ts`'s curated table of fourteen servers. The catalog the product shows is served by the backend from `app/core/catalog/mcp_servers.json`, is **fifty-nine** entries deep and has its own categories — this was a superseded copy that could only ever disagree with it. `logoDataUri` and the baked logos stay; `gen-mcp-logos.ts` takes its domains from the backend catalog, keyed on each entry's URL host because that is what the lookup asks for.
- `MCP_CATEGORIES`' four English headings, rendered nowhere.
- `summarizeIngestion`, called only by its own test.

## Testing

- `python3 scripts/check_i18n.py` — reports nothing where it reported 381; 3118 messages in `en.json`.
- `backend/tests/test_check_i18n_ts_files.py` — 16 new tests; 51 across the six guard files, all passing.
- **`make check` green** — `make lint`, backend at 100%, frontend 253 files / 3849 tests with the coverage gate met, `build-frontend`, `docs-build`, `audit`.
- `e2e` not run locally (needs a seeded backend). CI will say.
- **Frontend branch coverage is 97.62% against a 97.5% floor** — six branches of headroom, not a comfortable margin. Worth knowing before the next change lands near these files.

## Noticed, not fixed

- **#395** — the `ts.createSourceFile` rewrite the issue argues would make this free. It would, and it stays open and unblocked: a parser reads a `.ts` file by construction and would remove the looseness the `Translate` casts in the new tests work around. Deliberately out of scope here.
- **#603** (filed) — the BFF payloads and `tool-steps.ts`'s `in` join, which are exempted rather than fixed because neither can translate where it is written.
- **The `.tsx` clusters in #569** — `SCOPE_LABEL` in `mcp-server-list.tsx`, and `capability-settings.tsx`'s concatenated sentence and hand-rolled plural. All single-word or `.tsx`-shaped, so out of this issue's scope; the comment on #446 already names them.
- **`summarizeEmbedding`** in `ingestion-config.ts` is also read only by its own test. Left alone because nothing in this change forced the decision.
- **`mcp-logos.generated.ts` is unchanged — now #614.** Running `bun run gen:mcp-logos` produces host-based keys that match the lookups, which is a large diff and a network round trip per domain, so it is out of scope here. Filed because the current state is a defect rather than a cosmetic gap: measured, the intersection of the 14 baked keys with the hosts of every URL in `app/core/catalog/mcp_servers.json` is **empty**, so every badge falls through to Google's favicon service and the module's "renders OFFLINE" docstring is false. Pre-existing; this branch found it and fixed the generator.


DEENUU1:
CI: everything green except `e2e`, and that one is not this branch.

| | |
|---|---|
| `lint` · `test` · `test-frontend` · `docs` · CodeQL (×3) · Security Scan | pass |
| `e2e` | **fail, inherited** |

`e2e` died in **step 14, "Build the frontend"** — no product spec ran, and not in `[setup]` or `[seed]` either. The cause is `fonts.gstatic.com` answering **404** for the Geist Mono `.woff2` files, so `next/font/google` could not resolve them and Turbopack failed with 12 errors out of `src/app/layout.tsx`:

```
Received response with status 404 when requesting
  https://fonts.gstatic.com/s/geistmono/v6/or3NQ6H-...woff2
Module not found: Can't resolve '@vercel/turbopack-next/internal/font/google/font'
  Import trace: ./src/app/layout.tsx
```

Nothing here touches `layout.tsx`, the fonts or the build config, and `make build-frontend` passes locally — from a warm Next font cache, which is precisely why it does.

**#605 (`fix(frontend): self-host the three web fonts`) is open and fixes exactly this.** Until it merges, every branch's `e2e` build fails the same way. Not folding it in here: wrong scope, and it would put someone else's fix inside an i18n diff. Once #605 lands, this branch wants a rebase rather than a change.

DEENUU1:
Three things before this merges, none of them a change to the diff.

**The `e2e` failure was CDN flake, not #605's absence.** #610 built the frontend fine at
13:56 on a branch that does not self-host the fonts, so `fonts.gstatic.com` was answering
by then. The failed job is re-running on this same SHA
(`7c208936`) rather than waiting on #605 — which is still worth merging on its own terms,
since a green build off a public CDN is luck rather than a property of the repository.

**#610 deletes the file this PR's second commit edits.** `fix/i18n-guard-parser`, opened
an hour after this branch, replaces all 692 lines of `scripts/check_i18n.py` with
`frontend/scripts/check-i18n.ts`. The two overlap on 13 files — the script,
`frontend/messages/en.json`, `CHANGELOG.md`, `.claude/rules/frontend.md` and nine
components. Whichever lands second reconciles rather than rebases.

The order that keeps both fixes is **this one first**, and then #610 owes three things on
top — recorded on #168 so it is not lost in a PR comment:

1. **widen the parser's sweep to `.tsx?`.** `check-i18n.ts:954` filters to `.tsx` on
   purpose, with a comment saying the `.ts` offences are #446. After this PR that comment
   is stale in the worst way: the offences are migrated, so a `.tsx`-only sweep stays
   *green* while quietly reopening the hole that let 381 strings collect. The parser gets
   for free what the suffix gate here does by hand;
2. delete `backend/tests/test_check_i18n_ts_files.py`, whose subject it removes, porting
   the 16 cases — the near-misses especially — into `check-i18n.test.ts`;
3. union the `en.json`, `CHANGELOG.md` and `frontend.md` edits.

**#614 filed** for the item under "Noticed, not fixed": every baked MCP logo is keyed on a
brand domain the lookup never asks for. Measured after writing the PR body — the
intersection of the 14 keys in `mcp-logos.generated.ts` with the hosts of every URL in
`app/core/catalog/mcp_servers.json` is **empty**, `notion.so` and `github.com` are not
even the right domains — so this is a defect rather than a cosmetic gap and the module's
"renders OFFLINE" docstring is false today. Pre-existing; this branch found it and fixed
the generator without committing the 29 KB regenerated table. The PR body now points at
the issue.

PR #610, opened 2026-08-11
refactor(i18n): read the frontend with a parser, and fix the backlog it finds

### LINKED ISSUE #141: check_i18n.py passes over most real sentences, including the 404 page

### LINKED ISSUE #395: check_i18n.py misses a one-word template literal, and the reading model is why

DEENUU1:
## What this fixes

`scripts/check_i18n.py` has been patched for a new shape four times — #199, #246, #249,
#314 — and each fix was correct. The pattern is the problem: it reads a `.tsx` file as
text and decides per candidate whether it is looking at TypeScript or JSX, so every rule
carries a threshold standing in for a parse, and the next shape falls between two of
them. The last one was one word wide.

This replaces it with a parser and then fixes the whole backlog the parser can see:
**131 strings in 66 `.tsx` files** — 64 template literals, 62 text nodes, 4 strings and a
toast — measured by running the guard in this branch against the tree before the sweep,
plus **6 in 4 `.ts` files** once merging `main` restored the `.ts` sweep. 137 in 70 files
altogether; the two halves are counted separately because they were found by two
different versions of this branch, and the section at the bottom says why.

Closes #395. Closes #141 — the same mechanism, and the reason #395 said whoever takes
either should take both.

## What changed

- `frontend/scripts/check-i18n.ts` — the guard, walking `JsxText`, `JsxExpression`,
  `StringLiteral` and `TemplateExpression` through `ts.createSourceFile`. `MIXED`,
  `COUNT`, `LEAD`, `JSX_TEXT`, `mask_generics`, `readable`, `NOT_PROSE` and both
  word-count thresholds are **deleted rather than ported**: the four text rules become
  one *phrase* rule, because a node the formatter broke over three lines is one node.
- `frontend/scripts/check-i18n.test.ts` — 52 specs, replacing the five
  `backend/tests/test_check_i18n_*.py` files. Each `describe` names the issue it holds
  shut.
- `Makefile`, `package.json`, `.pre-commit-config.yaml`, `.github/workflows/ci.yml` —
  the guard runs from `make lint-frontend` (`bun run check:i18n`), which is where a
  TypeScript guard belongs. There is a pre-commit hook now; there was not one before.
- `frontend/messages/en.json` — **+134 messages, −34 dead keys**.
- 68 component files — the copy moved out. 128 of the 131 became messages; three took a
  reasoned `i18n-exempt` instead, and say why on the line above them.
- `.claude/rules/frontend.md`, `docs/channels.md`, `CHANGELOG.md`.

## How it failed before

Three shapes, all rendering in English under `pl`:

```tsx
frontend/src/components/kb/reusable-integrations.tsx:187
aria-label={`Remove ${source.name}`}          // below TEMPLATE's two-word threshold

frontend/src/app/not-found.tsx:23
<h1 className="…">
  Page not found                              // no bracket on the line, so no rule matched
</h1>

frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx:566
title={`Archive ${agent.name}?`}              // a bare `?` was on the machine-read list
```

Three policy changes, each a threshold the parser makes answerable rather than a rule
loosened:

- **whitespace, not a word count**, separates a sentence from an identifier being
  assembled — a space between a word and the `${` is prose, and nothing assembles an
  identifier with spaces in it, so `` `audience${key}Hint` `` still passes;
- **a bare `?` left the machine-read list.** It was there for query strings and it was
  exempting every confirm dialog in the tree — 8 of them;
- **a number and its unit is a formatter**, not a message. That is the fourteen
  `${bytes} KiB` shapes #395 left open: a unit is not translated, so they take one rule
  rather than fourteen exemptions.

## Testing

- `frontend/scripts/check-i18n.test.ts` — 52 specs, and **mutation-checked rather than
  assumed**, because #403 found that none of the ten Python tests exercised
  `mask_generics` at all. Stubbing `isTemplateCopy`, the exemption anchoring, the `?`
  change, or the phrase position each turns the suite red. The first version of the
  exemption spec did *not* discriminate, and a second case was added until it did.
- `make check` — green, except `db-check`, which needs a local Postgres. No model or
  migration is touched, so `alembic check` has nothing to say; CI runs it against a real
  database.
- `make lint` reports none of the 137 above, and `bun run check:i18n` answers in about a
  second.
- The 131 is the *final* guard's count for `.tsx`. An earlier revision of this branch
  reported 137 there — a collision with the total above, and worth disentangling: those
  six went away by the guard's own doing rather than the sweep's, three in
  `not-found.tsx` and `global-error.tsx` that the exemption fix below now covers, and
  three `${n} ctx` labels the unit rule answers.
- Two test breakages were mine and are fixed rather than worked around: a local
  `next-intl` mock that modelled `t` but not `t.rich`, and `chat.sourceNumber` holding
  `[{n}]`, which `messages/catalog.test.ts` rightly reads as a fragment of source.
- **e2e ran in CI and passed** (4m46s). Every e2e selector naming this copy —
  `Remove X`, `Rotate X`, `Delete X`, `Open X`, `Apply 1 change`, `Provider default` —
  resolves to the same rendered text, because the keys reproduce it verbatim. Seven
  messages had swapped `&apos;` for a curly `’`, which is not verbatim and not the
  catalog's convention (130 straight apostrophes to 8 curly); those are straight now.

## Also fixed, because the parser exposed it

An `i18n-exempt` applied to its own line and the next. The three exemptions in
`app/not-found.tsx` are written above an `<h1>` whose words are on the *third* line —
the opening tag carries four Tailwind classes — so the window covered the tag and missed
the copy. Nothing had noticed, because a text node alone on its line matched no rule at
all. An exemption now covers the element it opens, and a reason worth two lines covers
the code under the whole comment block.

## Merging `main`, and the guard it would have taken out

This branch was cut from `1f40a588`, and `a764e81f` (#609) landed after it — closing
#446 by putting `.ts` files in the offence sweep, gating the bracket rules on the
suffix, skipping `src/app/api/**`, and migrating all 381 offences. The port here had
narrowed the sweep back to `*.tsx` and said in three places that #446 was still open
with 406 offences to go, which was true of the merge base and not of `main`. Merging
that way would have removed the guard #609 added.

The sweep reads both suffixes now, by the same rules; `src/app/api/**` keeps its skip,
at the sweep rather than in a rule, because a route payload is a string a rule reads
perfectly well and what excuses it is where it lives (#603). `isSwept` is exported so
that has a spec. `backend/tests/test_check_i18n_ts_files.py` came in with #609 and is
ported to `check-i18n.test.ts` rather than dropped.

Four defects the review turned up, all fixed here:

- **`` `Bearer ${token}` `` was reported as copy** — an auth header value is the one
  header shape `MACHINE_READ` cannot see, holding no punctuation at all. Latent only
  while the sweep skipped `.ts`.
- **`missingKeys` checked a key against the wrong namespace** when a file held two
  translators, and unioning them hid eight keys rendering as raw key strings on the
  admin conversations page. A call now resolves to the nearest enclosing binding —
  by scope, because one page binds `getTranslations("pages.meta")` in
  `generateMetadata` and `getTranslations("pages.auth")` below it, both called `t`,
  and keying on the name alone reports 157 live keys as missing.
- **A ternary between two one-word labels in a readable prop was read by nothing** —
  #395's own defect wearing a ternary.
- **A toast holding a sentence was reported twice.** Four `UNITS` entries a
  two-character word regex can never reach are deleted, and eight source comments
  naming the now-deleted `scripts/check_i18n.py` are corrected.

## Noticed, not fixed

- **#621** — `timeAgo` formats a date older than a week as `en-US` in every locale.
  Reachable only after the sweep widened to `.ts`, and not fixable with a translator:
  `toLocaleDateString` needs the active locale, so it is eleven components either way.
- **`pl.json` gains nothing.** English is the source language and `src/i18n.ts` merges
  it under every locale, so these 131 render exactly as they did. Translating them is
  the separate, ongoing piece of work that `pl.json`'s 491-of-2985 coverage already is.
- **The `?` change surfaced two `confirm()` dialogs** in `mcp-server-list.tsx` that are
  browser-native rather than a component. They are translated now, but a native
  `confirm()` on a destructive flow is inconsistent with every other one in the product
  and untestable from Playwright — #611.
- **Twelve local `next-intl` mocks model `t` but not `t.rich`** — #612. All green today,
  because none of their components reads a rich message yet; one did here, which is the
  `kb-detail-sections` breakage above. Since the guard now pushes a split sentence
  towards `t.rich`, they will be hit. Left alone rather than folded in: twelve test files
  for no behaviour change on this branch.
- **#168, the issue map, has an empty body**, so "read the map before filing" has nothing
  to read. Left a comment there recording that this chain is closed and #446 is
  unblocked, rather than writing the map.


DEENUU1 (commented):
Reading this back as a maintainer rather than as the person who wrote it, and the core of it holds up: the parser is the right call, #141's plain multi-line node is genuinely caught now, and the 53 specs assert both directions with each `describe` pointing at the issue it holds shut.

Two things have to change before it merges, and both come from one place — this branch is cut from `1f40a588`, and `a764e81f` (#609) landed on `main` a few hours later. **#609 closed #446**: it put `.ts` files in the offence sweep, gated the bracket rules on the suffix, and migrated all 381 offences. So merging this as it stands quietly takes that guard back out, and takes the `src/app/api/**` skip with it. The PR is `CONFLICTING` anyway, so `main` has to come in first regardless — details on the two lines below, including the measured cost of putting `.ts` back (eight offences, not 406).

The rest is small: a `` `Bearer ${token}` `` false positive that goes live the moment the sweep is fixed, a `missingKeys` false positive in a file holding two translators, four entries in `UNITS` that can never match, and seven messages that quietly changed an apostrophe.

Two things for the body itself rather than a line:

- **Eight source comments still name `scripts/check_i18n.py`**, which this PR deletes: `frontend/messages/catalog.test.ts:15`, `frontend/src/i18n.ts:23`, `frontend/src/hooks/use-members.ts:14`, `frontend/src/components/rag/kb-stats-strip.tsx:15`, `frontend/src/components/whole-sentences.integration.test.tsx:25`, `frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx:79`, `frontend/src/app/[locale]/(dashboard)/rag/copy.integration.test.tsx:17`, `frontend/src/app/[locale]/(dashboard)/rag/[id]/delete-collection.integration.test.tsx:217`. `docs/` and `.claude/` are clean — these are the leftovers. A comment naming a file that does not exist is the failure mode `CLAUDE.md` calls out by name.
- **"e2e not run" is out of date.** CI ran it and it passed in 4m46s, so the by-hand check has a real answer behind it now. Worth correcting in the body, since the body is what survives as the record.

frontend/scripts/check-i18n.ts
```diff
@@ -0,0 +1,1017 @@
...
+  // not dead. `product` drops the playground, whose copy is nobody's to translate.
+  // `sweep` is `.tsx` alone - the offence rules are JSX rules, and the 406 offences a
+  // `.ts` file holds are #446 rather than part of this. Tests are out of all three: a
+  // test names its copy.
+  const sources: Source[] = sourceFiles(SRC).map((path) => ({
+    path,
+    text: readFileSync(path, "utf8"),
+  }));
+  const product = sources.filter(({ path }) => !SKIPPED_DIRS.some((dir) => path.includes(dir)));
+  const sweep = product.filter(({ path }) => path.endsWith(".tsx"));
```
DEENUU1: This is the line that has to change, and it is not the parser's fault — the branch went stale underneath it.

I cut this from `1f40a588`. `a764e81f` (#609) landed on `main` at 16:35 and did the thing the docstring at the top of this file says is out of scope: `.ts` files in the offence sweep, the four bracket rules gated on `path.suffix`, all 381 offences migrated, `src/app/api/**` skipped as untranslatable. **#446 is CLOSED.** So merging this takes that guard back out — every `toast.success("…")` in `src/hooks/**` goes back to invisible, and nothing stops the copy #609 just moved from coming straight back. `duplicatedInSource` catches a re-introduced *duplicate* of an existing message, but not a new sentence, so it is not a substitute.

The good news is that it is nearly free now, because the migration is already done. I ran this branch's `offences()` over `main`'s tree, `.ts` files only, skipping `src/app/api/`:

```
src/hooks/use-chat.ts:331: template '`\n\n❌ Error: ${message || t("unknownError")}`'
src/lib/admin-auth.ts:26: template '`Bearer ${accessToken}`'
src/lib/chat-sources.ts:37: template '`chunk ${item.chunk}`'
src/lib/ingestion-config.ts:384: template '`${kb.embedding_model} · ${kb.embedding_dim.toLocaleString()} dimensions`'
src/lib/platform-proxy.ts:120: template '`Bearer ${accessToken}`'
src/lib/utils.ts:78: template '`${Math.floor(diff / 60)}m ago`'
src/lib/utils.ts:79: template '`${Math.floor(diff / 3600)}h ago`'
src/lib/utils.ts:80: template '`${Math.floor(diff / 86400)}d ago`'
```

Eight offences in six files, not the 406 the docstring quotes. Six are real copy — `chunk {n}` is the shape `.claude/rules/frontend.md` names outright, and `{n}m ago` is a relative time Polish does not build by suffixing a letter. Two are the `Bearer` false positive I have left a note about on `MACHINE_READ`.

So: merge `main`, then widen the sweep and keep the route handlers out of it.

```suggestion
  const sweep = product.filter(({ path }) => !relative(SRC, path).startsWith("app/api/"));
```

The `.ts` bullet in this file's docstring and the "#446 is now reachable and still out of scope" paragraph in the PR body both have to go with it, and `.claude/rules/frontend.md` needs `main`'s version of the two bullets #609 wrote rather than this branch's.

frontend/scripts/check-i18n.ts
```diff
@@ -0,0 +1,1017 @@
...
+
+import ts from "typescript";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const FRONTEND = resolve(HERE, "..");
+const ROOT = resolve(FRONTEND, "..");
+const SRC = join(FRONTEND, "src");
+const CATALOG = join(FRONTEND, "messages", "en.json");
+
+const SKIPPED_DIRS = ["/dev/"];
```
DEENUU1, line 92: Worth saying explicitly here, because it is the mistake I would make next: `src/app/api/` must **not** join `SKIPPED_DIRS`.

`SKIPPED_DIRS` filters `product`, and `product` is what `duplicatedInSource` reads. A route handler is precisely where that rule earns its keep — it is the only rule that catches `{ detail: "Upload failed" }` sitting beside a catalog message, and there is a spec for exactly that at `check-i18n.test.ts:563`. Skipping the directory wholesale would turn that spec into a test of nothing.

`main` has it right: the exclusion is sweep-level (`UNTRANSLATABLE`, applied after `product`), so the offence rules skip a handler while the catalog rules keep reading it. That is why the fix on line 954 is a filter there rather than an entry here. A one-line comment saying so would stop the next person collapsing the two.

frontend/scripts/check-i18n.ts
```diff
@@ -0,0 +1,1017 @@
...
+/**
+ * A URL, a query string, a CSS value, a header - built by interpolation, read by a
+ * machine.
+ *
+ * A bare `?` used to be on this list, and it exempted every question in the tree:
+ * `` `Disconnect "${connection.name}"?` `` is a confirm dialog, not a query string, and
+ * it sat behind the character that was standing in for one. A question mark only says
+ * "machine" next to an `=`, which `[=]` already catches on its own.
+ */
+const MACHINE_READ = /[/&=<>#]|\b(?:px|rem|deg|vh|vw|attachment)\b/;
```
DEENUU1: The comment above this names "a header" as one of the four things the rule keeps out, and that is the one header shape it cannot see — because an auth header value has no punctuation in it at all.

Verified against the rule as written:

```
offences("x.ts", "const h = { Authorization: `Bearer ${token}` };")
→ template '`Bearer ${token}`'
```

`` `Token ${token}` `` the same. `Bearer` is a word, whitespace, then the hole — which is the prose shape the whitespace rule is looking for — and none of `[/&=<>#]` appears in the literal part, so nothing stops it. It is invisible today only because both call sites are in `.ts` (`src/lib/admin-auth.ts:26`, `src/lib/platform-proxy.ts:120`) and the sweep does not read those. Fix line 954 and it is live.

I would rather teach the rule than exempt two lines, since the next one will be written inside a component and the exemption will not be there:

```suggestion
const MACHINE_READ = /[/&=<>#]|\b(?:px|rem|deg|vh|vw|attachment|Bearer|Basic)\b/;
```

and a spec beside the URL and CSS one at line 304, so this does not have to be rediscovered.

frontend/scripts/check-i18n.ts
```diff
@@ -0,0 +1,1017 @@
...
+  "vh",
+  "vw",
+  "deg",
+  "ms",
+  "s",
+  "min",
+  "h",
+  "hr",
+  "d",
+  "B",
```
DEENUU1: Four of these can never match, and the comment above says "each entry is a token this product actually renders" — so they read as load-bearing.

`isFormatter` asks `UNITS.has(word)` for each result of `words()`, and `words()` is `WORDS = /[A-Za-z]{2,}/g`. A single-character token is never a word, so `"s"`, `"h"`, `"d"` and `"B"` are never looked up. `` `${n} s` `` passes for a different reason entirely — `isCopy` rejects it on `stripped.length < 2` before `isFormatter` is reached.

```suggestion
  "min",
  "hr",
```

Same for `"em"` further up, which is fine at two characters. Nothing changes behaviourally; it just stops the list claiming to do something it does not.

frontend/scripts/check-i18n.ts
```diff
@@ -0,0 +1,1017 @@
...
+ * that is not here is copy nothing reports, and it stays invisible until somebody
+ * notices it in English under `pl`. `noun` and `term` were added after exactly that
+ * (#362) - `<Pager noun="skills">` at six call sites and `<Fact term="Chunking">` at
+ * four, all rendered verbatim in every locale.
+ *
+ * A parser could read every prop's value instead of trusting its name, and that is
+ * not the same rule: `variant="destructive"`, `side="bottom"` and `type="submit"` are
+ * string-valued props nobody reads, so the widening would trade this list for a list
+ * of exclusions the same length. What the parser does buy is the *value*: a template
+ * literal or a ternary in a readable prop is read here, where `ATTR` matched a
```
DEENUU1: "a ternary in a readable prop is read here" is half true, and the half that is not is worth not claiming.

A ternary of two *sentences* is read, because `readString` picks up each branch. A ternary of two one-word branches is not:

```
offences("x.tsx", 'const A = () => <button aria-label={busy ? "Saving" : "Save"} />;')
→ []
```

`readString` wants `/^[A-Z][^\n]*\s/` — a capital and a space — so a single word in a readable prop only gets caught when it arrives as a bare string literal, which the `isJsxAttribute` branch handles, or inside a template literal. Through a ternary it falls between the two.

No regression — `ATTR` could not see it either — but this docstring is the reference for what the rule covers, so I would either narrow the sentence to "a template literal, or a ternary of sentences", or make the `isJsxAttribute` branch walk a `ConditionalExpression` in a readable prop and pin it with a spec. I lean towards the second; a one-word `aria-label` toggle is exactly the shape #395 was about.

frontend/scripts/check-i18n.ts
```diff
@@ -0,0 +1,1017 @@
...
+ * file whose translator is scoped to one namespace while its keys were filed under
+ * another, which renders the key itself on screen and reports an error nobody sees
+ * until a person opens that page.
+ */
+export function missingKeys(fileName: string, text: string, catalog: unknown): Offence[] {
+  const { namespaces, calls } = keyReads(fileName, text);
+  const scoped = [...namespaces].filter(Boolean);
+  if (!scoped.length) return [];
+  return calls
+    .filter(({ key }) => !scoped.some((namespace) => holds(catalog, `${namespace}.${key}`)))
```
DEENUU1: This reports a live key as missing when one file holds both a scoped and an unscoped translator, and that is new — the Python version read only calls named literally `t`, so an aliased root translator was invisible to it.

`keyReads` collects calls from every bound name and unions the namespaces; `scoped` then drops `""`. So a key read through the root translator is looked up under the *other* binding's namespace:

```
catalog = { agents: { title: "Agents" }, common: { cancel: "Cancel" } }

const t = useTranslations("agents");
const tRoot = useTranslations();
<p>{t("title")}{tRoot("common.cancel")}</p>

→ [{ line: 5, what: "common.cancel (in agents)" }]
```

`common.cancel` exists and renders fine. The build fails asking for `agents.common.cancel`, and the only way to make it pass is to add a key that must not exist — which is the #348 trap the module docstring warns about, arriving through the guard's own front door.

Latent rather than live: `add-model.tsx:116,120` and `chat-model-picker.tsx:71,75` both hold the pair, and both only pass `tRoot` to a helper. One literal call in either file and it fires.

The fix is to keep the binding rather than union it — `bound` becoming a `Map<string, string>` of name to namespace, so `calls` carries the namespace it was read through and `missingKeys` checks each key against its own. That also lets `unreadKeys` stop over-reading, which the `Reads` docstring currently defends as the safe direction; it would no longer need to be.

frontend/messages/en.json
```diff
@@ -1441,6 +1478,7 @@
     "strategy": "Strategy",
     "temperature": "Temperature",
     "use": "Use in…",
+    "useCollectionSettings": "Use the collection’s settings",
```
DEENUU1: This one changes what renders. `&apos;` is `'` (U+0027); this is `’` (U+2019), so the button reads `Use the collection’s settings` where it read `Use the collection's settings`.

Seven messages did the same on this branch — `useCollectionSettings`, `toolNameSteersModel`, `accountResponsibility`, `aiOutputDisclaimer`, `checkSpamFolder`, `operatingSystemHeading`, `serviceEnvironmentDescription`. The catalog's own convention is the straight one: 130 straight apostrophes against 8 curly, and 7 of those 8 arrived here.

Nothing selects on any of them — I grepped `e2e/` and the unit suites — so nothing is broken. But the PR body says the keys reproduce the copy verbatim, and for these seven they do not. `&apos;` only exists in the source because eslint forbids a bare `'` in JSX text; a JSON message has no such constraint, so the straight character carries over cleanly:

```suggestion
    "useCollectionSettings": "Use the collection's settings",
```

Either make all seven straight, or say in the body that the apostrophes were deliberately upgraded. Silent is the one option I would not take, since the next person diffing rendered copy has no way to tell which it was.

CHANGELOG.md
```diff
@@ -17,6 +17,52 @@ Two things are versioned separately from this file and worth knowing about:
...
+  existed to keep `` `audience${key}Hint` `` out. It is now
+  `frontend/scripts/check-i18n.ts`, walking `JsxText`, `JsxExpression`,
+  `StringLiteral` and `TemplateExpression` through `ts.createSourceFile`: a node the
+  formatter broke over three lines is one node, a type argument list is not JsxText
+  at all, and a comment is invisible rather than blanked. `MIXED`, `COUNT`, `LEAD`,
+  `JSX_TEXT`, `mask_generics`, `readable`, `NOT_PROSE` and both word-count
+  thresholds are deleted rather than ported; every policy rule carries over.
+  Runs from `make lint-frontend` (`bun run check:i18n`) and a new pre-commit hook,
+  with `frontend/scripts/check-i18n.test.ts` in place of the five
+  `backend/tests/test_check_i18n_*.py` files. Closes #395 and #141. (#597)
```
DEENUU1: Wrong number. There is no PR #597 — #597 is an open issue about the "could not be priced" caveat counting top-level runs. This is #610.

```suggestion
  `backend/tests/test_check_i18n_*.py` files. Closes #395 and #141. (#610)
```

Worth a look at the second entry's trailer too, in case the same number was copied twice.

frontend/scripts/check-i18n.ts
DEENUU1: Fixed. Merged `main` and widened the sweep: `product.filter(({ path }) => isSwept(relative(SRC, path)))`, so both suffixes are read by the same rules. Cost was the eight I measured minus the two `Bearer` false positives, so six real strings, all migrated - `timeAgo`'s three labels as ICU plurals taking the caller's translator, the stream-error prefix, `chunk {number}`, and `summarizeEmbedding` deleted rather than translated (no caller but its own test). The module docstring's `.ts` bullet and the PR body's "#446 still out of scope" paragraph are gone with it. `backend/tests/test_check_i18n_ts_files.py` came in with #609 and is ported to `check-i18n.test.ts`, nine specs.

frontend/scripts/check-i18n.ts
DEENUU1: Right, and taken as written: the skip is a sweep-level filter, not a `SKIPPED_DIRS` entry, so `duplicatedInSource` keeps reading the handlers and the spec at `check-i18n.test.ts` that pins `{ detail: "Upload failed" }` still tests something. Pulled it into an exported `isSwept` with the reasoning in its docstring, so it has two specs instead of a comment - one that a route payload is the shape that would otherwise be reported, one that the match is below `src` so `components/app/api/thing.ts` is not caught.

frontend/scripts/check-i18n.ts
DEENUU1: Fixed as suggested - `Bearer|Basic` on the machine-read list, with a paragraph saying why the character class cannot reach it. Spec added beside the URL and CSS one: `isTemplateCopy` false for both schemes, and `offences` clean on `{ Authorization: `Bearer ${accessToken}` }`. Checked it does not loosen anything - `${n} KiB` and `${n} min` still pass, `${n} files left` still refused.

frontend/scripts/check-i18n.ts
DEENUU1: Fixed - the four are gone and the docstring now says the constraint out loud, that every entry is two characters or more because `words()` matches `[A-Za-z]{2,}`. Confirmed the behaviour is unchanged: `${n} s` passes on `isCopy`'s length test, which is where it was passing before.

frontend/scripts/check-i18n.ts
DEENUU1: Took the second option: `labelChoice` reads a `ConditionalExpression` in a readable prop, and the docstring is narrowed to match what the code does. A label is capitalised or holds a space, which is what keeps `dir === "asc" ? "desc" : "asc"` out - `isCopy` says yes to `desc` on its own, so it needed the discriminator. Two specs, one each way.

frontend/scripts/check-i18n.ts
DEENUU1: Fixed, and the first attempt at it was wrong in a way worth recording. Keying `bound` on the name alone made this worse, not better: one page binds `getTranslations("pages.meta")` in `generateMetadata` and `getTranslations("pages.auth")` in the component below, both called `t`, so the second declaration won and **157 live keys across 31 files** read as missing. That is why unioning was load-bearing. It resolves by scope now - nearest enclosing binding of that name, falling back to every namespace that name takes where the walk finds none. Three specs, including one that the narrowing still reports a key its own namespace does not hold. It also found a real bug: eight keys on the admin conversations page were read through `useTranslations("admin")` while only `pages.admin` held them, so they rendered as their own key strings in every locale. Fixed here, and the four `2`-suffixed twins the union had kept alive are deleted.

frontend/messages/en.json
DEENUU1: Made all seven straight, so the catalog is 137 straight apostrophes and no curly ones. The PR body's "verbatim" claim now holds, and it says so where it lists the e2e selectors.

CHANGELOG.md
DEENUU1: Fixed - three occurrences of `(#597)` in the Unreleased section, all now `(#610)`. Good catch on the second entry.

DEENUU1:
The red `test` job here is not this branch: it failed at 99.98% on `app/core/catalog/__init__.py:79`, the `continue` in `custom_icon`, which is reachable only when `glob` yields a non-matching mark first — `scandir` order, hash order on the runners. The same failure hit #623, a commit that changes three version strings and nothing else. Filed as #625 and fixed in #626; once that lands this branch needs a re-run, not an edit.

Issue #194, opened 2026-08-05
spend_by_provider and spend_by_key cannot attribute a delegation's vendor or key


DEENUU1:
## What breaks

`GET /api/v1/runs/spend` answers "what did we spend at each vendor" from
`agent_runs.provider`, one row per run. A delegation to an agent on a *different*
provider spends money at that vendor and the split cannot say so.

A run has one spend ledger, so a delegate's tokens land in the parent run's
`cost_usd`, and that row carries the **parent's** provider. The child row is a
second copy of the same money, so it cannot be added:

* counting both, $1.00 of work reads as `openai 1.00` + `anthropic 0.40` = $1.40,
  which is more than the organization was billed and more than the cap enforces;
* counting only parents - what `spend_by_provider` now does, after #170 - reads as
  `openai 1.00`, and the $0.40 that genuinely went to Anthropic is attributed to
  OpenAI.

The total is right and reconciles with the bill. The attribution is wrong, silently,
and the screen gives a reader no way to tell.

## The sequence

1. An orchestrator on an OpenAI profile delegates to a published agent on an
   Anthropic profile.
2. The parent run finishes at $1.00, of which the delegate spent $0.40.
3. `GET /api/v1/runs/spend` → `by_provider: [{provider: "openai", cost_usd: "1.00"}]`.
   Anthropic does not appear at all.

`spend_by_key` has the same shape whenever a delegate runs on a different stored key.

## Where

- `backend/app/repositories/agent_run.py:299` — `spend_by_provider`
- `backend/app/repositories/agent_run.py:341` — `spend_by_key`
- `backend/app/agents/capabilities/budget.py` — one `SpendLedger` per run is the
  reason the child row cannot be added; see also `docs/governance.md`, "Delegation
  spends the parent's budget".

The limitation is stated in both docstrings rather than hidden, and
`tests/integration/test_delegated_run_spend.py::TestTheThreeBreakdowns` pins the
arithmetic that is correct today.

## How you would know it was fixed

An integration test where a $1.00 parent run on OpenAI contains a $0.40 delegate on
Anthropic, and `spend_by_provider` answers `openai 0.60` + `anthropic 0.40` — both
vendors named, still summing to the $1.00 the organization owes.

That needs the delegate's spend subtracted from the parent's row for attribution
purposes, or spend recorded per provider rather than per run. Neither is a one-line
change, which is why this is not folded into the branch that found it.

## Provenance

[...]

PR #613, opened 2026-08-11
docs(budgets): say which breakdown the pricing caveat measures

### LINKED ISSUE #597: The "could not be priced" caveat counts top-level runs, but governs by_provider and by_key too
## What breaks

`GET /api/v1/runs/spend` prints one "some runs could not be priced" caveat above
three breakdowns, and the schema says what it is for: *"Runs in the window whose
cost is a floor because some model in them had no price. **How much of everything
below is a fact**"*.

It is only a fact about one of the three. The caveat is summed from the per-agent
rows, which count **top-level runs only**:

- `backend/app/api/routes/v1/runs.py:441` — `partial_run_count=sum(row.partial_run_count for row in agents)`
- `backend/app/repositories/agent_run.py:620` — `func.count(AgentRun.id).filter(top_level, AgentRun.cost_is_partial)`

`by_provider` and `by_key` sum **every** row's own spend (`_own_cost()`,
`agent_run.py:658`) — delegated rows included, which is what #194 needed them to
do. So an unpriced *delegate* makes the vendor and key splits a floor while the
caveat above them reads `0` and the page says nothing.

Delegated rows do carry the flag — `agent_runner.py:2389` writes
`cost_is_partial=delegation.cost_is_partial` — so this is not a missing value, it
is a filter that no longer matches the rows underneath it.

## The sequence

1. An orchestrator on a priced OpenAI model delegates to an agent whose model has
   no entry in the price catalog.
2. The delegate's row is written with `cost_is_partial=True`, `parent_run_id` set.
3. `GET /api/v1/runs/spend` → `partial_run_count: 0`, and `by_provider` /
   `by_key` silently understate that vendor.
4. The Spend tab (`frontend/src/components/runs/spend-tab.tsx:70`) therefore
   suppresses `someRunsCouldNotBePriced` entirely, because the count is zero.

`by_agent` is unaffected: its `cost_usd` is the same top-level set the caveat
counts, which is why the mismatch is invisible on the row that is right.

## Where

- `backend/app/api/routes/v1/runs.py:441` — the page-level sum
- `backend/app/repositories/agent_run.py:620` — `filter(top_level, ...)`
- `backend/app/schemas/agent_run.py:350` — the description that claims more than
  the figure delivers
- `frontend/src/components/runs/spend-tab.tsx:70` — the line that renders it

An integration test where a priced parent run contains an **unpriced delegate**,
and the figure governing the vendor split says so — either by counting the rows
those two breakdowns actually sum, or by giving them their own caveat and
narrowing this one's description to `by_agent`. Whichever way, the invariant to
pin is that no breakdown is a floor without a figure on the same page saying it
is.

The comment at `runs.py:438` says the sum exists "so the figure and its breakdown
cannot disagree about which runs could not be priced". That is true of `by_agent`
and false of the two below it; whichever fix lands should make the comment true or
say which breakdown it is about.

## Provenance

Pre-existing, found while verifying #194 is fixed. Recorded at the time in
[#202](https://github.com/vstorm-co/agenticos/pull/202)'s commit body under "Known
and deliberately not folded in", but never filed — so it has been invisible to the
board since. Same cluster as #194 and #170: the arithmetic behind
`GET /runs/spend`, which #202 corrected for cost and did not revisit for the
caveat.

DEENUU1:
## What this fixes

`GET /api/v1/runs/spend` prints one "some runs could not be priced" caveat above
three breakdowns, and three places described it as measuring all three. It
measures one — `by_agent`, which counts the same top-level rows — and only
*marks* By provider and By key, which sum every row's own spend, delegated rows
included (#194). So one parent with three unpriced delegates reads `1` while
three figures below it are a floor.

The marker itself is sound, which is not what #597's body claims. Its sequence —
an unpriced delegate, `partial_run_count: 0`, "the page says nothing" — is not
reachable: a run tree shares one spend ledger (`BudgetGuard.for_delegate` passes
`ledger=self.ledger`), `SpendLedger.has_unpriced_models` is asked of every entry
in it, and the top-level row is written `cost_is_partial=ledger.has_unpriced_models`.
An unpriced delegate therefore flags its parent, `filter(top_level, cost_is_partial)`
counts it, and the caveat renders. That leaves a wording defect in a description
and a comment, which is what this changes — plus the invariant, which nothing was
pinning.

Closes #597.

## What changed

- `backend/app/schemas/agent_run.py:351` — `CostSummary.partial_run_count` says
  what it counts (top-level runs, one per tree), why it is never `0` while a
  figure below it is a floor, and which breakdown it measures versus marks
- `backend/app/schemas/agent_run.py:279` — `CostByAgent.partial_run_count`: "had
  a model with no price" was the run's own model; a delegate's counts too
- `backend/app/api/routes/v1/runs.py:420` — the comment claiming the figure and
  "its breakdown" cannot disagree, which is true of `by_agent` and overclaims for
  the two below it
- `frontend/src/components/runs/spend-tab.tsx:65` — same overclaim on the
  rendering side ("the breakdowns are a floor by exactly that many")
- `docs/governance.md:327` — the same sentence on the page
- `backend/tests/integration/test_delegated_run_spend.py` — two tests, below

No behaviour change: descriptions, comments and tests only.

## Testing

- `tests/integration/test_delegated_run_spend.py::TestTheCostScreen::test_an_unpriced_delegate_is_marked_above_the_splits_it_makes_a_floor`
  — a priced parent with an unpriced delegate: the figure the route sums reads
  `1` above an Anthropic split that is the delegate's own $0.40. This is the
  invariant #597 asked for — no breakdown is a floor without a figure on the
  same page saying so — and it was untested end to end
- `…::test_the_caveat_counts_trees_rather_than_the_rows_the_splits_sum`
  — one parent, two unpriced delegates, figure reads `1`; the delegate's own row
  is counted nowhere. Pins the magnitude the descriptions now state
- The other two links in the chain were already covered and still pass:
  `tests/test_delegation_seams.py` (an unpriced delegation makes the whole
  ledger a floor) and `tests/test_agent_runner.py::test_partial_pricing_is_flagged_on_the_run`
  (the ledger's answer is what `finish_run` is given)
- `make check` — green

## Noticed, not fixed

- The user-facing string is unchanged: `someRunsCouldNotBePriced` reads "N runs
  in this window could not be priced, so the figures below are a floor", which is
  true as written — N *runs* are a floor, and so is everything below. A reader
  could still infer that N rows of By provider are affected. Changing it is copy
  in two locales for no accuracy gain, so it stayed out of this diff.
- `.claude/worktrees/` is not in `.gitignore`, so a worktree in the main checkout
  shows up as untracked files — an easy way to stage something that should never
  be committed. Unrelated to this change; worth a one-line follow-up.

---

## Update after review — `1f84b15f`

The first commit's wording was itself an overclaim. *"It is never `0` while a
figure below it is a floor"* holds against the delegate route and not against
the window: `spend_by_agent` counts partials over `top_level` rows **inside**
the window, while `by_provider` and `by_key` price every row in it through an
unwindowed `_own_cost()` subquery. A parent that started before `since` and
delegated after it contributes its child's own spend to both splits and nothing
to the count. Against Postgres:

```
parent   started_at = since - 2h   cost_is_partial = True
delegate started_at = since + 30m  cost_is_partial = True, anthropic, $0.40

partial_run_count -> 0
by_provider       -> [('anthropic', 0.40, 1)]
```

Which is #597's step 3 reached through the window instead of the ledger — and
`_own_cost()`'s docstring already named the asymmetry, so shipping the absolute
would have left two docstrings in the repository disagreeing about whether the
case exists. Filed as **#620**; every copy of the claim now names it as the
exception.

- `schemas/agent_run.py` — the claim is bounded and names the case
- `repositories/agent_run.py` — `AgentSpendRow.partial_run_count` still carried
  the sentence this branch corrected on `CostByAgent`, and it is the row the
  schema copies the value off. Fourth copy of one drifting fact
- `runs.py`, `spend-tab.tsx` — both comments restated the schema description
  sentence for sentence; cut to what each call site needs plus a pointer to the
  canonical copy
- `docs/governance.md` — the same bound, with the straddling case spelled out
- `test_a_tree_that_straddles_the_start_of_the_window_is_marked_nowhere` —
  asserts `0` deliberately; inverting it is how #620's fix announces itself
- the docstring on `…_is_marked_above_the_splits_it_makes_a_floor` claimed the
  test proved the ledger link. Both its rows are written partial by hand, so it
  now says what it pins and points at `test_delegation_seams.py` for the link

Refs #620.

### DIFF
--- a/backend/app/api/routes/v1/runs.py
+++ b/backend/app/api/routes/v1/runs.py
@@ -420,15 +420,8 @@ async def get_spend(
         # How much of everything below is a fact. Summed from the per-agent rows
         # rather than queried again, so this figure and `by_agent` cannot
         # disagree about which runs could not be priced - they count the same
-        # top-level rows.
-        #
-        # It marks the two breakdowns underneath as well without measuring them.
-        # Those sum every row's own spend, delegated rows included, so an
-        # unpriced *delegate* makes a vendor and a key a floor while this counts
-        # trees: one parent with three unpriced delegates reads 1. The marker
-        # still fires, because a tree shares one spend ledger and the parent's
-        # row is therefore a floor too - which is why the figure is never 0
-        # while a figure below it is one.
+        # top-level rows. It marks the two breakdowns below without measuring
+        # them; see `CostSummary.partial_run_count`.
         partial_run_count=sum(row.partial_run_count for row in agents),
         by_agent=[
             CostByAgent(
--- a/backend/app/api/routes/v1/runs.py
+++ b/backend/app/api/routes/v1/runs.py
@@ -418,8 +418,17 @@ async def get_spend(
         to_date=to_date,
         month_to_date_usd=await service.monthly_spend(ctx),
         # How much of everything below is a fact. Summed from the per-agent rows
-        # rather than queried again, so the figure and its breakdown cannot
-        # disagree about which runs could not be priced.
+        # rather than queried again, so this figure and `by_agent` cannot
+        # disagree about which runs could not be priced - they count the same
+        # top-level rows.
+        #
+        # It marks the two breakdowns underneath as well without measuring them.
+        # Those sum every row's own spend, delegated rows included, so an
+        # unpriced *delegate* makes a vendor and a key a floor while this counts
+        # trees: one parent with three unpriced delegates reads 1. The marker
+        # still fires, because a tree shares one spend ledger and the parent's
+        # row is therefore a floor too - which is why the figure is never 0
+        # while a figure below it is one.
         partial_run_count=sum(row.partial_run_count for row in agents),
         by_agent=[
             CostByAgent(
--- a/backend/app/schemas/agent_run.py
+++ b/backend/app/schemas/agent_run.py
@@ -279,10 +279,11 @@ class CostByAgent(BaseSchema):
     partial_run_count: int = Field(
         default=0,
         description=(
-            "How many of those runs had a model with no price. The cost is a "
-            "floor by exactly that much, and '3 of 40 runs could not be priced' "
-            "is the difference between a figure a reader can act on and one "
-            "they have to take on trust"
+            "How many of those runs could not be fully priced - some model in "
+            "the run had no price, its delegates' included, because a tree "
+            "shares one ledger. The cost is a floor by exactly that many runs, "
+            "and '3 of 40 runs could not be priced' is the difference between a "
+            "figure a reader can act on and one they have to take on trust"
         ),
     )
     month_to_date_usd: Decimal | None = Field(
@@ -350,8 +351,15 @@ class CostSummary(BaseSchema):
     partial_run_count: int = Field(
         default=0,
         description=(
-            "Runs in the window whose cost is a floor because some model in "
-            "them had no price. How much of everything below is a fact"
+            "Top-level runs in the window whose cost is a floor because some "
+            "model in the run had no price - the run's own or any it delegated "
+            "to, which share one spend ledger. How much of everything below is "
+            "a fact: it is never 0 while a figure below it is a floor, because "
+            "an unpriced delegate is in its parent's ledger too. It counts "
+            "*trees* rather than the rows `by_provider` and `by_key` sum, so "
+            "one parent with three unpriced delegates reads 1 - it marks those "
+            "two breakdowns without measuring them, and measures `by_agent`, "
+            "which counts the same top-level rows"
         ),
     )
     by_agent: list[CostByAgent]
--- a/backend/app/schemas/agent_run.py
+++ b/backend/app/schemas/agent_run.py
@@ -354,12 +354,15 @@ class CostSummary(BaseSchema):
             "Top-level runs in the window whose cost is a floor because some "
             "model in the run had no price - the run's own or any it delegated "
             "to, which share one spend ledger. How much of everything below is "
-            "a fact: it is never 0 while a figure below it is a floor, because "
-            "an unpriced delegate is in its parent's ledger too. It counts "
-            "*trees* rather than the rows `by_provider` and `by_key` sum, so "
-            "one parent with three unpriced delegates reads 1 - it marks those "
-            "two breakdowns without measuring them, and measures `by_agent`, "
-            "which counts the same top-level rows"
+            "a fact: an unpriced delegate is in its parent's ledger too, so a "
+            "floor under `by_provider` or `by_key` is marked here even though "
+            "neither is measured here. It counts *trees* rather than the rows "
+            "those two sum, so one parent with three unpriced delegates reads "
+            "1, and it measures `by_agent`, which counts the same top-level "
+            "rows. A tree that straddles the start of the window is the one "
+            "case it misses: the delegate's row is inside the window and its "
+            "parent's is not, so the splits are a floor and this reads 0 "
+            "(agenticos#620)"
         ),
     )
     by_agent: list[CostByAgent]
--- a/backend/tests/integration/test_delegated_run_spend.py
+++ b/backend/tests/integration/test_delegated_run_spend.py
@@ -24,7 +24,7 @@ from __future__ import annotations
 
 import uuid
 from dataclasses import dataclass
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from decimal import Decimal
 
 import pytest
@@ -87,6 +87,7 @@ async def _run(
     cost: Decimal,
     secret: OrganizationSecret | None = None,
     partial: bool = False,
+    started_at: datetime | None = None,
 ) -> AgentRun:
     """A run somebody started, opened and finished the way every surface does.
 
@@ -97,6 +98,9 @@ async def _run(
     `partial` is what the run's ledger answered about pricing, which is a question
     about the whole tree: a delegate on a model with no price makes this row a
     floor too, because parent and delegate book into one ledger.
+
+    `started_at` is the column every spend window filters on, so it is what puts a
+    run inside or outside the one being asked about.
     """
     run = await agent_run_repo.create_run(
         db,
@@ -109,7 +113,7 @@ async def _run(
         model_label="gpt-4.1",
         provider="openai",
         secret_id=None if secret is None else secret.id,
-        started_at=datetime.now(UTC),
+        started_at=datetime.now(UTC) if started_at is None else started_at,
     )
     return await agent_run_repo.finish_run(
         db,
@@ -134,6 +138,7 @@ async def _delegated(
     task_id: str = "4f2a1b8c",
     secret: OrganizationSecret | None = None,
     partial: bool = False,
+    started_at: datetime | None = None,
 ) -> AgentRun:
     """A delegation, written the way `finish` writes one: complete, in one insert.
 
@@ -141,7 +146,7 @@ async def _delegated(
     while the run was still going - and the foreign key is the thing this proves:
     the parent's row has to exist by the time these are written.
     """
-    moment = datetime.now(UTC)
+    moment = datetime.now(UTC) if started_at is None else started_at
     return await agent_run_repo.record_delegated_run(
         db,
         run_id=uuid.uuid4(),
@@ -496,13 +501,14 @@ class TestTheCostScreen:
 
         The caveat counts top-level runs; By provider and By key sum every row's
         own spend, delegated rows included (#194). So an unpriced *delegate* makes
-        those two a floor through a row the caveat never looks at, and the figure
-        would read 0 above them if the delegation's row were the only one carrying
-        that unpriced request.
+        those two a floor through a row the caveat never looks at, and what keeps
+        the figure non-zero above the vendor it marks is the parent's row being
+        written a floor as well (#597).
 
-        It is not. A tree shares one spend ledger, so the request is in the
-        parent's ledger too and the parent's row is written a floor as well -
-        which is what keeps the count non-zero above the vendor it marks (#597).
+        Both rows are written partial here, which is the state that shared ledger
+        produces rather than a proof of it - `tests/test_delegation_seams.py` is
+        where the parent inherits its delegate's unpriced request. What this pins
+        is the arithmetic downstream of it.
         """
         org = await _org(db)
         orchestrator = await _agent(db, org, slug="orchestrator")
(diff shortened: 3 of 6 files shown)


backend/app/schemas/agent_run.py
```diff
@@ -350,8 +351,15 @@ class CostSummary(BaseSchema):
     partial_run_count: int = Field(
         default=0,
         description=(
-            "Runs in the window whose cost is a floor because some model in "
-            "them had no price. How much of everything below is a fact"
+            "Top-level runs in the window whose cost is a floor because some "
+            "model in the run had no price - the run's own or any it delegated "
+            "to, which share one spend ledger. How much of everything below is "
+            "a fact: it is never 0 while a figure below it is a floor, because "
+            "an unpriced delegate is in its parent's ledger too. It counts "
```
DEENUU1: This is the claim I could not make hold. The shared ledger closes the *delegate* route to a false zero, which is right - but not the *window* route. `spend_by_agent` counts partials over `top_level` rows **inside the window**; `by_provider` and `by_key` take every row in the window and price it with an unwindowed `_own_cost()`. A parent that started before `since` and delegated after it therefore contributes its child's own spend to both splits and nothing at all to this count.

Against the real database, parent at `since - 2h` and an unpriced delegate at `since + 30m`: `partial_run_count` reads `0` with `by_provider` reading `anthropic 0.40`. A floor with nothing on the page saying so - #597's step 3, arrived at the other way round.

It is narrow (a tree straddling the start of the window, so mostly the calendar-month and "today" presets), but this description is generated into `docs/reference`, so the absolute is worth bounding:

```suggestion
            "a fact: it is never 0 while a figure below it is a floor, because "
            "an unpriced delegate is in its parent's ledger too - except at the "
            "window's leading edge, where a delegate inside the window can have "
            "a parent outside it and nothing top-level to count. It counts "
```

And since it is a real gap rather than a wording one, I would file it: same cluster as #194 and #597, and #168 wants the link. A third integration test alongside the two here would pin whichever way it gets resolved.

backend/app/api/routes/v1/runs.py
```diff
@@ -418,8 +418,17 @@ async def get_spend(
...
+        # disagree about which runs could not be priced - they count the same
+        # top-level rows.
+        #
+        # It marks the two breakdowns underneath as well without measuring them.
+        # Those sum every row's own spend, delegated rows included, so an
+        # unpriced *delegate* makes a vendor and a key a floor while this counts
+        # trees: one parent with three unpriced delegates reads 1. The marker
+        # still fires, because a tree shares one spend ledger and the parent's
+        # row is therefore a floor too - which is why the figure is never 0
+        # while a figure below it is one.
```
DEENUU1: Two things about this block.

The last sentence - *"which is why the figure is never 0 while a figure below it is one"* - is the claim I flagged on `schemas/agent_run.py:357`. Whatever wording lands there should land here too, or better, not land here at all, which is the second thing.

This is now ten lines restating `CostSummary.partial_run_count` almost sentence for sentence, and that description is the canonical copy - it is what `docs/reference` is generated from. `code-style.md` is blunt about it (*"a multi-line essay is a docstring in the wrong place"*), and the repo's own line two files over is *"two copies of one fact drift"*. Which is what this PR exists to clean up. I would cut it to what a reader of this call site actually needs:

```suggestion
        # rather than queried again, so this figure and `by_agent` cannot
        # disagree about which runs could not be priced - they count the same
        # top-level rows. It marks the two breakdowns below without measuring
        # them; see `CostSummary.partial_run_count`.
```

frontend/src/components/runs/spend-tab.tsx
```diff
@@ -63,10 +63,14 @@ export function SpendTab() {
...
-          reader treating the totals as exact. Summed server-side from the same
-          rows, so the caveat and the breakdown cannot disagree about the count. */}
+          window's top-level runs could not be fully priced. A delegate on a
+          model with no price makes its parent's row a floor too - a tree shares
+          one spend ledger - so this is non-zero whenever By provider or By key
+          is a floor, which is what stops a reader treating the totals as exact.
+          It counts trees rather than the rows those two sum, so it marks them
+          without measuring them; By agent counts the same top-level rows and is
+          summed server-side from them, so the caveat and that breakdown cannot
+          disagree about the count. */}
```
DEENUU1: Same sentence as the backend one - *"so this is non-zero whenever By provider or By key is a floor"* - and the same hole in it (see my note on `schemas/agent_run.py:357`).

Separately: eight lines of backend arithmetic in a React component is a lot to keep true. Nothing in this file enforces any of it, and when the query changes this is the copy nobody greps. What a reader of *this* file needs is why the caveat renders once at the top instead of per breakdown; the rest belongs to the schema.

```suggestion
          window's top-level runs could not be fully priced. Saying it once at
          the top is what stops a reader treating the totals as exact. It marks
          By provider and By key without measuring them, and measures By agent -
          see `CostSummary.partial_run_count` for which is which. */}
```

docs/governance.md
```diff
@@ -324,10 +324,19 @@ this page's rule throughout:
...
+a figure wearing a plus sign is not.
+
+A run counts when any model in its tree had no price, **its delegates' included** —
+the tree shares one spend ledger, so an unpriced delegate makes the parent's row a
+floor as well. That is what lets one figure govern all three breakdowns: By
+provider and By key sum every row's own spend, delegated rows included, and the
+count is never `0` while one of their figures is a floor. It measures By agent,
+which counts the same top-level rows, and only **marks** the other two — it counts
+trees, so one parent with three unpriced delegates reads `1` while three figures
+below it are a floor.
```
DEENUU1: *"the count is never `0` while one of their figures is a floor"* is the strongest form of the claim, on the page a reader is most likely to act on. Same gap as inline on the schema: a delegate inside the window whose parent started before it makes By provider a floor while this reads `0`.

If you would rather bound it than fix it, the page has room to say so plainly, which the schema description does not:

```suggestion
A run counts when any model in its tree had no price, **its delegates' included** —
the tree shares one spend ledger, so an unpriced delegate makes the parent's row a
floor as well. That is what lets one figure govern all three breakdowns: By
provider and By key sum every row's own spend, delegated rows included, and the
count is `0` only when no run that **started in the window** is a floor. It
measures By agent, which counts the same top-level rows, and only **marks** the
other two — it counts trees, so one parent with three unpriced delegates reads `1`
while three figures below it are a floor.

A tree that straddles the start of the window is the one case the marker misses:
the delegate's spend is inside the window and its parent's row is not, so the
splits are a floor and the count above them reads `0`.
```

backend/tests/integration/test_delegated_run_spend.py
```diff
@@ -483,6 +491,83 @@ async def test_the_four_numbers_on_one_screen_agree(self, db):
...
+
+        The caveat counts top-level runs; By provider and By key sum every row's
+        own spend, delegated rows included (#194). So an unpriced *delegate* makes
+        those two a floor through a row the caveat never looks at, and the figure
+        would read 0 above them if the delegation's row were the only one carrying
+        that unpriced request.
+
+        It is not. A tree shares one spend ledger, so the request is in the
+        parent's ledger too and the parent's row is written a floor as well -
+        which is what keeps the count non-zero above the vendor it marks (#597).
```
DEENUU1: This paragraph asserts more than the test does. Both rows here are written `partial=True` by hand, so what gets pinned is *given* a parent flagged partial, the arithmetic downstream - not the thing the docstring says makes it non-zero, which is `for_delegate` sharing the ledger. If someone gave delegates their own ledger tomorrow, the parent would be written `cost_is_partial=False` in production and both of these tests would still be green.

Not arguing for a rewrite - `test_delegation_seams.py` covers that link and the PR body says so. Just that the docstring reads as if this test proves the join, and the assertions are only one side of it. Worth either trimming the second paragraph to a pointer at the seam test, or - better value for the same lines - adding the case that is genuinely covered nowhere: parent started before `since`, delegate inside it. That one comes back `0` today.

DEENUU1 (commented):
This is the right shape for the fix: the issue turned out to be three descriptions overclaiming rather than a silent zero, and withdrawing #597 with a verified refutation instead of quietly editing the body is exactly how it should have gone. The `by_agent` / `by_provider` / `by_key` split is now stated where a reader will actually hit it, and the two integration tests pin the magnitude the descriptions claim.

One thing to sort out before this lands, though. The diff replaces a vague overclaim with a sharper claim that is still not quite true: **"it is never `0` while a figure below it is a floor"** has a hole at the window's leading edge, and that sentence now sits in four places (schema description, route comment, `spend-tab.tsx`, `docs/governance.md`). The shared ledger closes the delegate route into it, which is the part this PR gets right - but the window closes nothing. I checked it against Postgres rather than by reading:

```
parent   started_at = since - 2h   cost_is_partial = True
delegate started_at = since + 30m  cost_is_partial = True, anthropic, $0.40

partial_run_count -> 0
by_provider       -> [('anthropic', 0.40, 1)]
by_key            -> [(None, None, 0.40, 1)]
```

That is #597's step 3 reached through the window instead of through the ledger: a split that is a floor, and no figure on the page saying so. `spend_by_agent` filters the partial count on `top_level` **inside the window**, while `spend_by_provider` / `spend_by_key` take every row in the window with an unwindowed `_own_cost()` subquery. And `_own_cost()`'s own docstring already names it: *"a child inside the window whose parent started before it is the one asymmetric case: the child's own spend counts and the parent's does not"*. So after this merges, two docstrings - one in `backend/app/repositories/agent_run.py`, one in `backend/app/schemas/agent_run.py` - disagree about whether that case exists. Either bound the sentence or file the gap; suggestions inline.

One more, outside the diff so I could not anchor it. **`backend/app/repositories/agent_run.py:557`** still carries the exact sentence this PR corrected in `CostByAgent` - *"partial_run_count: How many of them had a model with no price"*. That is `AgentSpendRow`, the row the schema copies the value off, so it is the fourth copy of the same drifting fact and arguably the most authoritative one. Worth folding in here, since fixing three of four copies of a sentence is roughly how this got filed in the first place.

Rest is small stuff.

backend/tests/integration/test_delegated_run_spend.py
```diff
@@ -86,12 +86,17 @@ async def _run(
...
 ) -> AgentRun:
     """A run somebody started, opened and finished the way every surface does.
 
     On OpenAI, because the delegate below runs on Anthropic: the provider split is
     where double-counting is least visible - the same money appears under two
     vendors at once, and each looks plausible on its own.
+
+    `partial` is what the run's ledger answered about pricing, which is a question
+    about the whole tree: a delegate on a model with no price makes this row a
+    floor too, because parent and delegate book into one ledger.
```
DEENUU1, line 100: Nice - this is what makes the helper honest. A bare `partial: bool = False` would read as "was this run priced", and the whole defect being fixed here is that it means something bigger. Same for the comment on the delegated row saying it carries the *other* answer.

backend/app/schemas/agent_run.py
DEENUU1: Done in 1f84b15f. The claim now names the exception and points at #620 rather than asserting an absolute the window does not hold: *"A tree that straddles the start of the window is the one case it misses: the delegate's row is inside the window and its parent's is not, so the splits are a floor and this reads 0."* Filed as #620 with the verified sequence, and `test_a_tree_that_straddles_the_start_of_the_window_is_marked_nowhere` pins the current `0` - inverting that assertion is how the fix announces itself.

backend/app/api/routes/v1/runs.py
DEENUU1: Done in 1f84b15f - cut to four lines and a pointer at `CostSummary.partial_run_count`, which is the copy `docs/reference` is generated from.

frontend/src/components/runs/spend-tab.tsx
DEENUU1: Done in 1f84b15f - same cut, same pointer. The component now says why the caveat renders once at the top and leaves the arithmetic to the schema.

docs/governance.md
DEENUU1: Done in 1f84b15f. The paragraph no longer claims the count is never `0`, and the straddling case gets its own paragraph with the link to #620 - the page has the room to say it plainly, which the field description does not.

backend/tests/integration/test_delegated_run_spend.py
DEENUU1: Done in 1f84b15f. The second paragraph now says what the test actually pins and sends the reader to `test_delegation_seams.py` for the ledger link. Added the uncovered case as its own test right below it: parent at `since - 2h`, unpriced delegate at `since + 30m`, and it comes back `0` (#620). The helpers take `started_at` for it.

backend/tests/integration/test_delegated_run_spend.py
DEENUU1: Kept as is.

PR #624, opened 2026-08-11
feat(onboarding): first-run tour, guided flows

### LINKED ISSUE #53: Add onboarding
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680fd8683c7c061fdb9b7) · Priority: **Low** · Category: Development

## The brief

From the Notion card:

- A **modal on `/dashboard`** for new users, walking them through the whole application — what is where and what it does
- **Tooltips on every page** explaining that page
- A **button/icon at the top of each page** that restarts the walkthrough — e.g. on `/agents`

## Sequencing

**Depends on #37.** The tour starts on `/dashboard`, which is currently a 23-line "Under construction" stub. A guided tour of a construction sign is worse than no tour. #37 also decides what the dashboard *shows per role*, and the tour has to match that — a Viewer must not be walked through controls they cannot see.

## What the repo already decides for you

- **Every string through `next-intl`.** A tour is a lot of copy, and hardcoding it is the single easiest way to break the i18n rule at scale. Budget for the message files.
- **Role-aware, not one script.** `.claude/rules/frontend.md:44` — a control the caller may not use is not rendered. A tour that points at a button a Member does not have is the same defect wearing a different hat. The step list has to be filtered by `usePermissions()`.
- **"Seen it" is per user and must persist.** `users.onboarding_at` already exists on the model (it came from the template). Check it before inventing a new column or a localStorage flag — localStorage means the tour reappears on every new browser.
- Components under ~100 lines; the tour is a component plus a step registry, not a switch statement in the layout.

## Worth deciding early

**Library or hand-rolled?** A tour library (driver.js, shepherd, react-joyride) buys positioning, focus trapping and keyboard handling — all of which are fiddly and all of which are accessibility requirements, not nice-to-haves. Weigh the bundle cost the way #37 has to weigh a chart library. Say which and why in the PR.

**Tooltips are a separate, larger job.** "A tooltip on each page explaining the page" is ~15 pages of copy that has to stay true as the product changes — the same maintenance problem as documentation, with none of the tooling. Consider whether the per-page explanation belongs in a help link to `docs/` rather than in a tooltip that will quietly go stale.

## Done when

- [ ] #37 merged first
- [ ] A first-run modal on `/dashboard`, dismissible, that does not reappear for a user who has seen it — persisted server-side
- [ ] The step list is filtered by permission; a Viewer is never shown a control they do not have
- [ ] A restart control on each page covered by the tour
- [ ] Every string via `next-intl`
- [ ] An integration test: a Viewer's tour skips the steps a Viewer cannot act on

### LINKED ISSUE #591: Agent draft can't be published by a builder who can't add a model
### What breaks

A member who holds `agents:edit` but **not** `connections:manage`, in an
organization with no runnable model profile, can create an agent draft they can
never make work. Adding a model writes a `/providers/model-profiles`, gated on
`connections:manage` (`frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx:670`,
`allowAdd={can(Perm.connectionsManage)}`), and publish is refused without a model
(`backend/app/services/agent_registry.py:670-671`). So they can create, but can
neither add a model nor publish.

### Sequence

1. Sign in as a member with `agents:edit` but not `connections:manage`, in an org
   that has stored no model profile.
2. Agents → New agent → create. The builder opens.
3. The model picker's "add" control is hidden.
4. Publish → validation refuses: "No model selected - pick one before publishing".

The only way out is a permission they don't have, or an admin adding a model
first — and nothing says so until publish fails.

### Onboarding side (already handled on `feat/onboarding`)

The create-agent guided flow now drops the "add a model" step for a caller
without `connections:manage` (`flows.ts`, `flow-agent-model-add` gated on
`connections:manage`), so the walkthrough no longer leads them to a hidden
control. The underlying product path still dead-ends.

### How you'd know it's fixed

The create-agent dialog (or the builder) tells such a user up front — "your
organization has no model yet; adding one needs `connections:manage`, or ask an
admin" — rather than only failing at publish.

Found working #53 (onboarding); pre-existing product behaviour, surfaced by the
onboarding flow. Part of the onboarding cluster (#53).

### LINKED ISSUE #592: Onboarding steps on the MCP page vanish when the server catalog is empty
### What breaks

Three onboarding steps on the MCP page — `mcp-add`, `mcp-filter`, `mcp-connect`
— live inside `<McpServerList>`, which the page renders only when
`rows.length > 0` (`frontend/src/app/[locale]/(dashboard)/mcp-servers/page.tsx:76-77`).
On a deployment whose server catalog is empty, the page shows its empty-state
card instead, so those `data-tour` targets never mount. The tour steps
(`frontend/src/lib/onboarding/tour.ts:285-297`) and the `create-mcp` interactive
flow (`frontend/src/lib/onboarding/flows.ts`) then spotlight nothing: the
coach/driver waits, times out, and lands centred with no highlight.

`mcp-connect` is doubly conditional — `data-tour={index === 0 ? "mcp-connect" :
undefined}` (`frontend/src/components/mcp/mcp-server-list.tsx:476`), so it also
needs at least one catalog card.

### How you'd know it's fixed

On an empty MCP catalog, the MCP "?" walk and the `create-mcp` offer either skip
the missing steps or point at the empty-state's own "add" affordance, rather than
spotlighting nothing.

Severity low: shipped deployments carry a curated catalog
(`catalog/mcp_servers.json`), so `rows` is normally non-empty. Found working #53;
same shape as the already-known `knowledge-add-integration` (needs
`connectors.length > 0`) condition. Part of the onboarding cluster (#53).

### LINKED ISSUE #593: Onboarding tour leaves Settings and the workspace file browser unexplained
### What's missing

The first-run tour and the header "?" replay cover the product's main sections
but leave two areas unexplained:

- **Settings** — `/settings/profile`, `/settings/account`,
  `/settings/notifications`, `/settings/slash-commands`. No tour step points at
  `/settings/*` (the `ROUTES.SETTINGS_*` constants exist but no step uses them).
- **The workspace file browser detail** — `workspaces/[id]`. The tour covers the
  list via `workspaces-browser` (`frontend/src/lib/onboarding/tour.ts:306`) but
  not the detail view where files are actually browsed and downloaded.

(The `admin/*` pages are intentionally uncovered — app-admin only.)

### Done when

A short tour stop, or a page "?" walk, exists for Settings and for the workspace
file detail, permission-filtered like the rest of the tour.

Found working #53. Low priority — orientation, not correctness. Part of the
onboarding cluster (#53).

OchnikBartek:
## What this adds

Onboarding for the dashboard, in two mechanisms that share one permission model
and one copy catalog:

- **A first-run walkthrough and a per-page "?" replay.** A driver.js spotlight
  tour that opens on first login and can be replayed from the header on any page.
  It steps across the dashboard, chat, the agent builder (one stop per tab), the
  knowledge-base detail, an organization's members and roles, the activity tabs,
  Settings, and the workspace file browser. Every stop is permission-filtered the
  same way the UI is — a Viewer never sees a stop for a control they cannot use,
  and the publish stop drops for anyone without `agents:publish`.

- **Guided interactive create flows (the coach).** An adaptive, field-by-field
  walk that offers to build the reader's first agent, skill, knowledge base or
  MCP connection on the real screen and runs it to a finished, usable result. The
  walk is **watertight** — it cannot leave the reader in a half-built state:
  - the guided create dialog is **frozen** while a walk is active (an outside
    click or Escape no longer dismisses the wizard out from under the coach);
  - the agent walk **gates on a model and on publish** — it will not reach chat
    with a draft that has no model or was never published, and the Publish step is
    always walked rather than skipped;
  - KB / skill / MCP walks **block submit** until every field tile has been
    visited, so nothing is created from a half-filled form. The submit guard
    cancels pointer clicks and both keyboard activations (Enter and Space on the
    guarded control), so Tab-then-Space cannot create the resource early either;
  - the MCP catalog step lets the reader pick **any** server (`roam`), not only a
    single spotlighted one, and the MCP tutorial closes after connect-and-check;
  - MCP completion is **scope-agnostic** (an org *or* a personal connection ends
    the walk);
  - a running flow is **torn down on an organization switch** (`session-reset`
    closes the coach and clears the offer), so it cannot route to the previous
    org's agent.

- **Full Polish translation of the whole onboarding.** A complete `onboarding`
  namespace in `pl.json` at parity with English (210/210 leaves), ICU plurals in
  correct Polish forms, interpolations 1:1, written for a non-technical reader.
  English remains the source language; `pl.json` holds only what is translated.

## Additionally resolved

Three issues found while building #53 are fixed here, each with its own commit
and test:

- **#593** — Settings and the workspace file browser now have a tour stop.
  Settings collapses its tabs onto one identity (`SETTINGS_DETAIL`); the workspace
  detail (`WORKSPACE_DETAIL`) points at the file browser. Both permission-filtered.
- **#592** — the MCP tour no longer spotlights an empty catalog. The catalog
  controls (`mcp-filter`, `mcp-add`, `mcp-connect`) are marked `optional` so the
  engine skips a target that never mounts, the empty-state card carries its own
  `mcp-catalog` target, and the `create-mcp` offer is suppressed when the catalog
  is empty (a cache read, no fetch).
- **#591** — a builder without `connections:manage`, in an org with no model, is
  now told up front where a model comes from (in the model picker's empty state)
  rather than only finding out when Publish refuses the draft.

## How it was verified

- `tsc`, eslint, prettier and the i18n guard (`scripts/check-i18n.ts`, 3500
  messages, no hardcoded copy) clean.
- `messages/catalog.test.ts` (en↔pl parity, no half-sentences) 11/11; full
  frontend suite 279 files / 4298 tests green.
- Frontend coverage gate meets its thresholds — 100% statements/functions/lines,
  97.5% branches — over the gated globs, which are `src/{app/api,lib,stores,hooks}`
  and the *finished* component directories only. **`src/components/onboarding/**`
  is deliberately outside the gate** (it is on the coverage plan to widen later):
  the tour, the coach and the freeze layer are covered by their integration tests
  and the e2e journeys, not by the 100% ratchet.
- Merged current `main` (behind 0). The only conflict was `pl.json`, where both
  sides appended new top-level namespaces (this branch `onboarding`, main `hosted`
  and `mcp`); resolved as a union and each namespace checked byte-for-byte against
  its source side. The diff is frontend + docs only, so CI's backend job
  legitimately skips.

Nothing is left red.

Closes #53
Closes #593
Closes #592
Closes #591


frontend/src/hooks/use-onboarding.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  // Auto-start fires at most once per page load, so a failed persist cannot loop
+  // the tour back open and a manual restart is never fought by the effect.
+  const hasAutoStarted = useRef(false);
+  const shouldAutoStart =
+    path === ROUTES.DASHBOARD &&
+    !permissionsLoading &&
+    !permissionsError &&
+    !!user &&
+    !user.onboarding_completed_at;
```
DEENUU1, line 136: **Blocking — this is what's failing CI right now.** The seeded e2e user never gets `onboarding_completed_at`, so this auto-start fires in every spec that lands on `/dashboard`, and with `allowClose: false` the overlay swallows every click. Six specs are red on this PR's head run (`auth.spec.ts:166`, `dashboard.spec.ts:43`, four in `sidebar.spec.ts` — all `locator.click: Test timeout` / `toBeHidden()` resolving to `driver-popover-title "Welcome to AgenticOS"`). Nothing in `frontend/e2e/` mentions onboarding.

The seed should mark onboarding complete (or `auth.setup.ts` dismiss the tour once), and while you're there, a small spec for the first-run modal itself would give this feature the one test layer it currently has none of.

frontend/src/hooks/use-onboarding.ts
```diff
@@ -0,0 +1,142 @@
...
+  // hand before the path moves. The first-run tour needs none of this: its list
+  // is permission-filtered, not path-derived, so navigation never changes it.
+  const [anchor, setAnchor] = useState<string | null>(null);
+  // React's "adjust state while rendering" pattern, not an effect: the guards
+  // make each branch fire at most once, so there is no loop and no extra commit.
+  if (isOpen && mode === "page" && anchor === null) {
+    setAnchor(path);
+  } else if (!isOpen && anchor !== null) {
+    setAnchor(null);
+  }
```
DEENUU1, line 83: **Blocking — one "?" press on an uncovered page kills the help button for the whole session.**

Sequence: `/channels` renders `PageHeader` (so the new "?") but has no `TOUR_STEPS` entry, so `stepsForPage("/channels")` is `[]`. Click "?" there: `openPage()` sets `isOpen: true`, this block freezes `anchor = "/channels"`, `step` is `undefined`, and the engine hits `tour.destroy(); return` — no popover, therefore no close button, and nothing ever calls `close()`. The anchor clears only when `isOpen` goes false, so every later "?" press on *any* page recomputes steps from the stale `/channels` anchor → still empty → the button is dead app-wide until a reload. Same failure for anyone whose permissions filter a page's steps to zero (e.g. "?" on `/skills` without `skills:view`). As a bonus, `useDetailTargets(isOpen)` keeps its three list queries enabled the whole time.

Fix shape — treat an open walk with no steps as closed, same render-adjust pattern you use for the anchor (or an effect):

```ts
if (isOpen && steps.length === 0) {
  close();
}
```

and ideally don't render the "?" at all when `stepsForPage(path, can)` is empty.

frontend/src/stores/onboarding-store.ts
```diff
@@ -0,0 +1,116 @@
...
+  dismissOffer: () => set({ offer: null }),
+  // Advancing here rather than leaving it to the coach keeps the widening of the
+  // step list and the move onto its first new step in one update: the recorded
+  // answer brings the detour into the flow, and `index + 1` lands on it. A
+  // question is never the flow's last step, so stepping past it always has
+  // somewhere to go.
+  answer: (questionId, value) =>
+    set((state) => ({
+      choices: { ...state.choices, [questionId]: value },
+      index: state.index + 1,
```
DEENUU1: The blind `index + 1` rests on "a question is never the flow's last step", which is true today — but it's a cross-file invariant that lives in `flows.ts` filtering and is enforced nowhere. I couldn't construct a real permission set that breaks it (built-in roles pair `agents:view` with `edit`), so this is hardening, not a bug: clamping here (or having the coach skip an already-answered question) makes the invariant local instead of trusting every future flow edit to preserve it.

frontend/src/components/onboarding/onboarding-coach.tsx
```diff
@@ -0,0 +1,499 @@
...
+          className="pointer-events-auto fixed z-[1000000001]"
+          style={{
+            top: blockRect.top,
+            left: blockRect.left,
+            width: blockRect.width,
+            height: blockRect.height,
+            cursor: "not-allowed",
+          }}
+        />
+      )}
```
DEENUU1: The guard blocks pointer events only — the keyboard walks straight through it. `create-kb-dialog` is a real `<form onSubmit>`, so on `flow-kb-field-name` pressing Enter in the name field submits: the KB is created three steps early, the later `flow-kb-field-create` step baselines its count *after* the creation, its `created` signal can never fire, and the reader dead-ends (closing the flow is the only exit). Same story for Tab+Enter through the full freeze — `pointer-events: none` doesn't remove anything from the tab order.

A capture-phase keydown that prevents Enter-submission while a `blockSubmit` step is showing covers both.

frontend/src/components/onboarding/onboarding-coach.tsx
```diff
@@ -0,0 +1,499 @@
...
+      // A dynamic step points at the very agent this flow created, resolved from
+      // the captured id; a plain one at its fixed `data-tour`. Without the id yet
+      // (it is captured on the builder, before any return leg) the dynamic
+      // selector falls back to the first such control rather than hunting forever.
+      const selector =
+        step.dynamicTarget === "createdAgentEdit"
+          ? `[data-tour="agent-card-edit"]${flowAgentId ? `[data-agent-id="${flowAgentId}"]` : ""}`
+          : `[data-tour="${step.target}"]`;
+      const target = await waitForElement(selector, signal);
+      if (signal.aborted || !(target instanceof HTMLElement)) return;
```
DEENUU1, line 377: If the target misses the 4s `waitForElement` window (cold query, slow builder after `router.push`), this effect returns with the rect never tagged for this step — `FreezeLayer` keeps the full-viewport freeze with no cut-out and nothing re-runs the wait, and a signal step renders no Next, so the only exit is abandoning the whole flow. The tour has a recovery path for exactly this (`optional` / centered popover); the coach has none.

Either drop the timeout for coach targets (the abort on step change already bounds the wait), or on timeout lift the freeze and keep the card showing.

frontend/src/components/onboarding/onboarding-coach.tsx
```diff
@@ -0,0 +1,499 @@
...
+      {/* A modal Radix dialog sets pointer-events:none on the body while it is open,
+          and this card sits outside that dialog — so it re-enables them for itself, or
+          its own Next and close inherit none and go inert while guiding a field inside a
+          create dialog. `onInteractOutside` on DialogContent keeps the click from also
+          dismissing that dialog. */}
+      <div
+        data-coach-card
+        role="dialog"
+        aria-label={t(`steps.${step.id}.title`)}
+        className="bg-popover text-popover-foreground pointer-events-auto fixed bottom-6 left-1/2 z-[1000000002] w-[min(28rem,calc(100vw-2rem))] -translate-x-1/2 rounded-xl border p-4 shadow-lg"
```
DEENUU1: `role="dialog"` without the dialog behaviors: no `aria-modal`, focus doesn't move to the card on step change, there's no trap, and Escape isn't handled — a keyboard user has to Tab through the visually-frozen page to reach the close button, and a screen reader is never told the page froze. driver.js covers all of this for the passive tour (it traps Tab — I checked the 1.8.0 dist); the coach is the gap.

Focus the card on step change, handle Escape → finish, and set `aria-modal` while a freeze is up.

frontend/src/hooks/use-onboarding-flow.ts
```diff
@@ -0,0 +1,278 @@
...
+  // adaptive step does not morph as the reader satisfies it: the "add a model"
+  // step must stay "add a model" long enough for the model appearing to advance
+  // it, not turn into "pick a model" the instant the profile is created and leave
+  // the reader on a step with no signal. Frozen per flow, and the hook remounts
+  // between flows, so a reopened flow reads the state afresh.
+  const [frozen, setFrozen] = useState<{ flowId: FlowId; state: OrgState } | null>(null);
+  if (flowId && liveState && frozen?.flowId !== flowId) {
+    setFrozen({ flowId, state: liveState });
+  }
+  const orgState = frozen?.flowId === flowId ? frozen.state : DEFAULT_STATE;
```
DEENUU1, line 241: While `liveState` is still `null` this falls back to `DEFAULT_STATE` ("the org has nothing"), so the coach can render the "no published agent — build one first?" fork in an org that *has* one, then swap it away when the lists settle — the exact morph the docstring above says freezing prevents. A fast "Yes" before it settles hands off to `create-agent` for an agent they already have. Usually masked by the react-query cache, but real on first navigation into chat.

Gating `isActive` on `liveState !== null` (render nothing until the snapshot lands) matches the documented intent.

frontend/src/components/agents/model-profile-picker.tsx
```diff
@@ -139,10 +139,17 @@ export function ModelProfilePicker({
...
+      // No models, and this caller cannot add one — the Builder hid the form
+      // above. Say why and where the fix is, rather than leave them to discover it
+      // at publish: an agent with no model is refused there, and the only ways to
+      // one are a permission they lack or an admin who holds it.
       return (
         <div className="border-border rounded-lg border border-dashed p-6 text-center">
           <KeyRound className="text-muted-foreground mx-auto h-5 w-5" />
           <p className="text-muted-foreground mt-2 text-sm">{t("organizationHasNoModels")}</p>
+          <p className="text-muted-foreground mx-auto mt-1 max-w-sm text-xs">
+            {t("noModelAskAdmin")}
```
DEENUU1, line 156: `noModelAskAdmin` asserts "a permission you don't have", but the component only knows `allowAdd`. In the builder that's `can(Perm.connectionsManage)`, so the sentence is true there — but `specialist-list.tsx` never passes `allowAdd`, so in a model-less org the delegation picker tells someone who *does* hold `connections:manage` that they don't. Pass `allowAdd={can(Perm.connectionsManage)}` in `specialist-list.tsx` (it already uses the permissions hook), or render this line only when the caller opts in.

frontend/src/components/mcp/mcp-server-list.tsx
```diff
@@ -544,6 +551,7 @@ export function McpServerList({ canManageOrganization }: McpServerListProps) {
                       <Button
                         size="sm"
                         variant="outline"
+                        data-tour="mcp-connect"
```
DEENUU1: This one is unreachable: the index-0 card already carries `data-tour="mcp-connect"` (line 474) and precedes every Connect button in document order, so `querySelector` always returns the card — the tour spotlights the whole card, never this button, and cards 2..N carry dead anchors. Keep exactly one; the card-level one is what actually runs today, so this attribute can just go.

```suggestion
```

frontend/src/app/[locale]/(dashboard)/mcp-servers/page.tsx
```diff
@@ -62,8 +62,10 @@ export default function McpServersPage() {
       ) : rows.length === 0 ? (
         <ServersCard count={0}>
           {/* Inline rather than an `EmptyState`: that component draws its own
-              bordered box, and inside a card it would frame one message twice. */}
-          <div className="px-6 py-12 text-center">
+              bordered box, and inside a card it would frame one message twice.
+              Carries the catalog's tour anchor so the "?" walk's first stop has
+              something to point at when the list it usually rides on is absent. */}
+          <div className="px-6 py-12 text-center" data-tour="mcp-catalog">
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/mcp-servers/page.tsx]: nit: this anchor is permanently shadowed — `ServersCard`'s header (`mcp-server-list.tsx:130`) carries the same `data-tour="mcp-catalog"` and precedes it in the DOM, so `querySelector` never returns this one, in either branch. The walk still works (the header always mounts), but the comment above claims a job this element never does. Drop the attribute here or fix the comment.

frontend/src/lib/onboarding/tour.ts
```diff
@@ -0,0 +1,418 @@
...
+
+  {
+    id: "skills-new",
+    page: ROUTES.SKILLS,
+    target: "skills-new",
+    permission: Perm.skillsEdit,
+    inTour: true,
+  },
+  { id: "skills-list", page: ROUTES.SKILLS, target: "skills-list", permission: Perm.skillsView },
+  { id: "skills-library", page: ROUTES.SKILLS, target: "skills-library" },
```
DEENUU1: `/skills` early-returns a refusal without `skills:view`, so `SkillLibraryGallery` never mounts — but this step is ungated while its sibling `skills-list` carries the permission. A caller without `skills:view` pressing "?" on `/skills` gets a step list of exactly `[skills-library]`, a 4s wait, then a centered caption describing a control that isn't there — the failure mode the doc at the top of this file says `permission` exists to prevent.

```suggestion
  { id: "skills-library", page: ROUTES.SKILLS, target: "skills-library", permission: Perm.skillsView },
```

frontend/src/lib/onboarding/tour.ts
```diff
@@ -0,0 +1,418 @@
...
+    page: ROUTES.RAG,
+    target: "knowledge-integrations",
+    permission: Perm.connectionsManage,
+  },
+  {
+    id: "knowledge-add-integration",
+    page: ROUTES.RAG,
+    target: "knowledge-add-integration",
+    permission: Perm.connectionsManage,
+  },
```
DEENUU1, line 297: The add-integration button renders only when `connectors.length > 0` (`reusable-integrations.tsx:100-102`) — the same data-conditional shape as the MCP controls, which got `optional: true` for exactly this reason. If the connector catalog is empty (or its request fails), this non-optional step waits 4s and pins its caption centered on nothing.

```suggestion
    id: "knowledge-add-integration",
    page: ROUTES.RAG,
    target: "knowledge-add-integration",
    permission: Perm.connectionsManage,
    optional: true,
```

frontend/src/components/onboarding/onboarding-tour.tsx
```diff
@@ -0,0 +1,240 @@
...
+        }
+      }
+      const element = step.target
+        ? await waitForElement(`[data-tour="${step.target}"]`, signal)
+        : undefined;
+      if (signal.aborted) return;
+      // An optional stop whose target never mounted — the MCP catalog's filter,
+      // add and connect controls when the catalog is empty — is skipped rather
+      // than pinned to the middle of the screen with nothing highlighted.
+      if (step.optional && step.target && !element) {
```
DEENUU1, line 243: The skip itself is right, but each skipped stop costs the full 4s `waitForElement` timeout — on an empty MCP catalog that's three in a row (`mcp-filter`, `mcp-add`, `mcp-connect`), 8–12s during which the previous popover just sits there looking hung. Worth a much shorter timeout for `optional` steps, or deciding emptiness synchronously from the catalog cache the way `CreationOffer` already does.

frontend/src/components/onboarding/detail-targets.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * For each detail pseudo-page a tour step can name, the concrete route the engine
+ * pushes to reach an example of it.
+ *
+ * A detail view has no route of its own, so the walkthrough opens a real one: the
+ * seeded demo where there is one, otherwise the first row. The list is fetched
+ * only while the tour is open, and a null `href` (an empty list) tells the engine
+ * to skip that section's steps rather than spotlight a route that would 404. The
+ * "?" replayed *from* a detail route never consults this — the reader is already
+ * on the row they mean, and the engine keeps them there.
```
DEENUU1: nit: stale contract — the engine deliberately does the opposite (`onboarding-tour.tsx:181-183`: "Describe the section where we are rather than skip it", `show(undefined)`). The engine's choice is the better one; this docstring, and the test name "gives a null href for an empty list so the engine skips that walk" in `detail-targets.test.tsx`, both describe a behavior that doesn't exist. Fix the wording in both.

frontend/src/app/globals.css
```diff
@@ -910,3 +910,131 @@
...
+ * (which would point at nothing from a fixed spot) is hidden.
+ *
+ * `display: block !important` is what keeps the caption from blinking as the
+ * spotlight glides. With `animate` on, driver hides the popover at the start of
+ * every transition (an inline `display: none`) and only rebuilds it at the
+ * halfway point, so a pinned caption would vanish for ~200ms each step. The
+ * `!important` overrides that inline hide, so the old caption holds until the
+ * rebuilt one replaces it in the same frame — an invisible swap, no blink.
+ */
+.driver-popover.tour-popover {
```
DEENUU1, line 1126: nit: only position is overridden here, so the tour popover keeps driver.css's `#fff` background and `#2d2d2d` text in dark mode — a light card next to a coach card that correctly uses `bg-popover` tokens. A `background: var(--popover); color: var(--popover-foreground);` block (plus button and progress-text colors) brings the two onboarding surfaces into agreement.

frontend/src/components/ui/dialog.tsx
```diff
@@ -29,13 +29,30 @@ DialogOverlay.displayName = DialogPrimitive.Overlay.displayName;
...
+        onInteractOutside={(event) => {
+          onInteractOutside?.(event);
+          // While the onboarding coach is guiding this dialog, its freeze is the whole
+          // point — the reader operates only the field being pointed at. Radix would
+          // dismiss on any outside interaction (the backdrop, or the coach's own card,
+          // which sits outside the dialog), stranding the walk on a step whose dialog
+          // is gone. Keep it open; the coach's close button ends the walk, and the
+          // dialog's own Cancel still closes it. A no-op when no coach is active
+          // (`[data-coach-card]` renders only while a flow runs).
+          if (document.querySelector("[data-coach-card]")) event.preventDefault();
```
DEENUU1, line 52: I like this mechanism — keying off live DOM presence of `[data-coach-card]` means a crashed or unmounted flow physically cannot leave a dialog stuck, and no existing caller passes these handlers (checked), so it's provably inert when no flow runs.

One consequence worth a short comment in the code: the check is global, so while *any* flow runs — including `roam` steps where the page is fully interactive — *every* dialog in the app stops responding to Escape and outside-click, not just the one being guided. X and Cancel still work, so nobody's trapped; just say it's deliberate.

frontend/messages/pl.json
```diff
@@ -677,5 +677,419 @@
...
+        "title": "Nie podłączono serwera MCP — podłączyć?",
+        "body": "Serwer MCP daje temu agentowi narzędzia z innego systemu, a jeszcze żadnego nie podłączyłeś. Podłączyć jeden teraz? To szybkie okno tutaj — bez wychodzenia z kreatora, i ruszymy dalej w chwili, gdy się połączy."
+      },
+      "flow-agent-mcp-connect": {
+        "title": "Otwórz katalog",
+        "body": "Wybierz Connect server — katalog otworzy się tutaj, w kreatorze, i wybierzemy jeden razem."
+      },
+      "flow-agent-limits": {
+        "title": "Ustaw budżet (opcjonalnie)",
+        "body": "Ogranicz, ile ten agent może wydać w miesiącu i ile kroków może zająć jedno uruchomienie, i wybierz, kto dostaje znać przy limicie. Tak agent, który się zapętli lub wymknie, zatrzymuje sam siebie. Wybierz Dalej, żeby przejść dalej."
```
DEENUU1: Grammar, twice in the limits copy: *„kto dostaje znać"* isn't Polish — the idiom is *dać komuś znać* and it can't be flipped this way. Also *„ile kroków może zająć jedno uruchomienie"* (*zająć* takes time, not steps) and *„który się zapętli lub wymknie"* (*wymknąć się* dangles without its complement).

For this line and `agent-limits.body` (line 959): „…i wybierz, kto zostanie powiadomiony przy limicie" / „…kto zostanie powiadomiony, gdy agent osiągnie limit albo zawiedzie", plus „z ilu kroków może się składać jedno uruchomienie" and „który się zapętli albo wymknie spod kontroli, zatrzyma się sam".

frontend/messages/pl.json
```diff
@@ -677,5 +677,419 @@
...
+        "title": "Co może każda rola",
+        "body": "Macierz uprawnień, rola po roli: czego każda może dotknąć i jak daleko sięga. Udostępnienie pojedynczego zasobu może poszerzyć dostęp ponad rolę, nigdy go nie zawęzi."
+      },
+      "vault-new": {
+        "title": "Sejf",
+        "body": "Klucze API i inne dane logowania żyją tutaj, zaszyfrowane w spoczynku. Agenci używają ich, żeby sięgnąć po potrzebne narzędzia, a zapisanego sekretu nikt już nie zobaczy."
+      },
+      "vault-keys": {
+        "title": "Co jest przechowywane",
+        "body": "Każde dane logowania w tej organizacji, wypisane po nazwie i rodzaju. Agent odwołuje się do klucza po nazwie; samej wartości po zapisaniu nikomu się nie pokazuje."
```
DEENUU1: *„Każde dane logowania"* — *dane* is plurale tantum, it doesn't take *każde*; and *dane logowania* mistranslates *credential* here (the vault stores API keys, not logins). Suggestion: „Wszystkie poświadczenia w tej organizacji, wypisane według nazwy i rodzaju. Agent odwołuje się do klucza po nazwie; samej wartości po zapisaniu nikt już nie zobaczy." Same term in `vault-new.body` („inne dane logowania" → „inne poświadczenia").

frontend/messages/pl.json
```diff
@@ -677,5 +677,419 @@
...
+      },
+      "explore-chat": {
+        "title": "Krótki przewodnik po czacie?",
+        "body": "Zobacz, jak zacząć rozmowę, przełączyć, który agent odpowiada, i zmienić model albo wysiłek myślenia dla pojedynczego czatu. Chcesz szybki, prowadzony przegląd?"
+      }
+    },
+    "steps": {
+      "flow-agent-create": {
+        "title": "Nazwij agenta",
+        "body": "Utwórz nowego agenta — nadaj mu nazwę i krótki opis, i utwórz go. Otworzy się kreator, gdzie dzieje się reszta. Przejdziemy dalej, gdy tylko powstanie."
```
DEENUU1: Two patterns in the pl flow copy worth one sweep:

1. The coach advances when the reader clicks a specific button, and under `/pl` those buttons render in English (`pages.agents.newAgent` etc. have no pl entries) — but six flow-create steps translate the name away (here: „Utwórz nowego agenta — … i utwórz go", which is also clumsy). The rest of this catalog gets it right („Kliknij Publish", „Kliknij Create", „zakładkę Knowledge"). Same fix pattern for `flow-agent-create`, `flow-skill-create`, `flow-agent-skills-create`, `flow-kb-create`, `flow-agent-knowledge-create`, `flow-org-create`: „Wybierz New agent — nadaj mu nazwę i krótki opis. Otworzy się kreator, gdzie dzieje się reszta."

2. Sixteen masculine-only past forms (`widziałeś`, `zbudowałeś`, `utworzyłeś`, `podłączyłeś`, `opublikowałeś`, „będziesz gotowy"…) — the rest of pl.json has two in 3k+ messages, and onboarding is the first copy a new user reads. Neutral rewrites read fine: „agenta, którego budujesz", „nowo utworzoną bazę wiedzy", „które są już podłączone", „gdy zechcesz".

docs/first-agent.md
```diff
@@ -6,6 +6,62 @@ a cent of tokens.
...
+no trip to another page, and none back.
+
+And building is not where it ends. After Publish lands, the flow carries the
+reader into the chat, has them pick the agent they just built, and closes only
+once they have sent it a first message - because a first agent nobody has run
+is a tour that stopped one step short of the point.
+
+Declining guides nobody; the offer returns at the end of the Agents **?** walk.
+Every other section's **?** ends the same way, offering to create that section's
+resource - a skill, a knowledge base, an MCP connection, a workspace - and the
```
DEENUU1: There's no workspace flow — `FlowId` has no `create-workspace` and `flowForPage("/workspaces")` returns `null`; the walks that end with an offer are agent, skill, knowledge base, MCP and organization.

```suggestion
resource - a skill, a knowledge base, an MCP connection, an organization - and the
```

DEENUU1 (dismissed):
This is a big one and most of it is genuinely well built. The state/engine split, server-side persistence through `PATCH /users/me` (exactly what #53 asked for instead of a localStorage flag), the frozen `OrgState` at flow start so steps don't morph mid-walk, the abort plumbing in `spotlight.ts`, and 210/210 en/pl parity with zero hardcoded copy across ~1.9k lines of new UI — that's careful work, and the `Rect.stepId` tagging trick for step transitions is genuinely elegant.

Two things block the merge:

1. **The e2e job is red, and this branch is the cause.** The tour auto-opens over the seeded e2e user and the overlay swallows clicks — 6 specs fail. Details inline at `use-onboarding.ts`.
2. **One "?" press on a page with no tour steps (e.g. `/channels`) kills the help button app-wide for the session.** Sequence and fix inline at `use-onboarding.ts`.

Points I can't anchor inline (files not in the diff):

- **Coverage transparency.** `src/components/onboarding/**` is not in the vitest coverage `include` globs, so the "100% / 97.67%" claim gates `lib` + `hooks` + `stores` (covered, and covered well) but not ~1,150 lines of component code — including the whole page-transition machinery in `onboarding-tour.tsx` (`router.push`, `detail.pending`, `activate`, flow-mode teardown), which no test executes, and the `optional`-skip branch, which would survive deletion with every #592 test still green (the integration tests mock `waitForElement` to always resolve). Either add the glob or say in the PR body that the component layer is e2e territory — right now the number reads as covering the feature.
- `frontend/src/stores/session-reset.ts` — the onboarding store is the only tenant-coupled store not reset on org switch; a flow surviving the switch will `router.push` to the old org's agent and land on a refusal.
- `rag/[id]` has four tour stops but no `PageHeader`, so no "?" — #53's "restart control on each covered page" misses exactly that page.
- Small copy consistency: `steps.flow-agent-mcp-connect.body` quotes "Connect server" but the button reads "Connect a server" (`pages.agents.connectServer`); "thinking effort" vs "reasoning effort" for the same control (`offer.explore-chat.body` vs `steps.chat-model-picker.body`); pl "komendy ze slashem" vs the product's own "Polecenia ukośnika" (`nav.slashCommands`).

On the bundled issues: #591 fixed (one caveat inline at the picker), #592 fixed with warts (two dead anchors and 4s of dead air per skipped stop, inline), #593 fixed cleanly.

frontend/src/components/onboarding/onboarding-coach.tsx
```diff
@@ -0,0 +1,555 @@
...
+  // (`isContentEditable`) and a Radix select keep Enter for newlines and choosing.
+  useEffect(() => {
+    if (!isActive || !blockSubmit) return;
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== "Enter") return;
+      const el = event.target as HTMLElement | null;
+      if (!el) return;
+      const submitsForm = el instanceof HTMLInputElement;
+      const activatesSubmit = el.closest(`[data-tour="${blockSubmit}"]`) !== null;
+      if (submitsForm || activatesSubmit) event.preventDefault();
```
DEENUU1: The Enter guard closes the door the comment above describes, but Space is the sibling key: a focused `<button type="submit">` also activates on Space, and the freeze only blocks pointers, not Tab. So Tab to the guarded Create button, press Space, and the resource is created steps early — the same dead-end this effect exists to prevent. Space has to stay free in text inputs (it types a space), so gate it on the submit target only:

```suggestion
      if (event.key !== "Enter" && event.key !== " ") return;
      const el = event.target as HTMLElement | null;
      if (!el) return;
      const submitsForm = event.key === "Enter" && el instanceof HTMLInputElement;
      const activatesSubmit = el.closest(`[data-tour="${blockSubmit}"]`) !== null;
      if (submitsForm || activatesSubmit) event.preventDefault();
```

frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx
```diff
@@ -146,16 +146,18 @@ export default function KBDetailPage({ params }: KBDetailPageProps) {
...
-        kb={kb}
-        mayEdit={mayEdit}
-        isLoading={isLoading}
-        isUploading={isUploading}
-        onRefresh={() => refresh()}
-        onEditParseOptions={() => setOverrideOpen(true)}
-        onChooseFiles={() => fileInputRef.current?.click()}
-        onDelete={() => setDeletingCollection(true)}
-      />
+      <div data-tour="kb-header">
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx:153]: Still the one covered page without a "?": four tour stops point in here, but this header is hand-rolled rather than `PageHeader`, so there is no replay control — while `agents/[id]` gets one through `PageHeader` and chat renders `RestartTourButton` explicitly. A `RestartTourButton` beside the actions in `KBDetailHeader` (the chat-container shape) closes the last gap in #53's "restart control on each covered page".

frontend/messages/en.json
```diff
@@ -2115,6 +2116,420 @@
...
+        "title": "Choose who answers",
+        "body": "Switch between your agents right here. Each brings its own instructions, tools and knowledge, so the same question can get a very different answer."
+      },
+      "chat-composer": {
+        "title": "Ask anything",
+        "body": "Type a message, drop in a file, or start with a slash-command. Enter sends; Shift+Enter adds a line."
+      },
+      "chat-model-picker": {
+        "title": "Tune the model",
+        "body": "Pick the model behind the agent and dial its reasoning effort or temperature — just for this conversation."
```
DEENUU1: nit: the control this points at is called "Thinking effort" (`chat.controls.thinkingEffort`), and `offer.explore-chat.body` says "thinking effort" too — this is the one place saying "reasoning effort".

```suggestion
        "body": "Pick the model behind the agent and dial its thinking effort or temperature — just for this conversation."
```

frontend/messages/en.json
```diff
@@ -2115,6 +2116,420 @@
...
+        "title": "Add MCP servers (optional)",
+        "body": "MCP servers give the agent tools from other systems you've connected. Attach any here, or choose Next to move on."
+      },
+      "flow-agent-mcp-ask": {
+        "title": "No MCP server connected — connect one?",
+        "body": "An MCP server gives this agent tools from another system, and you haven't connected one yet. Want to connect one now? It's a quick dialog right here — no leaving the builder, and we'll pick up the moment it's connected."
+      },
+      "flow-agent-mcp-connect": {
+        "title": "Open the catalog",
+        "body": "Choose Connect server — the catalog opens right here in the builder, and we'll pick one together."
```
DEENUU1: nit: the button reads "Connect a server" (`pages.agents.connectServer`), and the coach advances when the reader clicks it — quote it exactly, so the reader hunts for the right words.

```suggestion
        "body": "Choose Connect a server — the catalog opens right here in the builder, and we'll pick one together."
```

frontend/messages/pl.json
```diff
@@ -677,5 +677,419 @@
...
+        "title": "Rozmawiaj ze swoimi agentami",
+        "body": "Nowy czat otwiera świeży wątek, a Twoje wcześniejsze rozmowy układają się w tym panelu. To najszybszy sposób, żeby wypróbować agenta i zobaczyć, jak działa, krok po kroku."
+      },
+      "chat-agent-picker": {
+        "title": "Wybierz, kto odpowiada",
+        "body": "Przełączaj się między agentami tutaj. Każdy wnosi własne instrukcje, narzędzia i wiedzę, więc to samo pytanie może dostać zupełnie inną odpowiedź."
+      },
+      "chat-composer": {
+        "title": "Zapytaj o cokolwiek",
+        "body": "Wpisz wiadomość, upuść plik albo zacznij od komendy ze slashem. Enter wysyła; Shift+Enter dodaje wiersz."
```
DEENUU1: nit: the product's own term is „Polecenia ukośnika" (`nav.slashCommands`) — „komendy ze slashem" here and in `settings-tabs.body` (line 1091) is a second name for the same thing, in the copy that teaches it.

```suggestion
        "body": "Wpisz wiadomość, upuść plik albo zacznij od polecenia ukośnika. Enter wysyła; Shift+Enter dodaje wiersz."
```

DEENUU1 (dismissed):
Round 2. Both blockers from my last review are fixed, and fixed properly: the seed marks the owner's onboarding done so the tour no longer covers the e2e specs (all green now), and a "?" opened where nothing is walkable closes itself — with a guard against closing during the permissions-loading window, which I hadn't even asked for. The rest of the inline sweep landed clean too: the no-timeout coach wait (the pre-aborted-signal path in `waitForElement` was already handled — I checked), the `stateReady` gate with a regression test, the dark-mode popover tokens, and the pl grammar and masculine-form sweep, which was thorough.

Four points from my previous review body went unanswered, though, and two of them I need before approving:

1. **`src/stores/session-reset.ts` — the onboarding store is still not reset on org switch.** The store holds `flowId`, `choices` and `flowAgentId`, all tenant-coupled: a flow surviving a switch will `router.push` to the old org's agent and land on a refusal. `resetTenantState()` needs a `useOnboardingStore.getState().close()`. This branch introduces the store, so the fix belongs in it.
2. **`rag/[id]` is still the one covered page with no "?".** Details inline at the header div — #53 asks for a restart control on each covered page, and this PR closes #53.

The other two I can live with, but answer them on the thread rather than leaving them hanging:

3. **Coverage transparency.** `src/components/onboarding/**` is still not in the vitest coverage `include`, and the PR body still reads as if the gate covers the feature. This round actually widened the gap: the Enter guard, the focus move and the Escape handler are component-layer code no test executes. Either add the glob, or say in one sentence in the PR body that the component layer is outside the gate.
4. The `index + 1` clamp note on the store — skipping it as hardening is a defensible call; just reply on the thread so it reads as decided rather than missed.

Plus one new, small gap in the Enter guard itself (Space still activates the guarded submit — inline with a suggestion), and the three copy-consistency nits from last time, now anchored inline.

DEENUU1 (dismissed):
Round 3, on the Space-guard commit — and it's exactly right. Blocking both keys only on the guarded control, while Enter alone keeps its input-wide block, is the correct split: I checked all three `blockSubmit` anchors (`skill-dialog-create`, `kb-dialog-create`, `mcp-dialog-connect`) and each sits on a bare button, so the `closest()` scoping can never eat a Space typed in a field. The regression test drives both paths, and all three copy nits landed too.

And you're right about the KB detail "?" — I withdraw that one. I checked every leg of your citation: `rag/[id]/page.tsx` renders `KBDetailHeader`, which is built on `PageHeader` (`kb-detail-header.tsx:55`), which renders `RestartTourButton` (`page-header.tsx:78`), and `stepsForPage` collapses `/rag/<id>` onto `KB_DETAIL` (`tour.ts:388`) with its four stops. My comment said the header was hand-rolled; it isn't — I searched `components/kb/` and the header lives in `components/rag/`. My mistake, and thanks for answering with file:line instead of changing code that was already right.

That leaves one blocker, unchanged from round 2:

1. **`src/stores/session-reset.ts` still doesn't reset the onboarding store on an org switch.** `resetTenantState()` (session-reset.ts:23) empties every tenant-coupled store except this one, and the store holds `flowId`, `choices` and `flowAgentId` — all tenant-coupled: a flow surviving the switch will `router.push` to the previous org's agent and land on a refusal. `useOnboardingStore.getState().close()` is enough — closing stops the coach, and `openFlow` clears `choices` and `flowAgentId` at the next start — imported from `./onboarding-store` like the sibling stores. Clearing the pending `offer` in the same breath would be tidy: an offer minted from one org's caches shouldn't survive into another. This branch introduces the store, so the fix belongs in it.

And two threads that need answers, not necessarily code:

2. **Coverage transparency.** The PR body's "(over the gated `src/{app/api,lib,stores,hooks}` and components globs)" reads as if the component layer is gated, but `src/components/onboarding/**` is not in the vitest `include` list — ~1,150 lines outside the gate. Either add the glob or make the body say plainly that the onboarding component layer sits outside it, covered by the integration tests and e2e. One sentence closes this.

3. **The `index + 1` clamp.** Skipping it as hardening is fine by me — say so on the thread so it reads as decided rather than missed.

Fix 1, answer 2 and 3, and this is an approve — the feature itself has been merge-worthy since round 2.

frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx
Comment (DEENUU1) [reply frontend/src/app/[locale]/(dashboard)/rag/[id]/page.tsx:153]: Verified and withdrawn — `KBDetailHeader` is on `PageHeader` (`kb-detail-header.tsx:55`), which carries the `RestartTourButton`, and `stepsForPage` maps `/rag/<id>` to `KB_DETAIL`. I searched `components/kb/` for the header; it lives in `components/rag/`. My mistake.

DEENUU1:
@codex review

dev-vstorm:
@codex review

DEENUU1:
@codex review

dev-vstorm:
@codex review

DEENUU1:
@codex review

frontend/src/hooks/use-onboarding.ts
DEENUU1: Fixed, in two halves.

The seed marks it done rather than `auth.setup.ts` dismissing the tour: `seed.setup.ts` now PATCHes `onboarding_completed_at` for the bootstrapped owner before any product spec runs. Dismissing through the UI would have made every run depend on driver.js rendering and on the write it fires, which is a fixture step that fails for reasons that have nothing to do with the fixture — and it would have taught the suite that a returning user is one who happened to click something, not one the server says has finished. The step is idempotent (`if (me.onboarding_completed_at) return;`), so a run killed mid-suite self-heals on the next one.

The layer this feature had none of is now `frontend/e2e/onboarding.spec.ts`, three specs: the first-run walkthrough auto-opening for a user the *server* says has not finished and the dismissal being written through `PATCH /users/me` (waited on as a response, not as an overlay disappearing) and not returning on the next load; the header "?" replaying the current page's tips without recording completion; and the "?" being absent on a page the registry does not cover. It resets the flag itself and restores it in `afterEach` — safe because `workers: 1` and `fullyParallel: false` mean nothing else is loading a page while it is off.

All three pass locally.

frontend/src/hooks/use-onboarding.ts
DEENUU1: Both halves fixed.

The walk no longer stays open with nothing in it — `useOnboardingTour` closes one that opens empty, so the anchor clears and the button stays live app-wide. Not while the permission set is still loading, though: `can` answers false throughout that window, so every page reads as stepless until it settles and closing then would shut a walk that is about to have steps.

And the "?" is no longer offered where there is nothing to replay: `pageHasSteps` (`lib/onboarding/tour.ts`) and `RestartTourButton` returning null. It is deliberately blind to permissions — the button lives in a header twenty surfaces render, and making it wait on the permission query would drag react-query into every one of them for an answer that only differs on a page which *has* stops. `/admin/*` and `/dev/components` are the two pages this removes it from; a page whose stops a permission filters to zero is still caught by the close-on-empty above.

That cost eleven page-level integration tests a `usePathname` in their `next/navigation` mock, which is in the same commit.

`useDetailTargets` is gated too: `onboarding-tour.tsx` enables it only for a walk that actually names a fetched detail page (`FETCHED_DETAIL_PAGES`), so a "?" on the vault, runs or channels no longer costs three list requests to answer a question none of its steps asks.

frontend/src/stores/onboarding-store.ts
DEENUU1: You were right that nothing enforced it, and wrong that it holds — it is a bug, not hardening, and the permission set that breaks it is one an owner can build.

`create-agent`'s tail (limits, publish, the first chat run) is gated on `agents:publish`. A role holding `agents:edit` and `collections:edit` and nothing else therefore walks a flow whose last surviving step is `flow-agent-knowledge-ask`. Pressing **Skip** there records the choice, widens nothing, and steps one past the end; `clamped` in `useOnboardingFlow` puts the reader back on the question they just answered, where Skip re-answers it forever and the close button is the only way out. Built-in roles do pair `agents:view` with `edit` — but the roles catalog is a per-organization matrix, not a fixed set.

Fixed in the hook rather than the store, because the store cannot see the step list: `answer` now resolves the choice against the list it produces and ends the walk when nothing follows the fork. Computed rather than clamped afterwards, so there is no frame in which the answered question is back on screen.

Two tests: `use-onboarding-flow.test.ts` fails without the fix, and `flows.test.ts` pins the reachable permission set so a future flow edit that changes this shape says so.

frontend/src/components/onboarding/onboarding-coach.tsx
DEENUU1: Fixed, and Space with it (your follow-up). A capture-phase `keydown` on `document` cancels Enter and Space while a `blockSubmit` step is showing — Enter on any `HTMLInputElement`, and either key on anything inside the guarded control. Space stays free in a text input, where it types a space; a textarea, the source editor and a Radix select keep Enter for newlines and choosing.

The wider hole behind it is closed too: the freeze blocked pointers and left the tab order alone, so Tab walked into the dimmed page and Enter on whatever it reached acted. The card now traps Tab while the page is frozen — `roam`, `inOverlay` and any open dialog excepted, where the reader is meant to work the page or Radix owns the trap.

frontend/src/components/onboarding/onboarding-coach.tsx
DEENUU1: Fixed by dropping the timeout for coach targets, which is the first of your two suggestions. `waitForElement(selector, signal, null)` — the abort on step change or flow end already bounds the wait, so a fixed deadline bought nothing except the state you describe: full-viewport freeze, no cut-out, no Next, closing the flow the only way out.

The two remaining 4s waits in this file are harmless: the `activate` tab trigger falls through to the untimed target wait, and the `blockSubmit` box only leaves its guard unrendered. And the exits are real now regardless — the card's close button always renders, and Escape finishes the walk when no dialog owns it.

frontend/src/components/onboarding/onboarding-coach.tsx
DEENUU1: Three of the four are fixed; the fourth is now too.

`aria-modal` is set while a freeze is up, focus moves to the card on each step (yielding to a dialog or picker that manages its own), and Escape finishes the walk when nothing else owns it. The focus trap was the gap you named and it was still open: Tab walked straight past the freeze into the dimmed page, which is the keyboard half of the same hole the submit guard had. A capture-phase Tab handler now cycles the card's own buttons while the page is frozen — reversing the list for Shift keeps both directions one expression, since focus resting on the card itself reads as -1 either way. Excluded on `roam`, `inOverlay` and any open dialog.

frontend/src/hooks/use-onboarding-flow.ts
DEENUU1: Fixed, exactly as suggested. `useOrgSnapshot` freezes only once `liveState` is non-null, and `isActive` is gated on `stateReady`, so the coach renders nothing until the snapshot it will freeze on has landed. `DEFAULT_STATE` still feeds `steps` while the coach is inactive, so no fork built for an empty organization can reach the screen.

frontend/src/lib/onboarding/flows.ts
```diff
@@ -0,0 +1,793 @@
...
+        // The model *picker*, in the Build tab's Instructions card — not the
+        // Model settings card (`agent-model`), which is temperature and thinking.
+        target: "agent-model-picker",
+        activate: "agent-tab-build",
+        // Adding a model writes a `/providers/model-profiles`, which is
+        // `connections:manage`, not `agents:view` — the same gate the picker's
+        // "add" control carries. A caller with `agents:edit` but not that would
+        // otherwise be walked to a control the server hides from them.
+        permission: Perm.connectionsManage,
+        signal: { kind: "created", resource: "model" },
```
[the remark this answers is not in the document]
DEENUU1: Fixed — the step gates on the draft, not the list. `flow-agent-model-add` carries `signal: { kind: "modelSet" }`, which reads the stored agent's `draft_spec.model_profile_id` rather than the model-profile count, so the walk cannot leave the model step until the 1.2s builder autosave has actually landed. That is precisely the sequence you describe: the knowledge question navigates to `/rag`, unmounts the builder and cancels the save timer, and the later `published` signal could then never be met because Publish refuses a spec with no model. `flow-agent-model-pick` takes the same signal for the same reason.

frontend/src/components/onboarding/onboarding-coach.tsx
DEENUU1: Correcting myself: the trap I described above shipped broken, and `ba3e3e1f` fixes it.

Confining Tab to the card took the *spotlit control* out of the tab order — and the freeze deliberately leaves that control clickable, because a step carrying a signal renders no Next and operating the control is the whole advance. So a keyboard user reached a step with a `created`, `modelSet`, `arrived`, `published`, `selected` or `sent` signal — most of create-agent — with no route to the thing the step was waiting on, and Escape or the X as the only exits. Filed as closing an accessibility gap, it opened a wider one.

Tab now cycles the card *and* the control inside the cut-out, which is exactly what the freeze leaves usable. A fork cuts no hole, so there the card really is the whole of it, and `inOverlay` steps still yield to Radix.

The test missed it because its fixture was the one shape where trapping is harmless — a step with no signal, so the card had a Next. There is a test per shape now, and the new one fails against the commit above it.

frontend/src/stores/onboarding-store.ts
DEENUU1: Follow-up: the fix above was half-applied and `ba3e3e1f` finishes it.

It decided the destination in `useOnboardingFlow` from `clamped`, then left the store to advance from its own unclamped `index`. Those are the same number only while they agree — and `clamped` exists precisely because they need not: the step list is recomputed from `can`, which re-resolves on a permission refetch or an organization switch and can shrink the list under a live index. In that window the hook would read "not the last step, record it" and the store would step past the end anyway, reaching the same loop through a different door.

`answer` now takes the index it should land on, so the decision and the move are one number. The store's comment claiming a question is never a flow's last step is gone with it — this change proved it false and left it standing, which is worse than never having written it.

Issue #564, opened 2026-08-10
Collapse the backendFetch forwarder family (~40 routes) into platformProxy


DEENUU1:
~40 route files under `orgs/**`, `me/**`, `admin/**`, `sessions/**` each re-do the same block by hand (read cookie → 401, set bearer, stringify, `NextResponse.json`, map `BackendApiError`), while `platformProxy` already does all of it plus org-header forwarding, byte-accurate bodies, error-body passthrough and `no-store`.

**Cost today.** The duplication is exactly the bugs #546 (refusal message discarded) and #553 (no `no-store`) and #106 (org header dropped): a header/policy fix must be made ~40 times and has already drifted three ways.

**Target.** Replace the org/me/admin routes with `platformProxy()` mounts (as `kb`/`rag` were), keeping only the genuinely special ones hand-rolled — auth routes (set/clear cookies) and binary routes (custom `Content-Disposition`/CSP). Keep the `platform-proxy.test.ts` sweeps green.

**Size:** L. Each removed file is a net simplification.

Closes the root cause of #546, #553; refs #106, #436.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

OchnikBartek:
First cluster (sessions) migrated in #1176. Recording the blockers for the rest so the follow-up is scoped:

- **`orgs/**`** — the natural target is an `orgs/[[...path]]` catch-all, but the cluster keeps a binary `orgs/[id]/avatar` route (custom multipart/CSP handling). Next.js refuses a `[id]` segment and an optional catch-all at the same level, so `avatar` cannot stay hand-rolled beside the catch-all without a build conflict — it has to be migrated onto the proxy too (the proxy does forward `content-type`/`content-disposition`, so this is likely fine, but wants checking) or the cluster mounted a different way. Separately, the proxy forwards `X-Organization-Id` where the hand-rolled orgs routes drop it (the #106 fix) — a real behaviour change that should be confirmed against a backend / e2e before shipping.
- **`admin/**`** — these carry a `requireAdmin` frontend gate (`@/lib/admin-auth`), not the plain cookie check `platformProxy` does. Moving them onto the proxy drops the frontend pre-check and relies on the backend's `CurrentAppAdmin` 403. Defensible, but a behaviour change worth deciding deliberately.
[...]

Issue #553, opened 2026-08-10
The backendFetch route family omits no-store, so a refetch-after-write can be served stale


DEENUU1:
All `backendFetch` routes return `NextResponse.json(data)` with no `Cache-Control` (orgs/**, me/**, admin/**). `platformProxy` stamps `no-store` on any unmarked response (the #230 fix); `.claude/rules/frontend.md` states a hand-rolled route owes the same header.

**How it fails.** The members / invitations / integrations lists refetched right after a create/invite/revoke can be answered from cache and render a row as still-present or still-absent — the same class as #230, on the surface the shared proxy does not cover.

**Fix.** Route through `platformProxy` (#564), or add `{ headers: { "Cache-Control": "no-store" } }` to each `NextResponse.json`.

**Done when.** Each `backendFetch` response carries `no-store` (or is on `platformProxy`).

Refs #230, #564.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

PR #668, opened 2026-08-13
fix(api): answer a refusal with the backend's reason, not a generic

### LINKED ISSUE #456: Invitations never reach EXPIRED: the sweep has no caller, and a timed-out one is marked revoked
## What breaks

`InvitationStatus.EXPIRED` is never assigned to an invitation that expires. Two
halves, and each is wrong on its own:

**`invitation_repo.expire_stale` has no caller.** `backend/app/repositories/invitation.py:103`
flips every `PENDING` invitation past `expires_at` to `EXPIRED` and returns the
count. `rg expire_stale backend/` finds the definition and nothing else — no
service, no route, no Prefect flow, no test. It is the only thing in the codebase
that would ever write `EXPIRED`, so an invitation nobody clicks stays `pending`
for ever.

**And when one *is* clicked past its expiry, it is marked `revoked`.**
`backend/app/services/invitation.py:184`:

```python
if invite.expires_at and invite.expires_at < datetime.now(UTC):
    await invitation_repo.revoke(self.db, invite)
    raise BadRequestError(message="Invitation has expired")
```

`revoke` writes `InvitationStatus.REVOKED`. So the row says somebody withdrew the
invitation, when what happened is that it timed out — a different fact, and the
one an admin reads off the list when they ask why a colleague never got in.

## Sequence

1. Invite someone. `expires_at` is set, status `pending`.
2. Wait past it. Nothing runs, nothing changes: the row is still `pending`, and
   the pending list still offers it.
3. Have them click the link. It is correctly refused — access is **not** the
   problem here — and the row becomes `revoked` rather than `expired`.

## Where

- `backend/app/repositories/invitation.py:103` — `expire_stale`, uncalled.
- `backend/app/services/invitation.py:184` — `revoke` where `EXPIRED` belongs.
- `backend/app/db/models/invitation.py` — the enum that promises the value.

## Fixed when

An invitation past `expires_at` reads `expired` without anybody clicking it, and
one that is clicked past its expiry is `expired` rather than `revoked` — with a
test for each. The shape is now in the tree: `app/worker/tasks/approval_tasks.py`
and `ApprovalService.expire_stale` do exactly this for tool approvals, and
`prefect_app.py` registers the schedule. This is a second caller for the same
pattern, not a new one.

## Related

The same defect as #178 (`ApprovalStatus.EXPIRED` defined and never assigned),
found while fixing it — the queue for #178 asked whether `invitation.expire_stale`
was dead too, and it is. #178's fix does not touch invitations, deliberately.

Nothing in the map in #168 covers invitations; this is the first issue on them, so
it belongs to no existing cluster. Pre-existing, and no branch caused it.

### LINKED ISSUE #546: Proxy routes discard the backend's refusal reason, showing a generic 'Backend API error'
~35 `backendFetch` route handlers under `frontend/src/app/api/{orgs,me,admin,sessions,invitations}/**` map a failure as `{ detail: error.message }`. But `BackendApiError.message` is the constructor's literal — `server-api.ts:15` `super(\`Backend API error: ${status} ${statusText}\`)` — and the real backend body sits unused in `.data`.

**How it fails.** A Viewer tries to change a member's role → backend returns 403 with `detail: "Only Owner or Admin can manage members"` → the UI shows `Backend API error: 403 Forbidden`. Every refusal on the org / members / invitations / integrations / admin surface reads the same generic way. The three `auth/{login,register,oauth-callback}` routes already read `error.data.detail`; the rest drifted. `platformProxy` (the other forwarder) passes the body through verbatim.

**Fix.** Root cause is #564 (route this family through `platformProxy`). Cheap interim, matching the auth routes: `(error.data as {detail?: string})?.detail ?? error.message`.

**Done when.** A 403 on a member-role change shows the backend's message; a test with a route mock returning `{status:403, data:{detail:"…"}}` surfaces the detail, not the constructor string.

Refs #436 (fixed one instance of this class on /rag), #106 (same forwarder family).

---

DEENUU1:
> **This body was wrong and is being corrected.** It held PR #667's text verbatim —
> invitation expiry, `Closes #456` — on a diff that changes ~40 frontend proxy route
> handlers. The footer is fixed below so a merge cannot close the wrong issue; the
> narrative still needs writing by whoever finishes the branch. Found in the #683 map
> rebuild.

## What this fixes

The hand-rolled `backendFetch` route handlers under `frontend/src/app/api/**` discard
the backend's refusal reason and answer with a generic `Backend API error`, so a
refusal a person could act on ("that name is already taken", "you may not write that
collection") reaches the toast as a sentence that says nothing. That is #546.

## What changed

`gh pr diff 668 --name-only` is the list: the admin, auth, invitations, orgs, files,
me/* and rag route families, plus `admin-routes.test.ts`, `session-routes.test.ts`,
`cookie-routes.test.ts` and `docs/plans/frontend-coverage.md`.

## Still owed on this branch

- A written account of what changed and how it was verified, at the bar `CLAUDE.md`
  sets for a body — the failure prevented, concretely; how it was verified and how far
  that reaches; what was seen and deliberately not fixed.
- Whether this supersedes or only overlaps **#564** (collapse the `backendFetch`
  forwarder family into `platformProxy`), which is the structural version of the same
  change, and **#553** (the same family omitting `no-store`). If this touches all forty
  handlers anyway, doing it as #564 rather than as a per-file patch is the cheaper end
  state.

Closes #546

### DIFF
--- a/docs/plans/frontend-coverage.md
+++ b/docs/plans/frontend-coverage.md
@@ -93,19 +93,13 @@ between them dropped it. Fixed, and `admin-routes.test.ts` now pins each paramet
 the screen can send.
 
 
-Thirty-eight of the generated routes under `src/app/api` answer a refusal with
-`detail: error.message`. That is `BackendApiError`'s own string - `"Backend API
-error: 400 Bad Request"` - not the backend's `detail`, which is sitting right
-there in `error.data`. The routes written by hand for this platform
-(`login`, `register`, `oauth-callback`) read the envelope properly; the
-template's do not.
-
-The symptom is a user-visible one: an expired magic link, a stale password-reset
-token and a duplicate invitation all report a status code instead of the sentence
-that says what to do. The fix is one helper and thirty-eight call sites, so it
-wants its own change rather than riding along with the tests -
-`session-routes.test.ts` asserts the current behaviour with a comment naming it,
-because a test claiming the better sentence would be a test of nothing.
+The generated routes under `src/app/api` used to answer a refusal with
+`detail: error.message` - `BackendApiError`'s own string, `"Backend API error:
+400 Bad Request"` - while the backend's sentence sat unread in `error.data`.
+Fixed by #546: `backendErrorDetail` in `server-api.ts` reads the envelope, all
+fifty-five call sites go through it, and the refusal tests in
+`cookie-routes.test.ts`, `admin-routes.test.ts` and `session-routes.test.ts`
+assert the sentence as well as the status.
 
 ## Backend
 
--- a/frontend/src/app/api/admin/admin-routes.test.ts
+++ b/frontend/src/app/api/admin/admin-routes.test.ts
@@ -143,11 +143,16 @@ describe("the admin gate", () => {
   });
 
   it.each(GUARDED)("passes the backend's own refusal of %s through", async (_name, call) => {
-    vi.mocked(backendFetch).mockRejectedValue(new BackendApiError(409, "Conflict", null));
+    // Sentence included - `error.message` is the constructor's generic
+    // string, and answering with it was #546.
+    vi.mocked(backendFetch).mockRejectedValue(
+      new BackendApiError(409, "Conflict", { detail: "User already deactivated" }),
+    );
 
     const response = await call();
 
     expect(response.status).toBe(409);
+    await expect(response.json()).resolves.toEqual({ detail: "User already deactivated" });
   });
 });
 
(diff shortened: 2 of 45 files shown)


DEENUU1:
Superseded by #654, released in 0.0.134.

I checked rather than assumed. `Backend API error` no longer appears anywhere in `frontend/src`, and 38 of the route handlers under `src/app/api/**` now answer with `backendErrorDetail(error)` and the backend's own status — which is what this branch set out to do. The seven that do not call `backendErrorDetail` are `health` and the six `auth/**` handlers, and those were never part of the defect: they inline the same forwarding by hand with their own fallback (`(error.data as { detail?: string })?.detail || "Login failed"`), so they never answered generically.

Rebasing this branch would mean re-implementing what is already on `main` across 39 conflicting files, so it is closed rather than merged. #546 is fixed.

Issue #661, opened 2026-08-13
A channel turn refused before the run orphans its ChatFile rows


DEENUU1:
## What breaks

A channel turn refused before the run orphans its stored `ChatFile` rows.
`_receive_files` uploads the bytes and creates `ChatFile` rows before the run;
if `answer_default` raises an `AppException` (nothing exposed, an unlinked
sender, several agents to choose from), the router replies and returns
normally, the session commits, and the rows persist with `message_id` NULL and
the bytes on disk. `TranscriptService.record` in `_run`'s `finally` covers run
*failure*, but not a refusal *before* the run.

`chat_files` has no `organization_id`; scoping is `user_id` plus
message→conversation→org, which holds only once the row is linked — an orphaned
row is scoped by `user_id` alone.

## Where

`backend/app/services/channels/router.py` (`_receive_files` before
`answer_default`), `backend/app/db/models/chat_file.py`

## How I would know it is fixed

A turn refused before the run leaves no unlinked `ChatFile` row and no orphaned
bytes — either the files are received after the refusal checks, or a refusal
cleans them up.

Pre-existing, found reviewing #634.

Issue #660, opened 2026-08-13
A non-mention channel message with attachments stores every file twice


DEENUU1:
## What breaks

A non-mention channel message with attachments downloads and stores every file
twice. `_answer_mention` calls `_receive_files` (`router.py:370`) before
`parse_mention` runs inside `answer()`; on an `UnaddressedMessage` it returns
`False`, and `_route_inner` then calls `_receive_files` again (`router.py:249`)
on the default path. Two downloads, two `FileUploadService.upload` calls, two
`ChatFile` rows — only the second set linked to the turn.

Pre-existing on `main` (both call sites are older), but the branch that made a
channel file a visible row (`feat/agent-surfaces`, #634) is what turns the
duplicate into a user-visible artifact.

## Sequence

1. In a DM, or a channel where the bot is addressed with no `@slug`, send a
   message with a file and no handle.
2. `_answer_mention` receives the file, `answer()` raises `UnaddressedMessage`,
   it returns `False`.
3. The default path receives the same file again → a second download, a second
   `ChatFile` row.

## Where

`backend/app/services/channels/router.py:370` and `:249`

## How I would know it is fixed

One download and one `ChatFile` row per file on the default path. Cheap fix:
check `parse_mention(incoming.text)` before receiving in `_answer_mention`, or
receive once in `_route_inner`.

Pre-existing, found reviewing #634. Related: #547 (transports bypass
parse_incoming, dropping attachments).

Issue #658, opened 2026-08-13
run_now executes the agent synchronously in the HTTP request (504 + double-fire on slow runs)


OchnikBartek:
## What breaks

`POST /agents/{agent_id}/triggers/{trigger_id}/run` (`run_now`) executes the whole
agent run **synchronously inside the HTTP request**. `AgentTriggerService.run_now`
awaits `fire()`, which runs `AgentRunnerService.execute` to completion before the
route returns (`backend/app/services/agent_trigger.py:413`,
`backend/app/api/routes/v1/agent_triggers.py:59`).

For a fast agent this is fine. For a slow one, behind a default nginx
(`proxy_read_timeout` 60s) the caller gets a **504 while the run keeps going and
commits server-side** — which reads as "it failed" and invites a second press, i.e.
a second fire. The frontend's "Running now" toast also only appears *after* the run
has already finished, because the request it is tied to does not return until then.

The run itself is correct and its cost is bounded by the budget; this is a
request-lifecycle problem, not a run-accounting one.

## The sequence that triggers it

1. A trigger whose agent takes longer than the proxy read timeout to run.
2. A user clicks "Run now".
3. The request is held for the full run; nginx returns 504 before it completes.
4. The user, seeing a failure, clicks again → a second fire of the same schedule.

## The fix shape (deferred, not in #537)

The heartbeat already has the answer: `run_deployment(..., timeout=0)` submits the
run as its own flow and returns immediately (`backend/app/worker/tasks/trigger_tasks.py`).
`run_now` should do the same — dispatch the fire, return `202 Accepted`, and let the
trigger's run-log conversation in the sidebar show progress. The frontend toast
becomes "started" rather than "finished", and the endpoint stops returning the
advanced `last_run_id` inline (a contract change the widget must adapt to).

## How we'd know it's fixed

- `run_now` returns without awaiting the run to completion (a `202`, not a `200`
  carrying the finished run).
- A slow fired run no longer holds the request or its DB session open.
- A regression test that `run_now` dispatches rather than runs inline.

## Context

- Found reviewing #537 (agent triggers, Closes #44); the reviewer flagged it
  non-blocking and explicitly did **not** hold #537 for it.
- Adjacent to the fire-path hardening in #589, but a distinct change (that PR is the
  in-flight marker; this is the request lifecycle of a manual fire).
- Pre-existing in the triggers feature as merged — this branch found it, it did not
  introduce it.
[...]

PR #634, opened 2026-08-12
feat(channels): surface parity, the socket as an integration, and a hosted chat page

DEENUU1:
One branch for the three surface issues in the W2 milestone, and it grew.
They were one thread of work to begin with: #39 makes the existing surfaces
behave identically, #516 makes the most capable of them (the embed
WebSocket) a product option rather than a paragraph in the docs, and #517
builds the missing surface — a hosted page — on top of both. Doing them
together was deliberate: each one touches the same object (`agent_embeds`,
`RunSurface`, `docs/channels.md`), and doing them apart is how the surfaces
drifted in the first place.

Opening a bot in a real Mattermost channel and a page in a real browser then
found six more, all of them the same shape — a surface behaving unlike its
siblings — so they landed here rather than in six branches over the same
files. 62 commits, 128 files.

## What this closes

| | What makes it closed |
|---|---|
| **#39** Other channels setup | Every surface now runs the same turn loop, remembers its conversation, and is rate-limited. The widget was the only surface passing no `message_history`; the hosted page and the widget now stream the same frames the web chat does (`410fac8d`, `3865b4cb`, `a6c05f15`) |
| **#516** Offer the socket as an integration | `socket_url_for` beside `snippet_for`, published in the panel with the `Origin` rule and no `?token=` (`a2e8bef6`) |
| **#517** A hosted chat page | `/e/[publicKey]`, `hosted_config`, visitor continuity, the Builder controls, and editing a published one afterwards (`076513a7`, `a9a67a42`, `c606686c`, `e53edfc4`) |
| **#636 / #638** A thread past 200 messages sends the model its *first* 200 | `count_messages` + `skip`, so the window is the last 200. Two issues, one bug, one fix (`8b33b992`) |
| **#637** An update sending `null` answers 500 on a `NOT NULL` column | `app/db/updates.py` — the column decides, not a hand-kept list — applied to every `*Update`, with a guard over `app/services/**` **and** `app/api/**` (`3c9afd73`, `943635b7`, `f39b9e35`) |
| **#639** A group channel refuses every sender who never linked | A room admits an unlinked speaker and the turn runs as the binding's creator; a DM is unchanged; `require_link: true` is the opt-out and is now tested. `agent_runs.channel_identity_id` records who actually asked (`b64c6226`, `9a6bff9a`, migration `0024`) |
| **#640** Two copies of "an anonymous surface runs as its publisher" | One `access.publisher_context`, read by the embed session and by channels (`a5463c4d`) |
[...]

PR #685, opened 2026-08-13
fix(channels): gate a Mattermost webhook post the way the stream is

### LINKED ISSUE #662: Mattermost webhook transport answers every post, unlike the stream transport
## What breaks

The Mattermost **webhook** transport answers every delivered post, while the
docs present mention-gating as unconditional. `_from_webhook`
(`mattermost.py`) never sets `addressed`, so `IncomingMessage.addressed` is
`None` → `_names_the_bot` is `True` → the bot answers every post and posts
unknown-handle refusals out loud. The socket (stream) transport reads the
platform's mention list; the webhook one does not.

`docs/channels.md`'s addressing table ("A channel — only when it is named")
sits under the Mattermost section without distinguishing transports.

Mitigated in practice: Mattermost webhooks are trigger-word / channel-filtered
on the Mattermost side. But the rule should say it is stream-only, and the
webhook path should read the mention list if the event carries one.

Unverified and worth one real-server check alongside this: whether Mattermost
puts the bot's id in `data.mentions` for `@channel` / `@all` / `@here` — if it
does, `_addressed` reads every "@channel standup at 10" as addressing the bot.

## Where

`backend/app/services/channels/mattermost.py` (`_from_webhook`, `_addressed`),
`docs/channels.md` (Mattermost addressing table)

A webhook-delivered post that names nobody is not answered, and the docs say
which transport the mention rule applies to. `@channel` behaviour is pinned by
a test or documented as intended.

Pre-existing, found reviewing #634. Related: #547 (webhook bypasses
parse_incoming).

DEENUU1:
## What this fixes

Closes #662 — the Mattermost outgoing-webhook transport answered every post it
was handed, while the event stream had already stopped doing that.

## Stacked, deliberately — base is `feat/agent-surfaces` (#634), not `main`

Not an accident, and worth checking before merging. The addressing rule this
defect is *about* exists only on that branch:

| Symbol | On `main`? |
|---|---|
| `IncomingMessage.addressed` | No |
| `ChannelMessageRouter._is_overheard` / `_names_the_bot` | No |
| `MattermostAdapter._addressed` / `_own_ids` / `_own_user_id` | No |
| `tests/test_channel_addressing.py` | No |

`main`'s router has no addressing gate on *either* transport — it answers every
channel message on both — so there is nothing on `main` to fix, and a regression
test written there would assert behaviour that does not exist. Same shape as #658
finding the triggers feature only on PR #537.

**Consequence for the merge:** this is a stacked PR, so `gh pr merge --admin`
will refuse it. Land #634 first and this retargets to `main` cleanly; or merge
`main` into the base and de-stack.

## What changed

- `backend/app/services/channels/mattermost.py` — `_from_webhook` now sets
  `addressed`. An outgoing-webhook body carries no mention list, so what stands
  in for one is `trigger_word`: the word an operator told *this integration* to
  fire on, which is Mattermost's own record that the post was for us. Empty means
  the webhook fired on its channel filter, which delivers every post exactly as
  the socket does — `False`, for the same reason a `posted` event with no
  mentions is. The reasoning is in the docstring.
- `backend/tests/test_channel_addressing.py` — four cases for the webhook
  transport, asserting `_is_overheard` and not only the field.
- `docs/channels.md` — the addressing table now says it is the bot's rule and
  both transports obey it, with a second table for *how* each one reads "named".
  The trigger-word consequence is repeated once in the webhook setup steps,
  where an operator is standing when they choose them.

A direct message and an `@agent-slug` are untouched: both are decided before the
flag is read, so `@sales what is the refund window` still works over a
channel-filtered webhook and a DM still answers everything.

The cost, and it is documented: `@the-bot` is not readable on this transport,
because nothing in the body says which account the bot is. An operator who wants
that sets the trigger word to the bot's handle.

## How it failed before

`_from_webhook` left `addressed` at its default `None`. The router reads `None`
as "the platform did not say" — deliberately, so Slack and Telegram keep working
— and answers. So a bot on a channel-watched outgoing webhook replied to
colleagues talking to each other, and posted its unknown-handle refusals
("@ada is not available on this bot") out loud in a channel it was never
addressed in.

## Testing

The three new field assertions fail against the old parser with
`AssertionError: assert None is False` and pass with the fix; the overheard
assertion fails with them.

- `uv run pytest tests/test_channel_addressing.py tests/test_mattermost_channel.py tests/test_channel_mentions.py -q` → 85 passed
- `make lint-backend` → clean
- `make test` → 4622 passed, 7 skipped, coverage 100%
- `make docs-build` → clean

## CI is not fully green, and the red is not mine

`lint`, `docs`, `Security Scan` and `test` (the backend suite plus the 100% gate)
pass. `test-frontend` and `docker` skip — backend-and-docs-only paths, and a
skipped required check is a pass.

**`e2e` fails: 4 failed, 97 passed.** All four are `e2e/surfaces.spec.ts`, all
four are the same assertion in the `publishPage` helper —
`expect(submit).toHaveCount(0)` timing out at 15s with the Publish button still
on screen:

- `:64` A published surface › is editable after it exists, and the change survives a reload
- `:96` A published surface › takes a picture of its own once it exists
- `:128` What the page offers a visitor › starts a fresh thread without losing the link
- `:170` What the page offers a visitor › offers dictation only where the operator turned it on

Confirmed already red on the base ref rather than assumed. Run
[31692218637](https://github.com/vstorm-co/agenticos/actions/runs/31692218637)
on `feat/agent-surfaces` at sha `3f58155` — the head this branch was cut from —
reports `4 failed, 96 passed`: the same four specs, the same assertion. Mine has
one more pass, which is this branch's own added coverage.

The spec does not exist on `main` (`git ls-tree origin/main
frontend/e2e/surfaces.spec.ts` is empty) and nothing in this diff reaches a
hosted surface — the whole change is a Mattermost webhook parser's return value,
its docstring, tests and docs. Not touched here; it belongs to #634.

## Overlaps

#660 (a non-mention channel message with attachments stores every file twice)
and #661 (a channel turn refused before the run orphans its ChatFile rows) are
open against neighbouring code — `ChannelMessageRouter._receive_files` and its
two call sites — and are being worked in parallel. **Merge order matters.**

This branch does not touch either: the whole diff is `_from_webhook`'s return
value, its docstring, tests and docs. Nothing here needs what #660 or #661 owns.

## Noticed, not fixed

`@channel` / `@all` / `@here` on the **stream** transport. The bot answers a
broadcast wherever Mattermost lists it among the notified. #662 raised that as an
open question; this PR answers it by writing it down in `docs/channels.md` as
intended — a private notion of "who was this for" here would disagree with the
one the whole channel was shown. Nobody has checked a real server to see when
Mattermost actually puts a bot in that list, so if it turns out to be noisy in
practice that is a new issue rather than a silent filter added on a guess.

### DIFF
--- a/backend/app/services/channels/mattermost.py
+++ b/backend/app/services/channels/mattermost.py
@@ -766,6 +766,29 @@ class MattermostAdapter(ChannelAdapter):
         return response.content
 
     def _from_webhook(self, payload: dict[str, Any], bot_id: str) -> IncomingMessage | None:
+        """Normalise one outgoing-webhook body.
+
+        The addressing rule is the socket's, read off what this payload actually
+        carries. An outgoing-webhook body has no mention list - Mattermost sends
+        the post, not who it notified - so the bot's own account cannot be looked
+        for in it, and this path left `addressed` unset. The router reads unset as
+        "the platform did not say" and answers, which put the bot back in the
+        position the socket's rule took it out of: replying to colleagues talking
+        to each other in a channel it was merely invited to (#662).
+
+        What the payload does say is `trigger_word` - the word an operator
+        configured *this integration* to fire on, which is Mattermost's own record
+        that the post was for us. Empty means the webhook fired on its channel
+        filter alone, and a channel filter delivers every post exactly as the
+        socket does, so it is `False` for the same reason a `posted` event with no
+        mentions is.
+
+        The consequence worth knowing before choosing this transport: `@the-bot`
+        is not readable here, because nothing in the body says which account the
+        bot is. Set the trigger word to the bot's handle if that is how people
+        should reach it; an `@agent-slug` needs nothing, since the router reads a
+        slug out of the text.
+        """
         text = str(payload.get("text") or "").strip()
         if not text:
             return None
@@ -790,4 +813,5 @@ class MattermostAdapter(ChannelAdapter):
             platform_username=str(payload.get("user_name") or "") or None,
             platform_display_name=str(payload.get("user_name") or "") or None,
             message_id=str(payload.get("post_id") or "") or None,
+            addressed=bool(str(payload.get("trigger_word") or "").strip()),
         )
--- a/backend/tests/test_channel_addressing.py
+++ b/backend/tests/test_channel_addressing.py
@@ -21,6 +21,11 @@ regression, and the quietest to ship.
 *Silence where the platform did not say.* Slack and Telegram deliver on their own
 subscription rules, so reading "no mention data" as "ignore" would take a working
 bot on either of them off the air.
+
+And the rule belongs to the *bot*, not to one way of reaching it: Mattermost's
+outgoing webhook hands over a watched channel's posts the same way the socket
+does, and answered all of them for as long as it said nothing about who they
+named (#662).
 """
 
 from __future__ import annotations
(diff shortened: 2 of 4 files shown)


DEENUU1:
CI: everything green except `e2e`, and that red is inherited, not caused here.

Four failures, all `e2e/surfaces.spec.ts`, all the same assertion in
`publishPage` — `expect(submit).toHaveCount(0)` timing out with the Publish
button still on screen:

- `A published surface › is editable after it exists, and the change survives a reload`
- `A published surface › takes a picture of its own once it exists`
- `What the page offers a visitor › starts a fresh thread without losing the link`
- `What the page offers a visitor › offers dictation only where the operator turned it on`

The base branch has the same four at its own head. Run
[31692218637](https://github.com/vstorm-co/agenticos/actions/runs/31692218637)
on `feat/agent-surfaces` (#634, sha `3f58155`): `4 failed, 96 passed`, identical
specs and identical assertion. Mine is `4 failed, 97 passed` — the extra pass is
this branch's own suite, and nothing in the diff reaches a hosted surface.

`test` is green (the backend suite plus the 100% gate), as are `lint`, `docs` and
the security scan. `test-frontend` skipped: backend-and-docs-only paths.

Issue #656, opened 2026-08-12
check-i18n misses prose whose first word is hyphenated ("Sign-in failed")


DEENUU1:
## What breaks

The i18n guard's string-literal rule does not flag prose whose first word is hyphenated: `"Sign-in failed"` and `"Sign-in successful"` pass the sweep while `"Not authenticated"` is reported. That is how both strings survived in `src/app/api/auth/oauth-callback/route.ts` and `magic-link/verify/route.ts` through the #603 sweep (fixed in #654 by review, not by the guard).

## Sequence

1. Put `const x = "Sign-in failed";` in any swept component.
2. `bun run check:i18n` — nothing reported.
3. Replace with `"Login failed"` — reported.

## Where

`frontend/scripts/check-i18n.ts` — whichever predicate decides a literal "holds a sentence" (the capital-and-a-space test or its word segmentation) reads `Sign-in` as a non-word, leaving the string under the prose threshold.

## How you would know it was fixed

A spec in `frontend/scripts/check-i18n.test.ts` pinning `saidTs('const x = "Sign-in failed";\n')` to one offence, and the sweep still clean on the real tree.

## Where it sits

The reading-model chain: #199 → #246 → #249 → #314 → #141 → #395 → #446 → #603. Same class as the shapes #395's parser was built to end — a threshold standing in for a parse — but in the one rule that still has a heuristic (what counts as prose). Found reviewing #654; pre-existing.

PR #688, opened 2026-08-13
fix(frontend): detect prose after leading acronyms

### LINKED ISSUE #678: check-i18n misses prose whose first word is an acronym ("API keys are stored in the vault")
## What breaks

The i18n guard's string-literal rule does not flag prose whose first word is an
acronym. `"API keys are stored in the vault"` and `"MCP servers appear here once
connected"` both pass the sweep, while `"Provider keys are stored in the vault"` is
reported.

`NOT_A_SENTENCE`'s second alternative is `[A-Z]{2,}\s` — a leading acronym — and it
exempts the whole string rather than the token. This product's copy opens on `MCP`,
`API`, `RAG`, `KB` and `JWT` all over the place, so the hole is the width of a
vocabulary.

## Sequence

1. Put `const x = "API keys are stored in the vault";` in any swept component.
2. `bun run check:i18n` — nothing reported.
3. Replace `API` with `Provider` — reported.

## Where

`frontend/scripts/check-i18n.ts`, `NOT_A_SENTENCE`, the `[A-Z]{2,}\s` alternative.

A spec in `frontend/scripts/check-i18n.test.ts` pinning
`saidTs('const x = "API keys are stored in the vault";\n')` to one offence, a companion
spec keeping a bare acronym label (`"API"`, `"MCP server"`) out of the report, and the
sweep still clean on the real tree. Nothing in `src/` is currently hidden by it beyond
the `"AI agents"` SEO keyword in `src/lib/seo.ts`, so this is a hole rather than a
backlog.

## Where it sits

The same class as #656, in the same regex and the same rule — an alternative written to
exempt a *token* that exempts the *sentence* it opens. Found fixing #656; pre-existing,
and deliberately not folded into that branch, which is about the separator alternative.
The reading-model chain: #199 → #246 → #249 → #314 → #141 → #395 → #446 → #603 → #656.
(#168, the issue map, has an empty body, so there is no cluster entry to add this to.)

breezeFur:
## What this changes

The TypeScript string-literal sweep now reports sentence-like copy when it starts with an acronym, while still allowing compact labels such as `API` and `MCP server`. This closes #678.

## Why this way

The leading-acronym alternative was broad enough to exempt the rest of any matching string. Narrowing it to a complete two-word acronym label keeps the intended machine-label exception without hiding longer prose. The regression tests cover both sides of that boundary.

## Checks

- [ ] Platform-layer coverage is still 100%
- [ ] If a capability was added or changed, its `README.md` says why
- [ ] If a migration was added, `alembic downgrade base && alembic upgrade head` works

## Notes for the reviewer

- `bun run vitest run scripts/check-i18n.test.ts` — 70 passed
- `bun run prettier --check scripts/check-i18n.ts scripts/check-i18n.test.ts` — passed
- `bun run eslint scripts/check-i18n.ts scripts/check-i18n.test.ts --max-warnings 0` — passed
- `bun run type-check` — passed
- The whole-tree `bun run check:i18n` is not a reliable Windows baseline: backslash paths do not match the existing `/dev/` and `/app/api/` exclusions, so it reports 177 existing strings in excluded paths. This change does not add those offences.


frontend/scripts/check-i18n.ts
```diff
@@ -200,11 +200,11 @@ const UNITS = new Set([
...
 const MACHINE_WORD = /^[A-Z0-9_]+$/;
 /**
- * What `NOT_A_SENTENCE` kept out of the string-literal rule: a label built from title
- * case around a separator, a leading acronym, a CSS measurement.
+ * What `NOT_A_SENTENCE` keeps out of the string-literal rule: a label built from title
+ * case around a separator, a short acronym label, a CSS measurement.
  */
 const NOT_A_SENTENCE =
-  /^(?:[A-Z][a-z]+(?:\s[A-Z][a-z]+)*\s?[-/]\s?|[A-Z]{2,}\s|.*\b(?:px|rem|vh|vw|deg)\b)/;
+  /^(?:[A-Z][a-z]+(?:\s[A-Z][a-z]+)*\s?[-/]\s?|[A-Z]{2,}\s+[a-z][A-Za-z-]*$|.*\b(?:px|rem|vh|vw|deg)\b)/;
```
DEENUU1: The branch is behind `main` and the conflict lands on exactly this line. #677 landed while this was open: it added the `[A-Z]` after the separator in the *first* alternative, which is the fix for #656 (`Sign-in failed` was reading as a label). Resolve the conflict by taking this side and that `[A-Z]` disappears — the same class of bug as #678, one alternative to the left.

Your narrowing itself is right. This is both changes merged, plus a line saying why the anchor is there: every other narrowing in this regex carries its issue number, and the next person tempted to widen it needs #678 the way they need #656.

```suggestion
 * What `NOT_A_SENTENCE` keeps out of the string-literal rule: a label built from title
 * case around a separator, a short acronym label, a CSS measurement.
 *
 * The separator joins *two capitalised tokens* - `Model / Provider`, `Agent - Settings`
 * - and the `[A-Z]` after it is the whole of #656. Without it a hyphen inside the first
 * word was the separator, so `Sign-in failed` read as a label and every sentence whose
 * first word is hyphenated left the sweep silently: two of them survived #603 in the BFF
 * handlers and were found by review rather than by this (#654). Prose after a hyphen
 * carries on in lower case, which is the one thing a two-token label never does.
 *
 * The acronym alternative is anchored for the same reason (#678). `[A-Z]{2,}\s` exempted
 * the sentence its acronym opened, so `API keys are stored in the vault` left the sweep
 * while `Provider keys are stored in the vault` did not. Ending it on a single lower-case
 * word keeps the label it was written for - `MCP server`, `AI agents` - and lets the prose
 * through.
 */
const NOT_A_SENTENCE =
  /^(?:[A-Z][a-z]+(?:\s[A-Z][a-z]+)*\s?[-/]\s?[A-Z]|[A-Z]{2,}\s+[a-z][A-Za-z-]*$|.*\b(?:px|rem|vh|vw|deg)\b)/;
```

Checked against current `main`: prettier clean, 71/71 specs pass, and the whole-tree sweep reports nothing.

DEENUU1: One thing the narrowing takes away that I don't think you meant to. The old `[A-Z]{2,}\s` was accidentally covering separator labels whose first token is an acronym, because the first alternative only accepts a title-case token (`[A-Z][a-z]+`):

- `Model / Provider` — exempt before and after
- `URL / Endpoint` — exempt before, reported now

Nothing in `src/` is written that way today, so the sweep stays clean either way and this is not blocking. If you want it, teaching the first alternative about acronyms is enough:

```
/^(?:(?:[A-Z][a-z]+|[A-Z]{2,})(?:\s[A-Z][a-z]+)*\s?[-/]\s?[A-Z]|[A-Z]{2,}\s+[a-z][A-Za-z-]*$|.*\b(?:px|rem|vh|vw|deg)\b)/
```

I checked it does not reopen either hole — `Sign-in failed`, `Auto-refresh is on`, `API-key rotation failed` and `API keys are stored in the vault` all still report. Happy for you to leave it if you'd rather keep the diff tight, and I'll file it instead.

frontend/scripts/check-i18n.test.ts
```diff
@@ -481,6 +481,19 @@ describe("a .ts file", () => {
     expect(saidTs(source)).toEqual(["string 'Your profile'"]);
   });
 
+  it("refuses prose that opens with an acronym (#678)", () => {
```
DEENUU1, line 499: Good — both of the issue's examples pinned rather than one, and the `it` names say what broke instead of which function ran. That is what stops somebody widening this alternative back later.

frontend/scripts/check-i18n.test.ts
```diff
@@ -481,6 +481,19 @@ describe("a .ts file", () => {
...
+    expect(saidTs('const copy = "API keys are stored in the vault";\n')).toEqual([
+      "string 'API keys are stored in the vault'",
+    ]);
+    expect(saidTs('const copy = "MCP servers appear here once connected";\n')).toEqual([
+      "string 'MCP servers appear here once connected'",
+    ]);
+  });
+
+  it("keeps a short acronym label out of the report (#678)", () => {
+    expect(saidTs('const labels = ["API", "MCP server"];\n')).toEqual([]);
```
DEENUU1: Small thing about what this pins: `"API"` passes whether the alternative exists or not. It has no whitespace, so `readString`'s `/^[A-Z][^\n]*\s/` test never reaches `NOT_A_SENTENCE` at all. `"MCP server"` is the half doing the work here.

Worth adding the one string in the tree that actually depends on this rule, and the boundary the `$` buys you:

```suggestion
    expect(saidTs('const labels = ["API", "MCP server", "AI agents"];\n')).toEqual([]);
    // One lower-case word, not a phrase: the `$` is the whole of the narrowing.
    expect(saidTs('const label = "MCP server URL";\n')).toEqual(["string 'MCP server URL'"]);
```

Both pass on a rebased branch — I ran them.

DEENUU1 (dismissed):
Thanks for this — the diagnosis is right and the fix is the shape I'd have picked. The alternative was exempting the sentence its acronym opened rather than the acronym itself, and ending it on a single lower-case word is the smallest thing that separates the two.

I ran the check you couldn't. With your change on top of current `main`, `bun run check:i18n` over the whole tree reports nothing and all 71 specs pass — so the Windows caveat in your description is answered. Nothing new is flagged, and the one string in `src/` that actually leans on this rule (`"AI agents"` in `src/lib/seo.ts:40`, which carries no exemption) stays exempt.

Requesting changes for one reason, and it is mechanical rather than a problem with your code: the branch is behind `main` and conflicts on exactly this line. #677 landed the fix for #656 in the alternative next door, and resolving the conflict by taking your side reverts it silently. I've put the merged line in a suggestion below, plus one optional narrowing and a nit on the specs.

Separately, and not yours to fix: reviewing this I noticed the third alternative (`.*\b(?:px|rem|vh|vw|deg)\b`) has the same defect much wider — `.*` on both sides, so `"Use rem instead of pixels for spacing"` is exempt — and deleting it outright changes nothing on the real tree. I'll file that as its own issue.

DEENUU1:
Filed the unit-alternative one as #741 so it does not get lost — not something you need to touch here.

frontend/scripts/check-i18n.ts
DEENUU1: Resolved by merging current `main` into the branch — `origin/main` had moved four commits further (0.0.151, #587, 0.0.152) since this was opened. Took both sides as you described: the `[A-Z]` from #677 stays on the separator alternative, the `$` anchor lands on the acronym one, and the docstring carries both issue numbers.

DEENUU1: Taken rather than filed — a regression this branch introduces is cheaper to fix in it than to carry as an issue, and you had already checked the alternation. `987ba662`: the separator's left token is `(?:[A-Z][a-z]+|[A-Z]{2,})`, so `URL / Endpoint` is exempt again while `Model / Provider` is unchanged.

Pinned on both sides, since the point of the fix is that neither hole reopens: one spec for the label (`Model / Provider`, `URL / Endpoint`, `API - Settings`), one for the prose the two anchors exist for (`Sign-in failed`, `API-key rotation failed`). 73/73 pass and the whole-tree sweep is clean.

frontend/scripts/check-i18n.test.ts
DEENUU1: Already applied in `ab231683` — the array is `["API", "MCP server", "AI agents"]` with the `MCP server URL` boundary and the comment saying the `$` is the whole of the narrowing. `"AI agents"` is the one string in the tree (`src/lib/seo.ts`) that actually depends on the rule, so it is the half carrying the spec, as you say.

Issue #565, opened 2026-08-10
Lift the channel transport boilerplate into base.py helpers


DEENUU1:
The thread-split `x,_,y = id.partition(":")` appears at 6 sites (`slack.py:65,77,87`; `mattermost.py:140,183,231`); Telegram `Bot(token=...)+try/finally close` at 7 (a fresh TLS session per reply and per streamed edit); the Mattermost `httpx.AsyncClient`/auth header at 8 each; Slack's lazy `AsyncWebClient` at 7.

**Cost today.** A 4th channel copies it all again; a change to Telegram session handling, MM timeout/auth, or the thread split must be applied at 6–9 sites by hand and one will be missed.

**Target.** `split_thread()` in `base.py`; an `async with self._bot(token)` context manager for Telegram; `self._client()`/`self._headers()` for Mattermost; a Slack `self._web(token)` factory.

**Size:** M. Enables #547 (one normaliser per adapter).

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

DEENUU1:
One more site for the sweep, from review on #743: `supervisor.py:62` guards `remember_server` on truthiness, so clearing `api_base_url` on a polling bot leaves the old address in the adapter singleton map — the same stale-entry edge the webhook receiver just closed by mirroring the row unconditionally (`mattermost_webhook.py`, e966c24e). When the transport state lifts into `base.py`, mirror-unconditionally is the shape both call sites should keep.

Issue #547, opened 2026-08-10
Telegram polling and Mattermost webhook bypass parse_incoming, dropping attachments


DEENUU1:
Each channel adapter has two inbound paths that must produce the same normalised `IncomingMessage`. Slack funnels both through `parse_incoming`; Telegram and Mattermost hand-built a second copy, and it drifted.

- `telegram.py:414-438` (`_handle_update`, polling): `if not message.text: return`, `raw={}`, **no `_attachments()` call** — vs `telegram.py:305-349` (`parse_incoming`, webhook): `text or caption`, `_attachments`, `if not text and not attachments`.
- `mattermost.py:723-748` (`_from_webhook`): no `_attachments` call — vs `mattermost.py:633-669` (`_from_socket`): has it.

**How it fails.** A file dropped on a Telegram bot in polling mode, or a caption-only photo, is silently discarded and the agent answers about a document it never received — the exact regression `base.py:101-109` documents as fixed. #113 wired attachments through `parse_incoming`; these two paths were never brought along. (Reachability: Telegram polling is the dev/self-host mode; MM `_from_webhook` is the outgoing-webhook path.)

**Fix.** Make the alternate path reconstruct the raw payload and call `self.parse_incoming(...)` — one normaliser per adapter. #565 removes the duplication that let it drift.

**Done when.** Both paths produce attachments; a per-adapter regression test asserts a file-bearing update on the alternate path yields `attachments != []`.

Refs #113 (added attachment support; closed), #565.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

PR #743, opened 2026-08-13
fix(channels): record a Mattermost webhook bot's server

### LINKED ISSUE #692: A Mattermost webhook bot has no server URL, so its attachments cannot be fetched
A Mattermost bot in webhook mode never has its server URL recorded, so every
attachment on an outgoing-webhook post parses with an empty handle and cannot be
fetched. The reply names the file and says it could not be downloaded.

**Where.** `app/services/channels/mattermost.py:672-679` resolves the handle from
`self._base_urls[bot_id]`, and the only writer is `remember_server`, called from
`open_inbound_stream` (`app/services/channels/supervisor.py:61-63`). That runs for
polling bots only: `main.py:62` and `ChannelBotService` both reach it through
`get_active_polling_bots`, which filters `ChannelBot.webhook_mode.is_(False)`
(`app/repositories/channel_bot.py:71`). A webhook-mode bot therefore never appears,
`_base_urls` has no entry, and `_attachments` produces `handle=""`.

`download_attachment` (`mattermost.py:690-700`) refuses an empty handle by design -
guessing a Mattermost address is guessing which company's server to send a bot token
to - and `ChannelAttachmentService.receive` turns that into
`"report.csv: could not be downloaded."` in the reply.

**Sequence.**

1. Register a Mattermost bot with `webhook_mode=true` and an `api_base_url`.
2. Create the outgoing webhook in Mattermost's System Console, pointing at
   `/api/v1/mattermost/{bot_id}/webhook`.
3. Post a file to the watched channel.
4. The bot answers with the question, and `<name>: could not be downloaded.`

**Not a regression.** Before #547 the webhook parser read no `file_ids` at all, so
the file was dropped silently and the agent answered about a document it never
received. #547 makes the failure visible and honest; it does not make the file
reachable. `download_attachment`'s docstring on `main` already names this path.

**Fixed when.** A file posted to a webhook-mode Mattermost bot is downloaded and
stored, and a test asserts the handle resolves against the bot's own
`api_base_url` on that transport.

The bot row is already in hand at the receiver
(`api/routes/v1/mattermost_webhook.py:41` calls `bot_service.find_active`), so the
address is available where the parse happens; what needs deciding is whether the
route tells the adapter, or whether the address stops being adapter-singleton state.
The second is what #565 (lift the channel transport boilerplate into `base.py`) would
have to answer anyway.

Cluster: channels — #547 (this branch, PR pending), #565, #662, #660, #661, #690.
Nearest neighbour is #547; found while fixing it.

DEENUU1:
## What this fixes

A Mattermost bot in webhook mode never had its server URL recorded, so every attachment on an outgoing-webhook post parsed with an empty handle and could not be fetched — the reply named the file and said it could not be downloaded. Closes #692.

## What changed

- `backend/app/api/routes/v1/mattermost_webhook.py` — after the token check and before the parse, the receiver tells the adapter the bot row's `api_base_url` via `remember_server`. Per delivery, which also keeps the address fresh when an operator edits the bot.
- `backend/app/services/channels/mattermost.py` — two docstrings that described the old behaviour ("the outgoing-webhook path never calls `remember_server`") corrected.
- `backend/tests/test_channel_webhook_secret.py` — route-level regression test; `_sealed_bot` grows an `api_base_url` parameter.

## How it failed before

The only writer of the adapter's per-bot address map was `remember_server`, reached through `open_inbound_stream` — which runs for polling bots only (`get_active_polling_bots` filters `webhook_mode.is_(False)`). A webhook-mode bot never appeared there, `_base_urls` had no entry, and `_attachments` produced `handle=""`. `download_attachment` refuses an empty handle by design — guessing a Mattermost address is guessing which company's server to send a bot token to.

Not a regression: before #547 the webhook parser read no `file_ids` at all, so the file was dropped silently. #547 made the failure visible; this makes the file reachable.

## Testing

- `tests/test_channel_webhook_secret.py::TestTheReceiverRecordsTheServer::test_a_file_on_a_webhook_delivery_resolves_against_the_bots_own_server` — fails on `main` with `[''] == ['https://mm.example.com/api/v4/files/f1']`, passes here. Goes through the actual route with a signed body, so it pins the wiring, not just the adapter.
- `tests/test_channel_webhook_secret.py`, `tests/test_mattermost_channel.py`, `tests/test_channel_adapter_attachments.py`, `tests/test_channel_supervisor.py` — 78 passed.
- `make lint-backend` clean.

## Noticed, not fixed

- The issue names the alternative — the server address stops being adapter-singleton state altogether — as #565's question (lift the channel transport boilerplate into `base.py`). Deliberately not taken here; the route-tells-the-adapter shape is the scoped fix and #565 can subsume it.

### DIFF
--- a/backend/app/api/routes/v1/mattermost_webhook.py
+++ b/backend/app/api/routes/v1/mattermost_webhook.py
@@ -21,7 +21,7 @@ from app.api.deps import ChannelBotSvc
 from app.core.background import spawn
 from app.services.channel_bot import unseal_webhook_secret
 from app.services.channels import get_adapter
-from app.services.channels.mattermost import decode_webhook_body
+from app.services.channels.mattermost import MattermostAdapter, decode_webhook_body
 from app.worker.background.channel import process_channel_event
 
 logger = logging.getLogger(__name__)
@@ -52,6 +52,12 @@ async def mattermost_webhook(
     secret = unseal_webhook_secret(bot)
     if not secret or not adapter.verify_webhook_signature(dict(request.headers), secret, raw):
         raise HTTPException(status_code=403, detail="Invalid webhook token")
+
+    # A webhook-mode bot opens no stream, so this is the only place its server
+    # can reach the adapter - and the parser resolves attachment handles from
+    # it (#692). The isinstance narrows the registry's base type.
+    if isinstance(adapter, MattermostAdapter) and bot.api_base_url:
+        adapter.remember_server(str(bot_id), bot.api_base_url)
 
     incoming = adapter.parse_incoming(decode_webhook_body(raw), str(bot_id))
     if incoming is None:
--- a/backend/app/api/routes/v1/mattermost_webhook.py
+++ b/backend/app/api/routes/v1/mattermost_webhook.py
@@ -55,9 +55,11 @@ async def mattermost_webhook(
 
     # A webhook-mode bot opens no stream, so this is the only place its server
     # can reach the adapter - and the parser resolves attachment handles from
-    # it (#692). The isinstance narrows the registry's base type.
-    if isinstance(adapter, MattermostAdapter) and bot.api_base_url:
-        adapter.remember_server(str(bot_id), bot.api_base_url)
+    # it (#692). Mirrored even when empty, so clearing the row's address does
+    # not leave a stale one in the singleton map. The isinstance narrows the
+    # registry's base type.
+    if isinstance(adapter, MattermostAdapter):
+        adapter.remember_server(str(bot_id), bot.api_base_url or "")
 
     incoming = adapter.parse_incoming(decode_webhook_body(raw), str(bot_id))
     if incoming is None:
--- a/backend/tests/test_channel_webhook_secret.py
+++ b/backend/tests/test_channel_webhook_secret.py
@@ -25,7 +25,8 @@ from app.db.models.channel_bot import ChannelBot
 from app.main import app
 from app.schemas.channel_bot import ChannelBotCreate, ChannelBotUpdate
 from app.services.channel_bot import ChannelBotService, unseal_webhook_secret
-from app.services.channels import register_adapter
+from app.services.channels import get_adapter, register_adapter
+from app.services.channels.base import IncomingMessage
 from app.services.channels.mattermost import MattermostAdapter
 from app.services.channels.telegram import TelegramAdapter
 
@@ -332,7 +333,7 @@ class TestTheReceiverRecordsTheServer:
     ):
         secret = "the-real-secret"
         bot = _sealed_bot(secret, api_base_url="https://mm.example.com/")
-        captured: list = []
+        captured: list[IncomingMessage] = []
 
         async def _record(incoming) -> None:
             captured.append(incoming)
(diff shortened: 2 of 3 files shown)


backend/app/api/routes/v1/mattermost_webhook.py
```diff
@@ -53,6 +53,12 @@ async def mattermost_webhook(
     if not secret or not adapter.verify_webhook_signature(dict(request.headers), secret, raw):
         raise HTTPException(status_code=403, detail="Invalid webhook token")
 
+    # A webhook-mode bot opens no stream, so this is the only place its server
+    # can reach the adapter - and the parser resolves attachment handles from
+    # it (#692). The isinstance narrows the registry's base type.
+    if isinstance(adapter, MattermostAdapter) and bot.api_base_url:
+        adapter.remember_server(str(bot_id), bot.api_base_url)
```
DEENUU1: One edge the truthiness guard leaves open: if an operator clears `api_base_url` on the bot, the old address stays in the singleton map for the life of the process, so attachments on later deliveries keep resolving against a server the row no longer names. Every reader of `_base_urls` already treats an empty string as absent (`_attachments`, `_own_user_id`, `_run_stream`), so mirroring the row unconditionally closes it:

```suggestion
    if isinstance(adapter, MattermostAdapter):
        adapter.remember_server(str(bot_id), bot.api_base_url or "")
```

Non-blocking — the token is unchanged and belongs to that server, so nothing leaks anywhere new; it's just adapter state outliving what the row says. Note the supervisor (`supervisor.py:62`) has the same truthiness guard, so if you take this, that site is the sibling to sweep — probably under #565 rather than here.

backend/tests/test_channel_webhook_secret.py
```diff
@@ -316,3 +318,45 @@ async def test_an_unknown_bot_answers_200_without_running_anything(self, client:
...
+    and every attachment parsed with an empty handle - the file could be named in
+    the reply but never fetched (#692). The receiver holds the bot row, so it is
+    the one place the address can reach the adapter on this transport."""
+
+    async def test_a_file_on_a_webhook_delivery_resolves_against_the_bots_own_server(
+        self, client: AsyncClient
+    ):
+        secret = "the-real-secret"
+        bot = _sealed_bot(secret, api_base_url="https://mm.example.com/")
+        captured: list = []
```
DEENUU1: nit: bare `list` where the sibling test above annotates `processed: list[str]`. `captured: list[IncomingMessage] = []` (imported from `app.services.channels.base`) keeps the file fully typed and gives the `captured[0].attachments` access below a real type.

DEENUU1 (commented):
This is the right scoped fix for #692: the receiver already holds the bot row, and writing the address per delivery (instead of once at startup) is what makes this correct under multiple worker processes and keeps it fresh after an edit. The regression test going through the real route with a signed body is the strong part — it pins the wiring, not just the adapter — and correcting the two docstrings that described the old behaviour in the same change is exactly how those stay true. Two small non-blocking points inline: a stale-entry edge in the singleton map, and a typing nit in the test.

backend/app/api/routes/v1/mattermost_webhook.py
DEENUU1: Taken in e966c24e — the receiver mirrors the row unconditionally now, with a regression test priming the map with an old address and asserting the handle comes back empty once the row is cleared. The supervisor sibling is noted on #565.

backend/tests/test_channel_webhook_secret.py
DEENUU1: Fixed in e966c24e — `captured: list[IncomingMessage]`, imported from `app.services.channels.base`.

Issue #641, opened 2026-08-12
A channel thread stays visible to somebody removed from the channel


DEENUU1:
Deliberate, decided when the participant model was designed, and recorded here so it is not discovered as a surprise later.

`/chat` shows a channel thread to everybody whose linked chat account has written
in it. Participation is a `DISTINCT` over `messages.channel_identity_id` — the
record of who *spoke* — and it is never re-checked against the platform. So:

1. Somebody speaks in a Mattermost channel the bot is in, and links their account.
2. They are removed from that channel in Mattermost.
3. They keep reading the thread in AgenticOS, including everything said in it
   after they left.

That is a second access path to channel content, and it outlives the access the
platform grants. The alternative was rejected on cost: checking `channel_members()`
per listing is an API call per channel (cacheable in Redis), and it answers
`ChannelDirectoryUnsupported` on a platform that will not tell a bot.

**So the rule for anyone reading this later: `messages.channel_identity_id` says
who spoke. It does not say who may read.** Nothing may treat it as an
authorization check without adding the platform lookup first.

What fixing it looks like: `adapter.channel_members()` behind a short Redis cache,
consulted when the listing runs and when the thread is opened, with a decision
about what to do where the platform cannot answer (refusing is the safe default).

Sits with #639 (which admitted unlinked senders in a room) and #131 (channel-made
things invisible to the linked account, which the participant model answers).

PR #744, opened 2026-08-13
fix(channels): refuse writes to an unowned room thread from non-participants

### LINKED ISSUE #701: A room thread nobody linked an account in is writable by any member
Split out of the review of #634 (fixed there: a room's *participant* could delete the thread — `_may_write` now refuses them).

## What this is

`ConversationService._may_write` (`backend/app/services/conversation.py`) allows the write when a conversation has no owner recorded:

```python
owner = getattr(conversation, "user_id", None)
if owner is None:
    return True
```

A channel thread's `user_id` is the first *linked* speaker, so a room where nobody has linked an account has no owner — and any member of the organization can rename it, archive it, delete it, or append a `role: "assistant"` turn to it.

## Pre-existing, not caused by #634

This is exactly what `main` did before participation existed: the old check skipped the owner guard entirely when `user_id` was `NULL`. #634 narrowed the *read* to participants and deliberately left the write rule unchanged rather than inventing a new refusal in a security fix — the comment in `_may_write` points here.

## Why it needs a decision rather than a patch

Refusing everybody is not obviously right: somebody has to be able to tidy up a room thread, and there is no owner to be. The options are roughly

- a permission (`conversations:*` does not exist today — authority over conversations is ownership plus shares);
- the binding's creator, the way `publisher_context` already answers "who does this run as";
- participants only, which makes the read and the write agree again and accepts that a room with no linked speaker is nobody's to delete.

The third is the smallest and matches how the list already behaves. It is still a product call about who cleans up.

An integration test in `backend/tests/integration/test_channel_thread_participants.py`: a member of the organization who never spoke in an unowned room thread is refused the delete, and whoever the decision names is allowed it.

Found reviewing #634.

DEENUU1:
`ConversationService._may_write` answered `True` whenever `conversation.user_id` was `None`. A channel thread's owner is its first *linked* speaker, so a room where nobody linked an account had no owner — and any member of the organization could rename it, archive it, delete the transcript, or append a `role: "assistant"` turn the model is handed back as its own words on the next turn, including for threads their own list never showed them. Pre-existing behaviour; #634 narrowed the read to participants and deliberately left the write alone.

Of the options in #701 this takes **participants only**: an unowned thread is writable by exactly the set `_may_read` admits — a share, or somebody whose linked chat account spoke in it. It is the smallest change, it makes the read and the write agree again the way #634 made the list and the read agree, and it accepts that a room with no linked speaker is nobody's to delete. The two cases the rule must not touch are untouched:

- an **owned** thread still refuses its participants the write (#634's fix) — participation carries the write only where there is no owner it would be taken from;
- **channel ingestion** is unaffected, because the channel router writes turns through the repository, not through `ConversationService.add_message`.

`docs/channels.md` gains a paragraph on the ownerless case next to the existing "opens a thread rather than owning one" rule.

## Verified

- Unit (`tests/test_services_conversation.py`): a stranger's delete of an unowned thread is refused and the repo delete is never called; a participant's delete is allowed; the owned dashboard/share/owner cases unchanged.
- Integration (`tests/integration/test_channel_thread_participants.py`, real rows): a member who never spoke in an unowned room thread is refused the delete (row survives) and the assistant-turn append (transcript unchanged); a participant deletes the thread they were in; a normal owned conversation behaves exactly as before.
- `make lint-backend`, `make test` (4801 passed, platform layer at 100%), `make docs-build --strict` — all green.

Closes #701

### DIFF
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -805,11 +805,13 @@ it is about to wait.
     somebody, with no backfill: the turn points at the chat account and the
     account gains a person.
 
-    **That list says who spoke, not who may read.** Participation is never
-    re-checked against the platform, so somebody removed from the channel there
-    keeps the thread in their list here — deliberately, and [#641][641] is what it
-    would take to change. Nothing may use it as an authorization check without
-    asking the platform first.
+    **That list says who spoke, not who is still in the channel.** Participation
+    is never re-checked against the platform, so somebody removed from the
+    channel there keeps the thread in their list here — deliberately, and
+    [#641][641] is what it would take to change. Where participation carries an
+    access decision — opening a thread, changing an ownerless one — it carries
+    that staleness with it; anything needing the live channel roster has to ask
+    the platform, never this list.
 
     **And it opens a thread rather than owning one.** Speaking in a room admits
     you to reading it; renaming it, archiving it, deleting it or appending a turn
@@ -822,7 +824,8 @@ it is about to wait.
     the participants *are* who may change it — the same set that may open it.
     There is nobody the write would be taken from, and the alternative was the
     whole organization: any member could delete a transcript whose list entry
-    they had never seen ([#701][701]).
+    they had never seen ([#701][701]). The [#641][641] caveat carries over to
+    the write: this is who spoke, not who is still in the channel.
 
 [701]: https://github.com/vstorm-co/agenticos/issues/701
 
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -818,6 +818,14 @@ it is about to wait.
     whole transcript, or write a turn as the agent that everybody reads and the
     model is handed back as its own words on the next turn.
 
+    A thread whose first speaker never linked an account has no owner, and there
+    the participants *are* who may change it — the same set that may open it.
+    There is nobody the write would be taken from, and the alternative was the
+    whole organization: any member could delete a transcript whose list entry
+    they had never seen ([#701][701]).
+
+[701]: https://github.com/vstorm-co/agenticos/issues/701
+
 [641]: https://github.com/vstorm-co/agenticos/issues/641
 
     The refusal carries the way out. Message the bot and it answers with a URL;
--- a/backend/app/services/conversation.py
+++ b/backend/app/services/conversation.py
@@ -491,8 +491,9 @@ class ConversationService:
         `role: "assistant"` - to any conversation in the deployment, and it
         would render to its owner as the agent's own words. See `UNSCOPED`.
 
-        `user_id` narrows that to the owner or somebody the conversation was
-        shared with - `_may_write`, not `_may_read`, because a `role: "assistant"`
+        `user_id` narrows that to the owner, somebody the conversation was
+        shared with or, on a thread with no owner, somebody who spoke in it -
+        `_may_write`, not `_may_read`, because a `role: "assistant"`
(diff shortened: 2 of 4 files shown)


docs/channels.md
```diff
@@ -818,6 +818,14 @@ it is about to wait.
     whole transcript, or write a turn as the agent that everybody reads and the
     model is handed back as its own words on the next turn.
 
+    A thread whose first speaker never linked an account has no owner, and there
+    the participants *are* who may change it — the same set that may open it.
+    There is nobody the write would be taken from, and the alternative was the
+    whole organization: any member could delete a transcript whose list entry
+    they had never seen ([#701][701]).
```
DEENUU1: This paragraph now sits two paragraphs below "Nothing may use it as an authorization check without asking the platform first" — and it makes participation carry the write. The read already crossed that line in #634, but the write is the sharper case, so the page should own the trade instead of leaving the two paragraphs to disagree: somebody removed from the channel who said one thing still gets to delete the ownerless transcript, because participation is who *spoke*, never a live membership check. One sentence covers it:

```suggestion
    A thread whose first speaker never linked an account has no owner, and there
    the participants *are* who may change it — the same set that may open it.
    There is nobody the write would be taken from, and the alternative was the
    whole organization: any member could delete a transcript whose list entry
    they had never seen ([#701][701]). The [#641][641] caveat carries over to
    the write: this is who spoke, not who is still in the channel.
```

Worth also tightening the older "Nothing may use it as an authorization check" sentence while in here (it's just above this hunk, so I can't anchor on it) — as written it's contradicted by both this paragraph and the read rule one bullet up.

DEENUU1 (commented):
Self-review, since ai-review is still down (#311). The fix itself is right and I like the shape it took: `_may_write` now runs the same three checks in the same order as `_may_read` (owner, share, spoke_in), so the diff between the two methods is exactly the policy difference and nothing else. The integration tests assert consequences — the row survived, the transcript still has one message — not that a mock was called, and the owned case is pinned at both layers so #634's rule can't regress quietly.

Two small things, neither blocking:

1. The new docs paragraph should say out loud that the #641 caveat now carries the write — inline comment on `docs/channels.md`.
2. `backend/app/services/conversation.py:494` — the `add_message` docstring still says `user_id` narrows the write to "the owner or somebody the conversation was shared with". That's no longer the whole list: a participant of an ownerless thread is admitted too. It defers to `_may_write` in the same sentence so nobody will be badly misled, but since the reference docs are generated from these docstrings it's worth the one clause: "...or, on a thread with no owner, somebody who spoke in it."

I checked the blast radius: the channel router writes through the repository so ingestion is untouched; shares on an ownerless thread can't exist (`create_share` requires the owner), so in practice the write set is participants exactly; and the sidebar only shows an ownerless thread to its participants, so the UI controls and the backend refusal agree. Also verified the column-0 `[701]:` reference def inside the admonition renders fine — Python-Markdown strips reference definitions before block parsing, same as the existing `[641]:`.

docs/channels.md
DEENUU1: Both taken in 27d62059 — the suggestion as written, and the older sentence now says where participation does carry an access decision (opening a thread, changing an ownerless one) and that the staleness travels with it, instead of claiming nothing uses it. The add_message docstring gained the ownerless-participant clause in the same commit.

Issue #681, opened 2026-08-13
The chat tool_result frame relays a search provider's exception text to the client


DEENUU1:
**Severity: medium** · pre-existing, found reviewing the fix for
[#659](https://github.com/vstorm-co/agenticos/issues/659) (PR
[#675](https://github.com/vstorm-co/agenticos/pull/675)) · not caused by it

## What's wrong

#659 closed the `error` frame. The **`tool_result` frame** is the same leak one
frame over, and it is on the same socket:

`backend/app/services/agent_session.py:653-658`

```python
content = str(tool_event.part.content)   # a RetryPromptPart lands here too
await send_event(
    self.websocket, "tool_result", {"tool_call_id": ..., "content": content}
)
```

A `RetryPromptPart` carries whatever a tool raised as `ModelRetry`, and the web
research toolset raises one built straight out of a vendor exception:

- `app/agents/capabilities/web_research/_toolset.py:49` — `raise ModelRetry(str(exc)) from exc`
- `app/agents/capabilities/web_research/_search.py:143,164,184,208` —
  `SearchUnavailable(f"Tavily search failed: {exc}")`, and the same for Brave,
  Exa and DuckDuckGo, where `exc` is a raw `httpx` or SDK exception carrying the
  failing request URL and host.

So a search provider that 401s puts its own message — endpoint, host, and
whatever the query string held — into the chat panel and the browser console of
whoever is watching the run.

`ModelRetry(str(exc))` reaching the *model* is deliberate and correct
(`_toolset.py:45-48` explains why: an error shaped like a result makes the model
answer from memory). What is not deliberate is relaying that same string
verbatim to the client frame.

## The sequence

1. Give an agent the web research capability with a search key.
2. Break the key.
3. Ask it something that searches. The `tool_result` frame in the chat panel
   carries `Tavily search failed: Client error '401 Unauthorized' for url
   'https://api.tavily.com/search'` — the vendor's own text, unfiltered.

## How you would know it was fixed

The frame carries what the tool did and, for a failure, a controlled sentence —
the model still gets the retry text it needs to decide what to do next, and the
client does not. The two audiences are different and the code currently treats
them as one.

`_turn_failed` in `app/services/agent_session.py` (#659) and
`app/services/rag/failures.py` (#423) are the shape; the decision to make here is
whether the *retry* text is sanitized at the raise (in the capability, so the
model reads a controlled sentence too) or at the frame (so the model keeps the
[...]

Issue #659, opened 2026-08-13
Dashboard error frame sends str(e) of a provider exception to the client


DEENUU1:
## What breaks

`AgentSession._on_error` (`backend/app/services/agent_session.py:344`) sends
`str(e)` of an arbitrary exception to the dashboard client as the `error`
frame's message. `exceptions-security.md` is explicit that `str(exc)` from a
vendor SDK is not a controlled string — provider clients put the failing
request URL, and its query string, in the message — so a provider error can
put a key fragment or an internal URL in front of a signed-in member.

The embed surface got this right (commit history on `feat/agent-surfaces`):
it answers "Something went wrong" / "This assistant is unavailable" and logs
the detail. The dashboard did not, and `exceptions-security.md` does not carve
out member-facing surfaces.

## Sequence

1. A dashboard chat run hits a provider error whose message carries a URL with
   a token in the query string (OpenAI-compatible clients do this).
2. `_on_error` emits `{"type": "error", "data": {"message": str(e)}}`.
3. The member sees the raw provider message in the chat panel; it is also in
   any client-side log.

## Where

`backend/app/services/agent_session.py:344`

## How I would know it is fixed

The `error` frame carries a controlled sentence; the provider's own text
stays in the `logger.exception` line beside the raise, as it does on the embed
surface.

Pre-existing, found reviewing #634 (not caused by it). Related: the embed
surface's own error handling in `app/services/embed_session.py`.

DEENUU1:
Cross-linking from the #168 map rebuild — **the remedy this issue relies on does not currently work.**

Every fix in this cluster does the same thing: stop putting the vendor's own text in the response or the row, and keep it in the `logger.exception` beside the raise. That is the rule `.claude/rules/exceptions-security.md` sets, and it is only a fix if the log is scrubbed.

**#440 — reopened today — says it is not.** `PiiRedactionFilter` is attached to the root *logger* (`app/core/logging.py:73`), where it never sees a record from a module logger, and `setup_logging` has one caller (`app/main.py:217`), so the Prefect worker never installs it at all. Verified on `origin/main`. It was closed `COMPLETED` on 2026-08-07 unfixed; `app/services/rag/failures.py:19` has been documenting the failure in the present tense, citing #440, ever since.

[...]

Issue #440, opened 2026-08-06
PiiRedactionFilter redacts nothing: it is on the root logger, and the worker never installs it


DEENUU1:
**Severity: high** · pre-existing, found while fixing
[#423](https://github.com/vstorm-co/agenticos/issues/423) · **not fixed there** — see below

## What's wrong

`PiiRedactionFilter` scrubs nothing that the application actually logs.

`setup_logging()` (`backend/app/core/logging.py:69`) attaches the filter to the
**root logger**:

```python
root_logger = logging.getLogger()
if not any(isinstance(f, PiiRedactionFilter) for f in root_logger.filters):
    root_logger.addFilter(PiiRedactionFilter())
```

A filter on a logger runs in `Logger.handle`, which is reached only for records
logged *on that logger*. A record from `logging.getLogger(__name__)` propagates
to the ancestors' **handlers** (`Logger.callHandlers`) and never touches the
ancestors' filters. Every log line in this codebase comes from a module logger,
so every one of them bypasses it. Demonstrated on this checkout:

```
$ python -c "...; setup_logging(); \
    logging.getLogger('app.services.rag.ingestion').error('key sk-abc… for a@b.com'); \
    logging.getLogger().error('root: key sk-abc… for a@b.com')"
key sk-abcdefghijklmnopqrstuvwxyz123 for user a@b.com          # module logger: untouched
root: key [API_KEY_REDACTED] for user [EMAIL_REDACTED]          # root logger: redacted
```

**And the worker never calls it at all.** `setup_logging()` runs once, in
`app/main.py:193`, inside the FastAPI app factory. `app/worker/prefect_app.py`
does not call it, so even records that *would* be filtered are not filtered in
the process that does the ingestion, the syncs and the reports.

So the redaction a deployment believes is standing between its logs and Datadog
/ CloudWatch / Logfire — emails, JWTs, `sk-` keys, bearer tokens — is not there.
It has never been there.

## The sequence

1. Point the deployment's log shipper anywhere.
2. Fail anything that logs an exception with a credential or an address in it —
   a wrong embedding key, an SMTP failure, a connector 401.
3. The line arrives at the aggregator verbatim.

## How you would know it was fixed

A record logged through a module logger, in **each** of the three processes that
run this code — the API under uvicorn, a Prefect flow subprocess, and the CLI —
comes out redacted. That is three questions about when handlers exist relative
to `setup_logging()`, not one, which is why the fix is a change of its own:
[...]

PR #745, opened 2026-08-13
fix(chat): stop storing a tool's raw retry text on the transcript row

### LINKED ISSUE #695: The run transcript stores a search provider's exception text on a tool-call row
**Severity: medium** · pre-existing, found fixing
[#681](https://github.com/vstorm-co/agenticos/issues/681) (PR
[#694](https://github.com/vstorm-co/agenticos/pull/694)) · not caused by it

## What's wrong

#681 closed the chat `tool_result` **frame**. The transcript **row** is the same
leak on the run paths that do not go through the socket - the HTTP API and the
channel bots - and it is a stored column rather than a frame, so it is read again
weeks later by anyone who opens the run.

`backend/app/services/transcript.py:97`

```python
elif isinstance(part, ToolReturnPart | RetryPromptPart):
    results[part.tool_call_id] = str(part.content)
```

`settled_calls_in` at `transcript.py:131` does the same for the resume shape.
Both feed `TranscriptService.record(..., tool_calls=..., settled=...)` from
`app/services/agent_runner.py:3151,3157`, which writes the string onto the
message's tool-call row - rendered in the conversation and in run history to
every member who can read the run.

A `RetryPromptPart`'s content is written by whichever tool raised, and
`web_search` builds one straight out of the vendor exception it caught:

- `app/agents/capabilities/web_research/_toolset.py:49` — `raise ModelRetry(str(exc)) from exc`
- `app/agents/capabilities/web_research/_search.py:143,164,184,208` —
  `SearchUnavailable(f"Tavily search failed: {exc}")`, and the same for Brave,
  Exa and DuckDuckGo, where `exc` is a raw `httpx` or SDK exception carrying the
  failing request URL and host.

An MCP tool's retry text is a third party's entirely, so the choke point has to
be here rather than at each raise - which is the same conclusion #681 reached for
the frame.

## The sequence

1. Give an agent the web research capability with a search key, and expose it
   over the HTTP API or a channel (not web chat - that path is #681, now fixed).
2. Break the key.
3. Run it on something that searches. The tool-call row stored for that run holds
   `Tavily search failed: Client error '401 Unauthorized' for url
   'https://api.tavily.com/search'`, and it is still there on the runs page a
   month later.

The row says which tool failed and that the model was asked to retry; the
vendor's own text is in the `logger.warning` beside the write and nowhere else.
A `ToolReturnPart` - the success case - is untouched, and the model still reads
the retry whole, because Pydantic AI carries the part into the next request and
nothing on this path touches that.

`_tool_retry` in `app/services/agent_session.py` (#681) is the sentence to reuse
rather than re-invent; `app/services/rag/failures.py` (#423) is the older shape
of the same rule. Worth deciding at the same time whether the two live together -
`_turn_failed` (#659), `_tool_retry` (#681) and this one are three sentences on
one socket's worth of failures.

## Related

Cluster: **Security and tenancy** on the map in
[#168](https://github.com/vstorm-co/agenticos/issues/168), on the
#342 → #423 → #437/#659 → #676/#681/#686 spine - a refusal describes the refusal,
not the server. Nearest neighbour is #681, the same `RetryPromptPart` on the
socket rather than in a row; not a duplicate, that one never touched
`transcript.py`. Distinct from [#676](https://github.com/vstorm-co/agenticos/issues/676),
which is `agent_runs.error` - the run's own failure, not a tool's.

DEENUU1:
## What changed and why

#681 sanitized the chat `tool_result` **frame**; the transcript **row** was the same leak on the run paths that never open a socket — the HTTP API and the channel bots. `tool_calls_in` (`transcript.py`) and `settled_calls_in` both stored `str(part.content)` for a `RetryPromptPart`, and that content is written by whichever tool raised: `web_search` builds one straight from the vendor exception it caught, so a broken key put `Tavily search failed: Client error '401 Unauthorized' for url 'https://api.tavily.com/search'` — an endpoint, a host, and whatever the query string held — on a tool-call row rendered in conversation and run history to every member who can read the run, weeks later. An MCP tool's retry text is a third party's entirely, so the choke point is `transcript.py`, not each raise — the same conclusion #681 reached for the frame.

The stored row now holds the sentence the frame already sends — which tool failed and that the model was asked to retry — and the vendor's own text goes to a `logger.warning` beside the write, nowhere else. Concretely:

- `tool_retry_notice` (the sentence #681 built as `_tool_retry` in `run_stream.py`) **moved to `transcript.py` and became public**, because the reuse can only flow that way round: `run_stream` imports `agent_chat`, which imports `agent_runner`, which imports `transcript` — importing `run_stream` from `transcript` is a cycle. `run_stream` now imports the sentence back; the wire behaviour of #681 is unchanged.
- `_result_text` in `transcript.py` is the one place a part's content becomes a stored string on this path: a `ToolReturnPart` — the success case — is stored whole; a `RetryPromptPart` stores the notice and logs the text. Both `tool_calls_in` (streaming shape) and `settled_calls_in` (resume shape) go through it.
- The model still reads the retry whole: Pydantic AI carries the part into the next request itself, and nothing on this path touches `result.all_messages()` or the paused state — only what is stored changed.
- Mirrors `_tool_retry`'s behaviour exactly, including replacing our own controlled refusals: the frame already does, and one sentence at two choke points beats two rules.

## How verified

- Regression tests on both shapes **fail without the fix**: a `RetryPromptPart` whose content is a vendor-shaped exception string with a URL reaches the row as the notice naming the tool, not the URL, and the vendor text is asserted present in the log (`caplog`) — `test_a_call_the_model_was_told_to_retry_records_a_notice_not_its_text` (streaming shape) and `test_a_refusal_settles_the_call_with_a_notice_not_its_text` (resume shape).
- `test_a_return_holding_a_url_is_stored_whole` pins the boundary: a successful return with a URL in it is untouched.
- `make lint-backend` and `make test` green — 4797 passed, platform layer at 100%. Backend-only change; frontend targets not run. `scripts/docs_drift.py` reports nothing owed — no docs page describes the retry text (nor did #681 add one).

## Deliberately not taken

The issue asks whether `_turn_failed` (#659), this sentence (#681) and the run's own failure summary should live together — three sentences on one socket's worth of failures. They still live in three modules (`agent_session.py`, `transcript.py`, `agent_runner.py`'s `run_failure_summary`); consolidating them did not fall out of this diff naturally, so it is left as a possible follow-up rather than folded in here.

Closes #695

### DIFF
--- a/backend/app/services/transcript.py
+++ b/backend/app/services/transcript.py
@@ -133,6 +133,13 @@ def tool_calls_in(messages: Sequence[ModelMessage]) -> list[RecordedToolCall]:
     A `RetryPromptPart` counts as a result. A call the model was told to retry
     did happen and did fail, and dropping it would leave the transcript showing
     an argument list with no outcome - which reads as "still running" for ever.
+
+    A result whose call is not here is skipped rather than collected and then
+    dropped: it is :func:`settled_calls_in`'s to store, and reading it in both
+    places logged a settled retry's vendor text twice. Gating on `calls` is
+    sound because a result part sits in the `ModelRequest` that answers its
+    call's `ModelResponse`, so the call - when it is in these messages at all -
+    has always been seen first.
     """
     calls: dict[str, RecordedToolCall] = {}
     results: dict[str, str] = {}
@@ -144,7 +151,7 @@ def tool_calls_in(messages: Sequence[ModelMessage]) -> list[RecordedToolCall]:
                     tool_name=part.tool_name,
                     args=part.args_as_dict(raise_if_invalid=False),
                 )
-            elif isinstance(part, ToolReturnPart | RetryPromptPart):
+            elif isinstance(part, ToolReturnPart | RetryPromptPart) and part.tool_call_id in calls:
                 results[part.tool_call_id] = _result_text(part)
     return [
         RecordedToolCall(
--- a/backend/app/services/transcript.py
+++ b/backend/app/services/transcript.py
@@ -60,15 +60,65 @@ class RecordedToolCall:
         tool_name: What was called.
         args: What it was called with. Stored because "the agent sent an email"
             is not reviewable and "the agent sent an email to X" is.
-        result: What it returned, or the retry message when it failed. `None`
-            for a call that never returned - the run was parked on it, stopped,
-            or broke before it completed.
+        result: What it returned, or - when it failed - the notice that the
+            model was asked to retry, never the retry text itself
+            (:func:`_result_text`). `None` for a call that never returned - the
+            run was parked on it, stopped, or broke before it completed.
     """
 
     tool_call_id: str
     tool_name: str
     args: dict[str, Any]
     result: str | None = None
+
+
+def tool_retry_notice(part: RetryPromptPart) -> str:
+    """What a tool that asked the model to try again may tell a reader.
+
+    A retry's content is written by whichever tool raised: `web_search` turns a
+    search provider's failure into a `ModelRetry` built out of the `httpx` or
+    SDK exception it caught, so a broken key put "401 Unauthorized for url
+    'https://api.tavily.com/search'" - an endpoint, a host, and whatever the
+    query string held - in front of everyone watching the run (#681), and on
+    the tool-call row every member who can read the run opens weeks later
+    (#695). An MCP tool's retry is a third party's string entirely, which is
+    why this is one sentence at the two choke points rather than a rule at each
+    raise: `run_stream` sends it in the `tool_result` frame, and
+    :func:`_result_text` below stores it.
+
+    The model still reads the retry whole: Pydantic AI puts the part into the
+    next request itself and nothing here touches that, so the detail that
+    decides whether it retries, switches tack or gives up is unchanged. Only
+    what is shown and stored is trimmed, and the tool's own text stays in the
+    `logger.warning` beside the send or the write.
+
+    The tool's *name* still goes out, because it is what makes the frame worth
+    sending: a card that resolves saying which step failed is the difference
+    from one that spins for ever. `tool_name` is optional on the part - output
+    validation raises a retry that names no tool - hence the two forms.
+    """
+    called = f"The {part.tool_name} call" if part.tool_name else "A tool call"
+    return (
+        f"{called} failed and the model was asked to try again. The server log has the full error."
+    )
+
+
+def _result_text(part: ToolReturnPart | RetryPromptPart) -> str:
+    """What the row stores as the call's outcome.
+
+    A return is the tool's own answer and is stored whole. A retry's content is
+    written by whatever raised - `web_search` builds one out of the vendor
+    exception it caught, failing endpoint and query string included, and an MCP
+    tool's is a third party's string entirely - so the row holds the same
+    sentence the `tool_result` frame sends (#681) and the text itself goes to
+    the log beside the write, where run history read weeks later cannot reach
+    it (#695). The model is untouched either way: Pydantic AI carries the part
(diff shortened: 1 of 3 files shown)


backend/app/services/transcript.py
```diff
@@ -95,7 +145,7 @@ def tool_calls_in(messages: Sequence[ModelMessage]) -> list[RecordedToolCall]:
                     args=part.args_as_dict(raise_if_invalid=False),
                 )
             elif isinstance(part, ToolReturnPart | RetryPromptPart):
-                results[part.tool_call_id] = str(part.content)
+                results[part.tool_call_id] = _result_text(part)
```
DEENUU1: One retry on the resume shape gets logged twice from this module: `tool_calls_in` runs `_result_text` on the orphan part here and then drops the string (`results.get` finds no call to hang it on), and `settled_calls_in` runs it again on the same part at line 180. So every settled retry is two identical WARNING lines — and on web chat three, since `run_stream` logged it live already. That live copy I'd keep: it's the only one that survives a run that dies before `new_messages()` is read.

A return or retry part only ever sits in the `ModelRequest` that answers the call's `ModelResponse`, so by the time this loop scans it, its call is already in `calls` whenever it's in these messages at all. Skipping the orphans logs each part exactly once and removes a dead store at the same time:

```suggestion
            elif isinstance(part, ToolReturnPart | RetryPromptPart) and part.tool_call_id in calls:
                results[part.tool_call_id] = _result_text(part)
```

The trade is that the loop becomes order-dependent where it wasn't, though the message format guarantees that order. If you'd rather keep it order-independent, the duplicate is only log noise — fine to leave, but then worth a clause in `_result_text`'s docstring so the next reader doesn't chase the double line as a bug.

backend/tests/test_transcript.py
```diff
@@ -107,25 +108,48 @@ def test_a_call_and_what_it_returned_are_one_record(self):
...
+                ]
+            )
+
+        assert calls[0].result == (
+            "The web_search call failed and the model was asked to try again. "
+            "The server log has the full error."
+        )
+        assert vendor_text in caplog.text
+
+    def test_a_return_holding_a_url_is_stored_whole(self):
```
DEENUU1, line 142: Good boundary pin. The success case holding a URL is exactly the line #695 says must not move, and without this test the cheap over-fix — sanitizing both part types in `_result_text` — would have passed every other test in the file.

DEENUU1 (commented):
Takes #681's conclusion to the row it missed, and the shape is right: one sentence at the choke point, the vendor's text in the `logger.warning` beside the write, and regression tests on both shapes that fail without the fix.

I checked the blast radius before signing off: `agent_runner.py:3208,3214` is the only caller of both functions; web chat's rows already store the sanitized frame content (`agent.py` persists the collector's `result`, which `run_stream` fills post-#681), so no sibling writer still leaks; `subagents/_events.py` carries only an `ok` flag, never the retry content; and `paused_state` never reaches a schema or route, so "the model reads the retry whole" stays a server-side trade. The unnamed-tool branch of the moved sentence is still covered through `tests/test_agent_session.py:1431`. The docstrings also stay on the right side of the #440 caveat from the issue thread — they say where the text goes, not that the log is scrubbed.

One small thing inline: the resume shape now logs the vendor text twice from this module. Non-blocking, with a one-line fix if you want it.

backend/app/services/transcript.py
DEENUU1: Taken in 498d4bf7 — the elif gates on `calls`, with the order argument in `tool_calls_in`'s docstring and a regression test that reads resume-shape messages through both functions and counts the log records. `run_stream`'s live copy stays, for the run that dies before `new_messages()` is read.

Issue #746, opened 2026-08-13
An attachment the router cannot read is silence to the model, so the agent denies the file


DEENUU1:
## What breaks

Two routings in `backend/app/services/attachments.py` hand the model no text
at all about a file that arrived, so the agent denies receiving anything the
person just sent:

- **A non-image file with no `parsed_content`, on an agent with no workspace**
  (`_without_workspace`, `backend/app/services/attachments.py:238`) returns
  `AttachmentPlan(reference=None, inline=None)`. The model is told nothing, and
  answers "I don't see any file".
- **A routing failure** (`route`, `backend/app/services/attachments.py:208`)
  swallows the exception and returns an empty plan. Deliberate that it does not
  fail the turn — but the model is again told nothing, so the person who
  attached something is answered as if they had not.

`_too_large_to_show` in the same file already states the principle these two
violate: "Said rather than skipped: the person attached a picture and asked
about it, so silence reads as the model ignoring them."

Since #704's fix, the *transcript* names what arrived, which makes the mismatch
sharper: the conversation shows `Attached file: report.zip` as the user's turn,
followed by the agent claiming no file exists.

## The sequence

1. Send a channel bot (agent without a workspace) a file no parser reads —
   a `.zip`, say — with "what is in this?".
2. `_without_workspace` yields an empty plan; the prompt reaching the model is
   the bare question.
3. The agent answers that no file was attached, under a transcript that says
   one was.

## How I would know it is fixed

Both routings yield a reference naming the file and why it cannot be read —
the shape `_too_large_to_show` and `_unstored` already have — so the model can
answer honestly about what it cannot open. A test per routing asserting
`build_prompt` output names the file.

## Related

Pre-existing, found working #704 (the transcript half of the same silence);
belongs to the #660/#661/#690/#704 channel-attachments cluster recorded in
#168. Deliberately left out of #704's fix, which is scoped to the acceptance
criteria there (the transcript and the `ChatFile` link).

PR #747, opened 2026-08-13
fix(chat): name a caption-less turn's files in its user message

### LINKED ISSUE #704: A caption-less image on a workspace-less agent records no user turn
> **Corrected after filing.** The first version of this issue claimed any
> caption-less upload records no user turn. That is wrong, and the wrong version
> would have sent somebody to fix it in `transcript.py`, where the bug is not.
> `execute` replaces the caption with the *assembled* prompt, and
> `AttachmentRouter.build_prompt` prepends an `Attached file: …` reference - so
> a caption-less `.csv` does record a turn and is linked. The real case is
> narrower and is below.

## What breaks

A channel message whose attachment produces **no text at all** runs the agent and
records no user turn, so the conversation jumps to the answer and the `ChatFile`
the turn ran on keeps `message_id` NULL - there is no message for #690's fix to
link it to.

`TranscriptService.record` writes the user turn under `if prompt:`
(`backend/app/services/transcript.py:204`), and `prompt` is
`_prompt_text(user_prompt)` (`backend/app/services/agent_runner.py:1339`), which
joins only the *string* parts of an assembled prompt. Three routings leave that
empty when the caption is empty, all in `backend/app/services/attachments.py`:

- an **inline image within the ceiling**, on an agent with no workspace -
  `AttachmentPlan(reference=None, inline=BinaryContent)` (`attachments.py:233`),
  so `build_prompt` returns `["", BinaryContent]` and `_prompt_text` joins to `""`;
- a file with **no `parsed_content`** and no workspace (`attachments.py:238`);
- a **routing failure**, which returns an empty plan rather than failing the turn
  (`attachments.py:208`).

An agent *with* a workspace always gets a reference, so this is a bot whose agent
has no sandbox - which is the default.

`""` and the `None` a resume passes are treated identically, and they mean
different things: a resume genuinely has no new prompt, while a photo with no
caption is a turn somebody took and the one the run was billed for.

## The sequence

1. Send a Telegram bot a photo with no caption. Its agent has no workspace.
2. The agent answers about the image.
3. The conversation holds the assistant message and no user message.
4. `SELECT message_id FROM chat_files WHERE id = …` → NULL.

## Where

`backend/app/services/transcript.py:204`, `backend/app/services/agent_runner.py:1339`
(`_prompt_text`), `backend/app/services/attachments.py:208`/`:233`/`:238`.

A caption-less image turn leaves a user message in the conversation - naming what
arrived rather than an empty body, since a blank user message reads as somebody
sending nothing - and the `ChatFile` row carries that message's id. A resumed run
still writes no user turn: `None` and `""` must stop being one case.

## Related

Pre-existing, found reviewing my own fix for #690 (PR #703, which links a channel
turn's files to the user message the transcript writes and deliberately leaves
this remainder - there is nothing to link to). Belongs with the
#660 / #661 / #690 channel-attachments cluster; recorded in #168.

DEENUU1:
## What this fixes

Closes #704, and completes the remainder PR #703 deliberately left: a channel
turn whose attachment produced no prompt text now leaves a user message that
names what arrived, and the `ChatFile` row hangs off it.

Since #634/#703, `TranscriptService.record` already wrote the user turn under
`if prompt or attachments:` and linked the files — so the linking half of the
issue holds on `main`, and so does the `None`/`""` split the issue asks for: a
resume passes `None` and no attachments and writes no user turn, while an
empty caption always arrives with the file that makes it a turn. What remained
was the body. `content=prompt or ""` wrote a **blank** user message for a
caption-less upload, and a blank user message reads as somebody sending
nothing: the thread in `/chat` jumped straight to the answer, with the file
card as the only trace of the question.

## The choice, as the issue asked

Between "make the empty-text routings yield a textual reference" and
"distinguish the cases through the record path", this takes the **record
path**: since the `said` refactor (#634) the transcript records what the
person wrote, never the assembled prompt, so a reference added in
`AttachmentRouter.build_prompt` reaches only the model and would not have
touched the transcript at all. `record` is also the one write every
non-streaming surface reaches — the same reason the transcript and the #703
link live there.

`record` now composes the empty turn's body from its attachments —
`Attached image: photo.jpg`, one line per file — reusing the vocabulary
`build_prompt` already writes into the model's briefing, so the two compose
rather than diverge. A caption is never replaced; the naming fills only the
turn that said nothing. Web chat already does the equivalent in its composer
(`trimmed || t("analyzeFiles")`), so channels stop being the one surface with
blank turns.

## How it failed before

1. Send a Telegram bot a photo with no caption; its agent has no workspace.
2. The agent answers about the image.
3. The conversation shows a blank user bubble (the run and link were already
   recorded since #634/#703 — the issue's "no user turn at all" described the
   pre-#634 tree).

## Verified

- All new tests fail without the one-line fix in `transcript.py`.
- `backend/tests/test_surface_transcripts.py` — the issue's exact sequence
  through the real entry point: `ChannelAgentRouter.answer_default` with empty
  text and an inline image on a workspace-less agent (the router yields
  `["", BinaryContent]`), asserting the user message is named and the
  `ChatFile` links to *that* message.
- `backend/tests/test_transcript.py` — the named body for one file, for
  several (image and non-image), a caption surviving verbatim, and the
  existing `test_a_resumed_run_writes_no_user_turn` still pinning the resume
  half.
- `backend/tests/integration/test_transcript_savepoint.py` and
  `test_channel_attachment_rows.py` — both columns read back from a real
  Postgres: the message content and the `chat_files.message_id`.
- `make lint-backend` green. `make test` green: 4799 passed, platform layer at
  100%. The two tests that pinned the old blank body were updated to pin the
  named one.
- `docs/channels.md` § Files: one sentence saying a caption-less turn's
  message names its files.

## Noticed, not fixed

- **#746 (new, filed from this work)** — `attachments.py:208/:238` still hand
  the model nothing for a routing failure or an unparsed file on a
  workspace-less agent, so the agent denies receiving a file the transcript
  now names. Model-side, outside #704's acceptance criteria.

Closes #704

### DIFF
--- a/backend/app/services/transcript.py
+++ b/backend/app/services/transcript.py
@@ -134,6 +134,20 @@ def settled_calls_in(messages: Sequence[ModelMessage]) -> dict[str, str]:
     }
 
 
+def _what_arrived(attachments: Sequence[ChatFile]) -> str:
+    """The user turn's body when files arrived with no words around them.
+
+    A blank user message reads as somebody sending nothing, so a caption-less
+    upload names its files instead - the vocabulary `AttachmentRouter` already
+    uses for the model's briefing (#704). Only the names: the linked rows carry
+    the size and the type.
+    """
+    return "\n".join(
+        f"Attached {'image' if attachment.file_type == 'image' else 'file'}: {attachment.filename}"
+        for attachment in attachments
+    )
+
+
 class TranscriptService:
     """Writes a run's turns into the conversation the run belongs to."""
 
@@ -166,9 +180,13 @@ class TranscriptService:
         (/uploads/…, 43 KB, image)`. The dashboard's own uploads have been rows
         since they existed; a channel's became nothing at all.
 
-        An empty *prompt* is written when a file came with it, because a picture
-        posted with no caption is a turn: the row is what the file hangs off, and
-        without it the attachment belongs to nothing.
+        An empty *prompt* with a file beside it is a turn, because a picture
+        posted with no caption is one: the row is what the file hangs off, and
+        without it the attachment belongs to nothing. Its body names what
+        arrived rather than staying blank - an empty user message reads as
+        somebody sending nothing (#704). `None` is different from `""` here:
+        a resume passes `None` and no attachments, so it writes no user turn,
+        while an empty caption always arrives with the file that makes it one.
 
         An empty `answer` is written when the run called something, and skipped
         when it did not. A blank assistant message with nothing under it reads as
@@ -218,7 +236,7 @@ class TranscriptService:
                         self.db,
                         conversation_id=run.conversation_id,
                         role="user",
-                        content=prompt or "",
+                        content=prompt or _what_arrived(attachments),
                         run_id=run.id,
                         # Off the run rather than passed in: the run row already
                         # records which chat account asked, and a second route to
--- a/backend/tests/integration/test_channel_attachment_rows.py
+++ b/backend/tests/integration/test_channel_attachment_rows.py
@@ -145,6 +145,7 @@ class TestAFileThatArrivedWithATurn:
 
         The row is what the file hangs off, so skipping it because there were no
         words would lose the file - which is the original defect with an extra step.
+        The body names what arrived rather than staying blank (#704).
         """
         conversation = await _conversation(db)
         run = await _run(db, conversation)
@@ -156,5 +157,5 @@ class TestAFileThatArrivedWithATurn:
 
         written = await conversation_repo.get_messages_by_conversation(db, conversation.id)
         asked = written[0]
-        assert (asked.role, asked.content) == ("user", "")
+        assert (asked.role, asked.content) == ("user", "Attached image: report.png")
         assert [file.filename for file in asked.files] == ["report.png"]
--- a/backend/tests/integration/test_transcript_savepoint.py
+++ b/backend/tests/integration/test_transcript_savepoint.py
@@ -181,3 +181,52 @@ async def test_a_channel_turns_file_is_left_carrying_the_message_it_arrived_with
     ).scalar_one()
     linked = (await db.execute(select(ChatFile).where(ChatFile.id == stored.id))).scalar_one()
     assert linked.message_id == question.id
+
+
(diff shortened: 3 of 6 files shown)


backend/app/services/transcript.py
```diff
@@ -218,7 +236,7 @@ async def record(
                         self.db,
                         conversation_id=run.conversation_id,
                         role="user",
-                        content=prompt or "",
+                        content=prompt or _what_arrived(attachments),
```
DEENUU1, line 239: This closes the class for every surface that reaches `record` — channels, the embed widget, the HTTP API, and the unreachable-today `None` + attachments case falls out sensibly too.

One sibling write site is left with the same defect, though: the dashboard streaming socket. `agent_session.process_message` explicitly admits `message: ""` with `file_ids` (`agent_session.py:167`), and `persist_user_turn` stores that verbatim (`agent.py:244`) — the only thing preventing a blank bubble there is the composer's client-side substitution (`chat-input.tsx:153`, `trimmed || t("analyzeFiles")`), so any raw WebSocket client writes exactly the blank turn this PR removes from channels. It can't reuse `_what_arrived` as-is — `persist_user_turn` only has `file_ids` at that point, the `ChatFile` rows load later in `_attached_files` — so it's not a one-liner for this branch. I'd file it into the #704/#690 cluster as a follow-up, pre-existing and found reviewing this.

DEENUU1 (commented):
Reviewed the whole path, not just the diff. The record-path choice is the right one: since #634 the transcript records what the person wrote, so a reference added in `build_prompt` would never have reached it — and `record` is the one write every non-streaming surface passes through, which means the embed widget gets this for free too (`embed_session.py` accepts empty text with files and goes through `execute(said=text)`). The prompt × attachments matrix is fully pinned by the tests, the resume half (`approvals.py` passes `prompt=None`, no attachments) stays untouched, and the vocabulary really does match `attachments.py:108/:129`.

One thing this leaves, worth an issue in the #704/#690 cluster rather than this branch: the dashboard streaming socket is now the last write site with the same class of blank turn — see the inline comment on `transcript.py:239`.

PR #748, opened 2026-08-13
fix(chat): refuse linking a file the caller does not own

### LINKED ISSUE #706: A chat turn can link another user's file to its own message
## What breaks

A web-chat turn can attach **another user's file** to its own message by sending
that file's id. Nothing between the socket and the `UPDATE` checks who owns the
row.

`chat_file_repo.link_to_message` is a blind bulk update - no owner predicate, and
no check that the row is currently unlinked:

```python
# backend/app/repositories/chat_file.py:27
await db.execute(sql_update(ChatFile).where(ChatFile.id.in_(ids)).values(message_id=message_id))
```

The ids reach it from the client. `agent_session.py:485` reads `file_ids` off the
socket payload, `ConversationService.list_attached_files` loads them with
`get_many` (`conversation.py:526`) - which filters on id alone - and
`persist_user_turn` links them (`agent.py:244`). `chat_files` carries no
`organization_id`, so `user_id` is the only scope the row has, and none of the
three steps applies it.

Two consequences, both from the same missing predicate:

- **Metadata disclosure.** The victim's filename, MIME type and size are now on a
  message in the attacker's conversation, rendered by `MessageFileRead`
  (`backend/app/schemas/conversation.py:139`). The bytes are still refused -
  `GET /files/{id}` scopes on the caller (`api/routes/v1/files.py:47`) - so this
  is a leak of what a file is called, not of what is in it.
- **The victim's own transcript is rewritten.** The update overwrites a non-NULL
  `message_id`, so a file already attached to somebody else's message silently
  moves off it, and their thread loses the attachment it was about.

The same call is also reachable from a *published agent's* prompt id list, so it
is not only the dashboard.

## The sequence

1. User A uploads `salaries.xlsx`; note its id from `POST /files`.
2. User B, in their own organization, sends a chat turn with
   `file_ids: ["<A's id>"]`.
3. B's conversation now shows a `salaries.xlsx` chip. A's message, if it had one,
   no longer does.

## Where

`backend/app/repositories/chat_file.py:27` (`link_to_message`, no owner filter),
`backend/app/repositories/chat_file.py:18` (`get_many`, same),
`backend/app/services/conversation.py:519`, `backend/app/services/agent.py:244`,
`backend/app/services/agent_session.py:485`.

Linking a file id belonging to another user is refused rather than silently
applied - the update carries the owner in its `WHERE`, and a request naming a
file that is not the caller's gets a refusal rather than a message that quietly
holds someone else's filename. A regression test in `tests/integration/` that
tries it across two users and reads both rows back.

Worth deciding at the same time whether an already-linked file may be re-linked
at all; today it moves, silently.

## Related

Pre-existing, found reviewing PR #703 (#690, which adds a *second* caller of this
function for channel turns - that caller is safe, since its rows are created
server-side in the same turn for the sender, but it is what put the function under
review). Nearest existing thread is the security/tenancy cluster in #168 that
#342 → #423 sits on; not a duplicate of any of those - they are about what a
failure *message* discloses, this is about a missing `WHERE`.

DEENUU1:
## What

A web-chat turn could attach **another user's file** to its own message by sending that file's id. `chat_file_repo.link_to_message` was a blind bulk UPDATE — no owner predicate, no unlinked check — and `get_many` filtered on id alone, with the ids arriving straight off the socket payload. Two consequences from the one missing predicate: the victim's filename/MIME/size rendered in the attacker's conversation, and a non-NULL `message_id` was overwritten, so the file silently moved off the victim's own message.

`chat_files` carries no `organization_id`, so `user_id` is the only scope a row has — and now both the read and the update carry it:

- `chat_file_repo.get_many` and `link_to_message` take the owner in their `WHERE`; the UPDATE also requires `message_id IS NULL`.
- `ConversationService.link_files_to_message` reads the rows first and **refuses** rather than silently narrowing: a foreign or unknown id raises `NotFoundError` (deliberately indistinguishable, so an id cannot be probed for existence), an already-linked one raises `BadRequestError`. Both name only `file_ids` the caller sent — nothing of the victim's.
- The refusal escapes `persist_user_turn`'s swallow (which now covers only infrastructure failures), so the socket answers with an error frame instead of a turn that quietly dropped the attachment.
- The channel transcript (`TranscriptService._attach`, the second caller added by #703/#690) links rows as their own uploader; the embed's owner check moved from Python into the query, and a page whose publisher's account is gone (`owner_user_id` is `SET NULL`) reads no rows at all instead of passing `None` into the predicate.

## The re-link decision

**A file already on a message never moves, not even for its owner.** No caller legitimately re-links: web chat links fresh uploads once per send (no retry resends `file_ids`), channel rows are created server-side unlinked in the same turn, and the embed already documented and dropped a spent id. The silent move only ever rewrote history, so it is refused outright (`BadRequestError`) rather than allowed for the owner.

## How verified

- `backend/tests/integration/test_chat_file_ownership.py` — the issue's sequence against a real database, both rows read back: the attacker gets a refusal, the victim's message keeps its file; the unlinked-file disclosure case; the repo-level UPDATE skipping a foreign row even without the service's pre-read; the owner's own upload still linking; the re-link refusal leaving the row where it was; the scoped read returning nothing for a foreign id.
- Unit tests for the service refusals, the owner riding through `persist_user_turn`, `load_attached_files`, the embed session and the channel transcript, and the ownerless-embed drop.
- The pre-existing channel-turn integration test (`test_transcript_savepoint.py`) proves server-side rows still link through the new predicates.
- `make lint` and `make test` (100% platform gate) pass; `docs/architecture.md` and `docs/file-processing.md` updated in the same change.

Closes #706

### DIFF
--- a/backend/app/repositories/chat_file.py
+++ b/backend/app/repositories/chat_file.py
@@ -15,21 +15,46 @@ async def get_by_id(db: AsyncSession, file_id: UUID) -> ChatFile | None:
     return await db.get(ChatFile, file_id)
 
 
-async def get_many(db: AsyncSession, file_ids: Iterable[UUID]) -> list[ChatFile]:
-    """Batch-load multiple chat files by IDs."""
+async def get_many(db: AsyncSession, file_ids: Iterable[UUID], *, user_id: UUID) -> list[ChatFile]:
+    """Batch-load `user_id`'s chat files by IDs.
+
+    The ids come off client payloads, and `chat_files` carries no organization,
+    so `user_id` is the only scope a row has: another user's id resolves to
+    nothing here rather than to their row (#706).
+    """
     ids = list(file_ids)
     if not ids:
         return []
-    result = await db.execute(select(ChatFile).where(ChatFile.id.in_(ids)))
+    result = await db.execute(
+        select(ChatFile).where(ChatFile.id.in_(ids), ChatFile.user_id == user_id)
+    )
     return list(result.scalars().all())
 
 
-async def link_to_message(db: AsyncSession, *, message_id: UUID, file_ids: Iterable[UUID]) -> None:
-    """Link multiple chat files to a message by setting message_id on each."""
+async def link_to_message(
+    db: AsyncSession, *, message_id: UUID, file_ids: Iterable[UUID], user_id: UUID
+) -> None:
+    """Link `user_id`'s unlinked chat files to a message.
+
+    The WHERE carries both halves of the rule rather than trusting the caller:
+    without them this was a blind bulk update, so a turn naming another user's
+    file id put their filename on its own message and silently pulled the file
+    off the message it already hung on (#706). An id the predicates exclude is
+    skipped here; a caller that must refuse it reads the rows first
+    (`ConversationService.link_files_to_message`).
+    """
     ids = list(file_ids)
     if not ids:
         return
-    await db.execute(sql_update(ChatFile).where(ChatFile.id.in_(ids)).values(message_id=message_id))
+    await db.execute(
+        sql_update(ChatFile)
+        .where(
+            ChatFile.id.in_(ids),
+            ChatFile.user_id == user_id,
+            ChatFile.message_id.is_(None),
+        )
+        .values(message_id=message_id)
+    )
     await db.flush()
 
 
--- a/backend/app/repositories/chat_file.py
+++ b/backend/app/repositories/chat_file.py
@@ -33,20 +33,21 @@ async def get_many(db: AsyncSession, file_ids: Iterable[UUID], *, user_id: UUID)
 
 async def link_to_message(
     db: AsyncSession, *, message_id: UUID, file_ids: Iterable[UUID], user_id: UUID
-) -> None:
-    """Link `user_id`'s unlinked chat files to a message.
+) -> int:
+    """Link `user_id`'s unlinked chat files to a message, answering how many moved.
 
     The WHERE carries both halves of the rule rather than trusting the caller:
     without them this was a blind bulk update, so a turn naming another user's
     file id put their filename on its own message and silently pulled the file
     off the message it already hung on (#706). An id the predicates exclude is
     skipped here; a caller that must refuse it reads the rows first
-    (`ConversationService.link_files_to_message`).
+    (`ConversationService.link_files_to_message`) - and compares the count,
+    because a concurrent turn can take a row between that read and this UPDATE.
     """
     ids = list(file_ids)
     if not ids:
-        return
-    await db.execute(
+        return 0
+    result = await db.execute(
         sql_update(ChatFile)
         .where(
             ChatFile.id.in_(ids),
@@ -56,6 +57,9 @@ async def link_to_message(
         .values(message_id=message_id)
     )
     await db.flush()
+    # `execute` is typed to return `Result`, which has no `rowcount`; a DML
+    # statement actually returns `CursorResult`, which does (see resource_grant).
+    return result.rowcount  # ty: ignore[unresolved-attribute]
 
 
 async def delete(db: AsyncSession, *, db_file: ChatFile) -> None:
--- a/backend/app/services/conversation.py
+++ b/backend/app/services/conversation.py
@@ -601,12 +601,39 @@ class ConversationService:
             success=data.success,
         )
 
-    async def link_files_to_message(self, message_id: UUID, file_ids: list[str]) -> None:
+    async def link_files_to_message(
+        self, message_id: UUID, file_ids: list[str], *, user_id: UUID
+    ) -> None:
+        """Attach `user_id`'s own unlinked files to a message, refusing anything else.
+
+        The ids come off a socket payload, so each must resolve to the caller's
+        own *unlinked* row: a foreign id would render another user's filename in
+        this conversation and silently pull the file off the message it already
+        hangs on (#706). Refused, never narrowed - a turn that quietly dropped an
+        attachment would read as an agent ignoring the file it was asked about.
+        A file that is not the caller's answers exactly like one that does not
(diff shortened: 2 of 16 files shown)


backend/app/services/conversation.py
```diff
@@ -601,12 +601,39 @@ async def complete_tool_call(
...
+
+        The ids come off a socket payload, so each must resolve to the caller's
+        own *unlinked* row: a foreign id would render another user's filename in
+        this conversation and silently pull the file off the message it already
+        hangs on (#706). Refused, never narrowed - a turn that quietly dropped an
+        attachment would read as an agent ignoring the file it was asked about.
+        A file that is not the caller's answers exactly like one that does not
+        exist, so an id cannot be probed for whether it is taken.
+        """
+        ids = [UUID(fid) for fid in file_ids]
```
DEENUU1: One id shape still slips through the "refused, never narrowed" contract: a string that isn't a UUID at all. `UUID("garbage")` raises `ValueError` here, which `persist_user_turn`'s inner `except Exception` swallows as a transient persistence failure — the turn goes ahead, and then `list_attached_files` parses the same string again inside the run and dies as a generic "turn failed" frame, after the message was persisted and the `user_prompt` frame already went out. The ids come off the socket payload, so any client can send this.

The parse is pre-existing, but this PR is what drew the refusal/infrastructure line, and a `ValueError` lands on the wrong side of it. A malformed conversation id already gets a clean `BadRequestError` via `_conversation_uuid`; the file ids deserve the same:

```suggestion
        try:
            ids = [UUID(fid) for fid in file_ids]
        except ValueError:
            raise BadRequestError(
                message="Invalid file id", details={"file_ids": file_ids}
            ) from None
```

That makes it an `AppException`, so it rides the same early-abort path as the ownership refusal and the client gets one clear error frame instead of a half-persisted turn.

backend/app/repositories/chat_file.py
```diff
@@ -15,21 +15,46 @@ async def get_by_id(db: AsyncSession, file_id: UUID) -> ChatFile | None:
...
         return
-    await db.execute(sql_update(ChatFile).where(ChatFile.id.in_(ids)).values(message_id=message_id))
+    await db.execute(
+        sql_update(ChatFile)
+        .where(
+            ChatFile.id.in_(ids),
+            ChatFile.user_id == user_id,
+            ChatFile.message_id.is_(None),
+        )
+        .values(message_id=message_id)
```
DEENUU1, line 57: The pre-read in the service and this UPDATE run in different transactions when two turns race — two tabs, same user, both sending the same fresh upload id at once. Both pre-reads see the row unlinked, the first UPDATE wins, and under READ COMMITTED the second blocks on the row lock, re-evaluates `message_id IS NULL`, matches zero rows, and no-ops. That second turn then answers as if the attachment were fine — exactly the silent narrowing the service was rewritten to refuse, just moved into the race window.

The fix is cheap because the rowcount already tells the truth after the lock wait:

```python
result = await db.execute(
    sql_update(ChatFile)
    .where(...)
    .values(message_id=message_id)
)
await db.flush()
return result.rowcount
```

and `link_files_to_message` raises the same `BadRequestError` when it comes back short of `len(ids)`. Non-blocking — it needs two sockets naming the same fresh id in flight simultaneously, and the model still sees the file either way (the later `get_many` doesn't filter on `message_id`), so the visible damage is a missing chip. But it turns a race that lies into a race that refuses, which is the invariant the rest of the PR is built on.

backend/tests/integration/test_chat_file_ownership.py
```diff
@@ -0,0 +1,200 @@
...
+        mine = await _message(db, attacker)
+
+        with pytest.raises(NotFoundError):
+            await ConversationService(db).link_files_to_message(
+                mine.id, [str(fresh.id)], user_id=attacker.id
+            )
+
+        assert (await _reloaded(db, fresh.id)).message_id is None
+
+    async def test_the_update_itself_skips_a_foreign_row(self, db) -> None:
```
DEENUU1: This test is the one I'd fight to keep: it proves the predicate holds *without* the service's pre-read, which is precisely the guarantee the channel transcript relies on. If someone later "simplifies" the repo by trusting callers again, this fails and names why. Same for the class above reading both rows back — asserting the victim's `message_id` survived the attempt is what a mock-level test structurally can't do.

DEENUU1 (commented):
Self-review before merge, since the automated reviewer is still off (#311).

The shape is right: the owner predicate lives in the UPDATE itself, so the one caller that skips the service pre-read (the channel transcript) is still safe, and the integration tests read the rows back instead of trusting mocks. I traced every caller of `get_many` / `link_to_message` / `list_attached_files` / `load_attached_files` — all of them are in this diff, and `link_to_message` is the only writer of `ChatFile.message_id` anywhere, so the sibling audit is clean. The rollback path also holds: a refusal escapes `persist_user_turn`, the session context manager rolls the already-written user message back, and the socket handler turns it into an error frame.

Two non-blocking suggestions inline: a malformed (non-UUID) file id still falls on the wrong side of the new refusal/infrastructure line, and the UPDATE's rowcount is worth checking so a cross-session race can't silently narrow.

backend/app/services/conversation.py
DEENUU1: Fixed in 84784850 — malformed ids are parsed through str() and refused as a BadRequestError naming only the bad values, before any read.

backend/app/repositories/chat_file.py
DEENUU1: Fixed in 84784850 — link_to_message answers its rowcount and the service refuses a short count, so the race now fails loud. Not reproduced as a two-session test; covered at the service boundary with a mocked repository, and the integration suite asserts the count both ways (1 own, 0 foreign).

Issue #563, opened 2026-08-10
rating-buttons hand-rolls two fetch calls instead of a lib client and a hook


DEENUU1:
`components/chat/rating-buttons.tsx:68,109` — two raw `fetch` calls (POST/DELETE `/api/conversations/{id}/messages/{id}/rate`) with hand-rolled headers/`response.ok`/error parsing, reconciling counts manually instead of via query invalidation.

**Why it matters.** `frontend.md`: all API access goes through `src/lib` clients via a hook — no `fetch` in a component. The two error-parse blocks are duplicated and drift independently; the endpoints are invisible to `query-keys.ts`.

**Fix.** A `use-message-rating` hook with `rateMessage`/`removeRating` mutations over the lib client; delete the duplicated fetch blocks.

**Done when.** No `fetch` in the component; ratings invalidate through the query layer.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

Issue #502, opened 2026-08-08
An answered ask_user question leaves no trace in the transcript


DEENUU1:
An agent asks a question mid-turn, the person answers it, and the transcript
keeps no record that either happened. Scroll back and the turn reads as though
the agent simply knew.

## Where it goes

`ask_user` is a **WebSocket event**, not a tool call. `AgentSession._ask_user`
sends `{"type": "ask_user", "questions": [...]}`
(`app/services/agent_session.py:421`) and blocks on a future until the frame
dispatcher receives `ask_user_response`. On the frontend, `use-chat.ts:359`
puts the batch in `pendingQuestions`, `chat-container.tsx:365` renders the form
above the composer, and answering clears it (`use-chat.ts:728`). Nothing is
appended to the message: no part, no tool call, no text.

So the questions and the answers exist only for as long as the form is on
screen. A conversation reopened from history has neither.

## Why it surfaced now

`tool-results/ask-user.tsx` was a transcript view for exactly this - it read
`questions` out of a tool call's arguments and rendered the Q/A pair. It was
reached from `tool-call-card.tsx` on `toolCall.name === "ask_user"`, and **no
capability has ever registered a tool by that name**, in this repository or
before it. The component had never rendered. It was removed in #144 along with
the other names the frontend matched and the backend does not emit; the
drift test added there would refuse to let it back.

Removing it changed nothing a person can see. It did make the gap visible: the
code that looked like the answer to this was answering nothing.

## What it would take

Either end works, and they are different products:

- **Backend** - emit the question and its answer as parts of the assistant
  message, so history carries them the way it carries a tool call. Costs a
  message-part kind and a migration-free schema addition; gains a transcript
  that is true for anyone who reads it later, including an export.
- **Frontend only** - keep the answered batch in the chat store and render it
  as a part. Cheaper, and wrong the moment the conversation is reloaded, since
  the store is rebuilt from what the backend stored.

The backend one is the honest fix. A subagent's `ask_parent` reaches the person
through this same channel (`subagents/_toolset.py:312`), so whatever shape is
chosen has to name who was asking.

## Done when

- [ ] An answered question and its answer are visible in a conversation
      reopened from history
- [ ] A question a delegate asked says which delegate asked it
- [ ] A test that fails when an answered question leaves no part behind

---

Pre-existing; found while doing #144, which removed the unreachable renderer
rather than fixing this. Related: #145 (the other half of the tool-rendering
[...]

PR #759, opened 2026-08-13
Frontend standardisation: one table, one shell, admin on both, skills without install, map in four directions

### LINKED ISSUE #139: One table primitive, with sortable headers and filters, everywhere

### LINKED ISSUE #281: Skills are always available: drop the 'Ready-made skills' install step, one Skills list
## What

`/skills` splits into "Ready-made skills" (predefined, with an **Install** step) and the user's own skills. Drop that split and the installation model entirely: every skill — predefined and user-authored — appears in one **Skills** list, already available to bind. There is nothing to "install".

## Why

Installation is ceremony with no payoff. A skill is data the platform already ships or the user already wrote; gating a predefined skill behind an Install click just adds a step before it can be used, and forces two visual sections where one would do. Nothing downstream needs an "installed" flag that a bind-time reference could not answer.

## Scope

- Remove the "Ready-made skills" section and the Install action from `/skills`.
- One list showing predefined + user skills together, each directly bindable (the Builder's skill picker already resolves both — this is the library page catching up).
- Decide the fate of any `installed`/`is_installed` state on the model/API: if nothing reads it once install is gone, remove it (a migration) rather than leave a dead column; if a predefined skill needs a per-org row to be bindable, make that implicit rather than a user action. Say which in the PR.
- A predefined vs. authored skill can still be *labelled* (a badge), but both are present and usable from the same list.

## Done when

`/skills` shows one list, no Install step, and a predefined skill is bindable in the Builder without any prior action. Tests cover that a predefined skill resolves for an agent with no install row.

## Where it sits

Concrete product slice of the broad **#69 (Skills redesign)** bag. Related: **#73** (Seed Skills — where predefined skills come from), **#154** (skills e2e). Part of the W2 frontend-standardisation push alongside the sibling issues filed with it.

### LINKED ISSUE #282: A consistent list-page shell (container, pagination, filter bar) across resource pages
## What

`/skills`, `/mcp-servers`, `/sandboxes`, `/workspaces` (and the other resource-list pages) each hand-roll their own page shell — the outer container, the pagination control, and where the filter/search bar sits. Standardise the **shell** so every resource-list page has the same container, the same pagination, and the same filter-bar placement.

## Scope (the shell, not the table cells)

This is the layer *around* the table, complementary to **#139** (one table primitive with sortable headers and filters — the cells and sorting). Here:

- one list-page container (heading, description, primary action, the filter/search bar in one consistent place);
- one pagination control with one behaviour (page size, prev/next, "showing X–Y of N") — today some pages paginate, some render `list.length`, some do neither;
- one empty state / error state slot, so "nothing here yet" and "the request 502'd" are not the same pixels (see the empty-page-is-ambiguous rule in `.claude/rules/frontend.md`).

Roll it out across `/skills`, `/mcp-servers`, `/sandboxes`, `/workspaces` first; the shell should drop into any future resource page.

## Done when

The four pages use one shared shell component, paginate identically, and place their filter bar identically; a fifth resource page can adopt it without re-inventing the container.

## Where it sits

The shell half of the standardisation whose table half is **#139** (W2). Touches the per-resource redesign bags **#64** (MCP), **#69** (Skills), **#137** (Workspaces). Sibling of the admin-standardisation issue filed with it.

### LINKED ISSUE #283: Remove the Refresh button from /admin
## What

`/admin` has a **Refresh** button. Remove it.

## Why

A manual Refresh is a workaround for data that should refetch on its own — the query layer (TanStack Query) already invalidates and refetches on the mutations that change what the page shows, so a Refresh button is either redundant or a sign a mutation is not invalidating its query. If something on `/admin` genuinely goes stale without it, the fix is the missing invalidation, not a button that asks the user to do the framework's job.

## Scope

- Remove the Refresh control from `/admin`.
- If removing it exposes a place that then shows stale data, add the query invalidation that should have been there (say which in the PR) rather than restoring the button.

## Done when

`/admin` has no Refresh button and its data is current after the actions that change it, through query invalidation.

## Where it sits

A concrete quick-win slice of the broad **#63 (Admin redesign)** bag — `effort:s`, doable now without waiting for the full redesign. Part-of #63.

### LINKED ISSUE #284: Standardise the admin pages (users and tabs) to the shared table and list-page shell
## What

Standardise the admin pages — `/admin` and its tabs (`/admin/users` and the rest) — to the same shell, pagination, buttons, cards and tables as the other resource pages, so users, and everything else the admin area lists, are displayed the same way as skills, MCP servers, sandboxes and workspaces.

## Why

The admin area is its own visual dialect: its own table shape, its own pagination (or none), its own button and card styles, its own way of rendering a list of users. That is a second design to maintain and a second thing for a person to learn. It should adopt the shared primitives, not keep its own.

## Scope

- Adopt the shared **table primitive (#139)** and the shared **list-page shell** (the sibling shell issue: container + pagination + filter bar) for `/admin/users` and every other admin tab that lists rows.
- One button style, one card style, one pagination — the same ones the resource pages use.
- The users list in particular: same table, same pagination, same row/detail affordances as any other resource list.
- This is the concrete, do-able-on-W2 standardisation slice of the full **#63 (Admin redesign)**; the broader look-and-feel rework stays #63.

## Done when

`/admin/*` uses the shared shell and table primitive, paginates and styles buttons/cards identically to the resource pages, and `/admin/users` renders through the same table as any other list.

## Where it sits

Standardisation slice of **#63 (Admin redesign)**. Depends on **#139** (table primitive) and the sibling list-page-shell issue. Also relevant: **#63** stays the home for the non-standardisation redesign work (layout, IA).

### LINKED ISSUE #518: Visual map: four directions around the agent, and a node per subagent
## What is wrong with it now

The map's own subtitle says what it is for: *what reaches this agent, and what it
reaches for*. The layout does not say that. Everything except Channels and
Knowledge is stacked in one column on the right — model, toolbox, MCP, skills,
budget and delegation together — so a reader gets a list, in a diagram's
clothing, and the one relationship the picture exists to show is the one it does
not draw.

Two things are wrong beyond the arrangement:

- **Delegation is one box with names in it.** A delegate is another agent, with
  its own model, its own version pin and its own budget, and the map draws it as
  a line of text. `allow_dynamic` — an agent that may *invent* specialists —
  does not appear at all, and it is the setting with the widest consequences on
  the page.
- **Channels is one box saying "Jarvis bot".** An agent is reachable from the
  dashboard, the API, a widget, a WebSocket and a chat platform, and the map
  names one of the five.

## What it should look like

Four directions around `AGENT`, each answering a different question:

| Where | What | Why there |
|---|---|---|
| **Left** | Every surface it is reachable from: chat, the API, channel bindings, the widget, the raw socket | This is *what reaches the agent* — the subtitle's first half, and it is currently one box |
| **Top** | Model and budget | What the agent *is run as*: which model answers, and under what ceiling |
| **Right** | Toolbox, MCP servers, knowledge, skills | *What it reaches for* — the subtitle's second half |
| **Bottom** | One node per subagent, plus the delegation settings themselves (`allow_dynamic`, `allow_questions`, depth) | A delegate is an agent, so it deserves a node rather than a line — and a tree grows downward |

Each subagent node should be openable the way the current Delegation row already
is (`agent-map-detail.tsx`), because the question after "who does it delegate
to" is always "on which version, and with what budget".

## Where the work is

`frontend/src/components/agents/agent-map-view.ts` builds the node list and the
positions; `agent-map-nodes.tsx` draws them; `agent-map.tsx` is the dialog and
the zoom controls. The view is the interesting file — the rest follows from it.
The data is already there: the spec carries the capabilities, the pins and the
delegation config, and `use-exposures` carries the bindings.

Somebody who has never seen the agent can answer three questions from the
picture without opening a tab: where can people reach it, what does it stand on,
and what does it hand work to. Today the first is under-reported and the third
is a sentence.

## A defect the same screen is showing

The Toolbox lists **`channel_tools (missing)`**. That is not a rendering bug: the
draft genuinely carries a `channel_tools` binding, written during the window in
which the capability was still offered in the Toolbox, and it is now chosen per
bound bot instead — so the catalog no longer names it and publish refuses the
spec that carries it. Being fixed separately with a migration that clears the
binding from stored specs; noted here because this screenshot is where it is
visible.

Related: the surfaces column is the same subject as **#516** and **#517**, which
add two more of them.

### LINKED ISSUE #760: Activity: a date-range picker owned by the page, windowing all three tabs and the export
## What's needed

The Activity page has no date-range control. The backend answers a window on every
surface the page reads — `GET /runs` takes `started_from`/`started_to`
(`backend/app/api/routes/v1/runs.py:58-61`), the export *requires* one, and the spend
route takes `days` — but the UI only ever passes a window when one arrives on the URL
from the dashboard's p95 link (`frontend/src/app/[locale]/(dashboard)/runs/page.tsx:44-45`).
A person standing on the page cannot ask "what ran last Tuesday".

The export menu's own docstring records the gap: *"The window is a preset here rather
than the tab's own range picker, which the Activity page does not have"*
(`frontend/src/components/runs/export-menu.tsx:61`).

## The shape

- One range picker owned by the page (presets — 24h / 7d / 30d / 90d — plus a custom
  from/to), the way LangSmith and Logfire put the time window first among the controls.
- The window narrows all three tabs consistently and is mirrored into
  `?started_from=/?started_to=` so a narrowed view survives a reload and can be linked.
- The export's day presets collapse into "export what is on screen": the picker's
  window becomes the mandatory range the endpoint demands.

## Done when

A person can pick a window on the page, the run list / figures / export all answer over
exactly that window, and the URL round-trips it.

---
Part of the Activity-page (/runs) improvement push, worked in PR #759 — the successor
to #45 (plan + mockup, closed). Cluster: "The product's surfaces and shells" in the
issue map #168.

### LINKED ISSUE #761: Activity: an icon per run surface — Slack, Telegram and Mattermost get their marks
## What's needed

The Surface column renders the raw enum value as muted text
(`frontend/src/components/runs/run-table.tsx:87`), and the surface filter lists the
same six bare words (`run-history-tab.tsx:47`). `slack`, `telegram` and `mattermost`
are brands with instantly recognisable marks, and a column of lowercase words is the
slowest possible way to scan "which of these came in through Slack".

## The shape

- An icon per `RunSurface` value — the messaging channels get their brand marks (the
  MCP catalog already ships brand logos, see `src/lib/mcp-logos.generated.ts` for the
  pattern), `web`/`embed`/`api` get sensible lucide glyphs.
- The icon carries an accessible label (the surface name), and the same icon appears
  in the surface filter's options so the two read as one vocabulary.
- One mapping module (`surface → icon + labelKey`), not per-component tables — the same
  rule `tool-catalog.ts` exists for (#144's lesson).

## Done when

The run table and the surface filter draw a mark per surface, screen readers still
hear the surface name, and the mapping lives in exactly one module.

---
Part of the Activity-page (/runs) improvement push, worked in PR #759 — successor to
#45. Cluster: "The product's surfaces and shells" in the issue map #168.

### LINKED ISSUE #762: Activity: expose the filters the backend already answers — rated up, person, agent, version
## What's needed

The run history exposes a fraction of the filters the backend already answers.
`GET /runs` takes `user_id`, `environment_id`, `exposure_id`, `agent_version_id` and
`rated` in **both** directions (`backend/app/api/routes/v1/runs.py:56-70`), and the
frontend hook already types `rated: "down" | "up"`
(`frontend/src/hooks/use-runs.ts:54`) — but the tab renders only a status select, a
surface select and a one-way "Rated down" toggle
(`frontend/src/components/runs/run-history-tab.tsx:95-97`). There is no way to see the
runs people rated *up*, no way to narrow to one person's runs, and no way to compare
what version N did against version N+1 — the question the version strip raises and
then cannot answer with rows.

## The shape

- The rating filter becomes three-state: any / rated up / rated down (the hook and the
  route already speak both).
- An agent filter on the page itself — today `?agent=` only arrives via the Builder's
  link, so the reader cannot narrow to an agent without leaving the page.
- A person filter (`user_id`) and — when the list is already narrowed to an agent — a
  version filter (`agent_version_id`), each fed from data the page already loads or
  can cheaply ask for.
- LangSmith's filter bar is the reference: the filters read as one row of removable
  narrowings, not scattered controls.

## Done when

Rated-up runs are one click away, a reader can narrow to an agent / a person / a
version without hand-editing the URL, and every filter is a query the server answers
over the whole history (never a client-side pass over one page).

---
Part of the Activity-page (/runs) improvement push, worked in PR #759 — successor to
#45. Cluster: "The product's surfaces and shells" in the issue map #168.

### LINKED ISSUE #763: Run export drops the on-screen status/surface/rating filters it claims to carry
## What breaks

`ExportMenu`'s contract says the `params` prop carries "the filters currently applied
on the tab, sent verbatim so the file is what is on screen"
(`frontend/src/components/runs/export-menu.tsx:38-39`). The run-history tab breaks that
contract: it passes only `agent_id` (`frontend/src/components/runs/run-history-tab.tsx:160-170`),
so the status, surface, rating and slow-runs narrowings visible on screen are silently
dropped from the file.

The backend is not the gap — `GET /runs/export` accepts `status`, `surface`, `rated`,
`took_over_ms`, `user_id` and the rest, mirroring the list route
(`backend/app/api/routes/v1/runs.py:150-163`), and its docstring promises "the same
filters as the list route, so the file is what is on screen".

## The sequence

1. Filter the run history to `status=failed`, surface `slack`.
2. Export CSV, 30 days.
3. The file contains every run of every status on every surface. Anyone reading it as
   "the failed Slack runs" — the obvious reading, since that is what was on screen —
   acts on wrong data.

## Fix

Pass the tab's live filter state through `params` (status set, surface, rated,
took_over_ms), the same encoding `useRuns` puts on the wire. When the date-range picker
lands, the export window comes from it too.

## Done when

The exported CSV row set equals what `GET /runs` would list under the filters shown on
screen at the moment of export, proven by a test that sets a filter and asserts the
query string the export request carries.

---
Pre-existing, found preparing the Activity-page push in PR #759. Cluster: "The
product's surfaces and shells" in the issue map #168; the export itself landed in #211
(closed).

### LINKED ISSUE #764: Activity: paginate the run history — the page shows fifty rows and says nothing about the rest
## What's needed

The run history shows the first fifty rows and nothing else. `GET /runs` paginates —
`skip`/`limit`, `limit` capped at 100 (`backend/app/api/routes/v1/runs.py:75-76`) — and
answers `total`, but `useRuns` never sends `skip` (`frontend/src/hooks/use-runs.ts:74-94`)
and the tab renders no pager. An organization past its first afternoon cannot reach run
fifty-one except by narrowing filters until it surfaces; nothing on screen even says
the list is truncated.

The shared `PaginationBar` that PR #759 standardised across the resource pages (#282)
is the control to reach for; `/runs` stayed off the list-page shell deliberately, but
the pager itself is exactly this case.

## Done when

The run history pages through the whole narrowed set with the shared pagination
control, the row-count sentence says where the reader is ("51–100 of 1,204"), and
changing a filter or the sort resets to the first page.

---
Part of the Activity-page (/runs) improvement push, worked in PR #759 — successor to
#45. Cluster: "The product's surfaces and shells" in the issue map #168; the shared
control is #282's.

### LINKED ISSUE #765: Activity: open the conversation behind a run — the id is there and nothing links it
## What's needed

A run knows its conversation — `AgentRunRead.conversation_id`
(`frontend/src/types/runs.ts:90`, `backend/app/db/models/agent_run.py:174`) — and the
chat page already deep-links (`/chat?id=<conversation_id>`,
`frontend/src/app/[locale]/(dashboard)/chat/page.tsx:6`). Nothing in
`components/runs/` connects the two: from a run row, a focused run or the transcript
view there is no way to open the conversation the run happened in. The reader sees a
failed run, or one rated down, and cannot get to the exchange that produced it without
hunting the chat sidebar by memory.

This is the run-side twin of #137's "a link to the conversation" for workspaces, and
what LangSmith/Logfire do with every trace: the record links to the source.

## The shape

- An "open chat" action on a run row and on the focused-run view, rendered only when
  `conversation_id` is non-null (an API run with no conversation gets no dead control)
  and only for a caller the conversation will actually answer for.
- The transcript view (`useRunTranscript`) keeps its role — the run's own record —
  while the link answers "take me to the live thread".

## Done when

From any run with a conversation behind it, the conversation is one click away, and a
run without one shows no control rather than a disabled one.

---
Part of the Activity-page (/runs) improvement push, worked in PR #759 — successor to
#45. Cluster: "The product's surfaces and shells" in the issue map #168; adjacent to
#137 (workspace → conversation link).

### LINKED ISSUE #766: Activity: one control grammar and an observability-grade look across Approvals, Runs and Spend
## What's needed

The three Activity tabs — Approvals, Runs, Spend — work, but read as three stacked
cards with controls scattered where each feature landed: the export floats above the
Approvals and Spend tabs but inside the Runs card, the canned views / status / surface
controls sit in a loose row, the figures strip above the tabs does not react to any
narrowing, and the tables are dense text with no scannable hierarchy. Compared against
LangSmith's runs table and Logfire's live view — the two products whose job this page
shares — it reads as unfinished rather than as an observability surface.

## The shape (LangSmith / Logfire as reference)

- One consistent control row per tab: time window first, then filters, then sort,
  export on the right — the same grammar on all three tabs.
- The run table gets the observability-table treatment: status as a colored dot +
  label, surface as an icon (see the surface-icons issue), monospaced numerics
  right-aligned, relative timestamps with the absolute on hover, a row hover state,
  and the row itself clickable into the detail instead of detail-by-URL only.
- The figures strip reads as KPI tiles tied to the selected window, not three
  disconnected numbers.
- Approvals and Spend adopt the same card grammar so switching tabs does not feel
  like switching products.

Scope is presentation and layout; every behaviour change (date picker, filters,
pagination, chat link, export params) is its own issue, referenced from PR #759.

## Done when

The three tabs share one control grammar and one visual density, and a screenshot of
the Runs tab stands comparison with LangSmith's runs table without embarrassment.

---
Part of the Activity-page (/runs) improvement push, worked in PR #759 — successor to
#45 (the original "plan + mockup first", closed). Cluster: "The product's surfaces and
shells" in the issue map #168.

DEENUU1:
One branch for the W2 frontend-standardisation push: the shared table primitive, the shared list-page shell, the admin pages on both, the two product slices that fell out of them (no skills install step, no admin Refresh), and the agent map's four-direction layout.

## What changed

**One table primitive (#139).** `DataTable`'s `Column<T>` gains `sortable`/`sortValue` and `filter`/`filterOptions`/`filterValue`, each working in two modes: client-side over rows the caller holds whole, and server-side as a request (`sort`/`onSort`, `filters`/`onFilter`) — a client-side sort of page one, on a list with three pages, is worse than no header, so which mode a table uses follows where its rows live. The primitive owns the toggle reducer three pages had written verbatim, sets the `aria-sort` its `SortButton` docstring had promised and never delivered, and sorts a null value last in both directions. Every table migrated: the four sandbox tables, vault secrets, channel bots, `RunTable`, the admin organizations overview, plus the tables already on `DataTable` — and `components/ui/table.tsx` is deleted with its last caller. `teams/members-table` turned out to be rendered by nothing and was deleted rather than migrated. Sort state survives a reload: `useUrlSort` mirrors it into `?sort_by=/?sort_dir=`, validated against the same whitelist the backend route declares.

Where sorting or filtering was deliberately **not** added: sandbox sessions (a server-bounded live slice — a header sorting one page of it would lie), the KB documents table (a paginated append list under a Load-more), the org roles matrix (a permission grid whose columns are data, not fields), the channel bots list (a handful of rows, every question badge-comparative), and — caught reviewing the branch's own diff — the admin organizations overview and the org members table, which each hold at most one server page (`limit=50`), so a client-sort header over them would claim a whole-collection order the rows cannot deliver. The vault kind column carries the first select filter; server-mode filters are exercised by the primitive's tests and waiting for the first page that needs one in the header rather than the shell's filter bar.

**One list-page shell (#282).** `ListCard`/`ListCardEmpty` replace the five per-page card copies (Skills, MCP, vault, sandboxes, knowledge bases) and four inline empty-state blocks; search sits in the card header, secondary filters in a row at the top of the content (the MCP page's bar moved in from above the card), `Pager`/`PaginationBar` at the foot. Adopting the shell also closed three silent-error gaps: vault ignored the `listError` its hook documented loudly, the MCP page ignored its hook's `error`, and `useSkills` had no error to expose — each now renders `ErrorState` inside the card instead of an empty list dressed as reassurance (#32's shape). `/runs` stays off the shell on purpose: it is a tab container whose lists live one level down, each with its own states. Sandboxes, workspaces, vault and knowledge bases render unpaginated backend lists, so their shells carry no pager — the backend routes take no `skip`/`limit` to drive one.

**Admin standardised (#284).** `/admin/users` and `/admin/conversations` sort through the primitive (users' local `SortHeader` — no `aria-label`, restated header classes — deleted), all three list tabs paginate through one `PaginationBar` instead of three dialects, the overview's raw `<table>` becomes `DataTable`, the overview and system sections adopt `ListCard`, the system page's bespoke error card becomes `ErrorState`, and the users list gets an inline error state instead of a toast that expires over an empty table. Left as a follow-up: the conversations page reads the top-level `admin` i18n namespace where its siblings read `pages.admin` — a copy-only move.

**No install step (#281).** Every organization now starts with the shipped skill library: creation copies each bundled skill in as an ordinary org-visible row owned by the owner — the same shape `seed-skills` writes — so the Builder's picker and the runtime resolve them like anything else, and `/skills` shows one list with the existing `built-in` badge. The per-org row stays (rather than virtual library entries) because a spec binds skills by row id, and `resolve_for_agent`, publish validation and the audit trail all look rows up by id. `GET /skills/library`, `POST /skills/library/{key}/install`, the `installed` flag, the `LibrarySkill` schemas, `names_in_use` (its only reader), the gallery component and `useSkillLibrary` are gone; `install_from_library` survives as the seeding primitive and always lands org-visible. No data migration — existing organizations get missing bundled skills via the idempotent `seed-skills` command, and this deployment has none that predate the change. `docs/skills.md` updated in the same breath. Seeding tolerates a bundled skill whose name collides with another — nothing validates the shipped library's own uniqueness, and a packaging mistake there must not refuse every registration on the deployment.

**No Refresh button (#283).** The overview's Refresh refetched three read-only queries nothing on the page mutates; removing it exposed no stale data, so no invalidation was owed. The system page keeps its auto-refresh control, which is a polling cadence, not a workaround.

**Agent map in four directions (#518).** Surfaces on the left (dashboard chat, HTTP API, WebSocket, widgets, channel bindings — the old Channels box named one of five), model and budget on top, toolbox/MCP/knowledge/skills on the right, delegation at the bottom. Each subagent chip is a first-class bottom node with its own measured edge and the same openable detail as before, and the delegation policy gets a box of its own naming `allow_dynamic` — the setting with the widest consequences on the page — beside the default mode, depth and fan-out limits. (The issue's `allow_questions` does not exist on `SubagentsConfig`; the box carries what does.) The geometry generalises from a binary in/out to a `MapSide` with per-face anchors and a vertical curve twin.

## How it was verified

- `make check` green locally (lint, backend tests + the 100% platform gate, frontend coverage gate, build, docs `--strict`, audit).
- New tests: DataTable sorting/filtering in both modes (incl. null-last and `aria-sort`), `useUrlSort` (whitelist, mangled input, URL round-trip), `PaginationBar`, `ListCard`/`ListCardEmpty`, organization creation seeding the library (both org kinds), `useSkills` surfacing a failed read, the map drawing all four sides.
- Regression guards kept passing unmodified: the rag #32 error-vs-empty suite, the admin conversations error-state suite, the mcp-server-list integration suite, the sandbox loading/empty/error pins.

Closes #139
Closes #282
Closes #284
Closes #283
Closes #281
Closes #518

## The Activity page push (#760–#766)

The second wave on this branch: the /runs overhaul planned in the PR's TODO
comment, LangSmith and Logfire as the reference points.

**One window (#760).** A `PeriodControl` extracted verbatim from the dashboard's
FilterRow owns the page's time window - presets plus the custom calendar,
round-tripping through `?period=` exactly as the dashboard does. The figures,
the run table, the version strip, the Spend tab and all three exports read that
one window; the spend figure became the window's spend (summed from the
per-agent rows, so a delegation counts once); the p95 deep-link hands over
`period=` instead of resolved instants. The unfiltered empty state blames the
window, because with a default 30-day window "No runs yet" would tell an older
organization nothing has ever run.

**Every filter the backend answers (#762).** A `RunFilterBar` with status,
surface, a three-state rating (rated *up* existed in the API and hook all
along), agent, person, and - narrowed to an agent - version. The agent and
version selects fill from `/agents` and are withheld without `agents:view`;
picking a different agent drops the version narrowing. The agent narrowing is
page state mirrored into `?agent=`.

**Pagination (#764).** The shared `PaginationBar` under the table, `skip` on the
wire, "51–100 of 1,204", snapping to page one whenever the window, a filter or
the sort redefines the set being paged.

**The export tells the truth (#763, bug).** The tab passed only `agent_id`, so a
CSV exported beside a narrowed table silently contained everything. It now
carries the live filter state in the route's own names, and the window comes
from the period control instead of its own 7/30/90-day presets.

**Surface marks (#761).** Slack, Telegram and Mattermost wear their brand marks
(the same components the channels list draws), web/embed/api get glyphs - one
`SurfaceIcon` mapping shared by the table and the filter, aria-hidden beside the
name that carries the fact, rendering nothing for a surface it has no mark for.

**The chat behind a run (#765).** `conversation_id` was on `AgentRunRead` all
along; the frontend type now declares it and each row links to `/chat?id=` -
only when there is a conversation and only on the reader's own runs, because
the chat page lists its owner's threads.

**One control grammar (#766).** Window first, then per tab one control row -
filters left, export right - above a card that is the table and nothing else,
on all three tabs. Rows are clickable into the run detail (`?run=` became page
state through a new `useUrlState` hook, which also replaced the hand-rolled
`?agent=` copy of the same pattern); Started reads relative with the absolute
instant on hover; the narrowing notices' ways out are actions that clear the
page state rather than links that strand it.

Verified the same way as the first wave: scoped suites per piece (the tab's
sort/filter/paging wire-ups, the bar's permission gates, the export's query
string, the table's chat-link and row-click interplay, the hook's URL
round-trip), then `make lint-frontend`, `make test-frontend-cov` (100%
statements, 97.75% branches) and `make build-frontend` green locally.

## Third wave: the drawer, one approvals table, and the app-wide sweep

**Run detail in a drawer.** The detail opens in a right-side Sheet over whichever
tab opened it - the Sheet lives on the page, so a run row and an approval row are
both doors to the same view. Prev/next walk the run's *own conversation*
(server-resolved neighbours), the timeline replays the whole thread with tools
rendered raw, and a parked run offers a manual Resume.

**Approvals as one table.** The queue and the decided record are the same rows in
two states, so they share one table: pending rows keep their arguments expanded
with Approve/Reject in the row, decided rows fold them behind a disclosure.
Deciding the last outstanding call now also resumes the run - the queue used to
post only the decision, which left a run approved and parked forever.

**The shell, everywhere.** ListCard grows ControlsRow/FootRow and every resource
list moves its controls inside the container; tabs restyle once to the underline
form; the vault's For filter joins the controls strip (and DataTable's
filter-in-header feature, its only consumer gone, is deleted); workspaces rows
wear the agent's face and link the reader's own conversation.

**Skills top-up.** A bundled skill added to the catalog after an organization was
created never reached it - the listing now materializes the gap itself, so
`/skills` always shows the whole shipped catalog.

**Admin.** Organizations gets its own tab, Recent activity and the per-page
selectors are gone, the user drawer moves onto the shared Sheet dialect, and
`/admin/ratings` is deleted whole - ratings are read where the runs are, on the
Activity page.

Closes #760
Closes #761
Closes #762
Closes #763
Closes #764
Closes #765
Closes #766


DEENUU1:
## Activity page (/runs) — the improvement push, tracked

This branch continues into a proper overhaul of the Activity page. LangSmith's runs
table and Logfire's live view are the reference points. Filed today as #760–#766
(all on `W2 · Aug 10–14`, cluster added to the map #168); this comment is the
worklist.

### Already landed on this branch

- [x] Status + surface filters, a cost sort, export moved inside the history card (7f869eb6)
- [x] The table/shell primitives everything below builds on — #139, #282 (this PR closes both)

### To do, in working order — **all seven landed** (`bd2e010c`…`39c92644`)

- [x] **#760 — date-range picker.** Owned by the page, presets + custom from/to,
      windows all three tabs, mirrored into `?started_from=/?started_to=`, and becomes
      the export's mandatory range (replacing the 7/30/90-day presets).
- [x] **#762 — the filters the backend already answers.** Rating filter goes
      three-state (any / rated up / rated down — today only "Rated down" exists), plus
      an agent filter on the page itself, a person filter, and a version filter when
      narrowed to an agent. All server-side, like the existing ones.
- [x] **#764 — pagination.** `GET /runs` takes `skip`/`limit` and answers `total`; the
      UI shows the first fifty and says nothing. Shared `PaginationBar`, "51–100 of
      1,204", reset to page one on any filter/sort change.
- [x] **#763 — export carries the on-screen filters** *(bug)*. The tab passes only
      `agent_id`; status/surface/rating/slow-runs are silently dropped from the CSV
      while both docstrings promise "the file is what is on screen". The export
      endpoint already accepts every one of them.
- [x] **#761 — surface icons.** Slack/Telegram/Mattermost get their brand marks,
      web/embed/api get glyphs — in the table column and the filter, from one mapping
      module (the `tool-catalog.ts` rule, #144's lesson).
- [x] **#765 — open the chat behind a run.** `AgentRunRead.conversation_id` exists and
      `/chat?id=` already deep-links; nothing connects them. Link on the row and the
      focused-run view, rendered only when there is a conversation to open.
- [x] **#766 — one control grammar, observability-grade look.** Same control row on
      all three tabs (window → filters → sort → export), status dots, right-aligned
      monospaced numerics, relative timestamps, clickable rows, KPI-style figures tied
      to the window. Presentation only; behaviour is the issues above.

### Related open issues — adjacent, not this PR's scope

- #194 — spend_by_provider/by_key cannot attribute a delegation's vendor or key
  (backend; the Spend tab's split is silently wrong for cross-provider delegations).
- #502 — an answered ask_user question leaves no trace in the transcript (affects the
  run transcript surface, W4).
- #563 — rating-buttons hand-rolls fetch instead of a hook (ratings plumbing, W4).
- #3 / #15 / #235 — run accounting correctness (a failed run's row, resume budget
  double-count, terminal commit vs cancellation) — backend, W5; they change what the
  history *contains*, not how it reads.

Checked the closed ones too: #45 (the original plan+mockup), #210/#211 (duration
sort/filter, CSV export), #489 (version strip), #539/#620 (spend caveats) are all
done and superseded by the above.

DEENUU1 (commented):
Self-review before merge — ai-review is off (#311), so this sweep is the whole job. The primitives are in genuinely good shape: DataTable's error-beats-empty precedence, null-last in both directions and the single shared ordering site in `agent_run.py`; the skills top-up's per-install savepoint; the filter/export parity on the runs tab (#763 is really fixed — every param name matches the route literally); the URL plumbing composing without clobbering; and deletion hygiene is clean everywhere (no survivor of ratings, the gallery, `ui/table`, or the library endpoints).

Three things need fixing before this merges:

1. **The description claims all three Activity tabs export; only History does.** The approvals/spend export routes exist and `ExportMenu` was built to be reused — it's wired once.
2. **`admin/organizations` has the exact sort-a-truncated-page trap the description says was caught and removed.** Four `sortable` columns over `limit=50`, no pager.
3. **The `/runs` navigation proof passes on the loading skeleton** — the skeleton renders a real `<table>`, so the spec passes even when every query 502s, which is the #32 shape the spec's own comment says it avoids.

Things outside the diff's commentable lines, recorded here:

- `backend/app/schemas/agent_run.py:173-190` — `RunTranscript`'s docstring still says items are "narrowed to a single run by `messages.run_id`"; under `scope=conversation` that's untrue, and this docstring is the mkdocstrings source.
- Stale "gallery" wording survives in `backend/app/commands/seed_skills.py:3` and `backend/app/services/skill_library.py:60` and `:80` — the gallery and its Install button no longer exist.
- `nav.ratings` is orphaned in both catalogs (`en.json`, `pl.json`) — the tab left `ADMIN_TABS` but the key stayed; and the `pages.admin` subtitle still promises "ratings" while not mentioning the new organizations tab.
- Pre-existing, found reviewing this branch — worth filing rather than folding in: `useChannelBots` cannot report a list failure at all (channels renders "No channels yet" over a 502), and `useMcpServers` reports only the personal list's error, so a failed `/agents/mcp-catalog` still renders as an empty catalog — both contradict the "every resource list has an error state" claim, so the PR text should scope that claim or the branch should carry the fixes. Also pre-existing: the hardcoded English `"Approved"/"Rejected"` toast in `use-runs.ts:210`, and the hardcoded `built-in` badge in `skill-card.tsx:37` (rides the #656 hyphen blind spot).

frontend/src/app/[locale]/(dashboard)/admin/organizations/page.tsx
```diff
@@ -0,0 +1,108 @@
...
+    },
+  });
+
+  const columns = useMemo<Column<AdminOrganization>[]>(
+    () => [
+      {
+        key: "name",
+        header: t("name"),
+        className: "pl-5",
+        sortable: true,
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/admin/organizations/page.tsx]: **Blocking.** This is the exact trap the PR description says was caught and removed: four `sortable` columns client-sorting a `/admin/organizations?limit=50` fetch, no pager, under a card that says "every tenant". Past 50 orgs, "sorted by members" sorts an arbitrary first page — `data-table.tsx`'s own docstring calls that worse than no header. Either drop `sortable`/`sortValue` from all four columns (matching what the org members table already does), or read `total` and add `PaginationBar` + server-side sort like `admin/users`.

frontend/src/components/runs/run-history-tab.tsx
```diff
@@ -101,115 +172,161 @@ export function RunHistoryTab({
...
+                >
                   {t("showEveryAgent")}
-                </Link>
+                </button>
               </p>
-            )
+            )}
+          </div>
+          {canView && (
+            <ExportMenu
```
DEENUU1, line 250: **Blocking (or amend the PR text).** The description says all three exports read the shared window, but this is the only `ExportMenu` in the tree — `rg "spend/export|approvals/export"` over `frontend/src` returns nothing, while both backend routes exist and demand a range. The component was clearly built for reuse (`kind`, `rangeParams`, `permission` props). Wire it on Approvals (`created_from`/`created_to`, gated on the decide permission) and Spend (`from`/`to`, `runs:view`), or scope the claim down to the history tab.

frontend/e2e/navigation.spec.ts
```diff
@@ -54,11 +54,14 @@ const DASHBOARD_PAGES: {
...
-    // control on the page that only renders once the approvals query resolved.
+    // are asserted directly in `activity.spec.ts`; here the proof is the run
+    // history card's own empty state, which only renders once /runs resolved.
     path: "/runs",
     heading: "Activity",
-    proof: (page) => page.getByRole("heading", { name: "Nothing waiting" }),
+    proof: (page) =>
+      // Data-borne either way the query resolves: rows draw the table, none
+      // draws the window's empty state - chrome alone proves nothing (#32).
+      page.getByRole("table").or(page.getByText("No runs in this window")).first(),
```
DEENUU1: **Blocking.** This proof is satisfied by chrome, which is the thing the comment above it says it avoids: `LoadingState variant="skeleton-table"` renders a real `<table>` (`loading-state.tsx:197`), which carries the implicit `table` role and is on screen while `/runs` is still in flight — so `getByRole("table")` resolves on the skeleton and the spec passes even on a deployment where every `/runs` query 502s. In the seeded CI environment there are never runs, so the honest proof is the empty-state text alone:

```suggestion
      page.getByText("No runs in this window").first(),
```

backend/app/repositories/agent_run.py
```diff
@@ -179,6 +179,47 @@ async def get_run(db: AsyncSession, run_id: UUID, *, organization_id: UUID) -> A
...
+    A run that never started sits outside every timeline, and one with no
+    conversation has no thread to walk - both answer no neighbours at all.
+    """
+    if run.started_at is None or run.conversation_id is None:
+        return None, None
+    position = tuple_(AgentRun.started_at, AgentRun.id)
+    anchor = tuple_(run.started_at, run.id)
+    base = select(AgentRun.id).where(
+        AgentRun.organization_id == run.organization_id,
+        AgentRun.conversation_id == run.conversation_id,
```
DEENUU1, line 216: The walk filters on organization + conversation + position only — no `parent_run_id` term. Delegated runs are written with the parent's `conversation_id` and a real `started_at`, so "next" from a parent that fanned out steps through its own children before reaching the thread's next top-level turn — rows the default list (which excludes delegations) never showed. The integration tests never create a delegated run, so the case is unpinned. If the intent is level-locked neighbours:

```python
AgentRun.parent_run_id.is_(None) if run.parent_run_id is None
else AgentRun.parent_run_id == run.parent_run_id
```

(which also gives a delegated run sibling-neighbours). If stepping into delegations is intended, say so in this docstring — it currently frames the walk as the thread's turns.

backend/app/api/routes/v1/runs.py
```diff
@@ -216,9 +225,22 @@ async def get_run_transcript(
...
+        description=(
+            "`run` answers the run's own turns; `conversation` answers the whole "
+            "thread it sits in, each turn carrying its `run_id` so a client can "
+            "still tell which ones are this run's"
+        ),
+    ),
 ) -> Any:
     """The turns one run produced, for a run detail view to render as steps.
 
+    `scope=conversation` is the detail view showing the run in context: the whole
```
DEENUU1, line 240: The "this does not widen it" claim below isn't quite true: `POST /conversations/{id}/messages` writes turns with `run_id=None`, and the conversation-scope read serves them unfiltered. So a member's hand-appended note in their own thread becomes readable to any `runs:view` holder who opens any run in that thread, while the conversation route itself stays owner-scoped. Marginal in practice, but on this platform the refusal is the product — decide it consciously: either correct this docstring (and the service's) to say the scope reads the whole thread including non-run turns, or add `Message.run_id.is_not(None)` to the conversation-scope read and accept holes in the rendered thread.

backend/app/repositories/agent_run.py
```diff
@@ -399,7 +440,14 @@ async def list_runs(
         query = query.where(clause)
         count_query = count_query.where(clause)
 
-    column = _duration_ms() if order_by is RunOrder.DURATION else AgentRun.started_at
+    if order_by is RunOrder.DURATION:
+        column = _duration_ms()
+    elif order_by is RunOrder.COST:
```
DEENUU1, line 462: `cost_usd` and the token columns are `NOT NULL DEFAULT 0` and only written at finish, so a still-running run that has already burned tokens sorts as the *cheapest/lightest* row under `descending=false` — where duration deliberately excludes unfinished runs to the bottom via NULL. The 0 means "not recorded yet", not "free". The docstring above ("Both orders put nulls last") is also stale now there are four orders. Cheapest honest fix is the docstring; parity would take a CASE that nulls cost/tokens when `ended_at` is null so `nullslast()` applies to all four the same way.

backend/app/services/agent_runner.py
```diff
@@ -3350,11 +3350,35 @@ async def get_run(self, ctx: AuthContext, run_id: UUID) -> AgentRun:
             raise NotFoundError(message="Run not found", details={"run_id": str(run_id)})
         return run
 
+    async def neighbor_run_ids(
+        self, ctx: AuthContext, run: AgentRun
+    ) -> tuple[UUID | None, UUID | None]:
+        """The runs either side of `run` in its agent's history, for the detail view.
```
DEENUU1: "in its agent's history" — the repository deliberately walks the **conversation**, and its own docstring explains why agent-history was the bug. This wrong phrase ships into the generated reference docs, and it repeats in four more places introduced by this branch: `focused-run.tsx:38` and `:93`, `types/runs.ts:36-38`, and the "walking the agent's history" describe block in `delegated-runs.integration.test.tsx`. Concrete cost: an API-surface run has no conversation, so its arrows are permanently disabled — which these comments say cannot happen, so the next person fixes the wrong side. The schema field descriptions already say conversation; align the other five.

backend/app/services/skills.py
```diff
@@ -234,60 +233,50 @@ async def list_readable(
...
+            return
+        owner_ctx = AuthContext(
+            user_id=owner_id,
+            organization_id=ctx.organization_id,
+            role=OrgRoleName.OWNER,
+        )
+        for entry in missing:
+            try:
+                async with self.db.begin_nested():
+                    await self.install_from_library(owner_ctx, entry.key)
```
DEENUU1, line 274: The audit rows this writes record the first owner as the actor for writes the owner never made — triggered by any reader's GET, possibly years after that owner last signed in. Attribution is the audit trail's whole job. A distinguishing detail on the seeding-path entries (e.g. `details={"seeded": True}` or a dedicated `skill.seeded` action) would let a reader tell "the owner did this" from "the platform did this as the owner".

backend/tests/test_skills.py
```diff
@@ -683,6 +620,130 @@ async def test_deleting_a_skill_takes_the_grants_that_pointed_at_it(self):
...
+            patch(f"{SKILLS_PATH}.member_repo.first_owner_id", new=AsyncMock(return_value=None)),
+            patch.object(SkillService, "install_from_library", new=AsyncMock()) as install,
+        ):
+            listing = await SkillService(_db()).list_readable(ctx)
+
+        install.assert_not_awaited()
+        assert listing.total == 0
+
+    @pytest.mark.anyio
+    async def test_a_racing_twin_costs_that_row_and_never_the_readers_page(self):
```
DEENUU1, line 717: This mocks `install_from_library` and the session, so the load-bearing mechanism — `begin_nested()` recovering the session after a real asyncpg unique violation — is never exercised against Postgres. If the savepoint semantics were wrong, everything here would still pass and the first real race would 500 the reader's page. One integration test would close it: pre-insert a skill with a bundled name, patch `list_names` to report it missing, call `list_readable`, assert the page still answers.

frontend/src/components/runs/activity-figures.tsx
```diff
@@ -29,25 +16,33 @@ function monthStart(): string {
...
+  const { spend, isLoading: spendLoading, error: spendError } = useSpend(range);
   const {
     total: organizationRuns,
     isLoading: runsLoading,
     error: runsError,
-  } = useRuns(undefined, { startedFrom: monthStart() });
+  } = useRuns(undefined, { startedFrom: range.from, startedTo: range.to });
   // `total`, not the length of the page. `GET /approvals` answers fifty rows at
   // a time and nothing here asks for more, so a queue of a hundred and twenty
   // read "50" and went on reading it however long the queue grew - the same
```
DEENUU1, line 67: The skeleton gate below covers `spendLoading || runsLoading` only — `useApprovals`'s loading state is never read, so the Waiting figure renders its `?? 0` default while the query is in flight. That's exactly the "a nought here is a claim" defect this file's docstring forbids for the other two figures. Include the approvals loading state (when `canDecide`) in the gate.

frontend/src/app/[locale]/(dashboard)/runs/page.tsx
```diff
@@ -30,19 +45,28 @@ export default function RunsPage() {
...
-  const startedTo = searchParams.get("started_to");
+  // The page's one window, shared by the figures, all three tabs and the
+  // exports - the same `?period=` vocabulary the dashboard round-trips, so a
+  // narrowed view survives a reload and travels in a pasted link.
+  const [period, setPeriod] = useState<Period>(() => parsePeriodParam(searchParams.get("period")));
+  const changePeriod = (next: Period) => {
+    setPeriod(next);
+    setUrlParam("period", formatPeriodParam(next));
+  };
   const { can, isLoading: permissionsLoading } = usePermissions();
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/runs/page.tsx:70]: `useApprovals` is gated on `enabled: canDecide` with a paragraph explaining why — but the queries those controls narrow are not: `useRuns` in the tab and `useSpend`/`useRuns` in the figures fire unconditionally, and Member/Viewer hold no `runs:view`, so a deep link to `/runs` (nav hides it, the URL doesn't) makes two to three requests that predictably 403 and render as failures with "Try again" CTAs. Same `enabled: canView` treatment plus a no-access state closes it.

frontend/src/hooks/use-runs.ts
```diff
@@ -162,12 +199,23 @@ export function useApprovals(options?: { enabled?: boolean }) {
...
+      // The invalidation refetches the queue, so the read below is fresh.
       await queryClient.invalidateQueries({ queryKey: qk.runs.all() });
       toast.success(approval.status === "approved" ? "Approved" : "Rejected");
+      // Deciding the last outstanding call makes the resume possible; nothing
+      // performs it (the backend keeps the click and the execution apart). The
+      // chat's dialog has always followed its decision with the resume - the
+      // queue not doing the same left a run approved and parked forever.
+      const stillParked = queryClient
+        .getQueryData<ApprovalList>(qk.runs.approvals())
+        ?.items.some((item) => item.run_id === approval.run_id && item.status === "pending");
```
DEENUU1, line 222: `stillParked` reads the cached first page (50 oldest) of `/approvals` — a run whose other pending approval sits past those 50 reads as clear, `resume.mutate` fires, and the backend's refusal surfaces as an error toast for a flow that did nothing wrong. Either ask the server (a pending-for-run check) or swallow that specific refusal on this path.

frontend/src/components/runs/approvals-tab.tsx
```diff
@@ -3,113 +3,226 @@
...
+  period,
+  onFocusRun,
+}: {
+  period: Period;
+  onFocusRun: (runId: string | null) => void;
+}) {
   const t = useTranslations("pages.runs");
   const locale = useLocale();
   const { approvals, total, isLoading, error, decide, refetch } = useApprovals();
+  const history = useApprovalHistory({ from: periodStart(period), to: periodEnd(period) });
```
DEENUU1: `useApprovalHistory` takes no `skip`/`limit`, so the decided record silently truncates at the backend's 50 — the foot-note below covers the pending queue only, so "80 decided in this window" sits over 50 rows with nothing explaining the gap. Mirror the note for the decided half, or paginate it.

frontend/src/components/runs/run-history-tab.tsx
```diff
@@ -33,63 +56,111 @@ const SLOW_RUN_THRESHOLD_MS = 30_000;
...
   // Independent of the sort: "slow runs" is a filter, and the reader can still
   // re-sort the slow set by start time without it ceasing to be the slow set.
   const [minDurationMs, setMinDurationMs] = useState<number | null>(null);
 
-  const { runs, isLoading, error, refetch } = useRuns(agentId ?? undefined, {
-    startedFrom: startedFrom ?? undefined,
-    startedTo: startedTo ?? undefined,
+  // The version narrowing belongs to one agent's history: carried across a
+  // change of agent it would silently empty the next agent's list.
+  const changeAgent = (next: string | null) => {
```
DEENUU1, line 118: `versionId` resets only through `changeAgent` — an `?agent=` change arriving via navigation (the useUrlState "navigation wins" path) bypasses it, leaving the previous agent's `agent_version_id` on the wire and an inexplicably empty list. Narrow window (needs an in-page nav carrying `?agent=`), but the navigation-wins mechanism exists on purpose; key the reset the way paging is keyed on `narrowingKey`.

frontend/src/components/ui/data-table.tsx
```diff
@@ -19,6 +20,23 @@ export interface Column<T> {
    * Omit to keep the column always visible.
    */
   hideBelow?: "sm" | "md" | "lg";
+  /**
+   * Renders the header as a sort control. With `onSort` on the table the click
+   * is the caller's (server-side sorting); without it the table sorts the rows
+   * it holds, which needs `sortValue`.
+   */
+  sortable?: boolean;
```
DEENUU1, line 30: `sortable: true` without `sortValue` in client mode is a lying header: the click flips the arrow and sets `aria-sort`, the rows never move (the `visible` memo skips when `read` is undefined). Every current caller passes `sortValue` — nothing guards the next one. A discriminated column type (client-sortable requires `sortValue`) would make the mistake unrepresentable; a dev-time `console.warn` is the cheap version.

frontend/src/app/[locale]/(dashboard)/orgs/[id]/members/page.tsx
```diff
@@ -352,28 +354,36 @@ export default function OrgMembersPage({ params }: PageProps) {
...
-          skeletonRows={4}
-          getRowKey={(m) => m.id}
-          empty={t("noMembersYet")}
-        />
-      )}
+          a stand-in list could only approximate them. The shared list card
+          keeps the page's shape whether it is empty, loading or full. */}
+      <ListCard
+        title={t("membersCard")}
+        counted={isLoading ? null : t("memberCount", { count: members.length })}
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/orgs/[id]/members/page.tsx]: When the list query *fails*, `isLoading` is false and this states "0 members" as fact over a request that never answered — every org has at least its owner, so that reading is always false. `useMembers` exposes `error` and `refetch` and the page reads neither; this diff fixes the identical shape on vault and sandboxes with the same reasoning written on those pages. The `counted` line is new in this branch; the empty-on-error body below it is pre-existing but worth taking in the same motion.

backend/app/commands/seed_skills.py
```diff
@@ -86,14 +75,7 @@ async def _run(org_id: str | None, dry_run: bool) -> None:
...
-                    # Visible to the organization rather than to the person the
-                    # seed ran as, set at install rather than by a follow-up edit:
-                    # a bundled skill is for everybody, where private is the right
-                    # default for something somebody wrote and the wrong one for
-                    # something the platform shipped.
-                    installed = await service.install_from_library(
-                        ctx, skill.key, visibility=Visibility.ORG
-                    )
+                    installed = await service.install_from_library(ctx, skill.key)
                 except AlreadyExistsError:
```
DEENUU1: Nit: the service's top-up loop catches `(AlreadyExistsError, IntegrityError)`; this catches only the first. A listing top-up committing the same skill mid-command — between this command's pre-check and its flush — crashes the run and rolls back every org seeded before it in the shared session. Idempotent on re-run, so cheap, but one extra exception type makes the two loops agree.

frontend/src/hooks/use-url-sort.ts
```diff
@@ -0,0 +1,51 @@
...
+ * A `replaceState` rather than a router push, like the tab strips beside it:
+ * a sort is a view of the page, not a place Back should return through.
+ */
+export function useUrlSort<K extends string>(
+  allowed: readonly K[],
+  fallback: UrlSort<K>,
+): { sort: UrlSort<K>; setSort: (next: TableSort) => void } {
+  const params = useSearchParams();
+  const [sort, setSortState] = useState<UrlSort<K>>(() => {
+    const by = params.get("sort_by");
```
DEENUU1: Nit: the URL is read only in the `useState` initializer, so a navigation that changes `?sort_by=` under a mounted page is ignored — the sibling `use-url-state.ts` solves exactly this with the render-adjustment pattern and documents why. No page cross-links with `sort_by` today, so nothing breaks; noting the asymmetry so the first page that does isn't a surprise. Related: a mangled `?sort_dir=` silently becomes `desc` even when `fallback.dir` is `asc`.

frontend/src/app/[locale]/(dashboard)/admin/users/page.tsx
```diff
@@ -215,82 +174,42 @@ export default function AdminUsersPage() {
...
-        <div className="relative min-w-[240px] flex-1">
-          <Search className="text-muted-foreground absolute top-1/2 left-3 h-4 w-4 -translate-y-1/2" />
-          <Input
-            placeholder={t("searchByEmailName")}
-            value={search}
-            onChange={(e) => setSearch(e.target.value)}
-            className="pl-10"
+      <ListCard
+        title={t("usersCard")}
+        counted={t("totalCount", { count: total })}
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/admin/users/page.tsx]: Nit, two small ones: `counted` here (and on the conversations page) is always the formatted count, so the header reads "0 total" during the first request — `list-card.tsx` documents `null`-while-in-flight for precisely this, and rag/vault/skills/channels all honor it. And the `ListCardFootRow` below renders unconditionally while `PaginationBar` returns `null` at `total === 0`, leaving an empty bordered strip under an empty list.

frontend/src/components/runs/approvals-tab.tsx
```diff
@@ -3,113 +3,226 @@
...
+          )}
+        </div>
+      ),
+    },
+    {
+      key: "decision",
+      header: t("decisionColumn"),
+      cell: (approval) => (
+        <Badge variant={DECISION_VARIANT[approval.status] ?? "outline"}>
+          {t(DECISION_LABEL[approval.status] ?? "decisionExpired")}
```
DEENUU1: Nit: the `?? "decisionExpired"` fallback labels an *unrecognised* status as the specific state "Expired" — a mislabel rather than a fallback. A neutral humanized fallback (or just the raw status) lies less.

frontend/src/components/runs/run-history-tab.tsx
```diff
@@ -33,63 +56,111 @@ const SLOW_RUN_THRESHOLD_MS = 30_000;
...
+  const agentsById = useMemo(
+    () => new Map(agentRows.map((agent) => [agent.id, agent])),
+    [agentRows],
+  );
+  const membersById = useMemo(
+    () => new Map(members.map((member) => [member.user_id, member])),
+    [members],
+  );
+  const [filters, setFilters] = useState<RunFilters>(DEFAULT_RUN_FILTERS);
   const [sort, setSort] = useState<RunSort>(
```
DEENUU1, line 109: Nit: the dashboard's latency widget deep-links `?sort=duration`, the page reads it once at mount, and this local sort state is never mirrored back — so after the reader re-sorts, the stale `?sort=duration` stays in the URL and a copied link reasserts the duration sort. Clearing `?sort=` on the first user-initiated sort change would make the copied URL honest.

Issue #638, opened 2026-08-12
A channel bot past 200 messages is reminded of the thread's first 200, not its last


DEENUU1:
A channel bot in a long-running chat sends the model the *first* 200 messages
of the thread and never what was just said, so it answers as though the
conversation stopped hundreds of turns ago.

## Where

`backend/app/services/channels/router.py:728` — `_load_history`:

```python
messages = await conversation_repo.get_messages_by_conversation(
    db,
    conversation_id=conversation_id,
    skip=0,
    limit=200,
)
```

`get_messages_by_conversation` orders **oldest first**, so `skip=0` with a
`limit` takes the opening exchanges. The window is meant to be the most recent
200.

## The sequence that triggers it

1. Link a Slack, Telegram or Mattermost bot to an agent.
2. Hold one conversation in one chat past 200 messages — a support channel does
   this in days, and `channel_sessions` keys the conversation to the chat, so
   the thread does not roll over.
3. Ask a follow-up about the previous message.

The agent answers from turn 1–200. Nothing errors, nothing is logged, and the
transcript in the console looks correct — the messages are all there, and only
the slice handed to the model is wrong. That is why it has never been noticed:
the failure is invisible in every artefact except the answer's quality.

## Pre-existing, found reviewing #634

Not caused by that branch. #634 fixed the same bug from the other direction: the
embed surface passed **no** `message_history` at all, and the fix deliberately
took the most recent window rather than a bare `limit`, because the repository
orders oldest-first —
`app/services/embed_session.py:_history`. Reviewing that diff is what surfaced
the channel copy still having the original shape.

## Fix

The shape `embed_session._history` already uses:

```python
total = await conversation_repo.count_messages(db, conversation_id)
messages = await conversation_repo.get_messages_by_conversation(
    db,
    conversation_id=conversation_id,
    skip=max(0, total - HISTORY_MESSAGES),
    limit=HISTORY_MESSAGES,
)
```

Worth deciding at the same time whether 200 is still the number. The embed
surface uses 40 and says why: a public surface's model bill should not be a
function of how long somebody keeps typing. A channel is not public in the same
way, so 200 may well be right — but it should be a decision with a sentence
behind it rather than the number that was there.

There are two of these windows now, in two modules, with two constants. If a
third surface wants one, the slice belongs in `conversation_repo` as
`get_recent_messages(db, conversation_id, limit=...)` and the constant stays per
surface. That is the version that stops this drifting a third time.

## How you would know it was fixed

[...]

Issue #39, opened 2026-08-01
Other channels setup (websocket, frontend, etc.)


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576802d8748e022c5b587c9) · Priority: **High** · Category: Development

The Notion card is a title only. What it covers, based on the code: the surfaces that are *not* the three messaging platforms — the WebSocket chat, the embeddable widget, the public run API — and making them consistent with each other.

## Where we are

`RunSurface` already enumerates them and every run row is stamped with one. Existing surfaces:

- **WebSocket** — `app/api/routes/v1/agent.py` → `AgentSession` (`app/services/agent_session.py`). The most mature of them: one turn at a time, cancel-and-await teardown, session per turn.
- **Embed widget** — `app/api/routes/v1/embed.py` → `EmbedSession`. Origin-checked, token-admitted, rate-bucketed.
- **Public API** — `POST /api/v1/agents/{id}/run`.

## What "setup" should mean

The three surfaces have drifted, and the audit says where:

1. **Only web chat commits its run** (#3). API, embed and channel runs flush cost and then roll it back, so a failed run leaves no record and no budget impact. This is the single biggest inconsistency between surfaces and it is one line.
2. **Only web chat is un-metered for embeddings** (#16) — the mirror-image bug: `agent_chat.py:247` has no `metered_by`, so its knowledge-search cost is dropped.
3. **The embed socket holds a pooled DB connection for the whole browser session** (#12b) while `AgentSession` correctly opens one per turn. Fifteen widget visitors exhaust the pool and take the API down.
4. **Rate limiting is configured and never applied** (#7) — `SlowAPIMiddleware` is not on the stack, so every public surface is unmetered.

## Done when

- [ ] Every surface records a run identically — same commit boundary, same accounting, same metering. A test per surface asserting a failed run is still in history with its cost
- [ ] `EmbedSession` takes a session factory, not a session
- [ ] Rate limits apply to the public surfaces
- [ ] `docs/channels.md` describes what actually exists, surface by surface

DEENUU1:
Two more surface-consistency items, filed separately because they are channel-specific: #152 (a parked run dead-ends in a channel — the thread is told to "check the approvals queue" and never hears the answer) and #153 (per-channel reply formatting).

DEENUU1:
https://github.com/vstorm-co/agenticos/issues/516

DEENUU1:
#517

PR #775, opened 2026-08-15
feat(capabilities): add compaction, metered, with a real context window

### LINKED ISSUE #49: Add Context manager
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680adacc7d0f11caa2c4e) · Priority: **Low** · Category: Development

**Reference:**
- https://github.com/pydantic/pydantic-ai-harness/pull/465
- https://pydantic.dev/docs/ai/capabilities/compaction/

## Why it matters here

Context-window management: compaction, so a long conversation does not fall off the end of the model's window. The Notion card also carries a subtask that is really its own piece of work:

> **display costs and tokens per message and conversation**

That is a UI feature, not a capability — but it belongs to the same question ("what is this conversation costing me"), and the data is already there: `agent_runs` carries `input_tokens`, `output_tokens`, `cost_usd` and `cost_is_partial` per run, and runs link to `conversation_id`. Decide whether to split it into its own frontend issue or carry it here.

## What will bite

- **Compaction rewrites history, and history is persisted.** `agent_runner` dumps `result.all_messages()` into `paused_state` when a run parks. A compacted history that is restored on `resume` must still contain the tool-call ids the deferred results key on, or the replay breaks.
- Compaction costs a model call. Meter it (#16).
- Per-message cost display: `cost_is_partial` must be honoured — a token count shown next to an unpriced model is a number that lies (this is also #45's point 2).

## Done when

### LINKED ISSUE #771: Web chat never rehydrates a conversation's history, so a reopened thread runs with no context
The web chat's message history lives on the WebSocket and is never read back from
the database, so an agent answering a follow-up in an existing conversation is
answering with no context at all.

Found while scoping #49, which is about keeping a *long* history inside the
context window — this is the other end of the same question, and it is why
compaction in #49 is scoped to one run.

## What breaks

`AgentSession.conversation_history` (`backend/app/services/agent_session.py:80`) is
a plain list on the session object. It is appended to at
`agent_session.py:276` and `:285`, read at `:232`, and **loaded from nowhere**:

```bash
rg -n conversation_history backend/app   # four hits, all in agent_session.py
```

So `build_message_history(self.conversation_history)` answers `[]` for any socket
that did not itself produce the earlier turns, and the run is started with an
empty `message_history`. `ChatAgentRunner.execute` takes whatever the caller
passes (`agent_runner.py:2726`) and loads nothing of its own.

The client does not make up for it. `sendChatMessage`
(`frontend/src/hooks/use-chat.ts:563-581`) sends `message`, `conversation_id`,
optional file ids and the model overrides — never prior turns.

## The sequence

1. Have a conversation with an agent. Ask it to remember something.
2. Reload the page (or open the thread in a second tab, or wait for the socket to
   drop and reconnect).
3. Click the same conversation in the sidebar. The transcript renders in full,
   because the *transcript* comes from `GET /conversations/{id}/messages`.
4. Ask a follow-up that depends on anything said before step 2.

The agent answers as if the thread had started with that sentence. The user is
looking at the whole history on screen while the model is looking at none of it,
which is the part that makes it hard to notice and easy to disbelieve.

## Why it is worth more than it looks

- The two other surfaces do rehydrate. `channels/router.py:298` and
  `embed_session.py:595` both build a history from stored messages before running.
  Web chat — the surface the product is demonstrated on — is the one that does not.
- The symptom reads as a model problem ("it forgot"), not as a platform one, so it
  gets attributed to the agent's instructions.

An integration test that runs a turn on a fresh `AgentSession` against a
conversation that already has messages, and asserts the `message_history` handed
to the runner is not empty — and an e2e that asks a question whose answer is only
in an earlier turn, after a reload.

## Worth deciding while fixing it

`build_message_history` reconstructs a conversation as `UserPromptPart` /
`TextPart` and drops tool calls, tool returns and thinking. Restoring text is the
smaller fix and is probably the right first one; restoring the full
`ModelMessage` list is what would let #49's compaction reach across turns. Say
which one this issue is delivering rather than leaving it implied.

Related: #49 (scoped around this), #502 (a different hole in what the transcript
records).

### LINKED ISSUE #772: A message shows a cost with no sign that it is partial, and a conversation shows no total
A message renders `$0.0043` under an answer whose model could not be priced, and
nothing says the number is a floor. `agent_runs` carries `cost_is_partial` for
exactly this; `messages` does not.

Split out of #49, whose Notion card carried "display costs and tokens per message
and conversation" as a subtask. Most of that subtask already shipped — this is the
remainder. Same caveat #45 made about the Activity page, one surface further in.

## Where it stands today

Already there, and working:

- `messages.input_tokens` / `output_tokens` / `cost_usd`
  (`backend/app/db/models/conversation.py:180-197`)
- `MessageRead` serves all three (`backend/app/schemas/conversation.py:191-201`)
- `storedUsage` reads them back on reload, and answers `null` rather than zero when
  a message was never measured (`frontend/src/lib/message-usage.ts`)
- `MessageCost` draws them under a turn; `UsageStrip` draws the newest measured
  turn under the composer

Missing:

1. **No partial flag on a message.** `SpendLedger.has_unpriced_models` is what sets
   `agent_runs.cost_is_partial`, and it is not carried down to the turn. A run on a
   model `genai-prices` does not know books its tokens with `cost_usd = 0` and
   `priced = False`, so the message shows a total that is short by however much the
   unpriced requests cost — presented identically to one that is exact. That is the
   "number that lies" #45 named.
2. **No conversation total.** `ConversationRead` carries no aggregate, so "what has
   this thread cost me" is only answerable by adding up what is on screen.
3. **`MessageCost`'s docstring is now untrue** — it says usage "is not stored per
   message, so a reloaded conversation shows nothing here", which stopped being the
   case when the columns landed.

## What to do

- A `cost_is_partial` column on `messages`, written from the same ledger the run
  row's is written from, exposed on `MessageRead`, carried through `storedUsage`,
  and rendered as a caveat rather than a footnote — the number is a floor, and it
  should read as one.
- A conversation-level total. Decide whether it is a computed field on
  `ConversationRead` or a small aggregate endpoint; the messages are already there,
  so this is a `SUM` and a place to put it, not a new pipeline.
- Fix the `MessageCost` docstring in the same change.

A turn answered by a model with no price entry renders with the caveat and not as
a plain figure; a turn that was fully priced renders exactly as it does now. A
frontend test for both, since the whole point is that the two must not look alike.

Related: #45 (made the same point about `/runs`, closed), #49 (this was its
subtask), #539 and #620 (the other two places a cost is presented without the
gap in it being visible).

### LINKED ISSUE #773: Persist a model profile's context length, so compaction stops trusting a registry that is wrong
The `compaction` capability added in #49 triggers on a fraction of the model's
context window, and that window is resolved from `genai-prices`, which is wrong
for models this platform runs on — in the direction that breaks a run rather than
the one that wastes a summary. We already have a better number and throw it away.

## The two failures

**A spec with fallbacks resolves to nothing.** `ModelRequestSpec.build` returns
`build_with_fallbacks(...)` (`backend/app/agents/model_resolver.py:294`), a
`FallbackModel` whose `model_id` is a composite `fallback:...`. Nothing in the
registry matches it, so the fraction is taken of `fallback_context_window` —
200,000 by default. On a 1M-context model that compacts at a fifth of capacity and
pays for summaries it did not need.

**The registry over-records Sonnet-class Anthropic models.** Measured against
`genai-prices` 0.0.71 and documented by the harness: `anthropic:claude-sonnet-4-5`
records 1,000,000 against a real 200,000. `max_fraction=0.9` there resolves to a
900,000-token trigger, compaction never fires, and the provider refuses the request
instead. Over-recording is the direction that takes a run down.

`compaction`'s `context_window` field is the manual answer to both, and it is a
setting every author has to know to set — which means most will not.

## What we already have and do not persist

`ProviderModelRead.context_length` (`backend/app/schemas/model_profile.py:104`) is
served to the model picker from the live provider catalog, per
`app/services/model_catalog.py` — four of its provider adapters read a
`context_length` / `max_context_length` / `context_window` field straight from the
provider's own API. That is the number the provider itself publishes, which is
better than a bundled snapshot for exactly the cases above.

It is never written down. The model profile row carries no context column, so a
run has no way to ask, and no network call belongs in the request path.

## What to do

- A `context_length` column on the model profile, populated when a profile is
  created or its model changed, from the same catalog the picker reads.
- Carry it on `ModelRequestSpec` so `build_agent` can hand it to the capability.
- Have `compaction` prefer it over registry resolution, leaving `context_window` as
  the override for a deployment that knows better than both.

Decide explicitly what happens for a profile saved before the column existed, and
for a provider that publishes no length — falling back to today's behaviour is
fine, saying nothing is not.

An agent on a profile with fallbacks compacts at a fraction of its real window
rather than of 200,000, with no `context_window` set on the binding. A test that
builds a spec with a fallback and asserts the trigger the capability resolves to.

Blocks nothing, but it is what makes #49 correct by default rather than correct
when configured. Related: #49 (added the capability and the manual override), #194
(the other place a model's identity is not carried far enough to attribute
something).

### LINKED ISSUE #774: Show how full the context window is, beside what the turn cost
The chat says what a turn cost and how full the workspace is. It says nothing
about how full the *context* is — which is the ceiling a long run actually hits
first, and the one the user has no warning of until the answer degrades or the
provider refuses.

Split out of #49, which added the `compaction` capability. The number needed to
draw this is already computed there.

## What exists to build on

`pydantic-ai-harness` exports `ReportContextUsage`, a zero-LLM capability that
reuses the same estimator and the same resolved window as the compaction
strategies and only observes — it never edits history. Each reading carries
`used_tokens`, `window_tokens` and `resolved`, where `resolved` is `False` when the
window is the fallback rather than the model's real one. `on_usage` may be a
coroutine, so a gauge that pushes over a socket needs no sync bridge.

On our side the shape is already there. `UsageStrip`
(`frontend/src/components/chat/usage-strip.tsx`) draws tokens, cost, budget share
and a workspace bar from one `TurnUsage`; `UsageReport`
(`backend/app/services/usage_report.py:82`) is what fills it, and its
`sandbox.percent` is the precedent for "a percentage that is `None` when nothing
was measured, and drawn as nothing rather than as zero".

## What to do

- Bind `ReportContextUsage` for a run whose surface streams, ordered **after** the
  compaction capability so the reading reflects the corrected history rather than
  what triggered the compaction.
- Carry the reading on `UsageReport` / `TurnUsage` as a percent plus whether the
  window was resolved.
- Draw it in `UsageStrip` beside the workspace bar, amber and then red on the same
  thresholds. When `resolved` is `False` the percentage is a guess and has to say
  so — the alternative is a confident 41% against a window nobody knows.

## What to decide before building

Whether an agent with no `compaction` binding gets the gauge. Arguing for yes: the
warning is most useful to exactly the agent that will *not* compact. Arguing for
no: a reading whose denominator is the 200,000-token fallback is a number that
looks like measurement. #773 is what would settle it by making the denominator
real.

A long run shows the gauge climbing and it drops after a compaction pass; an
unresolvable model shows the reading marked as an estimate rather than as a fact.
A component test for both states, per the rule that an empty state and a failed
one must not be the same pixels.

Related: #49 (added the capability this reads from), #773 (makes the denominator
trustworthy), #772 (the other half of "what is this conversation costing me").

### LINKED ISSUE #776: Per-message cost is web-chat only, so a channel or API thread totals nothing
A conversation now reports what it has cost, and on three of four surfaces
that total is `null`.

Found while building the total in #772.

## What breaks

`TranscriptService._answer` (`backend/app/services/transcript.py:400`) writes the
assistant message with `model_name`, `agent_id`, `agent_version_id` and `run_id` —
and none of `input_tokens`, `output_tokens`, `cost_usd` or `cost_is_partial`. It is
the write path for channels, the public API and the embedded widget.

Only web chat fills them, through `persist_assistant_turn`
(`backend/app/services/agent.py:307`), which is handed a `UsageReport`.

So on a Slack thread, an API-driven conversation or a widget session:

- every message shows no cost, and
- `conversation_cost` counts nothing, so the strip reports no thread total at all.

The numbers exist — `agent_runs` carries all four and the message row already
carries `run_id`.

## What makes it more than a copy-paste

**A resumed run would double-count.** `record` is called once per execution, and a
run that parked and was resumed executes twice — so two assistant rows would each
be stamped with the run row's *cumulative* totals, and the thread total would count
the same money twice. The web chat avoids this on the client (`use-chat.ts` treats a
resumed segment as superseding the parked one, not adding to it); the database does
not.

So the fix needs to decide what a message-level cost means when one run writes two
of them: the delta since the previous row for that run, the whole run on the last
row only, or a `run_id`-aware aggregate that counts each run once. The third is
probably right and is the one that also fixes the web-chat path, which has the same
shape.

A Slack thread of three answers reports a non-null total, and a conversation whose
run parked and resumed reports that run's cost once rather than twice. An
integration test for the second, against real rows.

Related: #772 (added the total and the partial flag), #49 (the branch this came out
of).

### LINKED ISSUE #781: A turn cancelled by a closed tab is unrecoverable, and its spend is not
Closing the tab while a turn is running cancels it, and there is no way back to
it: the work is gone, the money is not.

Found while adding the compaction progress notice in #49 — a summary of a long
history takes seconds, and the tab closing during one is the case that made this
visible.

## What happens today

`AgentSession.shutdown` is called from the WebSocket route's `finally`
(`backend/app/api/routes/v1/agent.py:43`) and cancels the in-flight turn task.
`AgentRunnerService._run` catches the `CancelledError`, records the run
`cancelled` with what it spent, and re-raises
(`backend/app/services/agent_runner.py:3229`).

`resume` refuses anything that is not parked
(`backend/app/services/agent_runner.py:2884`), so a cancelled run cannot be
continued. The next message starts a fresh run, and whatever the cancelled one had
done — a summarised history, half an answer, a tool result — is discarded.

**The spend is kept, correctly.** Tokens the provider billed were billed;
`MeteredCompaction` books in a `finally` precisely so an abandoned summary is not
free. So a turn cancelled two seconds into a summary costs a summary and produces
nothing, and the next message pays for the same summary again.

## Why this is a feature and not a fix

A parked run can be resumed because it *finished* — `result.all_messages()` is a
coherent history and `paused_state` stores it
(`agent_runner.py:3216`). A cancelled run has no result: the cancellation
arrives inside `Agent.iter`, mid-request or mid-tool, and there is no
corresponding "the history as it stood" to write down without reaching into the
iteration's partial state and deciding what a half-executed tool call means.

There is also a product question this cannot answer on its own. Somebody who
closed the tab may have abandoned the turn deliberately. Silently continuing it on
their next visit — spending more money, acting on a tool call they walked away
from — is not obviously the behaviour they wanted, and `stop` is the same code
path as a dropped socket. So it needs to distinguish "I stopped this" from "my
laptop slept", which nothing currently records.

## What would count as done

Pick one and say which:

- **Nothing, deliberately.** Cancellation is correct, the cost is honest, and the
  waste is bounded by one turn. Close this with the reasoning written down so the
  next person does not re-derive it.
- **Make the waste visible.** A cancelled run says in the transcript that it was
  stopped and what it had spent, so the money is attributable to something a
  reader can see rather than appearing as a run with no answer.
- **Resume it.** Store enough of a cancelled run's state to continue, decide what
  a half-executed tool call means on replay, and separate a deliberate `stop` from
  a lost socket so only the second is offered a resume.

My own view is the second: the first is already true and only needs saying, and
the third buys back one turn of work at the cost of a new state machine over
partial runs.

## Worth knowing either way

The discarded summary is **not specific to cancellation**. Between turns the
history is rebuilt from the transcript as text
(`app.services.agent.build_message_history`), so a compaction never survives a
turn boundary at all — a successful turn re-summarises on the next message too.
Fixing that is the cross-turn replay problem, which is its own piece of work and
the reason #49's capability is scoped to one run.

Related: #49 (added compaction and the progress notice), #771 (the history that is
replayed as text), #776 (per-message cost on the surfaces that do not record it).

DEENUU1:
Ports the `pydantic-ai-harness` compaction strategies into the capability
registry — metered, with a real context window behind them — and closes the four
issues that came out of scoping it.

## What changed

**`build(deps)`** — `pydantic-ai-harness>=0.21.0`, which requires
`pydantic-ai-slim>=2.28.0` where this repo floored at 2.26.0. The lock resolves
`pydantic-ai-slim` and `pydantic-graph` 2.26.0 → 2.31.0. Its own commit, green on
the backend suite before any capability code existed, so a red job would be
attributable to the bump rather than to the feature.

**`feat(capabilities): add compaction, metered`** — `compaction`, registered with
`tools=()`, `side_effecting=False`, in `load_builtins()`. Four strategies behind
one `strategy` enum. Two things the port had to add to the library:

- **A budget that can see the summary.** `SummarizingCompaction` writes its
  summary through an `Agent` it constructs itself, so that request never passes
  `BudgetGuard.wrap_model_request` — the guard is a capability on *our* agent, not
  on the one the strategy builds. The tokens land in `ctx.usage` and would land
  nowhere else: the run under-reports its cost and no cap can stop a compaction
  loop. That is #16 wearing a different hat. `MeteredCompaction` measures the
  run's usage across the hook and books the difference through
  `record_ambient_usage`.
- **A scope somebody can reason about.** This reaches the messages of one run,
  because between turns the history is rebuilt from the transcript as text. That
  is the right scope — the history worth compacting is the hundred-step tool loop
  `DEFAULT_MAX_STEPS` allows — and it is documented rather than left to be
  discovered.

**`fix(chat): read a conversation's history from the transcript, not the socket`**
— #771. `AgentSession.conversation_history` was a list on the WebSocket, appended
to as turns went by and loaded from nowhere, so a conversation the socket did not
itself produce had no history at all: a page reload, a second tab or a dropped
connection was enough, and the agent answered the follow-up as though the thread
had started with it. The person was looking at the whole conversation on screen
while the model was handed an empty one. The list is gone; the history is read per
turn from the same rows the transcript renders, excluding the prompt row written
before the run. The tail-window offset moved into
`conversation_repo.get_recent_messages`, which is now the one place that turns
"the last N turns" into a `COUNT` and an offset — it had already been wrong from
opposite ends in #39 and #638, and this would have been a fourth copy.

**`feat(models): record what context window a model profile's model accepts`** —
#773. The trigger is a fraction of the window, and the only thing it could resolve
one from was the bundled price snapshot, which is wrong here in the direction that
breaks a run: it records 1,000,000 for `anthropic:claude-sonnet-4-5` against a
real 200,000, and a spec with fallbacks builds a `FallbackModel` whose composite
id resolves to nothing at all. The number was already in hand and thrown away —
ten providers publish it in the listing the model picker reads. `model_profiles`
grows a `context_length`, filled once at creation, carried on `ModelRequestSpec`,
and handed to capabilities as a resource rather than fetched by one.

**`fix(builder): show a capability's defaults as values, and say what a choice
does`** — a generated configuration form drew defaults as placeholder grey and
enum values raw, so `compaction` arrived as a row of empty boxes over a select
reading "Not set", offering `clear_tool_results` and `sliding_window` as choices
somebody makes by guessing. Defaults are now drawn as values (nothing is stored,
so the field keeps tracking a default that later changes in code), and a schema
may carry `x-enum-labels` saying what each value does.

**`feat(chat): mark a partial cost, total the thread, and show the context
fill`** — #772 and #774.

- A message showed a cost with no sign it was partial. `messages` grows
  `cost_is_partial`, the flag travels from the ledger to the stored turn and the
  live frame, and every figure a partial cost produces is drawn `≥ $x`.
- A conversation showed no total. Added up server-side, beside the page rather
  than from it, because the transcript pages at 100.
- Nothing said how full the context window was. Every agent now carries a gauge,
  and not only one that compacts — the warning matters most to the agent that
  will not, because that is the one the provider refuses.

**The gauge took three passes to get right, and each pass was a real defect.**

- *It was three times short.* The harness ships `ReportContextUsage`, which
  estimates the size of the history about to be sent — and a character heuristic
  cannot see the tool definitions, which the provider bills on every request. A
  real conversation on this branch measured 1,688 tokens against the 5,007 the
  provider charged for, on every turn. The reading is now `input_tokens` off each
  response: what the request actually occupied, instructions and schemas
  included, as the provider counted it.
- *It vanished on reload*, which is the one moment somebody asks how close they
  are to the ceiling. `messages.context_used_tokens` stores it, the same way the
  cost columns do.
- *The share was divided by the wrong model.* The chat lets somebody switch model
  between turns, and a 500,000-token history is half of a 1M window and 390% of a
  128K one — the second being a request the provider refuses outright, where a
  share frozen with the reading would still read "50%". So only the count is
  stored and the denominator is resolved from the model selected *now*: the
  picker's override, else `AgentRead.context_window_tokens`, which the listing
  reads off the profile the frozen spec names with the pricing registry behind
  it. Where neither can say, no share is drawn — one against an assumed window is
  a guess presented as a measurement, and it errs toward the ceiling.

That switch is also what compaction is for, and it already worked: the trigger is
a fraction resolved *per request* against the model the request is going to. The
same 250-message history passes untouched on a 1M window and is cut to 11
messages on a 128K one, before the request leaves. It is pinned by a test now,
because it is the whole reason the trigger is not a constant.

## Decisions worth reviewing

- **The default strategy is `summarize` at 0.9 of the window.** Summarising is the
  only strategy that keeps what the older turns *said*; the zero-LLM ones are
  cheaper because they throw information away, and an agent that silently forgets
  what it was told mid-run is a worse failure than a summary nobody asked for.
  `tiered` is one binding away for an agent whose runs are dominated by large tool
  results. **This costs one model call per run that crosses the threshold** — the
  trade is deliberate and it is the main thing to disagree with here.
- **The builder always returns a capability.** `thinking` is the precedent, and a
  builder answering `None` would drop out of
  `test_no_capability_escapes_the_drift_check`.
- **The config is flat scalars and one enum**, because the Builder renders string,
  number, boolean and enum. That is why the tiers are chosen by `strategy` rather
  than composed by the author.
- **The metering records, it does not refuse.** `BudgetGuard` refuses in
  `wrap_model_request`, which runs after this hook, so a compaction that crosses a
  cap stops the run on the request after it.
- **The context gauge is on every agent**, and costs one integer read per
  response now that it is the provider's own count rather than an estimate. The
  alternative is a warning that is missing exactly where it is needed.

## How it was verified

`make check` green: `make test` at the 100% coverage gate on the platform layer,
`make test-frontend-cov` at 97.52% branches against a 97.5% floor, `make db-check`
clean, and both migrations cycled forwards and back.

The tests worth naming, because they are the ones nobody writes by accident:

- a summary is booked against the run's ledger; a zero-LLM pass books nothing;
  tokens spent before a strategy raises are still booked; a model the registry
  cannot price is booked *unpriced* rather than dropped;
- **a resume regression** — a parked run's history is replayed with the approved
  call's result keyed on a tool-call id, so compaction must never leave a tool
  return without the call it answers. A 91-message tool history is compacted down
  and asserted to have no orphaned returns;
- the stored thread is what the model is told, and the turn being answered is not
  handed back as history as well as prompt;
- the conversation total against **real rows**, because `bool_or` over nulls,
  `COUNT` over nulls and a sum that ignores them are three things a mocked session
  cannot show.

**How far that reaches.** No test drives a real summarizing pass against a
provider, so what is asserted is that whatever lands in `ctx.usage` during the
hook is booked — not that the harness spends what it says. The strategies
themselves are the library's to test.

## Deliberately not done

- **No backfill of `cost_is_partial`** for messages already written. They read
  null, which is the honest answer, and their run rows still carry the flag.
- **No backfill of `context_used_tokens`** either: a turn that has already run
  cannot be re-measured, so the gauge lights up on the next message rather than
  retroactively.
- **`model_profiles.context_length` was backfilled by a one-off command that was
  then deleted rather than shipped.** There is one deployment, it has been run
  against it, and a maintenance command nobody will run twice is a file that goes
  stale in place. A profile added from here on records its window at creation.
## Then it turned out compaction never fired at all

Two settings were tried against a real agent - 0.5 of 5,000 tokens, then 0.4 of
10,000 - and neither produced a summary while the chat's own gauge sat at 77%.

**The trigger was counting characters.** The harness estimator anchors on the most
recent response carrying provider usage: that request's `input_tokens` counted the
instructions, every tool schema and every prior message, so anchored it reports the
provider's number and estimates only what came after. Replayed from the transcript
as bare text there was nothing to anchor on, and the heuristic read 9 tokens where
the provider had charged for 3,859 - the gauge and the trigger describing one
ceiling three orders apart.

An answer is now replayed with the size of the request it came out of. That is
`context_used_tokens` and deliberately not the `input_tokens` beside it, which is a
run's whole bill: anchored on a sum, a hundred-step tool loop would replay as
several times its own size and compact a history that fits.

**And a summary is now kept.** Compaction rewrites one run's messages, and the
thread between turns was rebuilt from the transcript - so the summary died at the
turn boundary and the next turn bought another over a history one turn longer. In a
real conversation, turns 4 and 5 each paid for a summary of the same five messages
and the second announced itself as summarising nine. `conversations.summary_messages`
holds the history as the model last saw it and `summary_ordinal` how far it reaches;
the next turn starts there and replays only what has been said since. Every surface
reads it, because a channel thread never rolls over and is where a per-turn summary
costs the most.

This is the cross-turn replay #781 flagged as its own piece of work. It landed here
because without it the capability this PR adds buys the same summary every turn.

**One setting cannot work, and now says so.** When the instructions and tool schemas
alone are past the trigger, no summary can get under them - so a paid strategy is
skipped rather than bought on every request for ever, and the chat says why. That
refusal rests on a number a *response* produces, so it is recorded on the
conversation (`overhead_tokens`) and a run starts from it; left per-run it was
`None` for the whole of a one-request turn and the refusal never fired at all. The
strategies that call no model keep running - there is no loop to prevent, and a
smaller request still beats an unbounded one.

The overhead correction in cca4e4ae is removed rather than kept: with an anchor it
subtracted the same schemas a second time.

## Still true after all that

- **A summary written by one agent replays into another** if the next turn of a
  thread goes to a different agent. No invalidation; named rather than fixed.
- **Channels get no compaction progress frames.** `on_compaction` is wired by web
  chat and the widget, so a bot goes quiet for the length of a summary.

Closes #49
Closes #771
Closes #772
Closes #773
Closes #774
Closes #776
Closes #781


DEENUU1 (commented):
Read this end to end — the backend is in very good shape. The two-estimator split (unanchored `estimate_token_count` for overhead vs anchored `estimate_context_tokens` for the trigger) is subtle and correct, the `attributed_to_run` difference is the right way to stop a parked+resumed run double-counting its cost, and the unscoped read in `model_history` is genuinely safe because `persist_user_turn` resolved the id a moment earlier — good that the docstring says so.

One thing to fix on the frontend: `compacting` and `compactionImpossible` aren't cleared when you switch conversations, so they get drawn under the wrong thread — the same class of staleness the layout effect just above them was written to prevent. Everything else is comments and praise.

frontend/src/hooks/use-chat.ts
```diff
@@ -138,6 +143,45 @@ export function useChat(options: UseChatOptions = {}) {
...
   // on its own `subagent_complete`.
   const [delegations, setDelegations] = useState<Delegation[]>([]);
+  // The summary in flight, or `null`. Cleared by the finishing frame and by the
+  // end of the turn: a `complete` that arrived without one - a run that failed
+  // between the two - would otherwise leave the notice up until the next message.
+  const [compacting, setCompacting] = useState<Compaction | null>(null);
+  // The window that cannot work, or `null`. Held apart from `compacting` because
+  // it outlives a turn: it describes what is configured, and clearing it when the
+  // turn ends would flash the one message that explains why nothing happened.
+  const [compactionImpossible, setCompactionImpossible] = useState<Compaction | null>(null);
```
DEENUU1, line 153: These two survive a conversation switch. The layout effect at line 757 already clears `delegations`, `pendingApproval` and `pendingQuestions` for exactly this reason — "the previous thread's ... drawn under the new one" — but it doesn't touch these, and `compactionImpossible` is deliberately never cleared by a `complete`, so it lingers.

Two concrete cases: switch away from a thread mid-summary and "Summarising the conversation…" shows over the thread you just opened until the old turn's `complete` lands (which, on a long summary, is exactly when it's on screen). And the "Context management cannot run…" warning describes one agent's overhead against its window — open a conversation on an agent that has room and it's still there, telling you the wrong thing.

Add them to the reset at line 757:
```ts
    setDelegations([]);
    setPendingApproval(null);
    setPendingQuestions(null);
    setCompacting(null);
    setCompactionImpossible(null);
    approvalOfferedForRef.current = new Set();
```

DEENUU1: Fixed in 86ec6481. Added `setCompacting(null)` and `setCompactionImpossible(null)` to the conversation-switch reset in the layout effect (beside the delegations/approval/questions clears), so neither frame is drawn under a thread it does not belong to. Regression test `leaves neither frame behind when another conversation is opened` in `use-chat.test.tsx` covers both cases.

backend/alembic/versions/0030_profile_context_length.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "0030_profile_context_length"
```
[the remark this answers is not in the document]
DEENUU1: False positive. `revision`, `down_revision`, `branch_labels` and `depends_on` are the module-level attributes Alembic reads by name to build the migration graph — every version file declares all four, and vulture is configured to ignore them repo-wide. (These migrations were also renumbered onto context's 0030 in the merge, so the chain is a single linear head.) Resolving.

Issue #175, opened 2026-08-05
A specialist create_agent kept does not survive the approval park that created it


DEENUU1:
## What breaks

`create_agent` registers a model-invented specialist so it can be addressed by
name with `task`. The registry it writes into is
`subagents_pydantic_ai.registry.DynamicAgentRegistry`, created inside
`SubAgentToolset.__init__` — that is, once per `build_agent` call, not once per
run. A run that parks for an approval is built again when it is continued, so
every registration made before the park is gone after it, while the replayed
transcript still carries the library's own
`Agent 'x' created successfully. ... Use task(description, 'x') to delegate tasks.`

The model then calls `task(..., 'x')` and reads
`Error: Unknown subagent 'x'. Available: <the pinned ones>` — recoverable, but it
was told otherwise.

## The sequence

1. An agent with `subagents` bound and `allow_dynamic: true`.
2. The model calls `create_agent('summariser', ...)`. It is gated by default, so
   the run parks; approving it resumes, and the tool registers into the *resumed*
   run's registry. `task(..., 'summariser')` works. So far so good.
3. Anything parks the run again — a second `create_agent`, a `delegate`, or a
   gated tool inside a delegate.
4. Approve, resume. `task(..., 'summariser')` now answers "Unknown subagent".

The same thing happens across conversation turns: each turn is its own
`prepare` -> `build_agent`, so a specialist created in one reply is unknown in the
next while the model can still see it in the history.

`backend/app/agents/capabilities/subagents/_capability.py` builds the capability
per run; `backend/app/services/agent_runner.py` `resume` -> `_assemble` ->
`build_agent` is where the second registry appears.

## Caused by #40 phase 6, which is what offered the tool at all

Not a pre-existing bug — `create_agent` was declared and not offered until the
dynamic entry points landed. What that change did about it is narrow the promise
rather than fix it: `create_agent`'s description now says the specialist lasts for
this reply and to create it again if `task` says it is unknown, and
`DynamicSpecialists`' docstring says the registry is per built agent rather than
per run. That keeps the transcript honest, which is the part that mattered; it
does not make the tool behave as its name suggests.

## Two ways to fix it

- **Carry the registrations in `paused_state`.** `PausedRunState` already holds a
  parked delegate's whole conversation for exactly this reason, so a list of
  `(name, description, instructions, model)` is the same shape of state, and
  re-registering on resume goes through the same `_specialist_factory` — so the
[...]

Issue #40, opened 2026-08-01
Add subagents-pydantic-ai


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680a6af12f86c96b47a60) · Priority: **High** · Category: Development

Add [`subagents-pydantic-ai`](https://github.com/vstorm-co/subagents-pydantic-ai) (our own OSS, PyPI `subagents-pydantic-ai`) as a capability, so an agent can delegate to named specialists instead of doing everything in one context.

The library's own API is already capability-shaped, which makes the integration small:

```python
from subagents_pydantic_ai import SubAgentCapability, SubAgentConfig

SubAgentCapability(subagents=[
    SubAgentConfig(name="researcher", description="...", instructions="..."),
    SubAgentConfig(name="writer",     description="...", instructions="..."),
])
```

It supports sync (blocking), async (background) and auto mode selection, plus token tracking and cancellation.

## The interesting question is not the wiring

It is **what a subagent *is* in a product where "an agent is data, not code."**

`SubAgentConfig` takes a name, a description and free-text instructions. That is a second, parallel way to define an agent — one that lives inside another agent's spec, is not versioned on publish, is not exportable as YAML, has no capabilities of its own, and is invisible to the permission model. Ship it naively and this platform has two incompatible notions of "agent".

Two coherent designs, pick one and write it down:

**A. Inline specialists.** `SubAgentConfig` becomes a field on the capability's config. Simple, matches the library, and a subagent is just a prompt fragment — no tools, no budget of its own, no separate history. Good enough for "researcher / writer" splits.

**B. Subagents are published agents.** The binding holds `agent_ids`, and each delegate is a real `AgentSpec` from the registry — versioned, permission-checked at publish, with its own capabilities and its own model. Strictly better product, and much more work: you need `resolve_access(..., AGENTS_RUN)` per referenced agent at publish time (see #6, where exactly this check is missing for `skill_ids`), plus a recursion guard.

B is where this ends up. A is a reasonable first cut if it is labelled as one.

## Things that will bite

- **Cost accounting.** A subagent run spends money. `agent_runner` records cost per `AgentRun` row; a delegate's tokens must land on the parent run or be their own row, and either way the budget guard has to see them *before* the request. Today it will not — this is the same class as #16, where web chat's embedding cost is dropped because no ledger is active. Look at `metered_by` in `app/services/agent_runner.py:722`.
[...]

PR #778, opened 2026-08-15
feat(capabilities): add a planning capability

### LINKED ISSUE #47: Add Planning
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680f08b40dbf6aa18dc6a) · Priority: **Low** · Category: Development

**Reference:**
- https://github.com/pydantic/pydantic-ai-harness/pull/404

## Why it matters here

A planning capability lets an agent decompose a task before executing it, which is what makes long multi-step runs survivable. It overlaps conceptually with #40 (subagents) — planning decides *what* the steps are, subagents decide *who* does them.

## What will bite

- **Decide the relationship with #40 before building either.** Two capabilities that both restructure a run will fight if designed apart.
- Planning adds model round trips. A plan that costs three requests before the first real one triples the floor cost of a run — the budget guard must see them.
- A plan is state. If a run parks on an approval mid-plan, the plan has to survive in `paused_state` and be restored by `resume`.

## Done when

DEENUU1:
## What this adds

The **Planning** capability — a checklist the agent keeps for itself over a
multi-step run, ported from `pydantic-ai-harness`. `write_plan`/`read_plan` plus
granular step tools, and under `enable_subtasks` a dependency-aware mode
(`add_subtask`, `set_dependency`, `get_available_tasks`, a `blocked` status). The
current plan is surfaced back each turn as a cache-safe tail reminder behind a
`CachePoint`, so the prompt prefix stays cacheable and the plan never lands in the
system prompt.

Closes #47.

## The three "what will bite" from the issue

- **Relationship with #40 (subagents).** Orthogonal, by construction: planning is a
  toolset plus a reminder, delegation is a toolset plus a run wrapper. Neither
  restructures the other, so an agent may bind both. Documented in both READMEs and
  `docs/reference/capabilities.md`.
- **Cost.** Does not apply here, and the commit says why: the tools are local
  checklist edits with no model or embedding request behind them, so there is no
  ambient usage to meter. The model's own round trips to call them are already
  counted by the budget guard.
- **Plan is state → must survive an approval park.** Done. The runner owns the
  store: it seeds one from `PausedRunState.plan` on resume, injects it through
  `PLANNING_STORE_RESOURCE`, and reads it back in `finish` when the run parks — the
  same shape as re-seeding a delegation's registry from its frames (#175/#254). No
  migration: `paused_state` is already a JSONB column; `plan` is a new field in the
  blob with an empty default, so a run parked before this stays resumable.

## Shape

- The capability wraps the library rather than returning it bare (like `subagents`,
  unlike `thinking`): the nine tool descriptions are declared in `_capability.py` and
  handed to the library through `descriptions=`, so the model, the Builder's approval
  UI and the tests read one string.
- `side_effecting=False`, no scope — none of the tools acts on the world.
- Registry-only: no `spec.py` change, no `SPEC_VERSION` bump. A stored spec references
  `planning` by id like any other capability.
- Adds `pydantic-ai-harness>=0.21.0` and floors `pydantic-ai-slim` at 2.28.0, which
  the harness requires.

## Verified

- `make test` — the platform-layer 100% gate holds (both new modules and the runner
  changes covered); `test_planning.py` and the runner park/resume tests.
- `make lint` — ruff, ty, vulture, the guard scripts, eslint, tsc, the i18n guard and
  `messages/catalog.test.ts`.
- Reviewed my own diff and swept the capability, runner-park and frontend-catalog
  siblings, since `ai-review` is off (#311).

## Deliberately out of scope

- **A plan panel in the chat UI.** The checklist is surfaced to the model; a
  first-class plan view is a separate piece of work. Each tool has a `tool-catalog.ts`
  row and renders with the generic renderer for now.
- **A persistent (SQLite/Postgres/Redis) plan store.** The library ships them; a
  plan's lifetime here is one run (which may park and resume), not a durable record
  across conversations. A follow-up if a use case appears.
- **Delegate-level plan park-survival.** A delegate that binds planning gets a fresh
  per-run store; only the run's own plan is carried across a park. Noted in the README.


backend/app/agents/capabilities/planning/_capability.py
```diff
@@ -0,0 +1,192 @@
...
+
+    `store` is the run's, injected by the runner so the plan survives an approval
+    park; `None` lets the library keep a fresh in-memory plan per run, which is the
+    right default for a build with no run behind it.
+    """
+    return Planning(
+        enable_subtasks=enable_subtasks,
+        cache_ttl=cache_ttl,
+        store=store,
+        descriptions={
```
DEENUU1, line 193: The rationale for wrapping instead of returning the library bare is "the tool text is this repository's" — but `descriptions=` only covers the tool descriptions. The library's `get_instructions()` guidance (the `write_plan`/granular/subtask sentences) also lands in the system prompt every turn, and this build leaves it at the library default. So the strongest, most-read prompt the capability contributes is actually the un-owned one, which is a bit at odds with the whole reason we're wrapping.

It's not a bug — the default guidance is fine wording. But with a floor-only pin (`>=0.21.0`), a harness release that rewrites that guidance changes our system prompt silently, and nothing here asserts on it. Worth either pinning the guidance text explicitly (`guidance=`) so it's this repo's string like the descriptions are, or at least a test that the guidance the model sees is what we expect.

backend/tests/test_planning.py
```diff
@@ -0,0 +1,207 @@
...
+
+    async def test_a_resume_re_seeds_the_stored_plan(self):
+        items = [PlanItem(content="Write the migration").model_dump(mode="json")]
+
+        store = await open_plan_store(items)
+        seeded = await store.get_items()
+
+        assert [item.content for item in seeded] == ["Write the migration"]
+
+    async def test_the_snapshot_round_trips_through_a_park(self):
```
DEENUU1, line 226: `test_the_snapshot_round_trips_through_a_park` covers a flat plan (completed / in_progress). The state that's actually fragile across a park is subtasks mode — `depends_on`, `parent_id`, and a `blocked` step. It does round-trip (the fields are on `PlanItem`, so `model_dump`/`model_validate` carry them), but that's the cell of the matrix nothing asserts. A second case with a blocked step that depends on another would pin it:

```python
async def test_a_blocked_dependency_survives_a_park(self):
    original = InMemoryPlanStore()
    await original.set_items(
        [
            PlanItem(id="a", content="Migration", status=TaskStatus.in_progress),
            PlanItem(id="b", content="Backfill", status=TaskStatus.blocked, depends_on=["a"]),
        ]
    )
    reopened = await open_plan_store(await dump_plan(original))
    seeded = {item.id: item for item in await reopened.get_items()}
    assert seeded["b"].status == TaskStatus.blocked
    assert seeded["b"].depends_on == ["a"]
```

DEENUU1 (commented):
This is clean, and the reasoning in the docstrings does most of the reviewer's work for me. I traced the three things the issue said would bite and they all hold: the reminder actually reaches the model (the capability goes into `capabilities=` in `factory.py`, so `wrap_model_request` runs — it's not just a harvested toolset), the store is owned by the runner and threaded through the one `PreparedRun` in `_assemble`, and `_delegate_resources` deliberately doesn't pass `PLANNING_STORE_RESOURCE` on, so a delegate gets its own fresh store exactly as the README claims. Park round-trip is tested and the full `PlanItem` (deps, `blocked`) survives it.

Two non-blocking notes inline. One on a prompt the wrapper *doesn't* own, one on a test that could close the subtask matrix.

(The `github-code-quality` bot's "PLANNING_TOOLS unused" is a false positive — it's the `tools=` argument to `@register` in `__init__.py` and is asserted in `test_planning.py`. Resolve with a pointer.)

backend/app/agents/capabilities/planning/_capability.py
DEENUU1: Fixed in 45ab6777. `guidance=` is supported (a `str | None` on the library's `Planning`), so the guidance is now this repo's own string like the descriptions are: `WRITE_PLAN_GUIDANCE`/`GRANULAR_GUIDANCE`/`SUBTASK_GUIDANCE` module constants (byte-faithful to the current default, verified via `get_instructions()`), assembled by `_guidance(*, subtasks)`. `TestGuidance` asserts the guidance the model sees equals the pinned string and that the subtask sentence appears only in subtasks mode — a harness release rewriting its default can no longer change our system prompt silently.

backend/tests/test_planning.py
DEENUU1: Fixed in 45ab6777. Added `test_a_blocked_dependency_survives_a_park` under `TestParkSurvival`: a `blocked` step with `depends_on` round-trips through `dump_plan`/`open_plan_store` with status and deps intact — the subtasks cell of the matrix the flat test left open.

backend/app/agents/capabilities/planning/_capability.py
```diff
@@ -0,0 +1,192 @@
...
+    "remove_task": REMOVE_TASK_DESCRIPTION,
+}
+_SUBTASK_DESCRIPTIONS: dict[str, str] = {
+    "add_subtask": ADD_SUBTASK_DESCRIPTION,
+    "set_dependency": SET_DEPENDENCY_DESCRIPTION,
+    "get_available_tasks": GET_AVAILABLE_TASKS_DESCRIPTION,
+}
+_DESCRIPTIONS: dict[str, str] = {**_CORE_DESCRIPTIONS, **_SUBTASK_DESCRIPTIONS}
+
+PLANNING_TOOLS: tuple[CapabilityToolInfo, ...] = tuple(
```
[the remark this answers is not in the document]
DEENUU1: False positive: `PLANNING_TOOLS` is the `tools=` argument to `@register` in `__init__.py` and is asserted in `test_planning.py`. Resolving.

PR #779, opened 2026-08-15
feat(capabilities): add guardrails that redact or block a run

### LINKED ISSUE #46: Add Guardrails
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576807fa8b2de3b5b3e9cd3) · Priority: **Low** · Category: Development

**Reference:**
- https://github.com/pydantic/pydantic-ai-harness/pull/478
- https://github.com/pydantic/pydantic-ai-harness/pull/474
- https://github.com/pydantic/pydantic-ai-harness/pull/470

## Why it matters here

Guardrails are the one capability on this list that is a **governance** feature, not a convenience. This platform's stated value is "mostly in what it refuses" — a guardrail is that sentence made executable: an input or output check that can stop a run before it does damage.

That makes it the highest-value item in W3 despite its Low priority, and it deserves a harder look than the others.

## What will bite

- **Where does a tripped guardrail land?** A refusal is a run outcome. Decide whether it is a new `RunStatus` (like `BUDGET_EXCEEDED`, which was given its own status precisely so an operator filtering for problems does not wade through it) or a `FAILED` with a distinguishable error. A guardrail trip is the platform working, not a malfunction — model it that way.
- **Cost of the check itself.** If a guardrail calls a model, that spend must reach the ledger *before* the next request. See #16.
- **Interaction with the approval gate.** Both can stop a run. Decide the order and write it down.
- Three PRs are referenced, not one — read all three before choosing a shape.

## Done when

DEENUU1:
Ports the `pydantic-ai-harness` guardrails into the capability registry as a
single `guardrails` capability, and gives a tripped guardrail its own run
outcome. Closes #46.

## What it does

A guardrail inspects the text flowing through a run at three edges and either
**redacts** a match or **blocks** the run:

| Edge | Reads | Redact | Block |
|---|---|---|---|
| input | the user's prompt | secrets, PII | keyword list |
| output | the agent's answer | secrets, PII | keyword list |
| tool result | what a tool returned, before the model reads it | secrets, PII | keyword list |

Redactors cover API keys/tokens/JWTs/PEM blocks and email/IBAN (mod-97)/card
(Luhn)/US-SSN. Tool-result screening is the headline: it is the only guard on
untrusted content entering the loop — a fetched page, a file, an MCP response —
where a prompt-injection payload would otherwise reach the model unread.

## Design decisions

- **Config, not callables.** An agent is data, so the harness's `guard=<fn>`
  cannot cross the wire. The config selects and parameterises the ready-made
  detectors: flat toggles + a delimited keyword string per edge, which is all the
  Builder's generated form renders (it has no array control). Shaped like the
  compaction capability — a non-tool, harness-wrapping capability.
- **A block is a run outcome, not a graceful answer.** The harness makes a block
  graceful (an input block completes the run with a refusal as the answer). For
  governance that is exactly wrong — a refusal should be visible. So the detector
  chain raises `GuardrailBlocked`, and the runner maps it to a new
  `RunStatus.GUARDRAIL_BLOCKED`, beside `BUDGET_EXCEEDED` and for the same reason:
  the platform working, not a malfunction. Verified that the raise escapes
  `agent.run()` cleanly from all three edges against a real `FunctionModel` run.
- **What was deliberately not ported.** The tool `approve` verdict duplicates the
  approval gate (which already parks a run per tool for a human) — dropping it also
  removes the guardrail-vs-approval ordering question entirely. Tool arguments have
  no text detector, so they are not an edge. Every detector is a pure function, so
  budget metering (#16) is not engaged; a future model-calling detector must book
  spend through `record_ambient_usage` first.
- **No migration.** `agent_runs.status` is a plain `String(24)` with no check
  constraint, so the new value stores as-is and `parse_csv`/the filter pick it up.
- **Frontend.** The new status is wired through the type union, the badge (warning
  tone, like a budget stop), its label, the delegation map, and the outcomes donut
  (folded into `other` — the ramp is five colours by design). Left alone on
  purpose: the "Recent failures" widget and the run-history "Problems" preset — a
  governance refusal is not a failure, and it stays visible and filterable under
  its own status in run history.

## Verification

- Backend: 100% line + branch coverage on the new module; the runner maps
  `GuardrailBlocked` → `GUARDRAIL_BLOCKED` without re-raising; registry drift and
  layout tests accept a toolless capability that builds `None` on empty config.
- `make lint-backend`, `make lint-frontend`, the frontend coverage gate (100%
  statements/functions/lines, 97.52% branches) and a strict docs build all pass
  locally. The full `make test` backend gate was still running when this was
  opened — CI runs it.
- Docs updated: `docs/reference/capabilities.md`, `docs/concepts.md`, and a design
  note in `docs/plans/`.

### DIFF
--- a/backend/app/services/agent_runner.py
+++ b/backend/app/services/agent_runner.py
@@ -91,6 +91,7 @@ from app.agents.capabilities.channel_tools import (
     CHANNEL_TOOLS_CAPABILITY_ID,
     ChannelDirectory,
 )
+from app.agents.capabilities.guardrails import GuardrailBlocked
 from app.agents.capabilities.sandbox import WORKSPACE_BACKEND_RESOURCE, WorkspaceIdentity
 from app.agents.capabilities.sandbox._identity import SessionScope
 from app.agents.capabilities.subagents import SubagentsConfig, acting_delegate
@@ -3240,6 +3241,13 @@ class AgentRunnerService:
             error = str(exc)
             budget_scope = exc.scope
             logger.info("Run %s stopped by budget: %s", prepared.run.id, exc)
+        except GuardrailBlocked as exc:
+            # A refusal, not a malfunction - its own status for the same reason
+            # `BUDGET_EXCEEDED` is. The message names the edge and the refusal, not
+            # the content that tripped it, so it is safe to store on the row.
+            status = RunStatus.GUARDRAIL_BLOCKED
+            error = str(exc)
+            logger.info("Run %s blocked by a %s guardrail", prepared.run.id, exc.edge)
         except Exception as exc:
             error = run_failure_summary(exc)
             logger.exception("Agent run %s failed", prepared.run.id)
--- a/frontend/src/components/dashboard/widgets/outcomes.tsx
+++ b/frontend/src/components/dashboard/widgets/outcomes.tsx
@@ -51,7 +51,11 @@ export function OutcomesWidget({ title, hint, period, seeAll, options }: Dashboa
           const failed = of("failed");
           const budget = of("budget_exceeded");
           const awaiting = of("awaiting_approval");
-          const other = of("running") + of("cancelled");
+          // `guardrail_blocked` joins `other`: the ramp is five colours by
+          // design and the four attention/terminal statuses spend them, so a
+          // sixth is "other" (see `seriesColor`). Without it a blocked run sits
+          // in `total` behind no arc, and the ring stops summing to the whole.
+          const other = of("running") + of("cancelled") + of("guardrail_blocked");
           const attention = failed + budget;
           // The order is the ramp's, and it is fixed by status rather than by
           // which statuses this window happens to hold: "failed" is the same
--- a/backend/app/agents/capabilities/_registry.py
+++ b/backend/app/agents/capabilities/_registry.py
@@ -596,6 +596,7 @@ def load_builtins() -> None:
         charts,
         clock,
         code_execution,
+        guardrails,
         knowledge,
         sandbox,
         skills,
--- a/backend/app/agents/capabilities/guardrails/README.md
+++ b/backend/app/agents/capabilities/guardrails/README.md
@@ -1,0 +1,63 @@
+# Guardrails
+
+An input/output/tool-result check that can **redact** content or **stop** a run.
+This is the one capability that is governance rather than convenience: the
+platform's value is "mostly in what it refuses", and a guardrail is that sentence
+made executable.
+
+## What it does
+
+Three edges, each configured independently:
+
+- **input** — the user's prompt, before the first model request.
+- **output** — the agent's final answer.
+- **tool result** — what a tool returned, before the model reads it. This is the
+  only guard on untrusted content entering the loop: a fetched page, a file, an MCP
+  server's response.
+
+On each edge, two kinds of check, drawn from `pydantic-ai-harness`'s ready-made
+detectors:
+
+- **redact** secrets (API keys, tokens, JWTs, PEM blocks) and/or personal data
+  (email, IBAN with mod-97, card with Luhn, US SSN). A match is rewritten in place
+  and the run carries on — an agent that quoted a key back has still done the work.
+- **block** on a keyword list. A match ends the run.
+
+## Why a block *stops* the run
+
+The harness makes a `block` graceful: an input block substitutes a refusal and the
+run *completes* with it, a tool block replaces the result and the model continues.
+Here a block instead raises `GuardrailBlocked`, which the runner maps to
+`RunStatus.GUARDRAIL_BLOCKED` — its own status, beside `BUDGET_EXCEEDED`, because a
+trip is the platform working and an operator filtering for problems should not wade
+through it. A refusal that reads like a normal completed answer is not a refusal
+anyone can find later.
+
+Redaction keeps the harness semantics: `replace` rewrites and the run finishes.
+
+## What it deliberately does not do
+
+- **No `guard=<callable>`.** An agent is data, not code. The config selects and
+  parameterises detectors; it cannot carry a Python function. A detector the config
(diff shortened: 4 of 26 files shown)


backend/app/services/agent_runner.py
```diff
@@ -3240,6 +3241,13 @@ async def _run(
             error = str(exc)
             budget_scope = exc.scope
             logger.info("Run %s stopped by budget: %s", prepared.run.id, exc)
+        except GuardrailBlocked as exc:
```
DEENUU1, line 3244: This clause is correct, but it's the **only** place `GuardrailBlocked` is caught, and the streaming web chat doesn't come through `_run`.

`ChatAgentRunner.run` (`agent_chat.py:382`) has its own settle try/except: it catches `BudgetExceeded`, and everything else - including a re-raised `GuardrailBlocked` - falls into `except Exception` -> `status = RunStatus.FAILED`, `error = run_failure_summary(exc)`, re-raise. Then `agent_session.py:362` catches `(AppException, BudgetExceeded)` and sends the plain reason, but `GuardrailBlocked` hits the generic `except Exception` below it -> `logger.exception(...)` (a traceback) and a generic "turn failed" frame to the user.

So on the primary surface a block is recorded as `failed` (not `guardrail_blocked`), logged like a crash, and the user sees a generic failure instead of the guard's safe message - the opposite of "an operator filtering for problems should be able to find it".

Mirror the budget handling in `agent_chat.py`, beside the `BudgetExceeded` clause:

```python
except GuardrailBlocked as exc:
    status = RunStatus.GUARDRAIL_BLOCKED
    error = str(exc)
    logger.info("Chat run %s blocked by a %s guardrail", prepared.run.id, exc.edge)
    raise
```

and add `GuardrailBlocked` to the `except (AppException, BudgetExceeded)` tuple in `agent_session.py:362` so the visitor gets the plain reason and it isn't logged as a crash. Worth a chat-stream regression test too - the runner test only exercises `execute`, so nothing here caught it.

frontend/src/components/dashboard/widgets/outcomes.tsx
```diff
@@ -51,7 +51,11 @@ export function OutcomesWidget({ title, hint, period, seeAll, options }: Dashboa
           const failed = of("failed");
           const budget = of("budget_exceeded");
           const awaiting = of("awaiting_approval");
-          const other = of("running") + of("cancelled");
+          // `guardrail_blocked` joins `other`: the ramp is five colours by
+          // design and the four attention/terminal statuses spend them, so a
+          // sixth is "other" (see `seriesColor`). Without it a blocked run sits
+          // in `total` behind no arc, and the ring stops summing to the whole.
+          const other = of("running") + of("cancelled") + of("guardrail_blocked");
```
DEENUU1, line 58: Good catch folding `guardrail_blocked` into `other` - without it the arcs would sum to less than `total` and the completed-share centre would drift. The comment explaining why it's `other` and not a sixth arc is exactly the kind that saves the next person a debugging session.

DEENUU1 (commented):
Really clean port. The design decisions are all written down where the next reader will look for them, tool-result screening is the right thing to put front and centre (it's the only guard on untrusted content coming back into the loop), and "a block is a run outcome, not a graceful answer" is exactly right for a governance feature. The `other`-fold in the outcomes donut is a nice catch too - the ring keeps summing to the whole.

One blocking thing, and it's the important kind: the new status is only wired into the non-streaming path, so on the web chat a guardrail block lands as `failed` - the exact bucket this whole change exists to stay out of.

**The block only becomes `guardrail_blocked` on the non-streaming path.** `GuardrailBlocked` is imported and caught in exactly one place - `agent_runner._run`. The streaming web chat doesn't come through `_run`: `ChatAgentRunner.run` (`backend/app/services/agent_chat.py:382`) has its own settle try/except that catches `BudgetExceeded` and drops everything else into `except Exception` -> `RunStatus.FAILED` + `run_failure_summary(exc)` + re-raise. Then `agent_session.py:362` catches `(AppException, BudgetExceeded)` and tells the user plainly, but a re-raised `GuardrailBlocked` hits the generic `except Exception` below it -> `logger.exception(...)` (a stack trace) and a generic "turn failed" frame. So on the surface most people actually use, a block is recorded as `failed`, logged like a crash, and the user sees a generic failure instead of the refusal. I left the full detail on the runner clause inline. It needs mirroring in both files, plus a chat-stream regression test - CI is green here only because those files reach 100% via an existing generic-failure test, so the guardrail outcome runs a covered line with nothing asserting it.

One minor: the docstring at `backend/app/repositories/agent.py:227` still says "The four terminal states - `completed`, `failed`, `cancelled` and `budget_exceeded` - never load it again." That's now five, with `guardrail_blocked`. The query itself is still correct (it only loads `running` / `awaiting_approval`), so it's just the docstring - but that file's docstrings feed the reference bar, so worth the one-word fix.

For what it's worth, the frontend treats `guardrail_blocked` differently from `budget_exceeded` (out of `attention`, out of the Problems preset and the Recent-failures query) - you called the Problems/failures exclusion out in the description, so I read that as intentional; flagging only so it's on the record beside the "beside budget_exceeded" framing.

backend/app/services/agent_runner.py
DEENUU1: Fixed in 2d1e32bb. Mirrored the budget handling on the streaming path: `agent_chat.py` now has a `GuardrailBlocked` clause beside `except BudgetExceeded` (sets `RunStatus.GUARDRAIL_BLOCKED`, `error = str(exc)`, logs at info with `exc.edge`, re-raises), and `GuardrailBlocked` joined the `except (AppException, BudgetExceeded)` tuple in `agent_session.py` so the visitor gets the guard's plain reason and it is no longer logged as a crash. Two regression tests assert the consequence: the streaming run records `GUARDRAIL_BLOCKED` (not `failed`) and the client sees the safe reason with no crash log.

PR #780, opened 2026-08-15
feat(capabilities): add tool search, deferring MCP tools behind it

### LINKED ISSUE #50: Add Tool Search
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d9957680b695aaf22633dc1732) · Priority: **Low** · Category: Development

**Reference:**
- https://pydantic.dev/docs/ai/capabilities/tool-search/

## Why it matters here

Tool search lets an agent find the right tool from a large set instead of carrying every definition in its context. This matters here more than in most projects: an agent can bind an arbitrary number of MCP servers, and every MCP tool's schema is context the model pays for on every request.

## What will bite

- **The approval gate keys on tool identity.** `approval_required_tools` reads `tool.id` for the decision and the effective name for the match, and `ApprovalGate` refuses rather than guesses when `capability_id is None` (which is how MCP tools arrive). A search layer that hands the model a tool by a different name must not break that pairing — `tests/test_approval_gate.py` and `tests/test_tool_overrides.py` pin the current behaviour.
- Tool renames from a binding (`ToolOverrides`) happen *after* `_describe`. Search has to sit on the right side of that.
- Searching tools may itself cost a model call or an embedding. Meter it.

## Done when

DEENUU1:
## What changed

Ports Pydantic AI's `ToolSearch` (2.26) into the capability registry as
`tool_search`, and pairs it with deferring the connected MCP toolsets so it
actually does something in this codebase.

**Why it matters here:** an agent can bind an arbitrary number of MCP servers,
and every tool a server exposes is a schema the model reads on every request
whether or not it ever calls it. Tool search hides those tools until the model
discovers the one it needs.

- `app/agents/capabilities/tool_search/` — wraps the library capability.
  Config: `strategy` (`auto` | `keywords` | `bm25` | `regex`) and `max_results`
  (1–50). `auto` maps to the library's `None`. `tools=()`.
- `factory._defer_for_tool_search` — marks `extra_toolsets` (the MCP servers)
  for deferred loading exactly when the capability is bound. The two are paired
  or not at all: `ToolSearch` is inert with nothing deferred, and a deferred
  tool with no search to find it is a tool the model can never call. An agent
  that does not bind it pays nothing.
- `docs/reference/capabilities.md` updated (table row + a "Tool search"
  section).

## The footguns the issue flagged, and why they're safe

- **Approval pairing** — deferral changes what the model *sees*, never a
  tool's identity. A discovered MCP tool arrives under its real prefixed name,
  so the approval gate still pairs on it and a binding's rename still reaches
  it.
- **Rename ordering** — `ToolSearch.get_ordering()` is `outermost`, so it reads
  the names a rename already applied rather than the ones underneath.
- **Metering** — no un-metered spend. The local strategies run in Python;
  native search runs inside the provider's own request the budget guard already
  meters; the discovery round-trips are ordinary model requests. The one shape
  that would escape the guard — a custom callable that itself calls a model or
  an embedding — is deliberately not exposed.

Also confirmed: two `ToolSearch` capabilities are a hard `UserError`, and
`AgentSpec` refuses binding a capability id twice, so there is always exactly
one — no conflict, and the library does not auto-inject a second.

## Done-when (issue #50)

- [x] Registered, `tools` declared (`()`), `side_effecting` correct, in `load_builtins()`
- [x] Config schema is a Pydantic model with sane defaults; a disabled/absent binding contributes nothing
- [x] 100% coverage on the new module (CI enforces it)
- [x] `docs/reference/capabilities.md` updated
- [x] A test that the capability contributes nothing when its config is absent

## Verification

`ruff`, `ruff format --check`, `ty` clean. Capability, registry-drift, layout
and factory-pairing tests pass in a scoped run (86 green). The full `make test`
coverage gate was running green through 50% when this was pushed at the
author's request; **CI is the wide net for the remainder.**

Closes #50


labmimors:
Non-blocking benchmark gap on [`f8dd6f2`](https://github.com/vstorm-co/agenticos/commit/f8dd6f2a7557c4f7cee435d7d3d6b303230dc3c3): the [current integration test](https://github.com/vstorm-co/agenticos/blob/f8dd6f2a7557c4f7cee435d7d3d6b303230dc3c3/backend/tests/test_agent_spec_and_factory.py#L157-L193) proves that a `DeferredLoadingToolset` wrapper exists, but it never observes the tools sent on the first model request or proves that a discovered, prefixed tool remains callable. A factory or Pydantic upgrade could therefore preserve these assertions while regressing the actual schema surface.

Would you accept a keyless regression fixture using the repo’s existing [`FunctionModel` capture pattern](https://github.com/vstorm-co/agenticos/blob/f8dd6f2a7557c4f7cee435d7d3d6b303230dc3c3/backend/tests/test_tool_overrides.py#L108-L140)? It could generate 12/100/1,000 deterministic tools and, through `build_agent`, record: (1) no binding → all N schemas visible on request 1; (2) `tool_search` → deferred N absent and local `search_tools` present; (3) a fixed query reveals the expected prefixed target and that target executes; and (4) canonical UTF-8 schema JSON bytes plus model-request count. I would label these strictly as bytes/steps—not tokens, cost, latency, or model quality.

Pydantic AI 2.26 already covers the search algorithm/provider adapters; this would pin AgenticOS’s [assembly seam](https://github.com/vstorm-co/agenticos/blob/f8dd6f2a7557c4f7cee435d7d3d6b303230dc3c3/backend/app/agents/factory.py#L229-L281), not duplicate upstream. Disclosure: I maintain `dsh-mcp-lens`, which uses a similar keyless 12/100/1,000-tool corpus; I can contribute a clean AgenticOS-native Python fixture with no external model call, network dependency, or dependency on that project.

backend/tests/test_agent_spec_and_factory.py
```diff
@@ -154,6 +154,45 @@ def test_disabled_capabilities_are_not_built(self):
...
+            organization_id=uuid.uuid4(),
+            extra_toolsets=[FunctionToolset()],
+        )
+
+    def test_the_mcp_toolsets_are_deferred_when_it_is_bound(self):
+        from pydantic_ai.toolsets.deferred_loading import DeferredLoadingToolset
+
+        built = self._agent_with_mcp(tool_search=True)
+
+        assert any(isinstance(ts, DeferredLoadingToolset) for ts in built.agent.toolsets)
```
DEENUU1, line 224: This pins that a `DeferredLoadingToolset` *wrapper* is present, but not the consequence the capability exists for: that the MCP tool schemas are actually hidden on the first model request and a `search_tools` function shows up in their place (and, conversely, that they're all visible when it's unbound). A `DeferredLoadingToolset` could exist and still leak every schema through, and this test would stay green.

@labmimors's offered fixture covers exactly this using the `FunctionModel` capture pattern already in `test_tool_overrides.py`. Worth folding in - it's the difference between testing the wrapper and testing the behaviour.

DEENUU1 (commented):
This is a clean port and the reasoning around it is the strongest part - the capability⇔deferral pairing, the "no un-metered spend" argument, and the rename/approval ordering all hold up when you trace them (`ToolSearch` really does sit `outermost`, discovered MCP tools keep their prefixed names, and `extra_toolsets` is only ever MCP). Docs and the "contributes nothing" cases are covered.

One thing worth tightening: the two factory tests assert that a `DeferredLoadingToolset` exists, but not the behaviour the capability is *for* - what the model actually sees on request 1. That's the gap issue #50 called out ("a factory or Pydantic upgrade could preserve these assertions while regressing the schema surface"), and it's what @labmimors is offering to close. I'd take that fixture.

DEENUU1:
@labmimors yes please - this is a good call and a well-scoped one. The `FunctionModel` capture pattern you point at is exactly right, and keeping it strictly to schema bytes and request counts (no tokens/cost/latency) is the correct line. A keyless, AgenticOS-native fixture with no external call would be very welcome. Thanks for the careful read.

backend/tests/test_agent_spec_and_factory.py
DEENUU1: Fixed in c0d7c0c2. Added two behaviour tests to `TestToolSearchDefersMcp` using the `FunctionModel` capture pattern from `test_tool_overrides.py`, asserting both directions on request 1: bound (`strategy: keywords` for determinism) → the model sees `["search_tools"]` and none of the MCP schemas; unbound → it sees the MCP tool names and no `search_tools`. The two `DeferredLoadingToolset` wrapper assertions stay — the new tests pin the consequence the wrapper exists for. Refs #50.

Issue #803, opened 2026-08-16
Prune tool_output_limits spills at run end on non-run-scoped workspaces


DEENUU1:
## What breaks

`tool_output_limits` (added in #57) spills an oversized tool return to the agent's
sandbox backend under `tool_output/{run_id}/{call_id}.{retry}` and never deletes it.
On the default `run` session scope this is harmless — the workspace *is* the run, so
the spills die with it. On a longer-scoped workspace (`conversation`, `user`,
`agent`) the workspace persists across runs, and every past run's spill blobs stay in
it forever.

Two consequences:

- **It contradicts #57's "must not outlive the run"** for those scopes.
- **On a `state` backend it can eventually break the agent's own writes.** Spills are
  persisted into the `agent_workspaces` JSONB row and count against
  `SANDBOX_STATE_MAX_BYTES`. Once accumulated spills approach the cap,
  `CappedStateBackend.write` starts returning an error — which fails not just new
  spills but the agent's own `write_file` / `edit_file`. It self-heals only by
  overwrite, and there is no delete tool.

## Sequence that triggers it

1. An agent binds `sandbox` with `session_scope: user` (or `conversation`/`agent`) and
   `tool_output_limits` with `action: spill`.
2. Over many runs it spills large tool returns; each lands at
   `tool_output/<that run's id>/…` and is never removed.
3. The `state` workspace fills toward `SANDBOX_STATE_MAX_BYTES`; a later
   `write_file` from the agent is refused because the document is over the cap.

## Where

- `backend/app/agents/capabilities/tool_output_limits/_store.py` —
  `BackendOverflowStore.write` (no cleanup).
- `backend/app/agents/capabilities/tool_output_limits/_capability.py` —
  `_build_store` (documents the shared lifetime and points here).

## How you'd know it was fixed

A `user`-scoped state workspace that spilled in run A shows no `tool_output/<run A>/…`
entries once run A has ended (asserted in a test), and the agent's own writes keep
working across many spill-heavy runs. The read-back within a run still works — the
pruning is at run end, after the last `read_tool_result` could fire.

## The work

Needs a way to delete a run's spill blobs at run end across both backend shapes (a
sync `state` dict and an async container filesystem), which the narrow `OverflowStore`
protocol does not currently expose. Either grow the store a `cleanup(run_id)` method
wired to a run-end hook, or delete the `tool_output/<run_id>/` prefix as part of
workspace close. The default `run` scope needs nothing; this is only for the
[...]

PR #804, opened 2026-08-16
feat(capabilities): add tool_output_limits, spilling to the run's backend

### LINKED ISSUE #57: Add Overflowing tool output
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576806087a7dbe8c38fa87f) · Priority: **Low** · Category: Development

**Reference:**
- https://github.com/pydantic/pydantic-ai-harness/tree/main/pydantic_ai_harness/tool_output_limits — the superseding module; port this
- https://github.com/pydantic/pydantic-ai-harness/tree/main/pydantic_ai_harness/overflowing_tool_output — the earlier module this issue was filed against

> **Updated 2026-08-16:** the harness has evolved `overflowing_tool_output` into
> `ToolOutputLimits`. Same problem — a tool return large enough to dominate the window,
> re-sent on every later request as a persisted `ToolReturnPart` — but a richer design:
> ordered size **bands**, each a `(over, action)` pair, with three actions and `then`
> fallbacks. `Truncate` (zero-LLM head/tail clamp), `Spill` (lossless — a handle +
> preview, read back through a bounded `read_tool_result(handle, offset, limit, from_end,
> pattern)` tool, the [pydantic-ai#4352](https://github.com/pydantic/pydantic-ai/issues/4352)
> design) and `Summarize` (one size-gated LLM call). Default with no bands is
> `Spill(then=Truncate())`. Both `return_value` and text `content` are reduced; overflowing
> non-text `content` is left alone with a warning. `per_tool` overrides the global bands.
> Port `tool_output_limits`, not the older module.

## Why it matters here

Handling a tool result too large for the context window — spill it somewhere the agent can page through instead of truncating it.

The repo currently truncates. `code_execution/_sandbox.py` clips at `MAX_OUTPUT_CHARS = 8000` with `…(output truncated)`; the web-search toolset does something similar. That is the right default and the wrong ceiling: a grep over a large repo or a big API response loses the part that mattered, silently from the model's point of view.

## What will bite

- **This pairs with #36.** `pydantic-ai-backend` brings a filesystem; overflow output has somewhere to *go* once it exists. Doing this before #36 means inventing a second store.
- Whatever holds the spilled output is tenant-scoped and must not outlive the run.
- Keep a truncation marker the model can act on. Silent loss is worse than a visible cut.

## Done when

DEENUU1:
## What this adds

A `tool_output_limits` capability, porting `pydantic-ai-harness`'s `ToolOutputLimits` into the registry. It reduces a tool return too large for the model's window **once, when it is produced**, so an oversized `ToolReturnPart` stops being re-sent in full on every later request of the run.

Three actions, freely chosen per binding:

- **`spill`** (default, lossless) — the full return is written to the agent's backend and replaced with a handle + preview; the model pages through it with `read_tool_result`.
- **`truncate`** — a cheap, lossy clamp with a marker saying what was cut.
- **`summarize`** — an LLM summary, the expensive option.

Closes the gap the repo has today: `code_execution` clips at 8,000 chars — the right default, the wrong ceiling, because the part that mattered is gone with nothing to act on.

## What changed

- `backend/app/agents/capabilities/tool_output_limits/` — the capability. `_store.py` (a `BackendOverflowStore` over the sandbox backend, sync or async), `_capability.py` (flat config → single harness band; a metering wrapper), `__init__.py` (registration).
- `backend/app/agents/capabilities/_registry.py:600` — added to `load_builtins()`.
- `backend/tests/test_capability_layout.py` — `tool_output_limits` joins `sandbox`/`planning` in `EXTERNAL_TOOLSET` (its `read_tool_result` tool comes from the harness `get_toolset()`).
- `frontend/src/lib/tool-catalog.ts` + `frontend/messages/en.json` — a `read_tool_result` catalog row (the registry-parity test enforces one per backend tool) and its two `chat.tools` keys.
- `docs/reference/capabilities.md` — the "What ships" row and a `## Tool output limits` section.
- `backend/tests/test_tool_output_limits.py` — the tests.

## Two platform additions on top of the library (the same two `compaction` makes)

- **A store the tenant owns.** The harness default keeps spills on shared disk forever — a cross-tenant leak that outlives the run. A spill instead goes to the agent's own sandbox backend (`state` / Docker / Daytona) under a `tool_output/` prefix, scoped and cleaned up exactly as that workspace is; an agent with no backend gets an in-memory backend built for the run and discarded with it, so nothing lands on shared disk. **This is the #36 pairing made real** — the issue was filed when #36 was open and warned against inventing a second store; #36 landed the backend, so the spill lands there. One adapter covers the sync `state` backend and the async container backends by awaiting a result only when it is awaitable. A spill the backend refuses (a `state` workspace at its byte cap) falls back to a truncation, never a silent drop.
- **A budget that can see a summary.** `summarize` runs through an `Agent` the harness builds itself, outside `BudgetGuard` — the #16 shape. `MeteredToolOutputLimits` books the `ctx.usage` delta through `record_ambient_usage`; `spill` and `truncate` call no model and cost nothing.

## Design notes

- Config is flat (one `action` at one `threshold`) because the Builder form cannot draw the harness's nested `bands` list — the same reason `compaction` picks a strategy rather than composing tiers. The single band carries the harness's recommended fallbacks: `spill → truncate`, `summarize → spill → truncate`.
- `summarize` inherits the run's model (and therefore its vault credential), the only safe choice on this platform — `compaction`'s config docstring explains why naming a model as a string is not.

## Testing

- `backend/tests/test_tool_output_limits.py` — 24 tests. **100% coverage** on the new module (105 stmts, 18 branches). Covers the spill round-trip through a real backend, the async-backend path, a refused write degrading to truncation, an unknown handle raising, and a `summarize` spend booked against the run.
- `make test` (backend 100% gate) passes clean at 100.00%. `make docs-build` (`--strict`), the frontend i18n guard, catalog test, `tsc`, `eslint`, `prettier`, and the backend guards all pass. Registry drift, layout and spec tests pass.
- Frontend `test:coverage` flaked on 5 specs unrelated to this diff (`dashboard`, `channels`, `runs/approvals-count`, `create-kb-dialog`, `command-palette`) — all `findByText` timeouts under coverage instrumentation, the mode `.claude/rules/testing.md` documents. All 5 pass under `test:run` (57 tests, 16s), and none touch the files in this diff. Flagging in case CI's `test-frontend` catches the same flake.

## Noticed, not fixed

- **Spill lifetime on non-run-scoped workspaces** — a spill shares its backend's session scope, so on the default `run` scope it dies with the run (what the issue asks for), but on a `conversation`/`user`/`agent`-scoped workspace it persists and can accumulate. Documented in the capability section and the `_build_store` docstring; pruning at run end is filed as **#803** (found reviewing this change).
- **`per_tool` bands and `tool_filter`** are not exposed: nested/callable config the Builder cannot render. The capability applies to all of a run's tools, the sensible default; per-tool thresholds are a follow-up if a real need appears.

## Scope

**All three actions** (spill/truncate/summarize), with `summarize` metered — the faithful port, agreed up front rather than just spill+truncate.

## Follow-up: review + #803

Two commits on top of the port, after reviewing the diff as a maintainer.

**Shared the ambient-usage delta (`refactor`).** `compaction` and this capability
each carried a byte-for-byte copy of the snapshot-and-diff that books a self-run
`Agent`'s tokens (`_counts`/`_spent`). The cache-token subtraction is exactly the
pricing detail that gets fixed in one meter and forgotten in the other, so both now
import `usage_counts`/`usage_delta` from `budget`, beside `record_ambient_usage`.

**Two smaller review fixes on the just-added capability.**
- The `summary_prompt` validator required both `{tool_name}` and `{output}`; the
  harness only needs `{output}` (it passes `tool_name=` as a kwarg a prompt may
  ignore), so requiring it refused a valid tool-name-agnostic prompt. Now only
  `{output}` is required.
- The metering test proved the wrapper books whatever lands in `ctx.usage` but did
  it through a hand-mutating stub. Added an end-to-end test running a real
  `summarize` over `TestModel`, so a harness that stopped passing `usage=ctx.usage`
  goes red here instead of silently booking zero.

**#803 — spills no longer outlive the run on a `state` backend.** The spill lifetime
this PR flagged as "noticed, not fixed" is now fixed for the backend where it
actually bites. A longer-scoped `state` workspace was persisting spills into the
`agent_workspaces` JSONB row every run; they count against `SANDBOX_STATE_MAX_BYTES`
and eventually make `CappedStateBackend.write` refuse the agent's own writes.
`SandboxWorkspaceService._flush_state` now strips the reserved `tool_output/` prefix
before persisting, so the stored document holds the agent's files and never its
spills, every run self-healing what a prior one left. Read-back is untouched — the
strip is at flush, after the last `read_tool_result` could fire. Regression test
included; it fails without the strip.

A longer-scoped **container** workspace still keeps its spills on the container
filesystem: that half has no flush boundary, needs the harness `ctx.run_id` at close
and a delete the backend protocol does not expose, and a blanket prefix delete would
race a concurrent run sharing the workspace. It stays scoped in #803 rather than
rushed into a cross-run deletion bug.

## Testing (updated)

- `make test` green at **100.00%**; the shared helpers are covered by both meters'
  suites and the new `state`-flush strip has its own regression test.
- `make lint` clean.

Closes #57
Refs #803


backend/app/agents/capabilities/tool_output_limits/_store.py
```diff
@@ -0,0 +1,86 @@
...
+
+if TYPE_CHECKING:
+    from pydantic_ai_backends import AsyncBackendProtocol, BackendProtocol
+
+# The directory spilled payloads live under, so they never collide with a file
+# the agent wrote and are obvious for what they are when it lists its workspace.
+OVERFLOW_PREFIX = "tool_output"
+
+
+class OverflowWriteError(Exception):
```
DEENUU1, line 41: Nice — degrading a refused spill to a visible truncation instead of dropping it, and raising `FileNotFoundError` on an unknown handle so the read-back tool can tell the model "that handle is gone" rather than handing it an empty slice, are exactly the two failure modes that would otherwise be silent. Both tested. This is the platform-shaped thinking the harness default doesn't do.

backend/app/agents/capabilities/tool_output_limits/_capability.py
```diff
@@ -0,0 +1,301 @@
...
+            return await self.wrapped.after_tool_execute(
+                ctx, call=call, tool_def=tool_def, args=args, result=result
+            )
+        finally:
+            spent = _spent(before, ctx.usage)
+            if spent is not None:
+                record_ambient_usage(ctx.model.model_name or "unknown", spent)
+
+
+def _counts(usage: RunUsage) -> tuple[int, int, int, int]:
```
DEENUU1: `_counts` and `_spent` here are byte-for-byte identical to the ones in `compaction/_capability.py:562` — only the docstrings differ. That's a shared, pricing-sensitive calculation (the cache-token subtraction in particular is the kind of thing you fix in one meter and forget in the other). Both modules already import `record_ambient_usage` from `budget/`, so that's the natural home. I'd lift both helpers next to it so the two meters can't drift:

```python
from app.agents.capabilities.budget import record_ambient_usage, usage_delta
```

Non-blocking, but it's the "fix the class, not the copy" case.

backend/tests/test_tool_output_limits.py
```diff
@@ -0,0 +1,276 @@
...
+            ToolOutputLimitsConfig(threshold=42_000, over_tokens=True, strip_ansi=True),
+            backend=StateBackend(),
+        )
+        assert limits.over_tokens is True
+        assert limits.strip_ansi is True
+        assert limits.bands[0].over == 42_000
+
+
+class TestMetering:
+    async def test_a_summary_is_booked_against_the_run_that_paid_for_it(self):
```
DEENUU1, line 218: This asserts the wrapper books whatever lands in `ctx.usage`, but it does it through `_Spender`, which mutates `ctx.usage` by hand. The thing that actually makes metering work — the harness `Summarize` running its agent with `usage=ctx.usage` — is only asserted in prose. If the harness ever switches to a fresh usage object, this test stays green and the summary silently books zero. `_run_context()` already hands the run a `TestModel`, so you can exercise the whole path for real:

```python
async def test_a_real_summarize_books_its_model_call(self):
    ledger = SpendLedger()
    limits = MeteredToolOutputLimits(
        wrapped=build_limits(
            ToolOutputLimitsConfig(action="summarize", threshold=500),
            backend=StateBackend(),
        )
    )
    with metered_by(ledger):
        await limits.after_tool_execute(
            _run_context(), call=_call(), tool_def=_tool_def(), args={}, result="x" * 5_000
        )
    assert ledger.entries and ledger.input_tokens > 0
```

Keep `_Spender` too — it isolates the wrapper. This one guards the seam between the wrapper and the harness.

backend/app/agents/capabilities/tool_output_limits/_capability.py
```diff
@@ -0,0 +1,301 @@
...
+    def _must_carry_the_output(cls, prompt: str) -> str:
+        """A summary prompt missing `{output}` summarises nothing; a stray field crashes it.
+
+        Refused here rather than mid-run, which is where it would otherwise
+        surface: the harness formats this string only when a `summarize` fires,
+        on the long returns an agent actually spills, so the mistake would be a
+        turn that summarised an empty string - or raised `KeyError` on an unknown
+        placeholder - after the run was already underway.
+        """
+        for required in ("{tool_name}", "{output}"):
```
DEENUU1: You require both `{tool_name}` and `{output}`, but the harness only needs `{output}` — it always passes `tool_name=` as a kwarg, so a prompt that never references it just works. Requiring it is a fine UX nudge ("your summary should name the tool"), but it does refuse an otherwise-valid tool-name-agnostic prompt. Intentional, or mirroring the harness's `.format(...)` call more strictly than it needs? Either answer is fine; just flagging it's stricter than the mechanism requires.

DEENUU1 (commented):
Read the whole thing plus the harness internals it leans on, and it holds up. The store-ownership and metering additions are the right two things to add on top of the library, they're the same two `compaction` makes, and the refusal paths (a write the backend rejects, an unknown handle) are actually tested rather than assumed. I confirmed the load-bearing harness facts myself: `Spill._spill` catches any exception so `OverflowWriteError` triggers the truncation fallback, and the summary `Agent` runs with `usage=ctx.usage` (harness `_capability.py:436`), so the metering delta really does see it. Three non-blocking notes below — a drift risk in duplicated pricing code, one test I'd add, and a small question on the prompt validator.

Issue #801, opened 2026-08-16
browser-use pins pydantic==2.12.5, so the browser_use engine can't be installed


DEENUU1:
**Found building #59** (the `browser_use` capability).

The `browse_web` capability is wired, registered, tested and documented, but its
engine cannot be installed alongside AgenticOS today.

## What breaks

`browser-use` 0.13.x hard-pins `pydantic==2.12.5` (an exact `==`, verified against
the 0.13.7 metadata on PyPI). This project pins `pydantic>=2.13.4`. uv's universal
resolver refuses to lock even an **optional** extra whose transitive pins conflict
with the base ones, so:

- adding `browser-use` to `[project.optional-dependencies]` makes `uv lock` fail
  (`make lint` failed here at the `uv lock` step until the extra was removed);
- force-installing it into the venv (`uv pip install browser-use`) drags
  `pydantic` back to 2.12.5 across the whole platform.

So there is no configuration in which the engine and AgenticOS coexist. The
capability ships registered and **fails loudly** at call time
(`app/agents/capabilities/browser_use/_toolset.py` raises a `RuntimeError` naming
the missing extra) rather than pretending to work.

## The sequence that triggers it

Enable `browser_use` on an agent, publish, run, call `browse_web` → `RuntimeError`,
because `pydantic_ai_harness.browser_use` cannot import without `browser-use`, and
`browser-use` cannot be installed without breaking `pydantic`.

## How we would know it is fixed

`browser-use` publishes a release whose `pydantic` requirement is a range admitting
`>=2.13.4` (track browser-use/browser-use). Then: add
`browser-use = ["browser-use>=<that version>,<0.14> ; python_full_version >= '3.11'"]`
to `[project.optional-dependencies]`, `uv lock` succeeds, `agenticos[browser-use]`
installs, and `browse_web` runs end to end against a real Chromium.

Pre-existing vs. found: **found while building #59**, not caused by this branch —
it is an upstream pin. The branch lands the integration so this is the only thing
between the capability and working.

DEENUU1:
Blocks the engine of #59 from running; the capability shipped fail-loud in PR #805 pending this.

Issue #33, opened 2026-08-01
Code health clean-up: nine small items from the 2026-08-01 audit


DEENUU1:
**Severity: Low** · AUD-028

Nine small items from the audit, batched so they do not bury the issues that matter. Each is independent and roughly sub-hour.

- [ ] **`spec_version` is write-only.** `rg spec_version app` returns two lines in `app/agents/spec.py` (`:46`, `:470`), neither a read; both migrations key on content shape, not the number. So a client's `spec_version: 3` YAML can be published carrying v7 constructs, and a spec from a newer deployment is accepted silently as long as its fields parse. Stamp it on publish and refuse `> SPEC_VERSION` in `from_yaml`, or delete the field. A version number nobody reads and that can be wrong is worse than none.

- [ ] **`spec.observability.token_secret_id` is not validated at publish.** `app/services/agent_registry.py:331-417` checks capabilities, collections, MCP servers, secrets and the model profile; it never touches `spec.observability`. A wrong-kind or cross-tenant id publishes fine, then `_instrument` logs `agent_logfire_token_unavailable` and the agent runs **untraced**. `app/agents/factory.py:307-322` claims *"publishing is where a missing secret is refused, and a run is far too late."* Run it through the same existence/tenant/kind checks. Test: `test_publishing_refuses_a_tracing_token_this_organization_does_not_have`.

- [ ] **`hint` is the whole secret for short values.** `ApiKeySecret(api_key='1234').hint` → `'1234'` (executed). `SealedStr` fields carry no `min_length` (`app/core/secret_kinds.py:114-118`, `:133-135`) and `seal()` refuses only the empty string, so a short proxy or test key is published verbatim in `SecretRead.hint` — whose field description reads *"Four characters of the secret — never the secret itself"* — to everyone with `secrets:view`, and into the audit entry at `organization_secret.py:203`. `min_length=8` on the sealed fields; it also catches a truncated paste. Test: `test_a_secret_too_short_to_hint_safely_is_refused`.

[...]

PR #805, opened 2026-08-16
feat(capabilities): add browser_use, delegating to a browser-use agent

### LINKED ISSUE #59: Add Browseruse
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d995768087a721da0ef06a3154) · Priority: **Low** · Category: Development

**Reference:**
- https://github.com/pydantic/pydantic-ai-harness/pull/419

## Why it matters here

Browser automation — an agent that can drive a real page rather than fetch its HTML.

## What will bite

- **This is the largest attack surface on the list.** A browser will follow whatever a page tells it, and the page is untrusted. Prompt injection here reaches a tool with side effects. It must be `side_effecting=True`, gateable, and documented as such.
- **SSRF.** A user-supplied URL fetched server-side is exactly what `app.core.sanitize.validate_webhook_url` guards against — and that guard currently has **no callers at all** (#33). Wiring it in is part of this task, not a follow-up.
- **Resource cost.** A browser is not a HTTP client: memory, a process, a timeout that actually fires. Where does it run — the API container, the worker, or the `sandboxd` service #36 introduces? Answer before building.
- Almost certainly should wait for #36, which brings the isolation this needs.

## Done when

### LINKED ISSUE #802: Meter the browser_use sub-agent's model spend through the run ledger
**Found building #59** (the `browser_use` capability).

The `browse_web` tool delegates to a `browser-use` agent that runs its own
perception-action loop, making one model request per step **on browser-use's own
model configuration**. Those requests happen inside the sub-agent's loop, outside
this platform's budget ledger — so the budget guard cannot see them, price them, or
stop a run that has blown its cap.

This is the exact failure `docs/governance.md` and the hard boundary in `CLAUDE.md`
exist to refuse: an unmetered model request. It ships in #59 gated behind
`side_effecting=True` (approval-gateable) and off by default (the engine is not even
installable yet — see the dependency blocker), but the gap is real and named here so
it is not discovered later.

## The fix, in the shape the codebase already has

`MeteredCompaction` (`app/agents/capabilities/compaction/_capability.py`) records
ambient model spend into the run's ledger through the `budget` capability's
context-var recorder, which the runner opens around the whole run
(`app/services/agent_runner.py`). The sub-agent's chat model needs the same
treatment: resolve the org's model, wrap it so each response reports its `Usage`
into the run ledger, and hand that to the harness `BrowserUse(llm=...)`. Model
resolution for a sub-agent is the `SubagentRuntime` pattern
(`app/agents/subagent_runtime.py`), placed in the build `resources`.

The pattern is prior art in #16 (metering the web chat surface, closed).

## How we would know it is fixed

A `browse_web` run's model spend appears in the run's ledger and counts against the
agent's budget; a run that would exceed the cap is refused before the sub-agent's
next step, with a test asserting the spend is recorded even when the browse run
fails. Depends on the engine being installable (the dependency blocker) to test end
to end, but the metering wrapper itself is testable against a fake model now.

DEENUU1:
## What this fixes

Adds the `browser_use` capability — a port of the `pydantic-ai-harness` capability
of the same name ([pydantic/pydantic-ai-harness#419](https://github.com/pydantic/pydantic-ai-harness/pull/419)).
One tool, `browse_web`, hands a self-contained natural-language goal to an
autonomous [browser-use](https://github.com/browser-use/browser-use) agent that
drives a real Chromium and returns its text result.

Every "Done when" box on #59 is ticked. The SSRF guard is wired (part of #33), and
the sub-agent's model spend is now metered against the run's budget (#802). One
thing is still blocked upstream and stays open — read *Still blocked*.

## What changed

- `backend/app/agents/capabilities/browser_use/**` — the capability. `_toolset.py`
  declares `browse_web` and builds its toolset **without importing browser-use**;
  the engine is reached through a factory seam (`BrowserDelegateFactory`) only when
  the tool is called. `_capability.py` is the `AbstractCapability`; `__init__.py`
  is the config schema + `@register` + `_build` + the `validate_cdp_url` SSRF check.
- `_registry.py` — `browser_use` added to `load_builtins()`.
- `agent_registry.py` — a new `web:browse` scope in `DEFAULT_GRANTED_SCOPES`, and
  `_browser_use_problems`, which SSRF-checks a remote `cdp_url` at publish, off the
  event loop, for every binding (a spec's own and each inline specialist's).
- `frontend/src/lib/tool-catalog.ts` + `messages/en.json` — the `browse_web` row
  and its two `chat.tools` keys (`TestFrontendToolCatalog` fails without them).
- `docs/reference/capabilities.md` — the Browser automation section, plus the
  *What ships* and *Scopes* rows.
- `backend/pyproject.toml` — a comment explaining why there is **no** optional
  extra (see below), not an extra.

## The decisions

- **Two modes, one seam** (as agreed): `mode='playwright'` launches a local
  headless Chromium; `mode='remote'` attaches over CDP to an operator-supplied
  `cdp_url` — the isolated browser service a self-hosted deployment points at
  rather than putting a browser process in the app container. Launch vs. attach is
  the only difference the harness sees.
- **The engine is an optional dependency the capability builds without.** The
  harness owns the browser (session lifecycle, allowlist, result rendering); this
  repo owns registration, config, the SSRF check, metering and the tool
  declaration. The capability registers, enumerates its tool and builds its
  toolset with the extra absent — which is the CI state — so a bound-but-unused
  capability pays nothing and CI needs no browser.
- **SSRF, at publish, off the loop.** A remote `cdp_url` is a URL this deployment
  connects to server-side, so it runs through `validate_webhook_url` (allowing
  `ws`/`wss` alongside `http`/`https`). This is that guard's **first production
  caller** — it had none, which was half of #33. The check resolves DNS
  (`getaddrinfo`), which blocks, and a capability is built on the event loop inside
  a tool call — so it runs in `agent_registry._browser_use_problems` via
  `asyncio.to_thread` when the spec is published, not on the build path. `_build`
  does no I/O.
- **The sub-agent's model is metered (#802).** `browse_web` runs the browser
  sub-agent on the host run's model — the one whose credential was resolved from
  the vault — wrapped in a `MeteredModel` that books each response against the
  run's ledger through the `budget` capability's ambient-usage recorder, the shape
  `MeteredCompaction` uses. Before this the loop ran on browser-use's own hosted
  model, one request per step, outside the budget guard.
- **`side_effecting` on the tool and the capability.** A browser follows an
  untrusted page, so `browse_web` is prompt injection reaching an action; the
  approval gate is what stands in front of it.

## Still blocked — the engine is not installable yet (#801)

The reference PR ships browser-use as an optional extra that "resolves away" when
absent. That does not work here: **browser-use 0.13.x hard-pins
`pydantic==2.12.5`** (verified again at 0.13.7, the current release), and this
project pins `pydantic>=2.13.4`. uv's universal resolver refuses to lock even an
*optional* extra with that clash, and force-installing it drags pydantic back a
minor across the whole platform. So there is no configuration in which the engine
and AgenticOS coexist today.

Per the decision on the issue, the capability therefore **ships registered and
fail-loud**: enabling and running it without the (currently un-installable) engine
raises a `RuntimeError` naming the fix. **#801 stays open** because only browser-use
loosening that pin closes it; the moment it does, adding the extra is the one
remaining step. This is why #802's metering is verified against a fake model here
and its end-to-end check waits on #801.

## Testing

- `backend/tests/test_browser_use.py` — config validation (both `mode` branches),
  the SSRF guard (`validate_cdp_url`) refusing a loopback and accepting a public
  `cdp_url`, `_build` doing no network I/O, the toolset offering only `browse_web`,
  delegation through a fake engine, the fail-loud path when the extra is missing,
  the kwargs mapping, and the `MeteredModel` booking a step against the ledger
  (and a no-op with no active ledger). **100% coverage** on the new module.
- `backend/tests/test_agent_registry.py` — publishing a `browser_use` agent with a
  loopback `cdp_url` is refused; a public one publishes.
- `make test` — 5275 passed, platform coverage 100%. `make lint-backend`,
  `make docs-build` — clean.

Closes #59
Closes #802
Refs #801, #33

### DIFF
--- a/backend/app/agents/capabilities/browser_use/__init__.py
+++ b/backend/app/agents/capabilities/browser_use/__init__.py
@@ -1,0 +1,109 @@
+"""Browser automation - delegate an open-ended web task to an autonomous browser agent."""
+
+from typing import Literal
+
+from pydantic import BaseModel, Field, model_validator
+from pydantic_ai.capabilities import AbstractCapability
+
+from app.agents.capabilities._registry import (
+    CapabilityBuildContext,
+    CapabilityToolInfo,
+    register,
+)
+from app.agents.capabilities.browser_use._capability import BrowserUse
+from app.core.sanitize import validate_webhook_url
+
+__all__ = ["BrowserUse", "BrowserUseConfig"]
+
+# A CDP endpoint is reached over HTTP(S) or a WebSocket, so the SSRF check allows
+# those four rather than webhook's http/https default. Everything else it enforces
+# - no userinfo, no private/reserved/loopback address, DNS resolved to a public IP -
+# applies unchanged.
+_CDP_SCHEMES = frozenset({"http", "https", "ws", "wss"})
+
+
+class BrowserUseConfig(BaseModel):
+    """How this agent browses the web.
+
+    `mode` is the whole decision. `playwright` launches a headless Chromium next to
+    the agent; `remote` attaches to a browser an operator runs elsewhere - a
+    hardened, isolated service a self-hosted deployment points at instead of giving
+    the app container a browser process. `remote` needs a `cdp_url`; `playwright`
+    forbids one.
+    """
+
+    mode: Literal["playwright", "remote"] = Field(
+        default="playwright",
+        description=(
+            "playwright: launch a headless browser locally. "
+            "remote: attach to a browser running elsewhere via its CDP endpoint."
+        ),
+    )
+    cdp_url: str | None = Field(
+        default=None,
+        description="Chromium DevTools endpoint to attach to; required by (and only valid in) remote mode.",
+    )
+    allowed_domains: list[str] | None = Field(
+        default=None,
+        description="Domains the agent may navigate to; null means no restriction. Globs like *.example.com are allowed.",
+    )
+    max_steps: int = Field(
+        default=25,
+        ge=1,
+        le=100,
+        description="Hard cap on the browser agent's steps per call; each step is one model request.",
+    )
+    use_vision: bool = Field(
+        default=True,
+        description="Send page screenshots to the browser agent's model; better on visual layouts, more tokens.",
+    )
+    headless: bool = Field(
+        default=True,
+        description="Run a locally launched browser without a visible window (playwright mode only).",
+    )
+
+    @model_validator(mode="after")
+    def _cdp_url_matches_mode(self) -> "BrowserUseConfig":
+        if self.mode == "remote" and not self.cdp_url:
+            raise ValueError(
+                "mode 'remote' requires a cdp_url pointing at a Chromium DevTools endpoint"
+            )
+        if self.mode == "playwright" and self.cdp_url:
+            raise ValueError(
+                "cdp_url is only valid in mode 'remote'; a playwright browser is launched locally"
+            )
+        return self
+
+
+@register(
+    id="browser_use",
+    name="Browser automation",
+    category="research",
+    description="Delegate an open-ended web task to an autonomous browser agent.",
+    tools=(
+        CapabilityToolInfo(
+            id="browse_web",
+            description="Delegate an open-ended web task to an autonomous browser agent.",
+            side_effecting=True,
+        ),
+    ),
+    config_schema=BrowserUseConfig,
+    side_effecting=True,
+    scopes=("web:browse",),
+)
+def _build(ctx: CapabilityBuildContext) -> AbstractCapability[object]:
+    config = ctx.config if isinstance(ctx.config, BrowserUseConfig) else BrowserUseConfig()
+    cdp_url: str | None = None
+    if config.mode == "remote" and config.cdp_url is not None:
+        # The one production caller of the SSRF guard (agenticos#33): a user-supplied
+        # URL this deployment connects to server-side is exactly what it exists to
+        # refuse - a metadata endpoint, a loopback debugger, an internal service.
+        cdp_url = validate_webhook_url(config.cdp_url, allowed_schemes=_CDP_SCHEMES)
+    return BrowserUse(
+        mode=config.mode,
+        allowed_domains=config.allowed_domains,
+        max_steps=config.max_steps,
+        use_vision=config.use_vision,
+        headless=config.headless,
+        cdp_url=cdp_url,
+    )
--- a/backend/app/agents/capabilities/browser_use/__init__.py
+++ b/backend/app/agents/capabilities/browser_use/__init__.py
@@ -13,7 +13,7 @@ from app.agents.capabilities._registry import (
 from app.agents.capabilities.browser_use._capability import BrowserUse
 from app.core.sanitize import validate_webhook_url
 
-__all__ = ["BrowserUse", "BrowserUseConfig"]
+__all__ = ["BrowserUse", "BrowserUseConfig", "validate_cdp_url"]
 
 # A CDP endpoint is reached over HTTP(S) or a WebSocket, so the SSRF check allows
 # those four rather than webhook's http/https default. Everything else it enforces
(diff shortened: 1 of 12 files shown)


backend/app/agents/capabilities/browser_use/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+    scopes=("web:browse",),
+)
+def _build(ctx: CapabilityBuildContext) -> AbstractCapability[object]:
+    config = ctx.config if isinstance(ctx.config, BrowserUseConfig) else BrowserUseConfig()
+    cdp_url: str | None = None
+    if config.mode == "remote" and config.cdp_url is not None:
+        # The one production caller of the SSRF guard (agenticos#33): a user-supplied
+        # URL this deployment connects to server-side is exactly what it exists to
+        # refuse - a metadata endpoint, a loopback debugger, an internal service.
+        cdp_url = validate_webhook_url(config.cdp_url, allowed_schemes=_CDP_SCHEMES)
```
DEENUU1: Two things about doing the SSRF check here.

First, `validate_webhook_url` ends in `socket.getaddrinfo` (`sanitize.py:128`, which even carries a TODO about being blocking I/O). `_build` runs inside `build_agent`, and that path documents itself as CPU-only and safe on the loop — `agent_runner.py:1220`: *"what the closure does is CPU work and Pydantic AI ... safe to call from inside a tool call"*. Both `build_agent` call sites (1231, 1759) run on the event loop with no offload, so a remote-mode agent with a hostname `cdp_url` now blocks the loop on DNS at every build.

Second, `validate_spec` validates the config schema but never calls the builder, so a `cdp_url` pointing at `127.0.0.1` passes publish and fails on *every run* instead — the "refused at run time, not at publish" shape we usually avoid.

Both point the same way: move the DNS resolution into `validate_spec`, which is `async` and can `await asyncio.to_thread(validate_webhook_url, ...)`. That rejects a bad endpoint once, at publish, off the loop — and you keep the cheap non-DNS checks (scheme, userinfo, IP-literal) wherever you like. The harness still does the real connect, so build-time re-validation was never the security boundary anyway.

DEENUU1 (commented):
Clean port, and the seam design is the right call — declaring `browse_web` and building the toolset without ever importing `browser-use`, reaching the engine only through `BrowserDelegateFactory`, is exactly what keeps a bound-but-unused capability free and CI browser-less. The untrusted-output instructions, `side_effecting` on both the tool and the capability, and wiring the SSRF guard as #33's first real caller are all solid, and the tests cover the refusals properly.

One thing worth a second look before merge: the SSRF check does a **blocking DNS lookup on the event loop, and only fires at run time** — details inline. Everything else is a note, not a blocker.

Smaller one for the body: the `harness_kwargs` → `pydantic_ai_harness.browser_use.BrowserUse(**kwargs)` mapping is the one contract CI can't exercise (the extra is absent), so if the harness ever renames `cdp_url` / `max_steps` / `use_vision` / `allowed_domains` / `headless` it only breaks at runtime with the engine installed. Worth pinning the harness version so that contract can't drift silently under us.

backend/app/agents/capabilities/browser_use/_toolset.py
```diff
@@ -0,0 +1,116 @@
...
+
+from pydantic_ai.toolsets import FunctionToolset
+
+
+class BrowserDelegate(Protocol):
+    """The browser-use engine, reduced to the one call `browse_web` forwards to."""
+
+    async def browse_web(self, task: str) -> str:
+        """Drive a browser to carry out `task` and return its text result."""
+        ...  # pragma: no cover
```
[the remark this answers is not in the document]
DEENUU1: `...` as a `Protocol` method body is the adjudicated pattern in docs/code-review.md — the ellipsis is the canonical stub for a structural type, and `# pragma: no cover` marks it deliberately unexecuted. No change.

backend/app/agents/capabilities/browser_use/__init__.py
DEENUU1: Addressed as suggested: the DNS-resolving SSRF check moved to publish time — `agent_registry._browser_use_problems` runs `validate_cdp_url` via `asyncio.to_thread` for every binding, and `_build` does no I/O. Covered by `test_agent_registry.py` (loopback refused at publish, public URL accepted).

Issue #55, opened 2026-08-01
Add media


DEENUU1:
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576802fb669d0b2c0b9b9a8) · Priority: **Low** · Category: Development

**Reference:**
- https://github.com/pydantic/pydantic-ai-harness/tree/main/pydantic_ai_harness/media

Port of a `pydantic-ai-harness` capability into the AgenticOS capability registry.

## Why it matters here

Media handling — images and other binary content in and out of a run. Partly present already: `agent_session.py:286-300` loads attached images per turn, and `file_upload.py` parses PDFs and DOCX.

## What will bite

- **The existing path blocks the event loop** (#25). `LocalFileStorage.save`/`load` are `async def` around `Path.write_bytes`/`read_bytes`, and `agent_session.py` loads every attached image *sequentially* on every turn. Fix that as part of this rather than adding more callers to it.
- Media is expensive in tokens and the cost is easy to lose. Image input tokens must reach the ledger.
- Storage is per-organization. An attachment must not be reachable across tenants — `files.py:54` scopes retrieval to `current_user.id` today; check that still holds for agent-side reads.

## The contract every capability in this repo obeys

- **The registry is the only door.** There is no `@agent.tool`. Register with
  `@register(id=..., name=..., category=..., description=..., tools=[...], config_schema=..., side_effecting=..., scopes=[...])`
  and add the module to `load_builtins()`. `app/agents/capabilities/clock/_capability.py` is the shape and the docstring bar.
- **`tools=` is not optional.** A tool absent from that list cannot be gated by the approval policy or renamed by a binding.
- **100% coverage and strict typing, automatically.** `app/agents/**` is already a glob in both `[tool.coverage.run] include`
  and `[[tool.ty.overrides]] include` in `backend/pyproject.toml`. No config edit needed, and no untested branch either.
- **The capability `id` is permanent.** Published specs reference it; changing it breaks agents nobody touched.
- **Docs.** `app/agents/capabilities/**` → `docs/reference/capabilities.md`, generated from docstrings. Reasoning goes in the docstring.
- **Cost.** Anything that spends tokens outside the main model request must go through `metered_by` / `record_ambient_usage`,
  or the budget guard cannot see it. #16 is the live example of forgetting this.

## Done when

- [ ] Registered, tools declared, `side_effecting` set correctly, in `load_builtins()`
[...]

Issue #25, opened 2026-08-01
Move blocking file work off the request event loop


DEENUU1:
**Severity: Medium** · AUD-023

## What's wrong

Two request-path functions declared `async` that do pure blocking work.

## 23a. PDF and DOCX parsing

`backend/app/services/file_upload.py:76-91` (`_parse_pdf_pymupdf`), `:107-117` (`_parse_docx_content`), called from `:137` inside `async def upload`; route at `api/routes/v1/files.py:17-24`.

`parse_content` is `async def` but every branch is blocking CPU. `pymupdf.open(stream=data)` plus `page.get_text("text")` over every page of a document up to `MAX_UPLOAD_SIZE` runs to completion without a single suspension point. One user uploading a large PDF freezes every other request and every in-flight agent WebSocket stream on that uvicorn worker — seconds, for a scanned or page-heavy file.

## 23b. `LocalFileStorage`

`backend/app/services/file_storage.py:112-125`. `save` and `load` are declared `async` and call `Path.write_bytes` / `Path.read_bytes` directly, up to 10 MB. Callers on the request path: `file_upload.py:140`, `rag_document.py:218`, `user.py:211`, `organization.py:195`, `agent_registry.py:684`, and per-turn in `agent_session.py:291` — which loads every attached image sequentially on every turn, so a conversation with three images blocks the loop three times per message.

## The codebase already knows this pattern

`agents/mcp.py:53` routes DNS resolution through `asyncio.to_thread` for exactly this reason, and `rag_document.py:239` switched a file write to anyio with the comment *"writing it synchronously stalls every other request on this worker until it lands."*

Sharpest evidence 23b is unintentional: `rag_document.py:218` calls the blocking `storage.save(...)` and then twenty lines later writes the *same bytes* to a temp path via `anyio.open_file`. One of the two writes got fixed.

## Fix

`await asyncio.to_thread(...)` in both, or `anyio.open_file` to match the sibling.

Same defect, lower severity because it is the worker rather than the API: `worker/tasks/rag_tasks.py:256` (`rglob` over a whole tree), `:289`, `:302` (`hashlib.sha256(filepath.read_bytes())`), `:323` (`filepath.stat()`).

## Acceptance criteria

- [ ] Both offloaded to a thread
- [ ] Regression test: a concurrent `asyncio.sleep(0)` task completes while `parse_content` runs against a fixture PDF

DEENUU1:
[...]

PR #807, opened 2026-08-16
feat(capabilities): add image generation

### LINKED ISSUE #58: Add Image Generation
**Source:** [VstormOS roadmap in Notion](https://app.notion.com/3adfa3d99576807c9f23fd4bf6661a2a) · Priority: **Low** · Category: Development

**Reference:**
- https://pydantic.dev/docs/ai/capabilities/image-generation/

## Why it matters here

Lets an agent produce images.

## What will bite

- **`side_effecting=True` and gateable.** Generating an image spends real money on someone else's key and produces content that may be published. It belongs behind the approval policy for at least some bindings.
- **Pricing.** `price_request` delegates to `genai-prices`; an image model may not be priced there. An unpriceable model records `0` with `priced=False` → `cost_is_partial` on the run. Verify that path rather than assuming, or image spend becomes invisible to budgets.
- Where does the image go? It needs the same tenant-scoped storage question as #55, and the two should agree.
- Provider credential: declare it as a `SecretRequirement` on the capability so publish-time validation refuses a binding whose key is missing — that is what `secret=` on `@register` is for.

## Done when

DEENUU1:
## What this adds

`Closes #58`. An `image_generation` capability: one tool, `generate_image`, that
draws an image from a prompt with a dedicated image model (OpenAI Responses or
Google), so it works whatever the agent's own model is.

Kacper chose the **full** scope (capability + native/local + tenant-scoped
storage), and to build the storage now so [#55](https://github.com/vstorm-co/agenticos/issues/55)
inherits its shape.

## What changed

- `backend/app/agents/capabilities/image_generation/**` — the capability:
  `generate_image`, `side_effecting=True`, an unconditional
  `SecretRequirement(API_KEY)`, config with sane defaults (`model` +
  optional `quality`/`size`/`background`/`output_format`/`aspect_ratio`), added
  to `load_builtins()`.
- `backend/app/services/generated_media.py` — organization-scoped storage over
  the existing file store (`generated/{org}`), hardened so a filename with `../`
  or from another tenant is a 404.
- `backend/app/api/routes/v1/generated_media.py` — `GET /api/v1/generated/{filename}`,
  serving **only** under the caller's own organization, via the shared
  `_workspace_bytes.file_response` (inline png/jpg/gif/webp).
- Frontend — a `generate_image` row in `tool-catalog.ts` (the backend drift test
  requires it), an image renderer, a same-origin proxy route
  `src/app/api/generated/[filename]/route.ts`, and the `chat.tools` i18n keys.
- `docs/reference/capabilities.md` — the row and a section.

## The three things #58 flagged, each handled not assumed

- **Metered.** The library's own image tool runs a subagent and discards
  `result.usage` — the invisible spend of #16. So the subagent is run in the
  capability and its usage booked to the run's ledger via `record_ambient_usage`
  (inside the `metered_by` block the runner already holds open). Image models are
  often unpriced by `genai-prices`; that path records zero and flags the run's
  cost partial rather than hiding it — the outcome #58 asked to **verify**, and
  there is a test for it.
- **Keyed.** The provider key is a `SecretRequirement`, so publishing without one
  is refused at the form.
- **Gateable.** `side_effecting=True`, so every call is a candidate for the
  approval gate and can be gated per binding.

## Storage, and how it meets #55

Organization-scoped, not per-user — wider than a chat upload, because there is no
record of *who* produced an image, so the tenant is the boundary enforceable
without one. #55 (a per-run record) is where it would tighten, inheriting the
`generated/{org}` convention and the serving route unchanged. When the run also
has a workspace open (the `sandbox` capability), the image is additionally
written under `/output`, so a later `execute` step can build with it — a PDF, a
slide, a page.

## Testing

- `make test` — backend, **100%** (5268 passed). New modules covered; the image
  model run itself is faked, so no live provider call is exercised.
- `make test-frontend-cov` — **100%** lines/statements/functions, 97.55% branches.
- `make lint` — clean. `make docs-build` (`--strict`) — clean.
- `backend/tests/test_image_generation.py`, `test_generated_media.py`,
  `tests/api/test_generated_media_routes.py`; frontend `tool-results.test.tsx`,
  `binary-routes.test.ts`, `tool-call-card.test.tsx`.

## Noticed, not fixed

- **`make db-check` fails locally** with `Can't locate revision
  '0033_conversation_summary'` — the shared dev DB is stamped at another branch's
  numbering of that migration (this branch has it as `0034_conversation_summary`).
  This branch adds **no** model or migration, so CI's fresh-DB `db-check` passes.
- **`LocalFileStorage` does its IO on the event loop** ([#25](https://github.com/vstorm-co/agenticos/issues/25)).
  This module awaits that API and adds no new blocking of its own; the class-level
  fix stays with #25 to keep this diff scoped.
- **Channel (Slack/Telegram) rendering** of a generated image is not wired here —
  charts render to PNG through a separate path; doing the same for images is a
  natural follow-up and not required by #58's acceptance criteria.
- **The model list** (`openai-responses:gpt-5.4`, `google:gemini-3-pro-image`) is
  a curated pair; more image-capable models can be added to the `Literal` as they
  land.

Refs #55, #25


backend/app/agents/capabilities/image_generation/_toolset.py
```diff
@@ -0,0 +1,196 @@
...
+            # unsupported setting. Hand it back for the model to rephrase rather
+            # than ending the turn on an error string.
+            raise ModelRetry(f"Image generation failed: {exc}") from exc
+
+        image = result.output
+        # Booked to the run's ledger, which the runner is holding open. Image
+        # models are frequently unpriced by `genai-prices`, in which case this
+        # records at zero and flags the run's cost as partial - the honest
+        # outcome, and the one #58 asked to be sure of rather than assume.
+        record_ambient_usage(model_name, result.usage)
```
DEENUU1: One thing on the metering: we pass the model name but not the provider, and we hand-roll the name with `model_id.split(":")[-1]`. The sibling path in `app/services/rag/image_describer.py:67` does it from the model object — `record_ambient_usage(self.model.model_name, result.usage, provider=self.model.system)` — and that's worth mirroring here for two reasons.

`price_request` tries `(provider, None)`, so a provider hint can only help (a bad hint just `LookupError`s and falls back to `None`). I checked what the model object reports for our two ids: `openai-responses:gpt-5.4` → `model_name="gpt-5.4"`, `system="openai"`; `google:gemini-3-pro-image` → `model_name="gemini-3-pro-image"`, `system="google"`. Those `system` values are exactly the genai-prices provider ids, so without the hint a priceable image model (Gemini especially) is more likely to fall through to zero-and-partial than it needs to. Capturing the model also drops the manual split.

```python
    model = _build_image_model(model_id, api_key)
    agent: PydanticAgent[None, BinaryImage] = PydanticAgent(
        model,
        output_type=BinaryImage,
        capabilities=[NativeTool(ImageGenerationTool(**tool_settings))],
        instructions=_SUBAGENT_INSTRUCTIONS,
    )
    ...
    record_ambient_usage(model.model_name, result.usage, provider=model.system)
```

`model.model_name` still resolves to `"gpt-5.4"`, so the ledger assertion in `test_image_generation.py` stays green. (And you can delete the `model_name = model_id.split(":")[-1]` line at the top of the builder.)

backend/tests/test_generated_media.py
```diff
@@ -0,0 +1,58 @@
...
+    with pytest.raises(NotFoundError):
+        await load_generated_image(intruder, filename)
+
+
+async def test_a_missing_image_is_not_found():
+    with pytest.raises(NotFoundError):
+        await load_generated_image(uuid4(), "deadbeef1234_image.png")
+
+
+async def test_a_traversing_filename_cannot_climb_out_of_the_org_directory():
```
DEENUU1, line 48: Nice — this is the test I'd have asked for. It's easy to write the tenant-isolation test as "another org gets a 404" and stop there; proving that a crafted `../generated_{other}/{leaf}` can't climb sideways inside the shared root is the one that actually pins the guard down.

DEENUU1 (commented):
Solid capability, and it does the three things #58 actually asked for rather than assuming them. Metering the subagent's usage onto the run ledger — the #16 lesson — is the crux, and there's a test that asserts the ledger entry rather than just running the path. The tenant-scoped storage with the `../` guard and the cross-org 404 test is exactly right, and extending the registry drift check so a secret-requiring capability still goes through `build()` is a nice catch. One suggestion on the metering side worth considering, otherwise this looks ready.

PR #813, opened 2026-08-16
feat(agents): pin a version from the environment's own row

### LINKED ISSUE #134: Environments: pin a version from the environment, not only from the version
Repointing an environment works, but only from the wrong end. The Versions list gives every row a **Promote to…** menu, so the sequence is "find the version you want, open its menu, pick the environment". An environment row itself offers a name, `serves v5`, and a delete button — `frontend/src/components/agents/environments-panel.tsx:48`.

That is backwards for the question people actually arrive with. Nobody thinks "what should v3 do"; they think "**dev should serve v3**". With six versions and three environments the current flow makes you scan a list of versions to change a property of an environment, and the row that states the property is the one you cannot edit.

**What it should do.** Each environment row gets a version control — a select of the agent's published versions, showing what it serves now:

```
dev        serves  [ v5  ▾ ]                            🗑
production serves  [ v6  ▾ ]  default
```

Picking another version is the same promote that already exists. **Promote to…** on the version rows stays: it is the right shape for "I have just published v7, put it on staging", and both being present is not duplication - they are two directions through the same edit.

**It is a UI change and nothing else.** `PATCH /agents/{id}/environments/{environment_id}` already takes `version_id` (`backend/app/api/routes/v1/agent_environments.py:49`, `EnvironmentUpdate.version_id`), and the client already has the mutation: `promote` in `frontend/src/hooks/use-agent-environments.ts:37`. The panel just never calls it.

**Three things worth getting right:**

1. **The default environment.** Publishing repoints it, so a version pinned there by hand is about to be overwritten by the next publish. The row should say so rather than pretending the pin is permanent - the panel's own description already explains the rule, and this is where somebody meets it.
2. **Renaming.** The same row is the obvious place for it and `EnvironmentUpdate.name` already exists, so leaving it out means coming back here. Worth doing in the same change.
3. **A version that no longer exists.** An environment pinned at a version that was deleted has to render as something other than an empty select - "v3 (removed)" or similar, so it is clear why the agent is not answering.

**How you would know it was fixed.** Open History on an agent with six versions and two environments, change `dev` from v5 to v3 in its own row, reload: the row says `serves v3`, and a bot bound to `dev` runs v3.

Size: **S**.

DEENUU1:
## What

Each environment row in the History tab now carries its own version control — a
select of the agent's published versions showing what the row serves — so "dev
should serve v3" is answered on the row that states the property, instead of
scanning the Versions list for v3 and opening its **Promote to…** menu. That menu
stays: the two are two directions through the same edit, and picking a version in
the row calls the same `promote` mutation (`PATCH …/environments/{id}` with
`version_id`) that already existed.

Two more things on the row, per the issue, because leaving them out meant coming
back here:

- **Rename**, via the `EnvironmentUpdate.name` the backend already accepts. The
  default environment is not offered one — the service refuses it, its name being
  part of the publish contract — and the new `rename` mutation sends the name and
  nothing else, so a relabel can never silently repoint a pin.
- **A pin at a removed version renders as `v9 (removed)`** in the select's
  trigger (a disabled item carrying the stored number), never an empty control —
  that stale pin is exactly why the agent is not answering, so it must stay
  legible.

The default row keeps saying *"serves v6 — what publish repoints"* next to its
select, so a hand-pin there is visibly temporary.

## Why

Nobody arrives thinking "what should v3 do"; they think "dev should serve v3".
With six versions and three environments the old flow made you scan a list of
versions to change a property of an environment, and the row stating the property
was the one you could not edit. (#134)

## Verification

- Panel unit tests: pinning from the row calls promote with the row's environment
  and the picked version; the in-flight state disables the control; the removed
  pin renders its number; rename validates the backend's slug rule before the
  request leaves, submits on Enter, abandons on Escape/Cancel; the default gets
  neither rename nor removal; a caller without manage rights gets no select, no
  rename, no removal — not rendered, not rendered-then-403.
- Integration test over the real hooks and a mocked API client: the acceptance
  path from the issue (dev serves v5 → its row picks v3 → the PATCH carries
  `version_id` only → the refetched row says *serves v3*), rename carrying only
  `name`, and the not-rendered rule.
- Hook test: `rename` PATCHes the name and nothing else, and surfaces the
  refusal like the other three mutations.
- `make lint`, `make test-frontend-cov` (100%/97.5% gates), `make check` — green
  locally.

Docs owe nothing: `docs/concepts.md`'s environments paragraph is behaviour-level
("promoted onto it") and stays true; the promotion *direction* in the UI was
never described.

Closes #134


frontend/src/components/agents/environments-panel.tsx
```diff
@@ -1,31 +1,87 @@
...
+    <Select value={environment.version_id} disabled={promoting} onValueChange={onPromote}>
+      <SelectTrigger className="w-32" aria-label={t("pinVersionFor", { name: environment.name })}>
+        <SelectValue />
+      </SelectTrigger>
+      <SelectContent>
+        {!pinExists && (
+          <SelectItem value={environment.version_id} disabled>
+            {t("removedVersion", { version: environment.version })}
+          </SelectItem>
+        )}
```
DEENUU1: `pinExists` reads an empty history as "the pin is gone", and `pinStatus` in `src/lib/agent-spec.ts:268` already adjudicated that exact reading — its first guard says treating an empty list as gone "would flash the worst verdict this function has onto every row on every load". That's what happens here: the environments query and the versions query race, and when environments resolves first every row's trigger renders "vN (removed)" until `/versions` lands. The unit tests can't see it because the mock always has the versions loaded.

The second half is worse because it isn't a flash: the backend caps the history at 50 (`backend/app/repositories/agent.py:431`, and the route passes no override), so a pin older than fifty publishes is absent from a list that is complete as far as it goes — and this row calls it removed, permanently. `VERSION_HISTORY_LIMIT`'s docstring names the distinction: "this version was deleted" and "this version is off the end of the page I could read" have different fixes, and only one of them is true.

The pin's number is stored on the row, so the fallback item can stay legible without the false verdict:

```suggestion
  const pinListed = versions.some((version) => version.id === environment.version_id);
  const pinGone = !pinListed && versions.length > 0 && versions.length < VERSION_HISTORY_LIMIT;
  return (
    <Select value={environment.version_id} disabled={promoting} onValueChange={onPromote}>
      <SelectTrigger className="w-32" aria-label={t("pinVersionFor", { name: environment.name })}>
        <SelectValue />
      </SelectTrigger>
      <SelectContent>
        {!pinListed && (
          <SelectItem value={environment.version_id} disabled>
            {pinGone ? t("removedVersion", { version: environment.version }) : <>v{environment.version}</>}
          </SelectItem>
        )}
```

with `VERSION_HISTORY_LIMIT` imported from `@/lib/agent-spec`. Worth two unit cases alongside: an empty history renders the trigger as `v9`, not "(removed)", and a 50-long history missing the pin renders the same.

frontend/src/components/agents/environments-panel.tsx
```diff
@@ -34,27 +90,92 @@ export function EnvironmentsPanel({ agentId, canManage }: { agentId: string; can
...
 
   async function add() {
     await create.mutateAsync({ name: name.trim() });
     setName("");
   }
 
+  async function saveRename(environmentId: string, next: string) {
+    await rename.mutateAsync({ environmentId, name: next.trim() });
+    setRenaming(null);
+  }
```
DEENUU1: Opening the editor and hitting Enter without changing anything sends `PATCH {name: "staging"}` — the backend writes it and records an `agent.environment_renamed` audit entry for a non-change (`writable` drops nulls, not unchanged values, and the service only skips the *duplicate* check when the name is equal, not the write itself). Cheap to guard here, where the old name is in scope:

```suggestion
  async function saveRename(environment: AgentEnvironment, next: string) {
    const trimmed = next.trim();
    if (trimmed !== environment.name) {
      await rename.mutateAsync({ environmentId: environment.id, name: trimmed });
    }
    setRenaming(null);
  }
```

(the two call sites then pass `environment` instead of `environment.id`). Non-blocking — the request is harmless to the row — but an audit line saying someone renamed something nobody renamed is noise in the one place noise costs.

DEENUU1 (commented):
This lands #134 the way the issue framed it: the row that states the property is now the row that edits it, the promote mutation is reused rather than duplicated, and all three of the issue's "worth getting right" points are addressed — with tests on the refusals, not just the happy path. The integration test asserting each PATCH body carries only what changed is exactly the right assertion, and the backend side holds up: rename-of-default is refused server-side and the button isn't rendered, so the two halves agree.

One blocking thing: `VersionPin` reads "absent from the versions list" as "removed", and we already adjudicated that reading in `pinStatus` (`src/lib/agent-spec.ts:268`) — an empty list is a request in flight, and a full 50-long list may be truncated, because the backend caps the history at 50. As written, every row flashes "vN (removed)" on every load until `/versions` resolves, and a pin older than fifty publishes reads "(removed)" permanently. Details inline.

frontend/src/components/agents/environments-panel.tsx
DEENUU1: Fixed in 05fb63d3 — the panel now applies the same reading as `pinStatus`: the stored number renders plain, and "(removed)" appears only when the history is non-empty and shorter than VERSION_HISTORY_LIMIT, so neither the query race nor a truncated fifty-long history can produce the false verdict. Two unit cases pin it: an empty history renders the trigger as `v9`, and a fifty-long history missing the pin renders the same.

frontend/src/components/agents/environments-panel.tsx
DEENUU1: Fixed in 05fb63d3 — `saveRename` now takes the environment and closes the editor without a request when the trimmed name is unchanged, so Enter on an untouched name no longer writes the row or mints an `agent.environment_renamed` audit entry. A unit case pins it: submitting an unchanged name issues no PATCH.

PR #814, opened 2026-08-16
feat(sandboxes): split connections from what is running, into tabs

### LINKED ISSUE #140: Sandboxes: split connections from what is running, into tabs
`/sandboxes` is three cards stacked down a page: the connections table, then **Running on {host}**, then a policy dialog. The second is the one that reads badly, and the fix is mostly the page's structure rather than the panel's styling.

## Two screens sharing one scroll

`app/[locale]/(dashboard)/sandboxes/page.tsx:114-162` renders `ConnectionsCard` and then `SessionsPanel` beneath it. They are answering different questions on different clocks:

- **Connections** is configuration. It changes when somebody registers a host, which is rarely. Gated on `connections:manage`.
- **Running on docker** is live. It refetches every ten seconds, it has its own toggle for memory and CPU sampling, and each row expands an activity log *inside* the card (`sessions-panel.tsx:167`).

Stacked, the live half sits below a table of unknown height, and the expanding activity log pushes the page around under whoever is reading it.

**Split them into two tabs.** `components/ui/tabs.tsx` exists and is used elsewhere. One tab per question, the tab in the URL so a link lands where it was sent, and the sessions query only running while its tab is on screen — which also stops the ten-second poll on a page nobody is looking at.

## What is wrong inside the panel itself

- **The heading names a host somebody has not chosen.** `watching` (`page.tsx:71-74`) is *the default active Docker connection*, picked silently. On a deployment with three Docker hosts the panel shows one of them, and nothing on screen says the other two exist or are unreachable. With tabs the second tab should carry a host selector, defaulting to the same connection but saying which one it is.
- **The table nests a table.** `ActivityLog` renders a bare `<table>` in a grey box below the rows (`:198-215`), so expanding a session's activity puts an unstyled log under a styled table. It should be a detail panel beside or under the row it belongs to, on the same table primitive as everything else.
- **`sessions.length` is stated but not bounded.** The description says "N of this organization's sandboxes, out of M allowed" and the table is unsorted and unfiltered. With twenty containers running, "which one is the runaway" means reading twenty rows. That is the shared-table issue, and this table is a good first caller for it — sort by idle and by memory is the whole reason somebody opens this screen.
- **Padding is doubled.** `CardContent` has `p-5` and the table wrapper inside it adds `px-5 pb-2` (`:113-115`), so the table is inset twice as far as the heading above it. `ConnectionsCard` gets this right — `p-0` on the content, padding on the wrapper.

## Watch out

- The page's own `isLoading` branch renders a second copy of the header and the connections card (`page.tsx:76-88`). Splitting into tabs is the moment to stop duplicating that, not the moment to duplicate it once more.
- `SessionsPanel` returns `null` when there is no connection. Under tabs, an empty second tab needs a real empty state saying no Docker connection is registered — a tab that opens onto nothing is worse than the current silence.
- Sessions are filtered to the organization by the backend, deliberately (`sessions-panel.tsx:60-64`). Nothing about this change may move that filter into the client.

## Done when

- [ ] `/sandboxes` is two tabs — connections and running sandboxes — with the active tab in the URL
- [ ] The live query does not run while its tab is hidden
- [ ] The running tab names its host and lets an operator switch when several are registered
- [ ] The activity log is styled as part of the table, not a bare `<table>` in a grey box
- [ ] The sessions table sorts, at minimum by idle time and memory
- [ ] An empty running tab explains that no Docker connection is registered

DEENUU1:
`/sandboxes` stacked two screens on two clocks down one scroll: the connections
table (configuration, changes when somebody registers a host) and **Running on
{host}** (live, polling every ten seconds, expanding an activity log inside the
card that pushed the page around under whoever was reading it). This splits the
page into two tabs and fixes what was wrong inside the running panel itself.

## What changed

- **Two tabs, the active one in the URL.** `?tab=running` deep-links; the
  default (connections) keeps the parameter off so `/sandboxes` stays one URL.
- **The live query only exists while its tab is on screen.** The sessions panel
  is unmounted with its tab, so the ten-second poll stops on a page nobody is
  looking at. A page test pins that the sessions hook is never constructed
  while the connections tab is active.
- **The running tab names its host and lets an operator switch.** It used to
  silently show the default docker connection while nothing on screen said the
  other hosts existed. With several container connections registered there is
  now a host selector; Daytona and switched-off connections are excluded since
  neither can be asked for sessions.
- **A real empty state.** The old panel returned `null` with no docker
  connection; the running tab now explains that no container connection is
  registered and where to register one.
- **The sessions table sorts and filters.** Sort by session id, idle time and
  memory (unsampled rows last, on the shared `DataTable` primitive from #139),
  and a search box narrows on id, runtime and state — with a distinct "none
  match" message so a filtered-out table cannot be read as idleness.
- **The activity log is styled as part of the table** — `DataTable` with
  labelled columns (Operation / Target / Detail / Duration) instead of a bare
  `<table>` in a grey box, failed operations still marked.
- **The loading branch no longer renders a second copy of the header**, and the
  hardcoded English scope words ("one conversation", "a single run") moved into
  the message catalog.

The organization filter stays server-side — nothing about the split moves it
into the client.

## Verification

- `sessions-panel.test.tsx` (28 tests) and a new page test (8 tests) cover the
  tab split, the not-mounted-while-hidden behaviour, the host selector, search,
  sorting, the empty running tab, and the permission-gated add button.
- `make check` green locally: lint, backend suite (5257 passed, platform layer
  at 100%), db-check (`alembic check` against a throwaway database — the shared
  dev database is stamped by an unrelated local branch), frontend coverage gate
  (100% lines/statements/functions, 97.56% branches), production build,
  docs-build, audit.
- `docs/configuration.md` updated where it described the panel as reading only
  the default host.

Closes #140


frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx
```diff
@@ -78,36 +75,70 @@ export default function SandboxesPage() {
...
+        <TabsContent value="running">
+          {isLoading ? (
+            <div className="space-y-3 p-5">
+              {[0, 1].map((row) => (
+                <Skeleton key={row} className="h-10 w-full" />
+              ))}
+            </div>
+          ) : (
+            <SessionsPanel connections={watchable} />
+          )}
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx:140]: When the connections request fails, `useSandboxConnections` answers `connections: []` with `error` set — so `watchable` is empty and this tab tells the operator "No container connection registered". That's a false statement about configuration when the truth is "the request didn't answer": hosts may well be registered and running things. The connections tab handles exactly this case with `ErrorState`; the running tab needs the same branch.

```suggestion
          {isLoading ? (
            <div className="space-y-3 p-5">
              {[0, 1].map((row) => (
                <Skeleton key={row} className="h-10 w-full" />
              ))}
            </div>
          ) : error !== null ? (
            <ErrorState description={error} className="m-5" />
          ) : (
            <SessionsPanel connections={watchable} />
          )}
```

And a page test to pin it: `state.error` set, open the running tab, assert the error is visible and "No container connection registered" is not.

DEENUU1 (commented):
The split itself is right and the tests pin the parts that matter — especially that the sessions hook is never even constructed while the connections tab is active, which is the whole point of the two-clocks argument. The tab-in-URL with the default kept off is a clean shape too.

Two real things to fix before merge, both small: the running tab renders the "no container connection registered" empty state when the connections request *failed* — the exact empty-vs-error ambiguity this repo's own frontend rule (and the connections tab, three lines up) guards against. And an open activity log survives a host switch, so the panel asks the new host for the old host's session.

One non-blocking follow-through: `src/lib/dashboard/sandbox.ts` already names these predicates (`watchableConnections`, `holdsSessions`, `primaryConnection`) for the #129 dashboard half — the page and panel re-inline both, which is precisely the drift the #129/#140 pairing was supposed to prevent.

frontend/src/components/sandboxes/sessions-panel.tsx
```diff
@@ -147,15 +196,38 @@ export function SessionsPanel({ connection }: SessionsPanelProps) {
               : t("sessionsCount", { count: sessions.length })}
           </CardDescription>
         </div>
-        <label className="flex items-center gap-2 text-xs">
-          <span className="text-muted-foreground">{t("sampleMemory")}</span>
-          <Switch checked={usage} onCheckedChange={setUsage} aria-label={t("sampleMemoryAndCpu")} />
-        </label>
+        <div className="flex items-center gap-4">
+          {connections.length > 1 && (
+            <Select value={connection.id} onValueChange={setChosenId}>
```
DEENUU1: `watching` holds a session id, and session ids are per-host — but it survives a host switch. Open the activity of `xc-1` on Host A, pick Host B here, and the panel below the table asks Host B for `xc-1`'s log: an error (or worse, a confusing answer) about a session that is no longer in the table above it, with no row relating the log to anything. Close the log when the host changes:

```suggestion
            <Select
              value={connection.id}
              onValueChange={(id) => {
                setChosenId(id);
                setWatching(null);
              }}
            >
```

(A session vanishing from the listing mid-watch has the same shape, but that one predates this PR — the host switch is what's new here.)

frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx
```diff
@@ -37,32 +44,22 @@ export default function SandboxesPage() {
...
-            ))}
-          </div>
-        </ListCard>
-      </div>
-    );
+  // Daytona holds no sessions of ours to enumerate, so only container
+  // connections can be watched at all.
+  const watchable = connections.filter(
+    (connection) => connection.is_active && connection.kind === "docker",
+  );
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx]: Non-blocking, but worth doing while the files are open: `src/lib/dashboard/sandbox.ts` already names these predicates for the dashboard half of this same data layer — `watchableConnections` (active), `holdsSessions` (`kind === "docker"`, with the note explaining why), and `primaryConnection` (default, else first active). This filter re-inlines the first two:

```tsx
import { holdsSessions, watchableConnections } from "@/lib/dashboard/sandbox";

const watchable = watchableConnections(connections).filter(holdsSessions);
```

and the panel's default-host pick (`sessions-panel.tsx:84-88`) re-inlines the third:

```tsx
const connection =
  connections.find((entry) => entry.id === chosenId) ?? primaryConnection(connections);
```

Two copies of "which host can be watched" is exactly the drift the #129 comment on #140 warned about — a third connection kind, or a change to what "watchable" means, now has to land in both places.

frontend/messages/en.json
```diff
@@ -3523,7 +3524,21 @@
       "sessionsOfLimit": "{count} of this organization's sandboxes, out of {limit} allowed. Refreshed every ten seconds.",
       "sharedBy": "Shared by",
       "show": "Show",
-      "state": "State"
+      "state": "State",
+      "host": "Host",
```
DEENUU1: nit: the new keys are appended after `state` while the rest of this block (and the `pages.sandboxes` block this same diff touches) is alphabetical. Nothing gates it, but the next person adding a key will follow whichever convention they see.

frontend/src/app/[locale]/(dashboard)/sandboxes/page.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+    expect(screen.queryByText("Local Docker")).toBeNull();
+  });
+
+  it("shows the refusal where the rows would be, not an empty collection", () => {
+    state.error = "Failed to load sandbox connections";
+    render(<SandboxesPage />);
+
+    expect(screen.getByText("Failed to load sandbox connections")).toBeVisible();
+    expect(screen.queryByText(/No sandbox connections yet/)).toBeNull();
+  });
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/sandboxes/page.test.tsx:154]: This is the right assertion — recording every construction of the hook and pinning the list empty proves *unmounted*, not merely hidden, which is the actual claim the PR makes about the ten-second poll. A visibility assertion would have passed with the poll still running.

frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx
Comment (DEENUU1) [reply frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx:140]: Fixed in e183ada9 — the running tab now shows the same ErrorState as the connections tab when the list is refused, with a page test asserting the false empty state never renders.

frontend/src/components/sandboxes/sessions-panel.tsx
DEENUU1: Fixed in e183ada9 — choosing a host now closes the open activity log, and the panel test records (connection, session) pairs so the cross-host ask is exactly what it refuses.

frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx
Comment (DEENUU1) [reply frontend/src/app/[locale]/(dashboard)/sandboxes/page.tsx]: Fixed in e183ada9 — both call sites now use the lib predicates. Semantics checked: watchableConnections + holdsSessions is exactly the inlined is_active && kind === "docker", and primaryConnection only differs by pre-filtering to active rows, which the page already guarantees for the panel's input.

frontend/messages/en.json
DEENUU1: Fixed in 54e45444 — the sessions block is alphabetical again; pages.sandboxes already was, and pl.json holds none of these keys so there was nothing to mirror.

Issue #138, opened 2026-08-04
All files: no search, no sort, six icons and no preview


DEENUU1:
The **All files** toggle on `/workspaces` answers the one question the per-workspace table cannot: who is holding a copy of that CSV. `FlatFiles` in `components/sandboxes/workspace-browser.tsx:205-290` renders it, and as a way to *find* a file it does not work.

## What a tile is today

A three-line row in a rounded border (`:228-271`): a 16-pixel icon, the full path in 12px monospace truncated to the column, then `agent · access · size` in 11px, and a download button. Two lines of content in a box tall enough for two lines of content, three across.

Three things are missing, and each one is the thing somebody is actually doing on this screen.

## 1. There is no search and no sort

`FlatFiles` has neither. `useAllWorkspaceFiles(true)` returns every file the reader can see, bounded by however many workspaces the backend got through, and they arrive in whatever order the listing produced. Finding `report.csv` means reading the grid.

Meanwhile `WorkspaceExplorer` — the single-workspace view, the one with *fewer* files — has a search box. The screen where search matters more is the screen without one.

Both exist as primitives already: `useListControls` + `SearchInput` in `components/ui/list-controls.tsx`, used by five galleries. Filter on the path, the agent name and the extension. Sort by name, size, modified time and agent.

`listing.truncated` is the part to get right: when the backend stopped early, a client-side filter is searching a *sample*, and the line at `:281-287` that admits the bound today has to stay visible while filtered, not be replaced by "3 results".

## 2. Every file gets the same six icons

`FileIcon`/`kindOf` (`components/sandboxes/file-tile.tsx`) maps ~30 suffixes onto `image | code | sheet | doc | archive | text`. So `.py`, `.ts`, `.json`, `.sql` and `.sh` are one monochrome `FileCode`, and anything unlisted — `.parquet`, `.ipynb`, `.env`, `.pdf` versus `.docx` — is a plain `FileText`. In a grid of thirty files, the icon column carries almost no information.

Either a wider, colour-coded set keyed on extension, or drop the icon and put the extension itself in a small badge — a `PY` chip says more than a grey document glyph and never has a missing case. Whichever way, it should be the *same* decision as the file icons in the chat panel and the composer; see the file-viewer standardization issue, which has to unify `FileIcon` and `FileKindIcon` anyway.

## 3. A tile shows no preview

[...]

PR #815, opened 2026-08-16
feat(api): carry a workspace file's modified time to the viewer

### LINKED ISSUE #500: A workspace file has no modified time, so the file viewer's header cannot show one
The shared file viewer's header reads `MD · 3 B · modified 2 minutes ago`, and
the third part only ever appears for a knowledge base document. A workspace
file — which is most of what people open, from the chat panel and from
`/workspaces` — shows `MD · 3 B` and nothing about when it changed.

Not a bug in the viewer: `ViewerFile.modifiedAt` exists, the callers that have
a time pass it (`kb/file-viewer.tsx` passes `doc.created_at`), and the header
renders it when it is there. The listing simply has no time to give.

## Where it stops

`WorkspaceFileRead` in `backend/app/schemas/workspace.py:14` carries `path`,
`size` and `is_dir`. It is built from `FileInfo` in `pydantic_ai_backends`,
whose fields are exactly `name`, `path`, `is_dir`, `size` — no mtime, for any
of the three backends. So this is not a column somebody forgot to serialize:

- `state` — the workspace is a JSONB document in this database. Whether a
  per-file mtime is even recorded there needs checking before anything else.
- `docker` / `daytona` — the file is on a host volume, and `sandboxd` would
  have to report `st_mtime` through `FileInfo` for the API to have it.

So the chain is: `pydantic_ai_backends.FileInfo` gains an optional
`modified_at`, the three backends fill it where they can, `WorkspaceFileRead`
carries it, and `WorkspaceFile` / `ConversationFile` in
`frontend/src/lib/{sandbox-workspaces,conversation-workspace}-api.ts` pass it
into `ViewerFile.modifiedAt`. The frontend half is one line per caller.

`pydantic_ai_backends` is upstream of this repo, which is why this is filed
rather than fixed in #136: it is a library change plus a schema change, not a
viewer change.

Open a file the agent wrote from the chat panel. The header reads
`MD · 3 B · modified 2 minutes ago` rather than stopping at the size — and an
integration test asserts the listing carries a time for a `state` workspace,
since that is the backend every deployment has.

Optional and worth deciding at the same time: `null` is the honest answer for
a backend that cannot report one, and the header already omits the part when
it is absent. Nothing needs a fallback.

Found finishing #136, which is where the header was built. Same cluster:
#137, #138 (the listing side of the same data), #133.

DEENUU1:
The shared file viewer's header reads `MD · 3 B · modified 2 minutes ago` — and the third part only ever appeared for a knowledge base document. A workspace file, which is most of what people open, stopped at the size. Not a viewer bug: `ViewerFile.modifiedAt` existed and the header rendered it; the listing simply had no time to give.

**Upstream first.** `pydantic-ai-backend` 0.2.26 ([vstorm-co/pydantic-ai-backend#104](https://github.com/vstorm-co/pydantic-ai-backend/pull/104), released) adds `FileInfo.modified_at` (ISO 8601, optional), filled where a real timestamp exists: a stored workspace records one on every write inside its JSONB document, a workspace archive reports `st_mtime` through the wire, and a live container's shell listing honestly reports none — `null` is the answer there, never a guess. `wire.FileEntry.modified_at` defaults to `None`, so a client and a sandboxd on either side of the release keep understanding each other.

**This PR carries it through:**

- Bumps the pin to `>=0.2.26`.
- `WorkspaceFileRead` (and `FlatFileRead` through it) gains `modified_at: datetime | None` — the flat all-files listing gets the field now because #138 sorts on it.
- The three listing routes pass `entry.get("modified_at")` through.
- `WorkspaceFile` / `ConversationFile` carry it into `ViewerFile.modifiedAt` in the chat panel, the workspace explorer and the flat browser.

**Verified:**

- An integration test proves the timestamp survives the JSONB round trip through Postgres for a `state` workspace — the backend every deployment has.
- An API test proves a row carrying a time serializes it and a legacy row serializes `null`.
- A panel test proves the header now says `modified …` for a file the agent wrote, opened from the chat panel.
- `make check` green (lint, backend 5281 tests at 100% platform coverage, frontend gate, db-check, builds, audit).

No docs page describes listing rows field by field; the schema docstring carries the reasoning.

Closes #500


backend/pyproject.toml
```diff
@@ -78,7 +78,7 @@ dependencies = [
     # `console` for the toolset, `remote` for the sandboxd client. Deliberately
     # not `docker`: this container never touches the socket, so the Docker SDK
     # has no business in the image. The service runs beside us as `sandboxd`.
-    "pydantic-ai-backend[console,remote]>=0.2.25",
+    "pydantic-ai-backend[console,remote]>=0.2.26",
```
DEENUU1, line 83: nit: the neighbouring pins document why their floor is where it is ("Floors at 0.2.16 for three fixes..."), and this one explains the extras but not the floor. Worth one line so the next person bumping or loosening it knows what 0.2.26 bought:

```suggestion
    # 0.2.26 is the floor because it adds `FileInfo.modified_at`, which the
    # listings carry to the viewer (#500).
    "pydantic-ai-backend[console,remote]>=0.2.26",
```

DEENUU1 (commented):
Clean carry-through: upstream field first, schema in the middle, one line per viewer call site, and tests at each layer (the JSONB round trip is the right one to prove — that's the backend every deployment actually runs). The null-honesty design is right too: a live container's listing reporting nothing beats it guessing "just now".

Two things the diff doesn't show, both found by sweeping the sibling surfaces:

1. **`frontend/src/components/chat/tool-results/workspace.tsx:96` — the tool-result file card still drops the time.** `WorkspaceFileCard` already has the listing row in hand (it reads `entry?.size` on the same line) but builds the viewer file without `modifiedAt`. So the most immediate surface — the "agent just wrote this file" card in the chat — keeps the old `MD · 3 B` header, while opening the same file from the workspace panel now shows the time. One line, same shape as the other three call sites:

```tsx
file={{ name, path: target, size: entry?.size ?? null, modifiedAt: entry?.modified_at ?? null }}
```

It's the exact class this PR fixes, so I'd fold it in here rather than file it — the test would mirror the new panel test.

2. **nit: `frontend/src/components/files/file-viewer.tsx:196` — the comment above `describe()` is now stale.** It says "a workspace listing carries no modification time at all", which this PR makes false for stored workspaces. The point it's making (every part can be missing on some surface) still holds via the live-container case, so it's a one-word-level reword, but worth doing in the same change since this PR is what invalidates it.

backend/tests/integration/test_workspace_visibility.py
```diff
@@ -220,3 +220,25 @@ async def test_and_never_another_organizations(self, db) -> None:
...
+        member = await _user(db)
+        agent = await _agent(db, organization, member)
+        stored = StateBackend()
+        stored.write("/report.md", "# findings")
+        workspace = await _workspace(db, organization, agent, files=dict(stored.files))
+
+        await db.refresh(workspace)
+        (entry,) = stored_entries(dict(workspace.files))
+
+        assert entry.get("modified_at") == stored.files["/report.md"]["modified_at"]
```
DEENUU1, line 244: Nice choice of layer — asserting equality against the backend's own recorded value (rather than a hardcoded string) means the test proves the round trip without coupling to the upstream timestamp format.

backend/pyproject.toml
DEENUU1: Fixed in 001fd730 — the pin now says what 0.2.26 bought: `FileInfo.modified_at`, which the listings carry to the viewer (#500).

DEENUU1:
Both review-body findings landed in 001fd730: the tool-result file card now hands the viewer `modifiedAt` from the same listing row it already reads the size from, with a unit test mirroring the panel one — and the `describe()` comment is reworded to blame the live container listing, which is the case that still holds.

Issue #676, opened 2026-08-13
A failed agent run still stores the model client's exception text on the run row


DEENUU1:
**Severity: medium** · pre-existing, found while fixing
[#659](https://github.com/vstorm-co/agenticos/issues/659) · the half of
[#437](https://github.com/vstorm-co/agenticos/issues/437) that never landed

## What's wrong

A failed agent run still stores the model client's own exception text on the run
row:

| | |
|---|---|
| `backend/app/services/agent_chat.py:391` | `error = str(exc)` → `finish(..., error=error)` |
| `backend/app/services/agent_runner.py:3186` | `error = str(exc)` → the same `finish` |

`agent_runs.error` is a `Text` column, it is on `AgentRunRead`
(`app/schemas/agent_run.py:51`), and it reaches the frontend's run types — so any
member who can read run history is served whatever the provider client said.
What raises there is a Pydantic AI model client, `httpx` underneath, whose
message routinely carries the failing request URL; for a profile with a custom
`base_url` that URL is a tenant's own endpoint, with the key still in its query
string. Same principle as
[#342](https://github.com/vstorm-co/agenticos/issues/342) (an HTTP body),
[#423](https://github.com/vstorm-co/agenticos/issues/423) (the ingestion
columns) and #659 (the chat frame): a refusal describes the refusal, not the
server.

`BudgetExceeded` on the line above is not the problem — that message is ours and
naming the ceiling is the point of it.

## Why this is a new issue rather than #437

#437 filed all three call sites together — the two run-row writes and the chat
socket. It was **closed on 2026-08-07 with no pull request and no comment**, and
the socket half was re-filed later as #659, which is fixed by
[PR #675](https://github.com/vstorm-co/agenticos/pull/675). The run-row half was
never touched and is now tracked by a closed issue, which means nothing surfaces
it. This is that remainder, and nothing more.

## The sequence

1. Give an organization a model profile with a custom endpoint and a key.
2. Break the key (rotate it, or let it expire).
3. Run the agent. `SELECT error FROM agent_runs WHERE status = 'failed'` has the
   provider's 401 verbatim, and the runs UI renders it.

## How you would know it was fixed

The run row carries a controlled sentence — what stopped the run and what the
person can do about it — with the client's own text in the `logger.exception`
already sitting on both call sites. `app/services/rag/failures.py` (#423) and
`_turn_failed` in `app/services/agent_session.py` (#659) are the shape to copy;
[...]

PR #816, opened 2026-08-16
fix(agents): compose a delegated run's stored error, never store the library's

### LINKED ISSUE #699: A delegated run's row still stores the subagent library's exception text
**Severity: medium** · pre-existing, found reviewing #676 (PR #691)

## What's wrong

#676 fixes the two places that write `agent_runs.error` for a run this platform
executes itself. A **delegated** run's row is written somewhere else and still
holds a foreign exception's text verbatim:

`backend/app/services/agent_runner.py:2360` builds `RecordedDelegation(...,
error=outcome.error)` and `:2429` writes that straight onto the child's
`agent_runs` row. `outcome.error` is `handle.error`, set at
`backend/app/agents/capabilities/subagents/_journal.py:940`, and the handle comes
from the `subagents-pydantic-ai` library, which composes it as:

```python
handle.error = f"Transient error (retry {attempt}): {exc}"
```

(`subagents_pydantic_ai/_execution.py:328`.) `{exc}` there is whatever the
delegate's model client raised - the same Pydantic AI / `httpx` message #676 is
about, whose text routinely carries the failing request URL, with the key still in
its query string for a profile on a custom `base_url`.

So the parent run row is now controlled and the child's is not, on the same page,
for the same failure. `run_failure_summary` in `agent_runner.py` is the sentence
this should reach; the string simply never passes through it.

## Why it is a separate issue rather than part of #676

The fix is in a different place and has a different owner. #676's two call sites
each catch an exception and compose the row's text; this path receives a
**string** that a third-party library already composed, so nothing here has the
exception to diagnose. The two candidate fixes are:

1. Sanitise at `_journal.py:940` - which means parsing a string the library built,
   or asking the handle for the exception it wrapped.
2. Fix it upstream in `subagents-pydantic-ai` so the handle carries the exception
   rather than a formatted string, and compose the sentence here.

(2) is the honest one and is a release of Kacper's own library, so it is not
something to fold into a branch about the run row.

## The sequence

1. Give an organization a model profile with a custom endpoint and a key.
2. Break the key.
3. Run an agent that delegates to a subagent on that profile.
4. `SELECT error FROM agent_runs WHERE parent_run_id IS NOT NULL AND status =
   'failed'` has the provider's text, and the runs UI renders it.

The delegated row carries the same controlled sentence its parent does, and the
library's text is in the log only. A test alongside #676's asserting a vendor
string does not reach `RecordedDelegation.error`.

## Related

Cluster: **Security and tenancy** on the map in #168, on the
#342 → #423 → #437/#659 spine. It is the fifth live descendant, beside #659 (chat
`error` frame, PR #675), #681 (chat `tool_result` frame, PR #694), #676 (the run
row, PR #691) and #695 (the transcript tool-call row). Same rule, five surfaces -
worth doing after the others land, since the sentences they each compose want to
end up in one helper rather than five.

DEENUU1:
## What

A delegated run's `agent_runs.error` and its closing `SubagentFinished` frame no longer store the subagents library's exception text. The settlement in `_journal.py` composes the same controlled sentence the parent's row gets — `run_failure_summary` — and the library's text goes to the server log with the original exception beside it.

- **Upstream first**: subagents-pydantic-ai 0.2.19 (vstorm-co/subagents-pydantic-ai#80, released to PyPI) adds `TaskHandle.exception`, so the platform receives the exception rather than parsing a string the library formatted. The pin moves to `>=0.2.19`.
- `run_failure_summary` moved from `app/services/agent_runner.py` to `app/agents/failures.py`: the capability layer could not import it from a service without inverting the layering. It now takes `BaseException`, which is what the library's retry callback receives. Both former importers (`agent_runner`, `agent_chat`) repoint; behaviour unchanged.
- `UsageLimitExceeded` keeps the library's text whole, deliberately: it is the delegation's `BudgetExceeded` — a ceiling doing its job, its message is pydantic-ai's own limit sentence, and the `usage limit exceeded` marker is the library's telemetry contract across both dispatch modes.
- `docs/governance.md` gains a paragraph in the child-row section saying what a failed child row may store.

## Why

#676 fixed the two places that write `agent_runs.error` for a run this platform executes itself. A delegated run's row was written from `handle.error`, which the library composes as `Transient error (retry N): {exc}` — `{exc}` being whatever the delegate's model client raised, whose text routinely carries the failing request URL with the key still in its query string on a custom `base_url`. The parent's row was controlled and the child's was not, on the same page, for the same failure.

## Verification

- Regression test (`test_a_provider_failure_reaches_the_row_as_our_sentence_not_its_own`) drives a delegation into a provider-style crash carrying a keyed URL and asserts the row and the frame store the controlled sentence while the vendor text lands only in the log. It fails without the fix.
- Full backend suite: 5279 passed, platform layer at 100%.
- `make lint`, `make db-check` clean; `make check` run before opening.

The upstream release is verified live: PyPI serves 0.2.19 and the lock resolves it.

Closes #699

### DIFF
--- a/backend/app/agents/capabilities/subagents/_journal.py
+++ b/backend/app/agents/capabilities/subagents/_journal.py
@@ -78,7 +78,7 @@ from subagents_pydantic_ai import (
 )
 from subagents_pydantic_ai.registry import DynamicAgentRegistry
 
-from app.agents.capabilities.budget import SpendShare, booked_to
+from app.agents.capabilities.budget import BudgetExceeded, SpendShare, booked_to
 from app.agents.capabilities.subagents._events import FrameLabels, frame_for
 from app.agents.deps import AgentDeps
 from app.agents.failures import run_failure_summary
@@ -165,13 +165,20 @@ def _recorded_error(handle: TaskHandle, subagent: str, task_id: str) -> str | No
     `run_failure_summary` - the one the parent's row gets - and the original goes
     to the log beside it.
 
-    `UsageLimitExceeded` keeps the library's text whole. It is the delegation's
-    `BudgetExceeded`: not a malfunction but a ceiling doing its job, its message
-    is pydantic-ai's own limit sentence - a number, never a request - and the
-    `usage limit exceeded` marker in front of it is the library's telemetry
-    contract across both dispatch modes.
+    The two ceilings keep the library's text whole. A ceiling is not a
+    malfunction, and either message is controlled end to end: `UsageLimitExceeded`
+    carries pydantic-ai's own limit sentence - a number, never a request - behind
+    the `usage limit exceeded` marker that is the library's telemetry contract
+    across both dispatch modes, and `BudgetExceeded` carries this platform's own
+    ceiling sentence behind the library's type-name prefix. `BudgetExceeded`
+    reaches this settlement because a delegate's requests are budget-checked
+    inside the delegate's own run, so the breach lands on the handle rather than
+    propagating to the caller the parent's row is written by - and composing it
+    away would tell the reader to retry against a breached budget.
     """
-    if handle.exception is None or isinstance(handle.exception, UsageLimitExceeded):
+    if handle.exception is None or isinstance(
+        handle.exception, UsageLimitExceeded | BudgetExceeded
+    ):
         return handle.error
     logger.warning(
         "Delegation %r (task %s) failed: %s",
--- a/backend/app/agents/capabilities/subagents/_journal.py
+++ b/backend/app/agents/capabilities/subagents/_journal.py
@@ -65,6 +65,7 @@ from typing import Any, Literal
 from uuid import UUID, uuid4
 
 from pydantic_ai.agent import EventStreamHandler
+from pydantic_ai.exceptions import UsageLimitExceeded
 from pydantic_ai.messages import AgentStreamEvent
 from pydantic_ai.tools import RunContext
 from subagents_pydantic_ai import (
@@ -80,6 +81,7 @@ from subagents_pydantic_ai.registry import DynamicAgentRegistry
 from app.agents.capabilities.budget import SpendShare, booked_to
 from app.agents.capabilities.subagents._events import FrameLabels, frame_for
 from app.agents.deps import AgentDeps
+from app.agents.failures import run_failure_summary
 from app.agents.spec import DelegationMode
 from app.agents.subagent_events import (
     SpecialistDefinition,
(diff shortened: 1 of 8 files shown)


backend/app/agents/capabilities/subagents/_journal.py
```diff
@@ -148,6 +150,39 @@ def _terminal_status(handle: TaskHandle, mode: Literal["sync", "async"]) -> Dele
...
+    `run_failure_summary` - the one the parent's row gets - and the original goes
+    to the log beside it.
+
+    `UsageLimitExceeded` keeps the library's text whole. It is the delegation's
+    `BudgetExceeded`: not a malfunction but a ceiling doing its job, its message
+    is pydantic-ai's own limit sentence - a number, never a request - and the
+    `usage limit exceeded` marker in front of it is the library's telemetry
+    contract across both dispatch modes.
+    """
+    if handle.exception is None or isinstance(handle.exception, UsageLimitExceeded):
```
DEENUU1: This branch forgets the second ceiling. A delegate's own model requests are budget-checked inside the delegate's run — `BudgetGuard.for_delegate` shares the ledger, and `wrap_model_request` calls `_assert_within_budget` before every request — so a delegate stopped by the parent's cap raises `BudgetExceeded` inside the library's `except Exception`. It isn't in `_ALWAYS_PROPAGATE` and it isn't an `AppException`, so it lands here as `handle.exception` and falls through to `run_failure_summary`, which stores:

> The run did not finish (BudgetExceeded) - retry it, and check the agent's model profile if it keeps failing.

That's a regression against main for this case: before this PR the row stored `BudgetExceeded: Agent monthly budget exhausted: $0.0500 spent of $0.05 limit` — every word controlled (the library's type-name prefix plus our own ceiling sentence), the numbers intact, and no "retry it" invitation on a breached budget. It also fires the WARNING with `exc_info` for a ceiling doing its job. The docstring above already makes the argument — "not a malfunction but a ceiling doing its job" — it just names only one of the two ceilings.

```suggestion
    if handle.exception is None or isinstance(handle.exception, UsageLimitExceeded | BudgetExceeded):
```

`BudgetExceeded` imports from `app.agents.capabilities.budget`, which this module already imports `SpendShare` and `booked_to` from. Wants a test beside the new one: a delegate driven into the ceiling, asserting the row and the frame keep the ceiling sentence with its numbers.

backend/app/agents/failures.py
```diff
@@ -0,0 +1,61 @@
...
+    callback receives (#699).
+
+    Ours is kept whole. An `AppException` raised inside the run is written in
+    this repository, and its message is the most useful thing an operator can be
+    shown - "No model profile is configured for this agent" beats any sentence
+    composed here. (The one place that folds a foreign `__str__` into an
+    `AppException` is `sandbox_workspace._reason`, deliberately and for a route's
+    answer; it runs outside both `try` blocks that call this.) `BudgetExceeded`
+    never reaches this function: it is caught above and its ceiling is the point
+    of it.
```
DEENUU1: Two claims here went stale with the move. "Both `try` blocks that call this" undercounts the callers now — the journal's settle path is a third, and it isn't a `try` block. And "`BudgetExceeded` never reaches this function" is no longer something the callers' shape guarantees: the delegation path *can* hand it in (see my comment in `_journal.py`), so once `_recorded_error` filters it, the sentence should say what each caller owes rather than assert a fact one forgotten caller can break:

```suggestion
    answer; it runs in a route service, outside every run a caller settles.)
    `BudgetExceeded` must not reach this function - each caller catches or
    filters it first, because its ceiling sentence is the point of it.
```

DEENUU1 (commented):
This is the right shape — upstream first, the handle carrying the exception instead of a string to parse, and the child's sentence composed by the same helper the parent's row uses. I verified the branch matrix against the 0.2.19 source: `finish()` overwrites `error` and `exception` on the winning transition, so a completed-after-retry handle can't leak the retry text, and all three `_recorded_error` branches have a test asserting the consequence.

One thing to fix before merge: `BudgetExceeded` is the second ceiling this path can settle, and it now gets mangled into "retry it" — a regression against main for that one case. Details inline at `_journal.py:174`.

Two pre-existing siblings, neither caused here: I filed #819 for `check_task`/`wait_tasks` returning `handle.error` verbatim into the parent's transcript (a `ToolReturnPart` is stored whole, so #695's scrub can't reach it — same keyed-URL payload, one surface over), and the `test` job was still pending when I reviewed, so worth a `gh pr checks 816` before merging.

backend/app/agents/capabilities/subagents/_journal.py
DEENUU1: Fixed in 3fe6efda — both ceilings keep the library's text whole now, and the regression test drives a delegation into `BudgetExceeded` and asserts the row and the closing frame keep the ceiling sentence with its numbers. It fails without the `_journal` change.

backend/app/agents/failures.py
DEENUU1: Reworded in 3fe6efda as suggested — the docstring now says what each caller owes instead of asserting a fact one forgotten caller could break.

PR #817, opened 2026-08-16
fix(capabilities): prune a run's spills off a container workspace at close

### LINKED ISSUE #803: Prune tool_output_limits spills at run end on non-run-scoped workspaces

DEENUU1:
## What this fixes

Closes #803 — the container half. #804 already stopped spills outliving the run on a `state` backend (the flush strips the reserved prefix); a longer-scoped **container** workspace (`conversation`/`user`/`agent`) still kept every past run's `tool_output/` blobs on its filesystem forever.

## How

- **The overflow store records what it writes.** `BackendOverflowStore` appends each handle (the backend's own normalized path) to a per-run spill log. The runner registers that log beside the workspace backend under `SPILL_LOG_RESOURCE`, and `_delegate_resources` shares it with delegates that share the parent's `sandbox` — their spills land on the same filesystem, so their handles belong in the same list. The in-memory fallback backend records nothing: it dies with the run and leaves nothing to delete.
- **`SandboxWorkspaceService.close` deletes exactly those paths.** For a container-backed workspace whose scope outlives the run, close runs `rm -f -- <handles>; rmdir -- <emptied dirs, deepest first> 2>/dev/null; true` through the backend's own `execute` — the backend protocol exposes no delete, and growing one belongs upstream. Every path is checked against the reserved-prefix invariant before it reaches the shell, even though only the store appends to the list.

## Why exact handles, not a prefix sweep

Two concurrent runs share a `user`- or `agent`-scoped workspace. A `tool_output/` prefix delete at one run's close would take the other run's spills mid-flight, while its `read_tool_result` can still be asked for them — the race #804 declined to rush. A run deleting only the handles its own store recorded cannot race anybody; the `rmdir` of the shared spill directory fails silently while a concurrent run still keeps files, which is the correct answer.

## Still true

- **Best-effort, like `_release`.** A run that crashes before its `finally` leaves its spills for the host; a prune the host refuses is logged (`workspace_spill_prune_failed`) rather than raised, because `close` runs in the same `finally` that records the run's cost.
- `run`-scoped containers die purged as before; `state` workspaces keep the flush-strip from #804.

## Testing

- `tests/test_sandbox_workspace.py` — the prune command (both handle forms, `rmdir` deepest-first ordering), the foreign-path guard running nothing, no spills → no command, a backend without `execute` left alone, and a failed prune logged rather than raised.
- `tests/test_tool_output_limits.py` — written handles recorded in the log (in write order, as returned), a refused write recording nothing, the in-memory fallback recording nothing, and the registry-level path: a spill through a built capability lands its handle in the resource-provided log.
- `make test` — 5422 passed, platform coverage **100.00%**. `make lint-backend`, `make docs-build` (--strict), `python3 scripts/docs_drift.py` clean.

Closes #803

### DIFF
--- a/backend/app/services/sandbox_workspace.py
+++ b/backend/app/services/sandbox_workspace.py
@@ -24,9 +24,11 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import shlex
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from pathlib import PurePosixPath
 from typing import Any
 from uuid import UUID
 
@@ -131,6 +133,17 @@ class OpenWorkspace:
     container-backed one, which keeps its files on the host rather than here.
     """
 
+    spills: list[str] = field(default_factory=list)
+    """Every spill handle `tool_output_limits` wrote to this workspace this run.
+
+    Registered beside the backend under `SPILL_LOG_RESOURCE`, appended by the
+    overflow store, and read back by `close`: a container-backed workspace whose
+    scope outlives the run deletes exactly these paths, so a spill never survives
+    the run that produced it (#803). Exact handles rather than the `tool_output/`
+    prefix, because concurrent runs share a longer-scoped workspace and a prefix
+    sweep would take another run's spills mid-flight.
+    """
+
 
 def sandbox_config(spec: AgentSpec) -> SandboxConfig | None:
     """This spec's workspace configuration, or `None` if it has no workspace.
@@ -164,6 +177,36 @@ def _without_spills(files: dict[str, Any]) -> dict[str, Any]:
         return relative == OVERFLOW_PREFIX or relative.startswith(f"{OVERFLOW_PREFIX}/")
 
     return {path: data for path, data in files.items() if not _is_spill(path)}
+
+
+def _under_overflow(path: str) -> bool:
+    """Whether this path is a file inside the reserved spill directory.
+
+    The delete in `_prune_spills` runs a shell command over paths from a list, so
+    even though only the overflow store appends to that list, every path is checked
+    against the one invariant that makes the command safe: it removes spills and
+    nothing else. Depth-agnostic because a backend normalizes handles its own way -
+    `/tool_output/…` from `state`, an absolute in-container `/workspace/tool_output/…`
+    from a service. The prefix must be a proper ancestor (`parts[:-1]`), which is
+    what every handle the store writes has, and what guarantees `_overflow_parents`
+    answers at least the spill directory itself.
+    """
+    return OVERFLOW_PREFIX in PurePosixPath(path).parts[:-1]
+
+
+def _overflow_parents(handle: str) -> list[str]:
+    """The directories above a spill handle, up to and including `tool_output/`.
+
+    What `rmdir` is offered once the files are gone, so a pruned run leaves no empty
+    run-directory behind. Ancestors above the reserved prefix are not returned:
+    the workspace root is not this function's to offer for deletion.
+    """
+    parents = []
+    current = PurePosixPath(handle).parent
+    while OVERFLOW_PREFIX in current.parts:
+        parents.append(str(current))
+        current = current.parent
+    return parents
 
 
 class SandboxWorkspaceService:
@@ -387,6 +430,8 @@ class SandboxWorkspaceService:
                 await self._flush_state(workspace)
             elif workspace.scope == "run":
                 await self._release(workspace)
+            else:
+                await self._prune_spills(workspace)
         except Exception:
             logger.exception("workspace_close_failed", extra={"scope_key": workspace.scope_key})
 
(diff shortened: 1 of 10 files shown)


backend/app/services/sandbox_workspace.py
```diff
@@ -438,6 +483,48 @@ def _warn_if_overtaken(workspace: OpenWorkspace, row: AgentWorkspace) -> None:
...
+        if execute is None:
+            return
+        directories = sorted(
+            {parent for handle in handles for parent in _overflow_parents(handle)},
+            key=lambda directory: directory.count("/"),
+            reverse=True,
+        )
+        files = " ".join(shlex.quote(handle) for handle in handles)
+        emptied = " ".join(shlex.quote(directory) for directory in directories)
+        command = f"rm -f -- {files}; rmdir -- {emptied} 2>/dev/null; true"
```
DEENUU1: The trailing `; true` makes the shell's exit status always 0. `DockerSandbox.execute` runs this through `sh -c` and returns the shell's exit code, so when `rm` fails (read-only filesystem, permissions), the compound command still exits 0 and `workspace_spill_prune_failed` never fires — the warning can only trigger on a transport-level failure, where the client itself fabricates `exit_code=1`. That silently re-opens the failure mode this PR exists to close: spills accumulate and nothing says so.

The test `test_a_failed_prune_is_logged_rather_than_raised` doesn't catch this because its mock returns `exit_code=1` directly — a real shell would have answered 0 for the same `rm` failure. Worth adding a test that runs the built command through a real `sh -c` so it proves the shell behavior, not the mock's.

Capture `rm`'s status before the `rmdir` (whose failure is expected and stays suppressed):

```suggestion
        command = f"rm -f -- {files}; status=$?; rmdir -- {emptied} 2>/dev/null; exit $status"
```

backend/app/services/sandbox_workspace.py
```diff
@@ -166,6 +179,36 @@ def _is_spill(path: str) -> bool:
...
+    The delete in `_prune_spills` runs a shell command over paths from a list, so
+    even though only the overflow store appends to that list, every path is checked
+    against the one invariant that makes the command safe: it removes spills and
+    nothing else. Depth-agnostic because a backend normalizes handles its own way -
+    `/tool_output/…` from `state`, an absolute in-container `/workspace/tool_output/…`
+    from a service. The prefix must be a proper ancestor (`parts[:-1]`), which is
+    what every handle the store writes has, and what guarantees `_overflow_parents`
+    answers at least the spill directory itself.
+    """
+    return OVERFLOW_PREFIX in PurePosixPath(path).parts[:-1]
```
DEENUU1: The guard accepts a handle with `..` after the reserved component: `/workspace/tool_output/../../etc/passwd` has `tool_output` in `parts[:-1]`, passes, and `rm -f` resolves it outside the spill directory. Today only the store appends backend-normalized paths, so this isn't reachable — but this function's whole reason to exist (per its own docstring) is to hold the invariant even if the list is polluted, and `PurePosixPath` doesn't resolve `..`. One line closes it:

```suggestion
    parts = PurePosixPath(path).parts
    return ".." not in parts and OVERFLOW_PREFIX in parts[:-1]
```

DEENUU1 (commented):
This closes the container half of #803 cleanly, and the core design decision is right: deleting exact recorded handles instead of sweeping the `tool_output/` prefix is what makes two concurrent runs on a shared `user`/`agent` workspace safe, and threading the same list to delegates that share the parent's sandbox closes the hole before anyone hits it. The reserved-prefix guard in front of the shell command is good defense in depth.

One thing needs fixing before merge: the trailing `; true` in the prune command means the shell always exits 0, so the `workspace_spill_prune_failed` warning this PR advertises can never fire for a refused `rm` — details inline.

backend/app/services/sandbox_workspace.py
DEENUU1: Fixed in b34cdebe — the command now captures rm's status before the rmdir and exits with it, and two new tests run the built command through a real /bin/sh: one proves the files and emptied directories are actually deleted, the other proves a permission-refused rm reaches workspace_spill_prune_failed (skipped for root, which is never refused).

backend/app/services/sandbox_workspace.py
DEENUU1: Fixed in b34cdebe — _under_overflow now refuses any handle carrying a `..` component, with a test that a traversal handle runs no command.

PR #818, opened 2026-08-16
feat(builder): render the delegation tree recursively on the agent map

### LINKED ISSUE #276: Render the delegation tree recursively on the agent map
## What

The agent map (#126) now draws an agent's direct delegates and inline specialists
as nodes, and a **published delegate is a click-through** to its own `/agents/[id]`
page — so today the delegation tree is walked one hop at a time, a page at a time.

This asks for the next step: render the delegation tree **recursively, inline** — a
parent, its delegates, *their* delegates, and so on, as one graph, without leaving
the page.

## Why it was left out of #126

#126 deliberately scoped to one agent's own map with click-through, because a
recursive view is a materially larger piece:

- it fetches each delegate's published spec (N+1 across the tree, needs batching or a
  server-side tree endpoint);
- it must bound depth and fan-out (the delegation policy already caps `max_depth` /
  `max_fanout` — the view should respect and surface them);
- it has to handle cycles, unreachable pins, and inline specialists that have no
  page of their own;
- the measured-edge layout that works for one hub does not obviously extend to an
  arbitrary-depth graph.

## How you'd know it's done

Opening the map on a parent shows, in one view, the delegates of its delegates to
the configured depth; each hop is still read-only; pins that are stale or gone are
still called out; and cycles/unreachable nodes render without breaking the layout.

## Links

- Follow-up to #126 (found while implementing it in #275).
- Cluster: Frontend surfaces (see the map in #168), and touches the delegation
  sub-cluster's shape.

DEENUU1:
## What changed

The agent map now renders the delegation tree **recursively, inline** — a parent, its delegates, their delegates, and so on, in one view — instead of one hop per page-walk.

**Backend** — `GET /agents/{agent_id}/delegation-tree`, answered by `AgentRegistryService.delegation_tree`, which walks the draft's pins with the same resolution publish uses:

- **One response instead of N+1.** The whole tree in one walk, with per-walk caches so a diamond (one delegate pinned by two parents) is one row read and one version read.
- **Depth is the runtime's bound.** A delegate gets `min(its own max_depth, what the branch has left)` — exactly what `subagent_runtime` hands down — so the tree shows what a run can actually reach, and a roster beyond that is marked `truncated` rather than drawn as reachable. A disabled or unparseable delegation policy delegates nothing, and reads the same way.
- **Access-checked against the caller, not the publisher.** A delegate the reader may not see answers `restricted` — no name, no children, indistinguishable from a pin at nothing — so a shared parent's map cannot probe the organization's private agents. Same no-probing rule `_resolve_pins` applies at publish.
- **Cycles are named, never followed** (`status: "cycle"`), and the walk stops at `_MAX_DELEGATION_NODES` and says so (`truncated` on the tree), the same two bounds the publish cycle check has.
- Stale pins carry `stale: true` plus the pinned version number; a pin whose version is gone answers `unpinned` with the delegate still named.

**Frontend** — the map's delegation section grows the recursive half:

- First-level delegates keep their measured edges to the hub; each subtree hangs off its parent with a **drawn connector** (`DelegateBranch`), the layout that cannot break at arbitrary depth — which is why the measured-edge approach was not extended.
- Every node at every depth is focusable, read-only, and a published delegate keeps its click-through. Nodes are keyed by **path**, so one delegate under two parents stays two nodes.
- What the walk could not resolve says so on the chip and in the detail panel: "No access", "Version gone", "Loops back", "Pin behind" (stale), "More below" (depth cap).
- The tree is fetched only while the map dialog is open (`useDelegationTree(id, { enabled: mapOpen })`). A failed fetch degrades to the one-hop map **with a sentence admitting the tree is partial** — never silently.
- The first level stays built from the on-screen draft (it may be an autosave ahead of the stored one); the server's answer is matched onto it by agent id, which `_one_pin_per_delegate` makes unique on anything publishable.

## How it was verified

- `tests/test_agent_registry_tree.py` (13 tests): two-hop expansion, restricted-equals-absent (tenant isolation through the tree), cycle naming, gone pins, `min(inherited, own)` depth, disabled/unparseable policies, root-cannot-delegate, the node bound, one read per diamond.
- `agent-map.test.tsx`: nesting to depth 3, first-level-only measured edges, nested focus + click-through, every badge and panel sentence, the partial-tree notice. `agent-map-tree.test.ts`: path keys, restricted naming, flag halves. `use-agents.test.tsx`: fetch gated on the dialog.
- `make check` green locally (lint, backend `test` + 100% platform gate, db-check, `test-frontend-cov` at 100/100/100/97.69, build-frontend, docs-build, audit). `docs_drift.py` reports nothing owed — no docs page describes the map (same answer #275 got).

## Notes

- The model-panel integration test's `@/hooks` module mock had to learn `useDelegationTree` — a mounting failure, not a behaviour change.
- Existing `Promote to…`/click-through behaviour is untouched; the tree is additive.

Closes #276


backend/app/services/agent_registry.py
```diff
@@ -1566,6 +1601,175 @@ async def list_versions(self, ctx: AuthContext, agent_id: UUID) -> list[AgentVer
...
+        index: int,
+        budget: int,
+        ancestors: frozenset[UUID],
+        walk: _TreeWalk,
+    ) -> DelegationTreeNode:
+        """One pin as a tree node, expanded only as far as a run could follow it."""
+        walk.count += 1
+        key = f"delegate:{ref.agent_id}:{index}"
+        row = await self._tree_row(ctx, ref.agent_id, walk)
+        if row is None:
```
DEENUU1, line 1688: This is the one I'd block on. The row is in hand here, and `row.status` is never looked at - so an **archived** delegate comes out of this walk as `status: "ok"`, with its name and its whole subtree drawn as reachable.

The concrete failure: A pins B at version 3, somebody archives B. `agent_repo.get` does not filter archived rows, `resolve_access` still passes, `_resolve_pin` still finds version 3 - so the map draws B and everything under it as a working delegation, while every run of A that reaches that hop raises `BadRequestError` at `agent_runner.py:2190` ("A delegate this agent is pinned to, 'B', is archived").

What makes me want it in this PR rather than a follow-up is that there are three places that decide whether a pin can run, and two of them already agree:

- `_resolve_pins` at publish - refuses an archived delegate;
- `_resolve_delegate` at run time - refuses it, and its docstring calls out this exact case: *"An agent archived after a parent pinned it would otherwise keep answering as that parent's delegate indefinitely - the caller hardest to notice"*;
- this walk - does not.

The map is the surface that is supposed to tell that hardest-to-notice caller. Drawing the hop green is the same class of thing the docstring above rightly refuses to do for the depth cap.

The fix is small - a fourth status, checked after the row and before the cycle test:

```python
if row.status == AgentStatus.ARCHIVED.value:
    return DelegationTreeNode(
        key=key,
        kind="delegate",
        status="archived",
        agent_id=ref.agent_id,
        name=row.name,
        mode=ref.preferred_mode,
    )
```

plus `"archived"` in the two `Literal`s (`DelegationTreeNode.status` and the TS `DelegationTreeNode`), and a row in `PROBLEM_LABEL` / `PROBLEM_DETAIL` - both are `Record<NonNullable<...>>`, so TypeScript will point at the two you need to fill in. It should be named, not hidden: the caller can see the row, and what they need to know is which pin to unarchive or drop.

Worth a test alongside the ones in `TestWhatIsRefused` - an archived delegate is named, not expanded.

backend/app/services/agent_registry.py
```diff
@@ -1566,6 +1601,175 @@ async def list_versions(self, ctx: AuthContext, agent_id: UUID) -> list[AgentVer
...
+                key=key,
+                kind="delegate",
+                status="unpinned",
+                agent_id=ref.agent_id,
+                name=row.name,
+                mode=ref.preferred_mode,
+            )
+        config = _subagents_config(pinned.spec)
+        roster = len(pinned.spec.subagents) + len(config.inline if config else [])
+        remaining = min(budget - 1, config.max_depth) if config else 0
```
DEENUU1, line 1729: This line is the one I checked hardest and it is right. Writing it out for whoever reads it next, because "is this the same bound the runtime uses" is not obvious from either side.

With root `max_depth = N`, delegate's own `M`, grandchild's own `P`:

- here: a delegate's children render when `min(N-1, M) >= 1`; a grandchild's when `min(N-2, M-1, P) >= 1`;
- `agent_runner.py:2232`: a delegate's delegates are resolved when `N-1 > 0`, a grandchild's when `min(N-2, M-1) > 0`.

Same conditions at every level - `budget` here is exactly `depth_remaining + 1` there, which is what makes `min(budget - 1, config.max_depth)` line up with `min(depth_remaining - 1, nested_config.max_depth - 1)`. Since `max_depth` is `ge=1`, the extra `P` in the tree's version never changes the answer.

So the map shows what a run can actually reach, which is the whole claim in the docstring. Nice.

DEENUU1 (commented):
This is a good piece of work and it does what #276 asked: one response instead of a page-walk, cycles named, restricted pins indistinguishable from absent ones, and a layout that does not depend on measuring edges at arbitrary depth. I traced the depth arithmetic three levels down against `agent_runner.py` and it agrees with the runtime exactly, which is the part I expected to be off by one.

One thing I want fixed before this merges: **the walk never checks whether a delegate is archived**, so an archived delegate renders as `ok` with its whole subtree drawn as reachable, while every run that reaches it raises `BadRequestError` at `agent_runner.py:2190`. Publish refuses an archived pin and the runner refuses an archived pin; the tree is the third site that decides "can this pin run" and it is the one that drifted. Details inline.

The rest is a suggestion (no loading state while the walk is in flight) and a nit (`max_depth` / `max_fanout` come back in the response and nothing reads them). I left the reasoning on each.

backend/app/services/agent_registry.py
```diff
@@ -1566,6 +1601,175 @@ async def list_versions(self, ctx: AuthContext, agent_id: UUID) -> list[AgentVer
...
+        nodes = await self._tree_level(
+            ctx,
+            spec,
+            budget=caps.max_depth if config else 0,
+            ancestors=frozenset({agent.id}),
+            walk=walk,
+        )
+        return DelegationTree(
+            max_depth=caps.max_depth,
+            max_fanout=caps.max_fanout,
```
DEENUU1: nit: `max_depth` and `max_fanout` come back in the response and nothing reads them. The map already renders both, from the local draft - `page.tsx:320-321` builds `mapDepthLimit` / `mapFanoutLimit` out of `config.max_depth` / `config.max_fanout` - so these two fields duplicate numbers the client had before it asked.

Either is fine by me, but pick one: drop them from `DelegationTree` (and from the TS type), or have the map read the server's numbers here and delete the draft-side pair. Two sources for one number is the thing that drifts later. Not blocking.

frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx
```diff
@@ -147,6 +149,9 @@ export default function AgentBuilderPage({ params }: PageProps) {
   const [confirming, setConfirming] = useState<"archive" | "delete" | null>(null);
   const [publishOpen, setPublishOpen] = useState(false);
   const [mapOpen, setMapOpen] = useState(false);
+  // Fetched only while the map shows it: the server resolves and
+  // access-checks every pinned version to build this.
+  const { tree, error: treeError } = useDelegationTree(id, { enabled: mapOpen });
```
Comment (DEENUU1) [inline frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx]: Suggestion: `isLoading` is dropped here, and while the walk is in flight the map is not honest about it. `tree` is `null`, so the first level renders with no children, no badges and no notice - a one-hop map that looks complete - and then the subtree pops in when the response lands.

It is the same failure the `delegationNotice` exists to prevent, just for a second instead of forever, and the notice slot is already there. Something like:

```tsx
const { tree, isLoading: treeLoading, error: treeError } = useDelegationTree(id, { enabled: mapOpen });
```

and then at the notice (line 664):

```tsx
delegationNotice={
  treeLoading
    ? tAgents("mapTreeLoading")
    : treeError
      ? tAgents("mapTreeUnavailable")
      : tree?.truncated
        ? tAgents("mapTreeTruncated")
        : null
}
```

with one new key saying the tree is still being walked. Cheap, and it means the map never shows a delegate as childless when it just does not know yet.

frontend/src/components/agents/agent-map-nodes.tsx
```diff
@@ -164,16 +184,69 @@ export function DelegateNode({
...
+/**
+ * The recursive half of the delegation tree: a delegate's own delegates,
+ * indented under it with a drawn connector instead of a measured edge.
+ *
+ * The hub's edges are measured because its nodes sit anywhere on a grid; a
+ * subtree is a list whose parent is always directly above, so a border is the
+ * honest connector and it cannot break however deep or wide the tree gets -
+ * which is what an arbitrary-depth graph needs from a layout (#276).
+ */
+export function DelegateBranch({ nodes, icons, focused, onFocus }: DelegateBranchProps) {
```
DEENUU1, line 230: Good call, and worth saying out loud: #276 named the measured-edge layout as the thing that would not extend to arbitrary depth, and instead of trying to make it extend you left it doing the one job it is good at (the hub) and drew a border for the rest. A subtree's parent is always directly above it, so there is nothing to measure and nothing to break.

The test that asserts exactly one `path.map-flow` is the right way to hold that shut - it fails the moment somebody "fixes" the nested nodes by registering refs for them.

backend/app/services/agent_registry.py
DEENUU1: Fixed in 2ee49099 — a fourth status, `archived`: the walk now reads the row status and names an archived delegate without expanding it, badged on the map with its own detail copy. One deliberate deviation: the check sits after the cycle test, not before, because `_resolve_delegate` refuses the loop before it reads the row — a pin that is both reads as the cycle, and `test_a_cycle_back_to_an_archived_agent_still_reads_as_the_cycle` pins that ordering.

backend/app/services/agent_registry.py
DEENUU1: Fixed in 2ee49099 — dropped the duplication on both sides: `max_depth` / `max_fanout` are gone from the backend `DelegationTree` model and from the TS type, and the tests that asserted them now assert what the walk answers instead. The map keeps building its limits from the local draft, which was the only reader all along.

frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx
Comment (DEENUU1) [reply frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx]: Fixed in 2ee49099 — the map now uses the notice slot for the in-flight walk: `treeLoading` shows a new `mapTreeLoading` line, then `mapTreeUnavailable` (already existed) on error, then `mapTreeTruncated`. `useDelegationTree` already exposed `isLoading`; the page just dropped it.

Issue #829, opened 2026-08-17
The Resend email provider imports an SDK that is declared nowhere


DEENUU1:
`app/services/email/providers/resend.py:14` imports the `resend` SDK inside
`ResendProvider.__init__`, and **`resend` is not declared anywhere** — not in
`dependencies`, not in an extra, not in the `dev` group. It is not a transitive of
anything else in `uv.lock` either.

So a deployment that sets its email provider to Resend gets `ModuleNotFoundError:
No module named 'resend'` the first time it tries to send — an invitation, a
password reset, an approval notification. Not at boot, not at configuration time:
at the first send, which is the point at which somebody is waiting for the mail.

```python
class ResendProvider:
    def __init__(self, api_key: str) -> None:
        import resend as resend_sdk   # <- undeclared, unguarded
```

`faker` in `app/commands/seed.py:20` has the same undeclared shape but is guarded
(`try/except ImportError` setting `HAS_FAKER`), so it degrades rather than raising.
Whatever is decided for one probably applies to the other.

**Found by** the `deptry` gate added in #155 — DEP001 (imported but missing from
the dependency definitions) is currently switched off in `[tool.deptry]` precisely
because these two are the only things it reports, and neither has an obvious answer
that belongs in a manifest-pruning PR.

## The decision

Three shapes, and the third is probably right:

1. **Declare `resend` as a required dependency.** Simple, and wrong for the same
   reason the rest of #155 was about: every deployment carries an SDK most of them
   never call.
2. **Guard it like `faker`** and fail with a domain exception naming the missing
   package. Honest, but it turns a configuration error into a runtime one, still.
3. **`[project.optional-dependencies] resend = ["resend>=..."]`**, imported behind a
   guard that raises an `AppException` naming the extra, and — the part that makes
   it a feature rather than a workaround — the provider is not selectable when its
   SDK is absent, the same shape #155 proposes for the long-tail model providers.

Whichever it is, DEP001 should come on in `[tool.deptry]` afterwards so the next
undeclared import fails in CI rather than in somebody's inbox.

## Done when

- [ ] Sending mail through Resend on a fresh `uv sync --no-dev` install either works
      or refuses with a message naming what to install — never `ModuleNotFoundError`
- [ ] `faker` in `app/commands/seed.py` is resolved the same way, or the difference
      is written down
- [ ] `DEP001` is removed from `ignore` in `[tool.deptry]` and `make lint` is green

[...]

PR #831, opened 2026-08-17
Drop three unimported dependencies and gate on the next one

### LINKED ISSUE #155: Backend dependencies: fastapi-cache2 is initialised and caches nothing, plus four more that are imported nowhere
`backend/pyproject.toml` declares 46 runtime dependencies, which resolve to **277 packages and a 788 MB venv**. Some of that is unavoidable — this is an agent platform and the model SDKs are large. Some of it is not. Below is what a pass over every declaration actually found, so the work can start from evidence rather than from a feeling.

## Dead: declared, imported nowhere

| Package | Evidence |
|---|---|
| **`fastapi-cache2`** | See below — the whole subsystem is dead, not just the dependency |
| **`slowapi`** | Configured, never applied (#7), and configured wrongly for a multi-replica deployment. See below |
| **`jinja2`** | Zero imports. `app/services/email/templates.py:1` — *"Template loader — reads pre-rendered HTML/text from `emails/compiled/`."* The templates are compiled ahead of time; nothing renders at runtime |
| **`pillow`** | Zero imports in `app/` or `tests/`. 13 MB installed. It is a transitive of the parsers anyway, so the declaration buys nothing |
| **`tabulate`** | Zero imports. Presumably intended for the CLI, which does not use it |
| **`itsdangerous`** | Zero direct imports. It is Starlette's requirement for `SessionMiddleware`, which `main.py:260` does use — so this is a correct transitive that is being pinned by hand. Either say so in a comment or drop it and let Starlette declare it |

## `fastapi-cache2` is worse than unused

`app/core/cache.py` is eleven lines. `main.py` calls `setup_cache(redis_client)`, which calls `FastAPICache.init(RedisBackend(...))`. **Nothing is ever cached.** There is no `@cache` decorator anywhere on a route — the two `@cache` hits in the codebase are `functools.cache` (`skill_library.py:26`, `capability_contracts.py:27`), which is the standard library.

So the library is installed, initialised on every boot, and holds a Redis backend that answers no question.

And the test that covers it, `tests/test_core.py:87-90`:

```python
def test_setup_cache_function_exists(self):
    """Test setup_cache function exists."""
    assert setup_cache is not None
    assert callable(setup_cache)
```

`.claude/rules/testing.md`: *"No test for a mock — if removing the implementation still passes, delete it."* This asserts that a name is bound and callable. It cannot fail while the import at the top of the file succeeds, and it is counted toward the 100% platform-layer gate. **Delete `app/core/cache.py`, the `setup_cache` call, the dependency and that test together.**

> **Corrected 2026-08-04.** An earlier version of this issue said "FastAPI does not ship a Redis SDK." That was wrong. [`redis/fastapi-redis-sdk`](https://github.com/redis/fastapi-redis-sdk) is real — official, from the Redis organization, `fastapi-redis-sdk` 0.8.0 on PyPI released 2026-07-31, requiring `redis>=6.0.0` and `fastapi>=0.115.0`. See the section below: it changes the answer for `slowapi` too, not only for the cache.

## `slowapi` is the same story, and the replacement covers both

`fastapi-cache2` at least *initialises*. `slowapi` does not even do that, and #7 is the issue for it: `app.state.limiter` is set and the `RateLimitExceeded` handler registered (`main.py:253-258`), but **`SlowAPIMiddleware` is never added**. In slowapi, `default_limits` are evaluated exclusively inside that middleware, and the alternative path — `@limiter.limit(...)` on a handler — has zero call sites. Neither do the three helpers the module exports:

```
rate_limit_low / rate_limit_medium / rate_limit_high  →  zero callers in app/ or tests/
```

So `core/rate_limit.py` is 47 lines of configuration nothing reads.

**And fixing #7 as written would ship the wrong thing.** `Limiter(key_func=get_remote_address, default_limits=[...])` is constructed with **no `storage_uri`**, so slowapi falls back to in-memory storage. Registering the middleware would give per-process limits: two replicas means double the allowance, a restart resets the counters, and "100 per minute" is a number that describes one container rather than the deployment. Nothing sets `storage_uri` anywhere in the repository.

`fastapi-redis-sdk` does distributed per-client rate limiting over Redis — which this deployment already runs, already opens through the lifespan, and already exposes as `app/clients/redis.py`. It also does DI-based caching with ETag / `304` / `Cache-Control`, which is what `fastapi-cache2` was installed for.

**So the shape of this change is: two dead dependencies out, one live one in, and #7 is fixed correctly rather than fixed into a per-process limiter.** Evaluate before committing to it — a 0.8.0 from July is young, and the `Depends()`-factory API is a different shape from slowapi's decorator, so the routes that eventually carry limits will read differently. But the alternative is maintaining two abandoned integrations for behaviour nobody has yet.

## Not removable, and worth recording so nobody tries

- **`psycopg2-binary`** looks dead — zero imports and every application URL is `postgresql+asyncpg`. It is not. `alembic/env.py:59` uses `engine_from_config` (a *sync* engine) against `settings.DATABASE_URL_SYNC`, which is bare `postgresql://` — SQLAlchemy resolves that to psycopg2. Removing it breaks every migration. Worth a comment beside the declaration, because this is the second time it has read as dead.
- **`rank-bm25`** is imported behind a `try/except ImportError` that logs and degrades (`retrieval.py:93-96`). That is the shape of an optional dependency declared as a required one. Either make it an extra and mean it, or drop the guard — the current arrangement is a fallback for a case that cannot happen, which CLAUDE.md calls a branch to delete. **It is also the sole reason `numpy` is installed** (24 MB, `uv.lock`: `rank-bm25 → numpy`, and nothing else in the tree asks for it), so this is a 24 MB decision rather than a tidying one.
- **`websockets`** has no import under `app/` by that name and looks redundant beside `uvicorn[standard]`, which vendors it. It is not: `app/services/channels/mattermost.py` lazily imports `websockets.asyncio.client.connect` for the Mattermost event stream, with an `ImportError` message naming the package. The explicit declaration is correct — the Mattermost adapter should not depend on an ASGI server's extras.
- **`google-auth`** shows zero hits for its distribution name because it provides the `google.oauth2` module, which has three import sites. Do not grep for the distribution.

## Duplicates in the declaration

```
pyproject.toml:23   "python-multipart>=0.0.20"
pyproject.toml:32   "python-multipart>=0.0.12"     ← weaker floor, silently ignored
pyproject.toml:34   "httpx>=0.28.0"
pyproject.toml:56   "httpx>=0.28.0"
pyproject.toml:81   "httpx>=0.28.0"                ← and again in the dev group
```

Harmless to the resolver, misleading to a reader, and the kind of thing that makes a floor bump miss a line.

## Where the weight actually is

Install size, largest first:

```
99M  googleapiclient    4 import sites — the Google Drive sync connector
66M  prefect            5 import sites — background work
59M  pymupdf            2 import sites — the default PDF parser
38M  grpc               transitive — see the correction below
26M  botocore           the S3 connector
25M  liteparse          the alternative parser
24M  numpy              transitive of rank-bm25, and of nothing else
19M  lxml               transitive of ddgs and python-docx
```

> **Corrected 2026-08-04.** An earlier version of this section attributed the 38 MB of gRPC to `google-api-python-client`. That is wrong, and it would send somebody looking for it in the wrong place. Per `uv.lock`, `google-api-python-client 2.198.0` depends on `google-api-core`, `google-auth`, `google-auth-httplib2`, `httplib2` and `uritemplate` — **no grpcio**. gRPC is a dependency of **`xai-sdk`**. See the section below.

**`google-api-python-client` at 99 MB for one sync connector** is still the single largest line and still worth its own decision. Drive's REST API is a handful of endpoints and `httpx` is already a dependency; `google-auth` handles the credential half separately and stays either way. Whether that trade is worth making depends on how much of the client library the connector actually uses — measure before deciding, and do not do it in the same change as the deletions above.

**Three document parsers ship at once** — pymupdf (59 MB), liteparse (25 MB) and `llama-cloud` — because a deployment picks one at runtime (`rag/config.py:183`). That is a real feature, not bloat. But it means an operator who only ever uses pymupdf carries the other two, which is an argument for extras (`agenticos[llamaparse]`) rather than for removal.

**`prefect` at 66 MB** is real: three `@flow` decorators (`worker/tasks/{mcp,report,rag}_tasks.py`) plus `aserve` with `IntervalSchedule` in `worker/prefect_app.py`. That is orchestration being used as orchestration, not a decorator borrowed from a large library. Not a dependency question — if it ever becomes one it is a background-work redesign, and #26 is the issue that would notice first.

## The model-provider extras are justified, and cost more than they look

`pydantic-ai-slim[openrouter,anthropic,google,duckduckgo,web-fetch,groq,mistral,cohere,huggingface,xai]` is eleven extras, and the reason is written down in `app/agents/model_resolver.py:32-36`:

> Three of them (`xai`, `cohere`, `huggingface`) need an SDK of their own, so they are in this repository's `pydantic-ai-slim` extras: **a provider selectable in the Builder that raises `ImportError` when a run reaches it would be worse than one nobody can pick.**

That is the right call and this issue does not argue with it. But the true cost is not the SDK sizes, because two of them drag something much bigger behind them:

| Extra | SDK | Brings with it |
|---|---|---|
| `xai` | 2.3M | **`grpcio` 38M**, `aiohttp` 2.7M, `protobuf` — and `aiohttp` is a second HTTP client in an image that standardised on `httpx` |
| `cohere` | 4.2M | **`tokenizers` 8.5M**, `fastavro` 3.1M |
| `duckduckgo` | — | `ddgs`, which brings `lxml` 19M and `primp` 10M |
| `openai` | 18M | (via other extras) |
| `google` | 15M | |
| `anthropic` | 12M | |
| `mistral` | 9.1M | |
| `huggingface` | 4.8M | |
| `groq` | 1.3M | |

**xAI support costs roughly 43 MB, of which its own SDK is 2.3.** Whether that is worth one entry in a provider dropdown is a product question, not an engineering one, and it is the kind of thing worth deciding deliberately rather than discovering. The honest alternatives are: keep it and record the cost; or make the long-tail providers an install-time extra (`agenticos[xai]`) and have the Builder hide a provider whose SDK is absent — which is a real feature, not a workaround, since a self-hosted deployment that will only ever use Anthropic should not ship four gigabytes of other people's clients.

Nothing here should be changed in the same pull request as the deletions above.

## The rule that would keep this from coming back

Nothing today notices a dependency that stops being imported. The deletions above are worth maybe 40 MB and two subsystems; the value is in not re-accumulating. A check in `make lint` comparing declared distributions against imported modules — `deptry` does exactly this for Python and understands the transitive-vs-declared distinction — would fail on `jinja2` today and on the next one automatically.

## Watch out

- Removing `app/core/cache.py` means editing **both** lists in `backend/pyproject.toml` — `[tool.coverage.run] include` and `[[tool.ty.overrides]] include` — verbatim and in the same order. `tests/test_coverage_gate.py` fails if they drift.
- `logfire[fastapi,asyncpg,redis,httpx]` is the product's observability and every one of its four extras corresponds to something instrumented. Do not touch it.
- The `pydantic-ai-slim[...]` extras are load-bearing for a reason stated in `model_resolver.py` — see above. They are a *product* decision to revisit, not a manifest to prune.
- **Grep for the module, not the distribution.** `google-auth` provides `google.oauth2`; `pydantic-monty` provides `pydantic_monty`; `websockets` is imported lazily inside a function in `mattermost.py`. A distribution-name grep called all three dead in the first pass of this audit and all three are live.
- Uncapped versions are deliberate and the comment at the top of the file explains why. A dependency audit is not an invitation to add caps.
- `make deps-upgrade`, the `dependency-freshness` workflow and Dependabot all read this file; a reshuffle should keep it readable for them.

## Done when

- [ ] `fastapi-cache2`, `app/core/cache.py`, the `setup_cache` call and its test are gone
- [ ] A decision in writing on `fastapi-redis-sdk` — adopted for caching and rate limiting, or rejected with a reason. If adopted, `slowapi` and `core/rate_limit.py` go with it and #7 is closed by this change
- [ ] `jinja2`, `pillow` and `tabulate` are gone, with a test run proving nothing imported them
- [ ] `psycopg2-binary` carries a comment saying Alembic needs it
- [ ] `rank-bm25` is either an extra with the guard, or required without it — and whichever it is, the 24 MB of `numpy` behind it is accounted for
- [ ] The duplicate `httpx` and `python-multipart` lines are gone
- [ ] A declared-but-unimported check runs in `make lint`
- [ ] The `google-api-python-client` question is answered in writing — kept with a reason, or replaced — but not bundled into this change
- [ ] The provider-extras question is answered in writing: every provider ships always, or the long tail becomes an install-time extra and the Builder hides what is absent. Not bundled into this change either
- [ ] `make check` green, and the venv measured before and after so the number is real

DEENUU1:
## What this fixes

`backend/pyproject.toml` declared three distributions nothing imports, declared two
of them twice, and initialised a cache subsystem that caches nothing. Nothing in
`make lint` would ever have noticed, which is the half of this worth more than the
bytes.

Closes #155.

## What changed

**Deleted, because nothing imports them**

- `backend/app/core/cache.py` — eleven lines that called `FastAPICache.init()` on
  every boot. There is no `@cache` decorator on any route; the two `@cache` hits in
  the tree are `functools.cache`. The `setup_cache(redis_client)` call in the
  lifespan and the `fastapi-cache2` declaration go with it.
- `tests/test_core.py::TestCacheSetup` — `assert setup_cache is not None`. It could
  not fail while the import at the top of the file succeeded.
- `jinja2` — zero imports. The email templates are compiled ahead of time and read
  off disk (`app/services/email/templates.py`). The only greps that look like hits
  are the `# ruff: noqa: I001 - Imports structured for Jinja2 template conditionals`
  line the project template left behind in twelve files.
- The duplicate `python-multipart>=0.0.12` (weaker floor, silently ignored) and the
  duplicate `httpx>=0.28.0`.

**`tabulate` stays — I deleted it, and CI caught me.** #155 lists it as
zero-import and so did my first pass, because both of us grepped `app/` and
`tests/`. `cli/commands.py:10` imports it, and removing it took the e2e **seed
step** down with `ModuleNotFoundError` before a single product spec ran. It is
restored, with its call site named in the manifest — and the gate below now reads
`cli/` and `alembic/` too, which is the fix that matters: a tree that ships and is
not scanned is a tree whose imports do not count. The first version of this PR
would have passed its own gate while breaking the CLI.

**Kept, with the reason written beside the declaration**

- `psycopg2-binary` — `alembic/env.py` builds a *sync* engine from
  `DATABASE_URL_SYNC`, which is bare `postgresql://`. Removing it breaks every
  migration. Second time it has read as dead, so it now says so in the file.
- `itsdangerous` — Starlette signs `SessionMiddleware` cookies with it, and the
  middleware is ours.
- `rank-bm25` — kept **required**, and the `try/except ImportError` around it in
  `app/services/rag/retrieval.py` is gone. It guarded a case that cannot happen,
  which is the branch CLAUDE.md says to delete. The import moved to module scope.
  The 24 MB of `numpy` behind it is the price of hybrid retrieval, taken
  deliberately and recorded in the manifest.

**`pillow` stays — the issue was wrong about it.** #155 lists it as zero-import.
It has one: `app/services/channels/chart_png.py:24` renders a chart tool's output
for a channel that cannot display HTML. That landed after the audit was written
(#157). The declaration now names its call site so the next sweep does not repeat
the mistake.

**Declared, because it was imported and undeclared**

- `anyio` — imported at module scope in `app/services/rag_document.py` and declared
  only in the `dev` group. `uv sync --no-dev` is what the image builds with, so
  this was relying on Starlette to pull it in. Found by the new gate on its first
  run.

**The gate**

`deptry app cli alembic` is now a step of `make lint-backend`, so it runs in CI's
`lint` job. It gates on DEP002 (declared, unimported) and DEP004 (imported,
declared dev-only).

DEP001 and DEP003 are off in `[tool.deptry]`, each with the reason in the file:
DEP003 fires 27 times on `starlette`, `mcp`, `yaml`, `openai`, `botocore`, `ddgs`
and `opentelemetry`, every one of them arriving through an extra we do control, and
declaring each by hand would double the manifest to restate what those extras
already promise. DEP001 fires on `faker` and `resend` — filed separately, see below.

Five dependencies are genuinely required and genuinely unimported (a driver a URL
scheme selects, SQLAlchemy's async bridge); they sit in `per_rule_ignores.DEP002`
with a one-line reason each. `uvicorn` was in that list in the first version of
this PR and is not now — `cli/commands.py` imports it, which the widened scan
found.

## The decisions #155 asked for in writing

**`fastapi-redis-sdk` — rejected.** Both halves of the case for it have gone away:

- *Rate limiting.* #155 assumed `slowapi` was still declared and still needed
  fixing. It is not — `app/services/rate_limit.py` (#39) is a Redis fixed-window
  limiter shared across workers, applied to all six unauthenticated surfaces, and
  `slowapi` is already out of the manifest. The per-process-limiter trap #155 warned
  about was avoided by that change, not left open. **#7 is functionally done and
  still open** — see "Noticed, not fixed".
- *Caching.* There are zero cache sites, and no obvious candidate for one: every
  answer on this surface depends on a cookie, a permission set and an organization
  header, which is why `platform-proxy.ts` stamps `no-store` on anything the backend
  leaves unmarked. Adopting a 0.8.0 library to cache nothing would be the same
  defect this PR deletes, in a newer package.

If a route ever genuinely wants caching, that is the moment to evaluate it, with a
call site to evaluate it against.

**`google-api-python-client` (99 MB) — kept.** Filed as a follow-up with the
measurement rather than bundled here, per #155. The connector and the sync source
use `build`, `Resource` and `MediaIoBaseDownload`; the last is chunked resumable
download, which is the part that is real work to reimplement over `httpx`, and
`build` reads Drive's discovery document rather than hard-coding endpoint shapes.
99 MB against a 634 MB runtime image, to avoid reimplementing resumable media
download and Drive pagination, is a trade worth keeping.

**Provider extras — keep them all shipping, cost recorded.** `xai` costs ~43 MB
(its own SDK is 1.4 MB; `grpcio` is 38 MB), `cohere` brings `tokenizers`,
`duckduckgo` brings `lxml` and `primp`. The alternative — install-time extras plus
a Builder that hides a provider whose SDK is absent — is a real feature, but it
trades a guaranteed-working dropdown for a deployment-dependent one on a platform
whose point is that any model works. Revisit if image size becomes an actual
constraint rather than a number in an audit. Both questions are filed so the
reasoning survives this PR.

## The size number, honestly

Measured on a **freshly created** venv (`rm -rf .venv && uv sync --no-dev`) at each
of the two manifests, which is what the image builds. The first measurement in this
PR compared reused venvs and was inflated ~100 MB by accumulated `__pycache__`;
these numbers replace it:

| | Before | After |
|---|---|---|
| Runtime venv | 634 MB | 634 MB |
| Distributions | 241 | 240 |

**One distribution, and no measurable bytes — not the ~40 MB #155 estimated.** The
estimate was wrong in a specific way worth recording: `jinja2` is a transitive of
`prefect`, `pillow` of the parsers, `numpy` of `rank-bm25`, and `tabulate` was never
dead at all. Removing a *declaration* only removes bytes when nothing else asks for
the package, and here only `fastapi-cache2` qualified.

So this PR is worth nothing on disk, and should be read that way. What it is
actually worth: a dead subsystem no longer initialised on every boot and no longer
holding a Redis backend, a test that could not fail, an undeclared runtime
dependency now declared, three deletions-that-would-have-been-wrong documented
where the next person will look, and a gate that fails on the next one.

## Testing

- `make lint-backend` — green, with `deptry` reporting no issues.
- `uv run pytest tests/test_core.py tests/test_rate_limit.py` — 41 passed.
- `uv run pytest tests/test_coverage_gate.py tests/test_ci_parity.py` — 29 passed.
  `test_ci_parity` is what confirms the new `lint-backend` step actually runs in CI.
- `make test` — the full backend suite plus the 100% platform-layer gate.
- `uv run agenticos --version` and `uv run agenticos cmd --help` on a
  `uv sync --no-dev` install — the check that would have caught the `tabulate`
  deletion locally, and did not exist in the first version of this PR.
- Proving nothing imports the removed package: `deptry` over `app`, `cli` and
  `alembic` is the proof going forward, and is what now covers the tree the first
  pass missed.
- `python3 scripts/docs_drift.py` — clean.
- Venv measured on a fresh `uv sync --no-dev` at each manifest.

`ty check` reports 66 warnings and exits 0, unchanged from the base ref.

**`e2e` is red on `main` too, for an unrelated reason.** The release commit for
0.0.171 (run `31974018906`) fails at `journey.spec.ts:52` after ~36s of product
specs. The failure on the first push of this branch was mine and is fixed; if `e2e`
is red again it is worth checking against `main` before reading it as this PR's.

## Noticed, not fixed

- **#7 is done and still open.** `app/services/rate_limit.py` closed it in #39 —
  its own module docstring says so — but the issue was never closed, which is how
  #155 came to be written on the assumption that `slowapi` still needed fixing.
  Left open deliberately: closing somebody else's issue is not this PR's call.
- **`resend` is imported unguarded inside `ResendProvider.__init__`** and is not
  declared anywhere (`app/services/email/providers/resend.py:14`). A deployment that
  configures Resend gets `ModuleNotFoundError` at first send. `faker` has the same
  shape in `app/commands/seed.py` but is properly guarded. Filed.
- The `# ruff: noqa: I001 - Imports structured for Jinja2 template conditionals`
  line survives in twelve files and now names a dependency that is gone. Cosmetic,
  and out of scope for a manifest change.
- **`vulture` reads `paths = ["app"]` only**, so `cli/` and `alembic/` are unscanned
  for dead code by the same blind spot that made me delete `tabulate`. Not widened
  here — it is a different tool with its own false-positive profile and deserves its
  own change rather than a drive-by.


DEENUU1:
Note for whichever of us merges these two: `DEP001` can come off the `[tool.deptry] ignore` here once #829 lands.

#829 turned out to be a bit more than an undeclared import — `ResendProvider` was never reachable at all (no `case "resend"` in the factory, no `RESEND_API_KEY` setting), and the match ended `case "log" | _`, so `EMAIL_PROVIDER=resend` silently resolved to the development provider and reported every invitation as sent. The provider is deleted there rather than wired up.

That leaves `faker` as the only DEP001 report, and it stays deliberately: guarded, degrading to `random`, with `uv add faker --group dev` in its own docstring. So after both branches are in, the ignore is one line to remove — happy to do it in whichever merges second.

DEENUU1:
@codex review

PR #832, opened 2026-08-17
fix(agents): keep a delegate's provider text out of a status answer

### LINKED ISSUE #819: check_task and wait_tasks hand a delegate's provider text to the parent's transcript
**Severity: medium** · pre-existing, found reviewing #816 (the fix for #699) · not caused by it

## What's wrong

#816 keeps a delegate's provider text off the child `agent_runs` row and its
closing `SubagentFinished` frame. The same text still reaches the **parent's
stored transcript** through the library's own task-status tools, because those
tools hand `handle.error` to the model verbatim and a tool *return* is stored
whole.

In `subagents-pydantic-ai` 0.2.19, `toolset.py`:

- `:1256` — `check_task` on a `FAILED` task: `f"Error: {handle.error}"`
- `:1265` — `check_task` on a `RETRYING` task: `f"Retry {handle.retry_count}: {handle.error}"`
- `:1406` — `wait_tasks` on any finished-not-completed task: `f"{handle.status.value.upper()} - {handle.error}"`

`handle.error` for a failed or retrying delegation is
`f"{type(exc).__name__}: {exc}"` / `f"Transient error (retry {n}): {exc}"` —
`{exc}` being the delegate's model client, whose message carries the failing
request URL with the key still in its query string on a custom `base_url`
(#699's exact payload).

The string returned by `check_task` becomes a `ToolReturnPart` in the parent's
messages. `_result_text` in `backend/app/services/transcript.py:121` stores a
`ToolReturnPart` whole — deliberately, "a return is the tool's own answer" —
so #695's `tool_retry_notice` never touches it, and the row is rendered in the
conversation and in run history to every member who can read the run. The live
`tool_result` frame streams the same string.

Note the `RETRYING` case leaks for delegations that eventually **succeed**:
`TaskHandle.finish` clears `error` on completion, but the model may have polled
`check_task` mid-retry and the returned text is already on a transcript row.

## The sequence

1. Give an organization a model profile with a custom endpoint and a key; break the key.
2. Run an agent that delegates with `mode="async"` to a subagent on that profile.
3. The parent polls `check_task` (or `wait_tasks`) after the delegate fails.
4. The tool-call row for `check_task` stores the provider's text, keyed URL included.

`check_task`/`wait_tasks` answers for a failed or retrying task name the
exception's type (the handle carries `exception` since 0.2.19, so the library
has it to name) rather than embedding `{exc}`, and a test in the capability
suite asserts a vendor string handed to a crashing delegate does not appear in
any `ToolReturnPart` of the parent's messages.

The honest fix is upstream again, like #699's: compose those three strings in
`subagents-pydantic-ai` from `type(handle.exception).__name__` (keeping the
`usage limit exceeded` marker whole), release, and bump the pin here.

## Related

Checked the map in #168: this is the **Security and tenancy** cluster, on the
#342 → #423 → #437/#659 spine — the sixth surface, beside #659 (chat `error`
frame), #681 (chat `tool_result` frame), #676 (the run row), #695 (the
transcript tool-call row) and #699 (the delegated run's row, PR #816). #695
scrubbed `RetryPromptPart` rows; this is the `ToolReturnPart` half it could not
reach, because here the vendor text is the tool's own answer. Same caveat as
the rest of the cluster: the log copy is only safe once #440 is fixed.

DEENUU1:
## What

`check_task` and `wait_tasks` no longer hand a delegate's provider text to the
parent's stored transcript.

Both composed their failure lines from `handle.error`, which embeds the
exception's own message — and a model client's message carries the failing
request URL with the key still in its query string on a custom `base_url`
(#699's exact payload, arriving by another route).

What those tools return becomes a `ToolReturnPart`, and `_result_text`
(`app/services/transcript.py`) stores one **whole**, on purpose: a return is the
tool's own answer rather than something this platform composed. So #695's
`tool_retry_notice` never touches it. The key reached a stored tool-call row,
rendered in the conversation and in run history to every member who can read the
run, and streamed live as `tool_result`.

The sixth surface in the #342 → #423 → #437/#659 cluster, beside #659, #681,
#676, #695 and #699 — and the one a scrubber could not have reached.

## Upstream first, like #699's

The honest fix was in the library, so that is where it went:
**subagents-pydantic-ai 0.2.20** ([#81](https://github.com/vstorm-co/subagents-pydantic-ai/pull/81),
merged and released to PyPI) composes all four status lines — `Error:`,
`Retry N:` and `Outcome:` in `check_task`, and the terminal line in
`wait_tasks` — from `TaskHandle.exception`, the field 0.2.19 added for exactly
this. `handle.error` is untouched and still carries the whole thing for a host to
log. The pin here moves to `>=0.2.20`.

The `usage limit exceeded` marker survives upstream: it is the library's own
text, and it is what tells a model to stop delegating rather than retry.

## The retry case is the nastier half

The `RETRYING` line leaked for delegations that eventually **succeed**.
`TaskHandle.finish` clears `error` on completion, so the handle ends up clean —
but a model that polled `check_task` mid-retry already has the answer on a
transcript row, and nothing goes back to remove it.

## Verified

`make check` green — lint, the backend suite at 100% coverage, `db-check`, the
frontend gates, `docs-build` and the audit.

The regression test (`TestTaskLifecycle`) crashes a delegate on a model whose
message carries `api_key=sk-live-9f2c`, polls both tools, and asserts that
neither the key nor the host appears in what they answer. Against the old pin it
fails on exactly that string:

```
E  AssertionError: assert 'sk-live-9f2c' not in 'Task result...-9f2c failed'
```

Upstream: 1192 tests, 100% coverage, green on Python 3.10–3.13. Two of its
existing tests asserted the leak (`Task crashed`, `Search API down` appearing in
the answer) and now assert the class is named instead.

`docs/governance.md` already described the delegated run's row and its closing
frame refusing provider text; the parent's own transcript is a third place that
rule now reaches, so the section says so.

## Not fixed here

The log copy is only safe once #440 is fixed — the same caveat the rest of the
cluster carries.

Closes #819


DEENUU1:
@codex review

Issue #109, opened 2026-08-02
TypeScript 7 and ESLint 10 are blocked, each by a different plugin


DEENUU1:
## Where this came from

Dependabot's #97 grouped three bumps: `eslint-config-next` 15 → 16, `typescript` 5.9.3 → 7.0.2, `eslint` 9.17 → 10.8. A group lands whole or not at all, so it sat. `eslint-config-next` 16 has now landed on its own (#105, which was most of the work — 31 React Compiler hook-rule violations). The other two are each blocked by a different third party, and both blockers were confirmed by installing them here rather than read off a changelog.

## TypeScript 7 — blocked by `typescript-eslint`

```
Error: typescript-eslint does not support TS 7.0.
  at node_modules/typescript-eslint/dist/index.js:52:11
  at node_modules/eslint-config-next/dist/index.js:5:64
```

A hard refusal at load time, and `eslint-config-next` imports it, so lint dies before it reads a file. Nothing in this repo can work around it.

The compiler itself is fine: with the two changes in `chore/typescript-7-prep` — dropping the removed `baseUrl`, and the `scrollMargin` the DOM lib added to `IntersectionObserver` — `tsc --noEmit` is clean under 7.0.2. So when `typescript-eslint` ships support, the bump is a version number.

## ESLint 10 — blocked by `eslint-plugin-react`

```
TypeError: Error while loading rule 'react/display-name':
  contextOrFilename.getFilename is not a function
  at node_modules/eslint-plugin-react/lib/util/version.js:31
```

ESLint 10 removed the legacy context API that `eslint-plugin-react` still calls when it sniffs the React version. This one is independent of TypeScript — it reproduces with `typescript@5.9.3` — and `eslint-plugin-react` also arrives through `eslint-config-next`, so it is not ours to pin around either.

## What to do

Nothing, until both plugins ship. Worth re-testing when either of these moves:

- [ ] `typescript-eslint` announces TypeScript 7 support
- [ ] `eslint-plugin-react` announces ESLint 10 support

Re-testing is `bun add -d typescript@7 eslint@10 && bun run type-check && bun run lint` on a scratch branch; both failures are immediate and unambiguous.

Close #97 in favour of this: the group's third member has landed, so a fresh Dependabot pull request for whichever of the two unblocks first is more useful than the stale group.

DEENUU1:
Re-tested against what is published today, and both blockers are somebody else's peer range rather than anything this repo can pin around:

[...]

PR #835, opened 2026-08-17
refactor(frontend): draw every brand mark from one generated glyph set

### LINKED ISSUE #156: Frontend dependencies: date-fns is unused and three icon libraries answer one question
`frontend/node_modules` is **1.2 GB**. Most of that is Next, React, Playwright and the toolchain, and is the cost of the stack. What follows is what a pass over every declaration in `package.json` actually found.

**One caveat up front, because it changes what the numbers mean.** Everything below is *installed* size, which is CI time and disk. It is **not** bundle size — `lucide-react` at 40 MB ships a few kilobytes to the browser because it tree-shakes per icon. The two costs are different and both matter, so this issue asks for `bun run analyze` (the script already exists) before and after, and treats the install number as the install number.

## Dead: declared, imported nowhere

**`date-fns`** — 26 MB, and `rg "date-fns"` across the whole repository returns exactly one hit: the line in `package.json`. Every date in this product is formatted by hand (`workspace-browser.tsx:45` computes "3 days ago" with arithmetic on `Date.now()`, and there are at least three more copies of that helper). Either it was for something that never landed, or it lost an argument with a one-line function. Delete it — and separately decide whether the hand-rolled `used()` / `timeAgo()` helpers should become one shared thing (see #138, which already lists the duplicate `size()` helpers).

That is the only completely dead one. The rest of the list is real but worth questioning.

## Three icon libraries, for icons

```
85M  react-icons        6 icons imported, all in components/icons/brand-icon.tsx
59M  @lobehub/icons     22 icons imported, all in components/vault/provider-icon.tsx
40M  lucide-react       149 import sites — this is the UI icon set, keep it
32K  src/lib/mcp-logos.generated.ts   a fourth source, generated by `bun run gen:mcp-logos`
```

`lucide-react` is the design system and is not in question. The other three all answer the same question — *what is the logo for this third-party service* — and they answer it in three different places with three different mechanisms.

- **`react-icons` ships 85 MB to provide `FaAws`, `FaMicrosoft`, `FaSlack` and three `Si*` icons.** Six SVGs. `brand-icon.tsx`'s own comment says the point is "never hand-authored SVG paths", which is a good principle and does not require the entire Font Awesome and Simple Icons catalogues to be on disk.
- **`@lobehub/icons` ships 59 MB for 22 model-provider logos**, each imported individually from a deep path (`@lobehub/icons/es/Anthropic/components/Mono`).
- **`mcp-logos.generated.ts` already demonstrates the alternative** — a generator that pulls what is needed at build time into one checked-in file, with a `bun run gen:` script and a "do not hand-edit" rule in `.claude/rules/frontend.md`.

**The obvious move is to extend that generator to cover all three** and drop both packages. One mechanism, one place, 144 MB off the install, no runtime dependency, and the icon set stops being three answers. Cost: the generator has to handle two more sources, and adding a provider means running a script instead of adding an import — which is already true for MCP logos.

Do measure the bundle first. If both libraries already tree-shake to near-nothing, this is a CI-time and consistency argument rather than a payload one, and it should be argued on those terms rather than on the 144 MB.

## Small, and worth a sentence each

- **`nanoid`** — one call site, `use-chat.ts`, for a client-side message id. `crypto.randomUUID()` is in every browser this product supports and in Node 19+. A dependency for one id generator.
- **`yaml`** — one call site, `version-history.tsx`, for `stringify` when showing a spec. That is a real feature and YAML is not something to hand-roll. Keep, but it is worth knowing it is 1.2 MB for one function on one page, and that the *backend* already exports YAML (`GET /agents/{id}/spec.yaml`) — so the frontend may be re-serializing something it could fetch.
- **`@opentelemetry/api` + `@vercel/otel`** — 5.8 MB together, and genuinely used: `instrumentation.ts` calls `registerOTel({ serviceName: "agenticos-frontend" })`. The real question is not whether they are imported but **whether anything receives the traces.** The backend's Logfire setup is thorough; if the frontend is emitting spans into a void on every deployment, that is either a gap to close or two dependencies to remove, and right now nobody knows which.
- **`cmdk`** (116 KB) powers ⌘K, **`class-variance-authority`** three UI primitives, **`clsx` + `tailwind-merge`** the `cn()` helper, **`recharts`** the charts. All small, all load-bearing, all fine.

## What would keep it from coming back

`knip` finds unused dependencies, unused exports and unused files in a TypeScript project, and it would have flagged `date-fns` the day it stopped being used. Wiring it into `make lint` is the difference between doing this once and doing it every six months.

Note **#109**: TypeScript 7 and ESLint 10 are each blocked by a different plugin. Any tool added here becomes a fourth thing to keep compatible — pick one that is actively maintained.

## Watch out

- The frontend coverage gate is 100% and currently **failing on `main`'s CI** (#143). Deleting a file changes the denominator; run `bun run test:coverage`, not `bun run test:run`.
- `.claude/rules/frontend.md` forbids hand-editing `mcp-logos.generated.ts`. Extending the generator means extending the rule's wording too.
- Removing an icon package means every icon it provided has to still render — a missing brand logo is a silent visual regression that no test catches. Screenshot the vault, MCP and agent-builder icon rows before and after.

## Done when

- [ ] `date-fns` is gone
- [ ] `bun run analyze` output recorded before and after, so the bundle claim is measured rather than asserted
- [ ] A decision in writing on the three icon sources — one mechanism, or a reason for three
- [ ] `nanoid` replaced with `crypto.randomUUID()`, or kept with a reason
- [ ] The OpenTelemetry question answered: something consumes the frontend's traces, or the packages go
- [ ] `knip` (or equivalent) runs in `make lint` and fails on an unused dependency
- [ ] Every icon that rendered before still renders, verified visually
- [ ] `bun run type-check && bun run lint && bun run test:coverage` clean

### LINKED ISSUE #836: gen:mcp-logos writes an unformatted file, so make lint fails until the commit hook rewrites it
`bun run gen:mcp-logos` writes `frontend/src/lib/mcp-logos.generated.ts` unformatted, so the next command you run fails on the file the generator just wrote:

```
$ bun run gen:mcp-logos
Wrote 11 logos → src/lib/mcp-logos.generated.ts
$ make lint
Checking formatting...
[warn] src/lib/mcp-logos.generated.ts
make: *** [lint-frontend] Error 1
```

It looks formatted in the repository today only because `lint-staged` runs `prettier --write` over staged `.ts` files, so the shape it has on `main` was applied by the commit hook rather than by the generator. Anybody who regenerates and runs `make lint` before committing hits a failure that has nothing to do with their change, and the obvious reading of it — "prettier disagrees with a generated file" — points at the wrong fix.

**Where:** `frontend/scripts/gen-mcp-logos.ts:107` — `await writeFile(OUT, body, "utf8")` is the last thing it does.

**Fix:** the same two lines `scripts/gen-brand-icons.ts` already carries, added in #156:

```ts
execFileSync("bunx", ["prettier", "--write", "--log-level", "warn", OUT], { stdio: "inherit" });
```

**Done when:** `bun run gen:mcp-logos && make lint-frontend` is green from a clean checkout, with nothing staged and no commit in between.

Pre-existing — found while writing the brand-icon generator in #156, which hit the identical failure and solved it this way. Same cluster as #156 and #155 (the tooling half of the dependency work); no behaviour changes, and nothing in the product reads the formatting.

DEENUU1:
## What this fixes

`frontend/node_modules` was **1.2 GB across 1012 packages**. `date-fns` was declared and imported nowhere, `nanoid` duplicated a helper the chat store already had, and three separate mechanisms answered one question — *what is the logo for this third-party thing*. This makes it one mechanism, removes four dependencies, and puts a gate under the one that went unnoticed longest.

Closes #156
Closes #836

## What changed

**One glyph set, generated.**

- `frontend/scripts/gen-brand-icons.ts` — fetches every mark from the sets that own them (Simple Icons + Font Awesome for services and connectors, `@lobehub/icons-static-svg` for model providers) and writes the glyph data. Adding a mark is a row in `BRANDS`/`PROVIDERS` plus a re-run — never an import, never a hand-authored `d`.
- `frontend/src/lib/brand-glyphs.generated.ts` — 89 marks as raw SVG path data, 97 KB.
- `frontend/src/components/icons/glyph.tsx` — the one thing that turns glyph data into an `<svg>`, so `BrandIcon` and `ProviderIcon` draw identically rather than agreeing by accident.
- `brand-icon.tsx`, `provider-icon.tsx`, `channel-platform-icon.tsx`, `surface-icon.tsx` — all four read the generated set. `react-icons` and `@lobehub/icons` are gone.

**Two dead dependencies.**

- `date-fns` — 26 MB, zero imports, *and listed in knip's own `ignoreDependencies`*, so the report that would have found it had been told not to look.
- `nanoid` — four call sites, all wanting a client-side id. `chat-store.ts` already had `newPartId()` doing the same job with a non-secure-context fallback. Both now call `clientId()` in `src/lib/ids.ts`; the fallback stays, because `crypto.randomUUID` is undefined on plain HTTP off `localhost` — which is where an embedded widget on an internal host runs.

**Three tools that were dead or unguarded.**

- `bun run analyze` produced **nothing**. `@next/bundle-analyzer` is a webpack plugin, Next 16 builds with Turbopack, so every run printed "not compatible with Turbopack builds, no report will be generated" and exited 0. It is `next experimental-analyze` now, and the dependency is dropped.
- `make lint` now fails on a declared dependency nothing imports (`bun run lint:deps` — knip narrowed to that one question). knip moves from `bunx knip@5` to a pinned devDependency, and its config to `knip.jsonc` so each ignore carries its reason on the line above it.
- Nothing consumed the frontend's OpenTelemetry spans — see below.

## How it failed before

Three mechanisms, three answers to one question, and no gate under any of them:

```
85M  react-icons        6 icons + 63 Si*, imported into one record
114M @lobehub/icons     21 provider marks, deep-imported one file each
26M  date-fns           imported by nothing at all
```

`@lobehub/icons` also dragged in a **second copy of `lucide-react`**, `antd-style` and `es-toolkit` as transitive dependencies, none of which this app uses.

## Testing

Measured, before and after, on a clean install:

| | before | after |
|---|---|---|
| `node_modules` | 1.2 GB / 1012 pkgs | **729 MB / 722 pkgs** |
| client JS, raw | 4623.9 KiB | 4577.4 KiB |
| client JS, gzip | 1866.9 KiB | 1847.8 KiB |
| `react-icons` in the bundle | 76.0 KiB raw / 30.5 gzip | 0 |
| `@lobehub/icons` in the bundle | 60.9 KiB raw / 27.6 gzip | 0 |

**The issue's caveat was right**: both packages tree-shook, so 137 KiB of library code became 90 KiB of checked-in path data and the payload win is ~19 KiB gzipped. The argument is the install — **471 MB and 290 packages** off every CI job and every checkout — and having one mechanism instead of three.

**"Every icon that rendered before still renders" is measured, not eyeballed.** Each of the 89 marks was rendered from the removed package (`react-icons@5.6.0`, `@lobehub/icons@5.15.0`) with `renderToStaticMarkup`, and its `viewBox` and every `d` diffed against the generated glyph:

```
89 identical, 0 differ
```

Automated tests:

- `src/components/icons/brand-icon.test.tsx` — every name in the set draws a path; no mark carries a literal fill; a Font Awesome viewBox is not forced to 24×24.
- `src/components/icons/glyph.test.tsx` — every path of a layered mark is drawn, in order, with its `fill-opacity`; the fill rule is carried only where the mark needs it.
- `src/components/vault/provider-icon.test.tsx` — extended from one provider to **every** provider in the set.
- `src/lib/ids.test.ts` — the secure-context branch and the fallback.
- Six specs that identified a mark by its source SVG's `<title>` now use `src/test-utils/brand-marks.ts`, which names a mark by its path data. Stricter than the title was: a title proved *a* mark rendered under that name; this proves the drawn shape is the one the set holds for that id.

`make check` — every CI job except e2e — passes. Frontend suite: 308 files, 4696 tests, coverage gate green (statements 100%, branches 97.6%, functions 100%, lines 100%).

**Not verified:** e2e (needs a seeded backend) — CI will run it. No screenshots were taken; the byte-identical diff above is the stronger check and covers all 89 marks rather than the handful a screenshot would show.

## The OpenTelemetry question, answered

`instrumentation.ts` registers the SDK on every boot. Off Vercel, `@vercel/otel` attaches an exporter only when `OTEL_EXPORTER_OTLP_ENDPOINT` is set — and **no compose file, Dockerfile or CI job passed it through**, so the spans were built and dropped in-process.

Kept rather than removed, because the packages cost **0 KiB in the client bundle** (server-only, and measured above) and the capability was one variable away from working. So the variable is passed through in `docker-compose-prod.frontend.yml`, and `instrumentation.ts` and `frontend/README.md` now say plainly what unset means instead of leaving the next reader to find out.

## Docs

`.claude/rules/frontend.md` and the `frontend-feature` skill gain the "a brand mark is one row in a generator's table" rule and the knip gate. `docs/commands.md` records what `make lint` and `make dead-code` each run now. `NOTICE` gains the attribution the vendored path data needs — Font Awesome's icons are CC BY 4.0, and the licence banner its SVGs carry is stripped from the data and reproduced in the generated file's header.

## Noticed, not fixed

Filed rather than folded in:

- **knip 6 is available; this pins 5.** Upgrading is its own change, and per #109 every tool here is a compatibility constraint.

## Also here — #836

`gen-mcp-logos.ts` had the problem this generator solves: it wrote `src/lib/mcp-logos.generated.ts` unformatted, so `bun run gen:mcp-logos && make lint` failed on the file the script had just written, and it only looked formatted on `main` because `lint-staged` rewrote it on commit. It now runs prettier on its own output, the same two lines `gen-brand-icons.ts` carries. Verified from a clean checkout, nothing staged: `bun run gen:mcp-logos && make lint-frontend` is green.


DEENUU1:
@codex review

Issue #554, opened 2026-08-10
mcp-server-list still uses native confirm(): no busy-guard, so a double-click can double-DELETE


DEENUU1:
`components/mcp/mcp-server-list.tsx:360-372` (`handleDisconnect`) uses native `window.confirm()`. `rg "\bconfirm\("` across the scope returns only this site; `rag/[id]/page.tsx`, `agents/**`, `skills/page.tsx`, `embeds-panel.tsx` all migrated to `ConfirmDialog`.

**How it fails.** The destructive-confirm pattern was deliberately migrated off `window.confirm` everywhere (rag/[id] documents why). This lone holdout carries the three regressions the others fixed: untranslatable English prompts, a non-accessible native modal, and — the real bite — **no `confirmBusy` guard**, so a double-click can fire a second DELETE where the migrated sites cannot.

**Fix.** Replace with `ConfirmDialog` + `t()` + a `confirmBusy` flag, mirroring `rag/[id]`.

**Done when.** Disconnect uses `ConfirmDialog`; a second click during the in-flight DELETE is a no-op.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

DEENUU1:
Consolidating #611 into this one — same file, same function, same line (`mcp-server-list.tsx:360`, `handleDisconnect`), same fix. Keeping this one because its title names the part that actually bites: no `confirmBusy` guard, so a double-click can fire a second DELETE where every migrated site cannot. `severity:medium` / `effort:m` here is the more accurate triage than #611's `severity:low` / `effort:s`.

Three things from #611 worth carrying, because they are not in this body:

- **It cannot say what it costs.** `ConfirmDialog` has a `description`, which is where "agents bound to it keep running, without its tools" belongs. Today the whole warning is crammed into the one string `confirm()` renders, so `mcp.disconnectOrgWarning` should split into a title and a description rather than move across whole.
- **`confirm()` is synchronous and blocks the event loop**, so the disconnect path is untestable from Playwright without a dialog handler — and `e2e/` has no coverage of it at all. Every other destructive flow is covered precisely because the dialog is in the DOM. So the acceptance criteria gain one: an e2e spec that opens the dialog, cancels, and asserts the connection is still there.
- **The copy is already translated** either way — #395 moved both strings into `mcp.*` — so this is about the control, not the words, and the i18n guard stays green through the change.

Consolidated during the #683 map rebuild.

PR #844, opened 2026-08-17
refactor(rag): move KB detail to React Query; type admin keys; audit cleanups

### LINKED ISSUE #557: useKBDetail holds server data in useState off React Query; two registered query keys are dead
`hooks/use-knowledge-bases.ts:150-524` (`useKBDetail`) holds the collection, documents, sync sources and connectors in seven `useState` slots and refetches by hand. `lib/query-keys.ts:188-189` registered `kb.detail`/`kb.documents` for exactly this data — `rg` finds zero references outside the registry.

**How it fails.** Every sibling resource hook lives in the query cache; `useKBDetail` alone does not. Nothing outside the page can invalidate it — a rename in the list, or an ingestion-status change, never reaches the detail page — and it forced a bespoke tenant-switch clearing block that `queryClient.clear()` already does for every other hook.

**Fix.** Move the KB + documents onto `qk.kb.detail(id)` / `qk.kb.documents(id)` (pagination via `useInfiniteQuery` or a page param); sync sources/connectors as their own keyed queries. The org-switch drop then covers it and the dead keys become live.

**Done when.** An external mutation invalidates the detail page; the two keys are referenced.

---

### LINKED ISSUE #558: Admin query-key factories type params as unknown; admin.ratings is keyed three ways
`lib/query-keys.ts:296-299` (`admin.users/conversations/ratings/system`) all take `params?: unknown`. `admin.ratings` is keyed three incompatible ways: `use-dashboard-data.ts:138` `{summary:true,from,to}`; `admin/ratings/page.tsx:62` `{summary:30}`; `:72` `{page,filter,commentsOnly}` — and the page's request at `:63` sends **no params**, so `{summary:30}` is a magic discriminator matching nothing on the wire.

**How it fails.** The registry exists so callers cannot drift; `unknown` defeats it, and a future typo (`{summry:…}`) silently mints a fresh cache key instead of failing.

**Fix.** A real param type per factory; at minimum `admin.ratings: { from?: string; to?: string }`, and drop the `{summary:30}` fiction so the page and widget share one window.

**Done when.** The factories are typed; the ratings page and widget key the summary the same way.

---

OchnikBartek:
## Summary

The query-keys cluster of the 2026-08-10 frontend duplication audit — three
issues that share `src/lib/query-keys.ts`, worked in one branch so the shared
file has one owner.

## Related issues

- Closes #557 — `useKBDetail` onto React Query
- Closes #558 — typed admin query-key factories
- Part-of #569 — the non-MCP items of the audit batch (the MCP-cluster items,
  which share `mcp-server-list.tsx` with #554, land in a separate PR)

## Changed

- **#557** — `useKBDetail` moved off seven `useState` slots onto
  `qk.kb.detail(id)` / `qk.kb.documents(id)` (documents via `useInfiniteQuery`)
  plus three keyed section queries. The bespoke `useChanged` tenant-clearing
  block is gone — `useTenantCacheReset`'s `removeQueries()` covers it — so the
  two dead keys are live and an external mutation can invalidate the page.
  `loadFailed` keeps a cold first-load failure (whole-page error) distinct from
  a failed refresh (stale under a banner).
- **#558** — `admin.conversations` / `admin.ratings` typed to their real params;
  the dead `admin.users` factory removed; the `{ summary: true }` discriminator
  dropped (the ratings page it distinguished against no longer exists).
- **#569 (non-MCP)** — dead `conversations.count` key removed; `patchKB`
  rethrows like its siblings; sandbox `usage` discriminator moved into the
  `sessions(id, usage)` factory; `api/files/[id]` encodes its path segment;
  the members table waits on the permission answer before drawing its action
  column; the message-body render IIFE extracted into `MessageBody`.

## Testing

- `bun run test:coverage` green — 53 KB-hook + 6 KB-detail integration tests
  rewritten to the query idiom, plus the sandbox / dashboard / members / files
  suites. Global gate met (branches 97.5%).
- `bunx tsc`, ESLint, Prettier, the i18n guard all clean (pre-commit on each
  commit).

## Notes for reviewers

- **Request volume**: documents refetch under `useInfiniteQuery` re-reads each
  loaded page at its own skip, so a poll over a long list issues one request per
  loaded page rather than a single `limit=100`. Deliberate — flagged in the #557
  commit body.
- **`patchKB`**: made to rethrow per #569, but it has **no live caller** — there
  is no KB rename dialog in the tree. Kept (contract-consistent primitive for a
  future dialog) rather than deleted; say the word if you'd prefer it removed.
- **Local production build**: could not run in the worktree (Turbopack rejects
  the shared `node_modules` symlink used for the fast test loop). `tsc` is clean
  and there are no new RSC boundaries; CI's `build-frontend` is the build gate.


DEENUU1:
@codex review

DEENUU1:
@codex review

PR #847, opened 2026-08-17
fix(api): stamp no-store on every backendFetch route response

### LINKED ISSUE #553: The backendFetch route family omits no-store, so a refetch-after-write can be served stale

OchnikBartek:
## Summary

The `backendFetch` route family returned `NextResponse.json(data)` with no
`Cache-Control` (orgs/**, me/**, admin/**, sessions/**, auth/**). `platformProxy`
already stamps `no-store` on any unmarked response (the #230 fix), and
`.claude/rules/frontend.md` says a hand-rolled route owes the same header.
Without it, the members / invitations / integrations lists refetched right after
a create/invite/revoke could be served from cache — the same staleness class as
#230, on the surface the shared proxy does not cover.

## Related issue

Closes #553 · Refs #230, #564

## Changed

- New `bffJson<T>(data, init?)` in `server-api.ts` stamps `Cache-Control:
  no-store`, leaving an explicit policy alone the way `platformProxy` does;
  `bffRefusal` stamps it too.
- Every `NextResponse.json(...)` in the **44** backendFetch route files now goes
  through `bffJson` — one chokepoint, so no route can forget the header. Applied
  by a reviewed script (dropped the now-unused `NextResponse` import where a file
  no longer referenced it; kept it in the 11 files that still `new NextResponse`).

## Testing

- `bun run test:coverage` green — 391 API-route tests pass. New `server-api.test.ts`
  cases cover `bffJson` (no-store, status preserved, explicit policy left alone)
  and `bffRefusal`. The binary routes that set their own `Cache-Control`
  (private/public `max-age`) keep it — their existing header assertions still pass.
- `tsc`, ESLint, Prettier all clean.

## Notes for reviewers

- Scope is `no-store`, not a `platformProxy` migration (#564) — the issue's
  alternative. The chokepoint helper is the lighter of the two and leaves the
  proxy migration free to happen later.
- Local production build not run in-worktree (Turbopack rejects the shared
  `node_modules` symlink); CI's `build-frontend` is the build gate.


DEENUU1:
@codex review

PR #918, opened 2026-08-19
fix(rag): check secret-view before binding an embedding key

### LINKED ISSUE #912: rag: binding an embedding secret skips the caller's secret-view check
`KnowledgeBaseService._check_embedding_secret` (`backend/app/services/knowledge_base.py`) looks a chosen embedding key up scoped only to the organization and never runs the caller's secret-view check:

```python
row = await organization_secret_repo.get(self.db, secret_id, organization_id=organization_id)
if row is None:
    raise ...
```

So a member with `collections:edit` who supplies the UUID of another member's **private** OpenRouter secret binds a key `secrets:view` would refuse them. The picker only offers keys they can see, but the API takes an id and an id is guessable.

This is the exact gap fixed for the rerank key in #142 / PR #911 (commit `117e8f7d`): `_check_rerank_secret` now calls `resolve_access(..., SECRETS_VIEW)`. `_check_embedding_secret` is its mirror and predates that branch, so it is filed separately rather than folded into a rerank PR.

**Trigger**: member A creates a private (visibility=private) OpenRouter secret; member B (same org, `collections:edit`, no share/grant on A's secret) POSTs `/kb` with `embedding_secret_id` = A's secret id. Expected: refused as "not in this organization's vault". Actual: accepted; embeddings later bill A's key.

**Fixed when**: `_check_embedding_secret` runs `resolve_access(self.db, ctx, row, Perm.SECRETS_VIEW, resource_type=SECRET)` exactly as `_check_rerank_secret` and agent secret bindings (`agent_registry.py:1223`) do, with a regression test.

Found reviewing #142 (PR #911). Same RAG cluster as #142 in the issue map #168.

OchnikBartek:
## Summary

`KnowledgeBaseService._check_embedding_secret` looked the chosen embedding key up scoped only to the organization and never ran the caller's secret-view check. A member with `collections:edit` who supplied the UUID of another member's **private** OpenRouter secret bound a key `secrets:view` would refuse them — the picker only offers what they can see, but the API takes an id and an id is guessable. Embeddings then billed the other member's key.

## Changed

- `_check_embedding_secret` now takes `ctx` and runs `resolve_access(self.db, ctx, row, Perm.SECRETS_VIEW, resource_type=SECRET)` on the fetched row, exactly as the agent secret bindings do (`agent_registry.py:1226`).
- A key the caller cannot view is phrased as one the vault does not hold, so a refusal cannot enumerate another member's private secrets.

## Testing

- New regression test in `tests/test_kb_scoping.py`: a member is refused another member's private secret supplied by id (accepted without the check), and a viewable secret is still bound.
- `uv run pytest tests/test_kb_scoping.py tests/test_ingestion_embedding_key.py tests/test_embedding_resolution.py` green; pre-commit (ruff/ty/vulture) green.

Closes #912


DEENUU1:
Merged `main` in (it was four commits behind) and added the docs half this was missing: `docs/file-processing.md` listed the two refusals creation already made — another organization's key, the wrong purpose — so a reader acting on that page would still have expected `collections:edit` alone to be enough to bind a key. It now names `secrets:view` and says why the check exists rather than only that it does.

Reviewed the rest of the surface for the same hole and there isn't one: `KnowledgeBaseUpdate` carries no `embedding_secret_id`, so creation is the only path that binds, and `embedding_resolution.py` only reads back what was stored.

`tests/test_kb_scoping.py tests/test_ingestion_embedding_key.py tests/test_embedding_resolution.py` green (64), `make docs-build` clean.

PR #920, opened 2026-08-19
feat(runs): what the model was given, and where the time went

DEENUU1:
The run detail could say what happened and what it cost. It could not say
what the model was **given** — which prompt, which tools, described how,
under which settings — and it was dropping half of what the wire already
sent it. Three commits, each standing on its own.

## What was broken, and was not a gap in the API

A question with a document attached rendered as a question with nothing
attached. `MessageRead` has carried `files` since attachments existed — the
repository eager-loads them on every transcript read — along with the
per-turn model, token split, cost and context size. The timeline rebuilt its
argument to `conversationMessageToChatMessage` field by field and dropped all
of it on the way, on the one surface whose job is to say what reached the
model. The stored row is now handed over whole, which is what that mapping
was written for.

So a turn shows its files — openable through the shared `FileViewer`, because
"the agent answered badly" and "the agent was handed a scan with no text
layer" are the same transcript until somebody can look at the file — and
beside them the model that answered, what the turn cost, and how much history
it carried.

## What the model was given, recorded rather than reconstructed

None of it is derivable afterwards. The prompt the model saw is the spec's
instructions plus the platform's, plus whatever a channel binding appended,
plus the bound skills, plus whichever reminder fired; the tool schemas come
from the capability registry plus the organization's MCP servers minus
whatever tool search hid. Rebuilding that from the stored spec would be a
second implementation of the builder, and a second implementation disagrees.

`RecordingModel` wraps the model the agent was built with — the trick
`MeteredModel` already uses to book a sub-agent's spend — and writes down each
request as it passes: instructions and system parts,
`ModelRequestParameters.function_tools` exactly as the provider received them,
the settings, one entry per request with its duration, tokens and what it
asked to call next, and the last request's whole message list. What is stored
is what was sent.

`GET /runs/{id}/manifest` reads it back, authorized the way the transcript is:
existence resolved against the caller's organization first, then `runs:view`.
A second tab on the run drawer renders it, with the requests as bars scaled
against the slowest one — a run that took forty seconds is either one slow
request or nine quick ones with eight tool calls between them, opposite
problems the run row cannot tell apart.

Three decisions worth reviewing:

- **A table, not a column on `agent_runs`.** That table is the most-listed in
  the product; a JSONB document holding every tool's JSON schema would be read
  by run history, spend, the dashboard and the export to answer a question
  none of them asks.
- **Provider passthrough is never recorded.** `extra_headers` and `extra_body`
  are where a provider credential rides, and the vault is the only place a
  secret is kept.
- **The write is guarded *and* in a SAVEPOINT.** Swallowing the exception is
  not enough: a failed flush leaves the session unusable, so the run's own
  terminal write would be lost to a record nobody asked for. That is not
  hypothetical — it took `test_delegation_row_failures` down before the
  SAVEPOINT went in, and the integration test now forces the same failure with
  a NUL byte Postgres genuinely refuses.

## Stepping between runs

Each run is a query key of its own, so an arrow press dropped the whole view
to a skeleton and rebuilt it. Both queries now hold their previous answer (the
pattern `query-freshness.ts` already sets out), the neighbours are prefetched
on arrival so the step is a cache hit, and the timeline is left standing — two
runs of one conversation are the same turns, so what moves is the anchor,
which glides. The arrow keys do what the buttons do. Nothing navigates from a
held row: its neighbours are its own, and acting on them would undo the step
just made.

## Verified

`make check` — lint, the backend suite (5772 passed) at 100% on the platform
layer, the frontend gate (100% lines/statements/functions, 97.67% branches),
the frontend build, the docs build and the dependency audit. The migration
round-trips and `alembic check` is clean against a database at its own head.

Not verified: nothing here was exercised through the browser. `make db-check`
also fails on my machine, and only there — see below.

## Merge order (needs doing before this lands)

This branch's migration is `0037_run_manifest`, chained onto `0036`, which is
correct against `main` today. #914 carries a `0037_deployment_settings` off
the same parent. **Whichever merges second must renumber**, or `main` ends up
with two alembic heads and `alembic upgrade head` fails outright. #914 is
green and ahead, so the expectation is that this one becomes `0038` chained
onto it — a three-line change I will make once #914 lands.

Closes nothing; this is new behaviour rather than a filed defect.


DEENUU1:
Wrong branch on my part — this work belongs on #914 and has been moved there, with the migration renumbered to 0038 so the two no longer collide.

Issue #956, opened 2026-08-19
The ingestion flow's _update_status has a branch nothing can reach, and a status parameter with one value


DEENUU1:
`_update_status` in the ingestion flow takes a status and branches on two
values. Every one of its four callers passes the same one.

## Where

`backend/app/worker/tasks/rag_tasks.py:465-494`, called from `:212`, `:297`,
`:318` and `:331` — all four with `"error"`.

```python
async def _update_status(
    rag_document_id: str, status: str, error_message: str | None = None
) -> None:
    ...
            if status == "error":
                if (await doc_svc.get_document(rag_document_id)).status == "error":
                    return
                await doc_svc.fail_ingestion(...)
            elif status == "done":
                # vector_document_id required for complete_ingestion; callers
                # set status="done" directly in _run_ingestion with the ID.
                pass
```

So the `elif` is a `pass` that cannot be reached, the `if` is always true, and
the parameter is a constant dressed as a choice. The comment under the dead
branch explains why nothing takes it, which is the tell: it documents an
absence rather than a behaviour.

## Why it is worth taking

- CLAUDE.md: "If a branch cannot be reached, delete it; if it can, test it."
  This one cannot, and `vulture` does not see it because the parameter is read.
- The function's real job is narrower than its name: it records the **first**
  failure and refuses to overwrite it (#423). `_fail_document(rag_document_id,
  error_message)` says that; `_update_status(..., "error")` hides it behind a
  vocabulary the caller must know.
- The status literals here are strings, while the column's vocabulary is now
  `DocumentStatus` (#148). Two spellings of one set of values is how #148
  happened in the first place.

## Fixed when

`_update_status` is `_fail_document`, taking no status; the four call sites drop
their `"error"`; the dead branch is gone; and the "first failure wins" behaviour
that #423 established still has its test.

Found while fixing #148 — the same status vocabulary, one module over.
Deliberately not folded into that branch, which was about the count the listing
reads. Nearest open siblings: #566 (collapse the duplicated document lookups in
the same subsystem) and #948 (the vector store engine the same worker leaks per
document). Neither overlaps this function.

PR #957, opened 2026-08-19
fix(rag): count the status the pipeline writes, not one nothing writes

### LINKED ISSUE #148: rag: KB stats count status "completed", but ingestion writes "done", so indexed_count is always 0

DEENUU1:
`indexed_count` was `0` on every knowledge base in the product, however many
documents had finished ingesting.

## What was wrong

The listing's aggregation filtered on a status value nothing writes —
`repositories/rag_document.py:203`:

```python
func.count().filter(RAGDocument.status == "completed"),
```

The pipeline writes `processing`, `done` and `error`, which is what the model
default, `docs/file-processing.md` and the frontend's `rag-status.ts` all say.
The `FILTER` clause therefore could not match a row: `documents` and `indexed`
disagreed on every collection, and a collection where everything succeeded read
as entirely unindexed.

## What changed

The literal is now an enum. `DocumentStatus` lives beside the column in
`app/db/models/rag_document.py` — the shape `RunStatus`, `AgentStatus` and
`InvitationStatus` already take — and the writer (`RAGDocumentService`) and the
reader (`counts_by_collection`) name the same member rather than two strings
that were free to drift. `create` and `update_status` take `DocumentStatus`
instead of `str`, so the next typo is a type error rather than a count that
silently reports nothing.

`docs/file-processing.md`'s status row now says the three values are the
`DocumentStatus` members and that the indexed count filters on `done`.

## Why it survived

The test that asserted this behaviour set the statuses by hand, and it set
`"completed"` and `"failed"` — neither of which anything in the product writes.
It agreed with the query instead of with the pipeline, so it passed for as long
as the bug existed.

It now sets `DocumentStatus` members, and a second test names no status at all:
it runs `complete_ingestion` and then reads `counts_by_collection`, so the writer
and the reader have to agree or it fails. Both fail against the old filter —
verified by putting `"completed"` back and re-running (`2 failed, 4 passed`).

## Verified

- `tests/integration/test_platform_flows.py -k WhatACollectionReportsItHolds` —
  6 passed, against a real `pgvector/pgvector:pg16`.
- Every unit and API test touching `rag_documents` — 268 passed, 2 skipped.
- `make lint-backend` clean, with the same 61 `ty` diagnostics `main` reports.
- `make db-check` **fails identically on `main`** in this checkout: the dev
  database is stamped at a migration renumbered before merge
  (`0047_trigger_fire_in_flight`). It says nothing about this change; CI runs it
  against a fresh database. No DDL changed here — the enum is a Python-side
  default.
- No frontend path changed, so `test-frontend` is not expected to run.

## Found and not fixed

`_update_status` in `app/worker/tasks/rag_tasks.py` branches on `"error"` and
`"done"`, and all four callers pass `"error"` — so the `done` branch is a `pass`
that cannot be reached. Filed as **#956** rather than folded in here.

Closes #148

### DIFF
--- a/backend/app/db/models/rag_document.py
+++ b/backend/app/db/models/rag_document.py
@@ -1,5 +1,6 @@
 """RAGDocument model - tracks documents ingested into RAG collections."""
 
+import enum
 import uuid
 from datetime import datetime
 
@@ -9,6 +10,22 @@ from sqlalchemy.dialects.postgresql import UUID as PG_UUID
 from sqlalchemy.orm import Mapped, mapped_column
 
 from app.db.base import Base, TimestampMixin
+
+
+class DocumentStatus(enum.StrEnum):
+    """Where a tracked document is in its ingestion.
+
+    Three values, and the column holds nothing else: a row is created
+    `PROCESSING` by the upload, and the pipeline moves it to `DONE` or `ERROR`.
+    It is an enum because it was not one - the listing's `indexed` count
+    filtered on a fourth value, `"completed"`, that nothing has ever written, so
+    every knowledge base in the product reported `indexed_count: 0` however many
+    documents had finished (#148).
+    """
+
+    PROCESSING = "processing"
+    DONE = "done"
+    ERROR = "error"
 
 
 class RAGDocument(TimestampMixin, Base):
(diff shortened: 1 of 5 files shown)


DEENUU1:
@codex review

Issue #545, opened 2026-08-10
Backend duplication & quality cleanup (audit 2026-08-10)


bll-kacper-wlodarczyk:
Batched small backend items from the 2026-08-10 duplication audit. Each is Low / effort:s.

- [x] `InvitationCreate.email` (`schemas/organization.py:105`) drops the `max_length=255` its `UserCreate`/`UserUpdate` siblings carry → over-length address 500s at the DB instead of a clean 422.
- [x] `RAGCollectionList` (`schemas/rag.py:52`) and `ConnectorList` (`schemas/sync_source.py:86`) omit the `total` the `*List` convention (and `CatalogIconList`) carry.
- [x] `parked_calls` (`agent_runner.py:2769,2784`) compares status with raw string literals where every sibling uses `RunStatus`/`ApprovalStatus` — a rename breaks it silently.
- [x] `skills.py:372,567` / `repositories/skill.py:188,243` take `update_data: dict` (bare) instead of `dict[str, Any]` / the dumped `*Update`.
- [ ] `agent_runner.py` widens `extra_toolsets`/`message_history` to `list[Any]` where the callees type them precisely.
- [x] `commands/doctor.py:49,93,148` use `Any` as a shortcut where the models exist.
- [ ] `agents.py:179-196,259-275` (`get_agent`/`get_version`) hand-map the response field-by-field; the JSONB→`AgentSpec` parse belongs in a `model_validate` on the Read schema.
- [ ] `cost_usd` crosses the chat socket as a `float` (`usage_report.py:284`) but as a Decimal string on REST — pick one wire representation.
- [ ] The MCP tool-name prefix normaliser is copy-pasted (`agents/mcp.py:158` ↔ `lib/tool-steps.ts:80`) with hand-copied test expectations — add a parity fixture (the #144 shape).
- [ ] The visibility disjunction is hand-copied in `repositories/skill.py:104` and `knowledge_base.py:52,74`, with `false()` vs Python `False` drift — one shared predicate helper.
- [ ] The `parent_doc_id` filter is sanitised inline in `retrieval.py:260` and again as `_sanitize_id` in `vectorstore.py:121`, then serialised and regex-parsed back (`:390`) — pass it as a typed argument.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

OchnikBartek:
[...]

PR #965, opened 2026-08-19
fix(api): serialize both upload responses the same way

### LINKED ISSUE #560: Two upload routes serialize the same schema differently (response_model_exclude_none)
`api/routes/v1/rag.py:299-305` (`ingest_file`, `response_model_exclude_none=True`) vs `knowledge_bases.py:149-153` (`upload_kb_document`, no such flag). Same `RAGIngestResponse` (`document_id: str | None`), same operation, both feed the KB upload UI. `response_model_exclude_none` appears in exactly one route in the tree.

**How it fails.** One route omits `document_id` when null; the other emits `"document_id": null`. A client normalising the response gets a present-but-null key on one surface and a missing key on the other. (The two route bodies are near-duplicates — the shared `Form(...)` description has also drifted; batched in the FE/BE cleanup issue.)

**Fix.** Pick one (dropping `exclude_none` is simplest — `null` is honest here).

**Done when.** Both upload routes serialize `document_id` identically.

---

DEENUU1:
The same schema, the same operation, two shapes on the wire.

`POST /rag/collections/{name}/ingest` carried `response_model_exclude_none=True`
and `POST /kb/{kb_id}/documents` did not — both returning `RAGIngestResponse`,
both feeding the same upload UI. `document_id` is `str | None` and is `None` on
every accepted upload, because the vector store's id does not exist until the
worker has indexed the file. So one address omitted the key and the other sent it
as `null`, and a client normalising the answer got a different shape depending on
which it had called.

## The fix

The flag is gone rather than added to the other route. `null` is the honest answer
here — the id is pending, not absent — and it was the **only** use of
`response_model_exclude_none` in the tree, so keeping it would have meant
defending a convention with one member.

## Verified

Two tests at both addresses, in the file that already parametrizes over them:

- `document_id` is present and `null` on a 202.
- The response's keys are exactly `RAGIngestResponse.model_fields` — asserted
  against the **schema**, not against the other route's answer, so it fails at
  whichever address drifts rather than only when the two disagree, and a field
  added to the schema that one route drops is caught by it too.

Both fail with the flag put back, and only on the `/rag` path — which is the
asymmetry itself.

- `make lint` clean, same 61 `ty` diagnostics as `main`.
- `make test`: **6140 passed, 15 skipped, 100%** on the platform layer.
- **No client is affected.** The frontend never reads `document_id` from an
  upload response — `grep` finds only `vector_document_id` on tracked documents,
  a different field on a different schema — so the added key changes nothing on
  screen. No frontend path changed, so `test-frontend` is not expected to run.

`docs/file-processing.md` now names both addresses and says they answer 202 with
every field of the schema, `"document_id": null` included.

## Deliberately not fixed

The `Form(...)` description for `ingestion` still differs between the two routes.
#560 says that half belongs to the batched FE/BE cleanup in #545, not here.

Closes #560

### DIFF
--- a/backend/app/api/routes/v1/rag.py
+++ b/backend/app/api/routes/v1/rag.py
@@ -299,7 +299,6 @@ async def delete_document(
 @router.post(
     "/collections/{name}/ingest",
     response_model=RAGIngestResponse,
-    response_model_exclude_none=True,
     status_code=status.HTTP_202_ACCEPTED,
     dependencies=[Depends(require(Perm.COLLECTIONS_EDIT))],
 )
--- a/backend/tests/api/test_ingestion_override_routes.py
+++ b/backend/tests/api/test_ingestion_override_routes.py
@@ -28,6 +28,7 @@ from app.core.permissions import AuthContext, OrgRoleName
 from app.db.models.knowledge_base import KBScope, KnowledgeBase
 from app.db.models.resource_grant import Visibility
 from app.main import app
+from app.schemas.rag import RAGIngestResponse
 from app.services import rag_document
 
 pytestmark = pytest.mark.anyio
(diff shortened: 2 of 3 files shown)


DEENUU1:
@codex review

Issue #548, opened 2026-08-10
Two RAG document-lookup helpers disagree on precedence; get_documents has no ORDER BY


DEENUU1:
`ingestion.py:54-68` (`_find_existing_by_source`) and `:173-188` (`get_existing_hash`) both answer "which stored doc is this `source_path`", with different precedence.

- `_find_existing_by_source`: **two full passes** — every doc checked for `source_path` first, then a second pass for `filename` — so a source-path match always wins globally.
- `get_existing_hash`: a **single interleaved pass** — a `filename` hit sets `doc_id` (no `break`) and the `if doc_id is None` guard then blocks any later `source_path` match.

`PgVectorStore.get_documents` runs `SELECT parent_doc_id, metadata FROM {table}` with **no `ORDER BY`**, so row order is arbitrary.

**How it fails.** Given two docs where a filename-only match precedes a source-path match in row order, `find_existing` returns doc B (source) while `get_existing_hash` returns doc A's hash. The worker's `new_only`/`update_only` sync (`rag_tasks.py:388,394,407`) then compares the live `file_hash` against the **wrong document's** `content_hash` → an unchanged file is re-ingested (needless re-embedding spend every sync) or a changed file is skipped (stale index).

**Fix.** One helper returning `(document_id, content_hash)` with a single explicit precedence (source over filename), used by both call sites; give `get_documents` a deterministic `ORDER BY` as defence. See #566.

**Done when.** The id path and the hash path return the same document for the same `source_path`; test: two docs (filename-match ordered before source-match) → both agree.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

Issue #27, opened 2026-08-01
Paginate GET /rag/documents, and stop full-scanning on ingest


DEENUU1:
**Severity: Medium** · AUD-025

## What's wrong

`GET /api/v1/rag/documents` selects every `rag_documents` row across all the caller's readable collections and serializes them in one response.

## Where

- `backend/app/repositories/rag_document.py:51-58` (`get_all` — no `limit`, no `offset`)
- `backend/app/services/rag_document.py:82-86` (`total=len(docs)`)
- `backend/app/api/routes/v1/rag.py:340-357` (no `skip`/`limit` query params)

## How it fails

Grows without bound with tenant data. A tenant with 50k ingested documents gets a multi-second query and a response of tens of MB, held entirely in memory. The sibling `get_for_kb` at `:61-76` pages properly, so the pattern exists — this path just does not use it.

## Fix

Give `get_all` `skip`/`limit` with a `COUNT` for `total`, and add the standard `skip: int = Query(0, ge=0), limit: int = Query(50, ge=1, le=100)` per `.claude/rules/api-conventions.md`.

## Related, worse, and worth doing together

`services/rag/vectorstore.py:412-429` (`get_documents`) runs `SELECT parent_doc_id, metadata FROM rag_<col>` with **no `WHERE` and no `LIMIT`** and groups in Python — and `services/rag/ingestion.py` calls it at three sites (`:55`, `:72`, `:150`) to answer "is there already a document with this source path". On a 200k-chunk collection that is a full table read into worker memory, once per ingested document.

Fix: a predicate query (`SELECT DISTINCT parent_doc_id ... WHERE metadata->>'source_path' = :p LIMIT 1`) with a supporting expression index.

In the same module, `vectorstore.py:308-324` inserts chunks one round trip at a time in a Python loop — a 500-page PDF is thousands of sequential statements where one `executemany` would do.

## Acceptance criteria

- [ ] The route respects `limit` and reports a `total` larger than the page
- [ ] Ingestion issues O(1) statements for the existence check

DEENUU1:
The RAG cluster this belongs to, previously linked only from #139:

- **#61** — the `/rag` page is the caller. A 1009-line page reading an unbounded endpoint is the same problem twice; whichever is done first should not assume the other
[...]

PR #967, opened 2026-08-19
refactor(rag): one document lookup, one scan, both answers

### LINKED ISSUE #566: One RAG document-lookup helper; collapse the 3-4 full-collection scans per ingest
`ingestion.py:54,70,173` each call `get_documents` (a full-collection read) and hand-roll a near-identical match loop. The worker's `new_only` path does up to **four** full scans to ingest one file, and the four match predicates have already drifted (#548).

**Cost today.** Four full-collection reads per ingested file, and four copies of the matching predicate that must stay in agreement (they do not).

**Target.** One private helper that fetches `get_documents` once and returns the matched `(id, hash)` with source-over-filename precedence stated in one place; the public methods project from it.

**Size:** M (S for the correctness half in #548).

Closes the root cause of #548.

---

DEENUU1:
Three lookups walked the same document listing with three predicates —
`_existing_by_source` (source_path, then filename), `_find_existing_by_hash`
(content hash), and two public methods each projecting one field from the first.
Every one of them read the whole collection.

## The count

Ingesting one changed file in `new_only` mode read the collection **four times**:
the sync asked for an id, then for a hash, and `ingest_file` then asked for both
again. An unchanged file cost two. It is now **one** for the sync's decision and
**one** inside the ingest.

## The reason both answers come back together

That is the smaller half. While the id and the hash were computed separately they
could **disagree** — the id lookup checked every document for a `source_path`
match before falling back to `filename`, while the hash lookup interleaved the
two. A caller then compared a live file's hash against a different document's
`content_hash` than the one it was about to replace, and either re-embedded an
unchanged file every night or skipped a changed one as current (#548).

`existing_document` returns a frozen `StoredDocument` carrying both, so a caller
**cannot ask for one without the other** and cannot pair them wrongly. That makes
#548 unrepeatable rather than merely fixed.

`find_existing` and `get_existing_hash` are gone rather than kept as wrappers: two
methods whose whole purpose was to project half of one answer are how the answer
got split in the first place. Both sync callers — the worker flow and `rag-sync`
in the CLI — did the identical two-call dance and now do one.

## Behaviour is unchanged

Precedence included: `source_path`, then `filename`, then `content_hash`. A store
that refuses answers "no match", as before, and that is now stated where it
happens: a listing this cannot read is not evidence the document is absent, but
treating it as a match would delete a document on the strength of a failed query.

## Verified

`tests/test_rag_document_lookup.py` keeps the five precedence tests #548 left,
rewritten onto the one call, and adds three:

- the hash fallback sits **behind** the filename match — a file that moved keeps
  its identity, a file that was copied does not take over the original's;
- an unhashed file does not match a document that has no hash either;
- two count `get_documents` awaits, which is what stops a fourth lookup being
  added later.

10 passed. `make lint` clean, same 61 `ty` diagnostics as `main`. `make test`:
**6144 passed, 15 skipped, 100%** on the platform layer. No frontend path changed.

## Not addressed

**#27** — paginating `get_documents`. One read per file is still one read too many
on a collection of thousands, and this change does not pretend otherwise.

Closes #566

### DIFF
--- a/backend/app/commands/rag.py
+++ b/backend/app/commands/rag.py
@@ -176,25 +176,23 @@ async def ingest_path_async(
 
             source_path = str(filepath.resolve())
             if sync_mode in ("new_only", "update_only"):
-                existing_id: str | None = await ingestion.find_existing(collection, source_path)
+                # One scan for both answers - see `existing_document` for why
+                # they are not two calls (#548, #566).
+                existing = await ingestion.existing_document(collection, source_path)
 
                 if sync_mode == "new_only":
-                    if existing_id:
+                    if existing.document_id:
                         file_hash: str = hashlib.sha256(filepath.read_bytes()).hexdigest()
-                        existing_hash: str | None = await ingestion.get_existing_hash(
-                            collection, source_path
-                        )
-                        if existing_hash and file_hash == existing_hash:
+                        if existing.content_hash and file_hash == existing.content_hash:
                             skipped_count += 1
                             continue
 
                 elif sync_mode == "update_only":
-                    if not existing_id:
+                    if not existing.document_id:
                         skipped_count += 1
                         continue
                     file_hash = hashlib.sha256(filepath.read_bytes()).hexdigest()
-                    existing_hash = await ingestion.get_existing_hash(collection, source_path)
-                    if existing_hash and file_hash == existing_hash:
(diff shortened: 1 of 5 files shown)


DEENUU1:
@codex review

Issue #939, opened 2026-08-19
Tabs for the RAG pages: Reusable integrations as /rag's third tab, and a knowledge base's three sections as three


DEENUU1:
Both RAG pages stack everything vertically. `/rag` has two tabs and then a third
section below the fold; a knowledge base has three sections and no tabs at all.
Make the structure match what is on the pages.

## `/rag`: Reusable integrations as the third tab

`rag/page.tsx:95` renders two triggers - **Knowledge bases** and **Search** - and
then, at `:176`, `<ReusableIntegrations>` under the base grid, with the reason in
the comment above it:

> Below the collections, because it is the thing they are fed from: a connector
> configured once and cloned into each base that needs it.

The relationship is right and the placement is not. On an organization with a
dozen bases the integrations are below a grid three rows deep, and they are also
*only* reachable from the Knowledge bases tab - a page-level concern hidden inside
one of two tabs. **Integrations** as the third trigger, beside the two that exist.

## A knowledge base: three sections, three tabs

`rag/[id]/page.tsx` stacks, in order: the documents table (`:221`), the ingestion
panel (`:239`, "How documents are read"), and the sync sources (`:242`). Each
carries a comment explaining why it is under the one above it - and each argument
is about *reading order on first visit*, which is not the same as where somebody
returns to. Somebody adjusting a parser scrolls past every document to reach it.

Three tabs: **Documents** · **How documents are read** · **Sync sources**.

Two things to keep while doing it:

- **The document count and the status strip stay above the tabs**, where they are
  now (`KBStatsStrip`) - they describe the collection, not one tab.
- **The override banner** ("Parsed with overrides… Review · Clear", `:200`) belongs
  with Documents, since it is about the next upload, and it must not disappear
  when somebody is on another tab without something saying it is still in force.

## And put the tab in the URL

`/rag` already holds its tab in local state (`useState<RagTab>`), so a link to the
Search tab does not exist and a reload loses the choice. Whatever tabs the KB page
grows should be `?tab=` from the start, and `/rag`'s existing two should join them
- the same fix **#934** makes for `/runs`, for the same reason: a tab that is not
addressable cannot be linked to from anywhere, including from the product's own
emails and onboarding.

## Done when

- `/rag` has three tabs and no section below the fold.
- A knowledge base has three tabs, with the stats strip and the override banner
  handled as above.
- Both pages carry `?tab=`, surviving a reload and travelling in a pasted link.
[...]

PR #976, opened 2026-08-20
feat(rag): a markdown editor, one width scale, and a mark on each parser

### LINKED ISSUE #940: The RAG dialogs: a markdown editor for the image prompt, one width scale, and a mark on the parser choices
Three complaints about the same dialogs, and they are one piece of work.

## 1. The image-description prompt is a bare textarea

`ingestion-settings.tsx:455` renders the **Describe images → Prompt** field as a
plain `<Textarea>`. It is a model prompt, several sentences long, and the product
already has the control for that: `MarkdownEditor`
(`src/components/ui/markdown-editor.tsx`) - source/preview toggle, the same one the
Builder uses for an agent's instructions, an exposure prompt, a page field and a
capability's generated form (`page-fields.tsx`, `exposure-prompt.tsx`,
`schema-form.tsx`, `embed-form.tsx`).

It is a primitive with `invalid` / `describedBy` already wired precisely so a
generated refusal can be shown under it - which this field needs, since
`INGESTION_LIMITS.prompt.maxLength` is enforced and reported. Swap the control;
nothing else about the field changes.

The same field appears in the **Create knowledge base** dialog, which embeds
`IngestionSettings` whole - so one change fixes both.

## 2. The Create knowledge base dialog is half the width of the same form elsewhere

| Dialog | Width | Holds |
|---|---|---|
| `create-kb-dialog.tsx:159` | `sm:max-w-lg` (512px) | name, description, scope, embedding secret **and all of `IngestionSettings`** |
| `ingestion-dialog.tsx:90` | `sm:max-w-3xl` (768px) | `IngestionSettings` |
| `upload-override-dialog.tsx:73` | `sm:max-w-3xl` | `IngestionSettings` |
| `sync-source-wizard.tsx:223` | `sm:max-w-2xl` (672px) | the wizard's four steps |

The same component is given 768px in two dialogs and 512px in the one that also
carries four fields above it. That is the cramping: not a judgement call, a
disagreement. Create should be at least `2xl`, and the four should agree on a scale
rather than each picking one - `max-w-lg` for a confirm, `2xl` for a form, `3xl`
for a form with a preview beside it, decided once.

## 3. The PDF parser choices have no mark

`PDF_PARSERS` in `src/lib/ingestion-config.ts:90` is three `Choice` rows, and
`Choice` carries `value`, `labelKey`, `hintKey` - no icon. So "Parse the next
upload differently" is three lines of text where every other picker in the product
(providers, connectors, MCP servers) draws a mark.

**Only one of the three has a brand**, and pretending otherwise is how a menu ends
up with one logo and two blanks:

- **LlamaParse** → LlamaIndex's mark, one row in `BRANDS`/`PROVIDERS` in
  `scripts/gen-brand-icons.ts` plus `bun run gen:brand-icons`. Never a
  hand-authored path and never an icon-package import - `frontend.md` is explicit,
  and #156 is what that rule cost.
- **PyMuPDF** and **liteparse** have no public mark. They take a lucide icon, the
  way `brandMark(name)` already lets a table mix the two.

So the `Choice` type grows an optional mark, and the three rows fill it - one
brand, two icons - rather than one row getting special treatment.

## Done when

- The prompt field is a `MarkdownEditor` in both dialogs, with its length refusal
  still shown under it.
- The four dialogs agree on a width scale, and the create dialog is not the
  narrowest thing holding the widest form.
- Each parser choice draws something, generated for the one that has a brand.
- A vitest spec asserts the editor renders its preview toggle in the create dialog
  - the regression that matters, since the field is easy to swap back.

## Related

- **#939** - the tabs on the same two pages. Different change, same screens;
  whichever lands first, the other rebases.
- **#568** (open) - unify the two backend-schema form renderers (`SchemaForm` vs
  `ConfigureStep`). `IngestionSettings` is a third hand-written form over the same
  kind of config, so if #568 is taken first this dialog may be one of its callers.
- **#156** (closed) - one generator for every brand mark, and the reason a new one
  is a row rather than an import.

DEENUU1:
Three complaints about the same dialogs, and they are one piece of work.

## The image-description prompt is a `MarkdownEditor`

It is a model prompt, several sentences long, and it was a bare
`<Textarea rows={3}>` — while the product already has the control for that: the
one the Builder uses for an agent's instructions, an exposure prompt, a page field
and a capability's generated form. `invalid` and `describedBy` were already wired
on that primitive precisely so a generated refusal can be shown under it, which
this field needs since `INGESTION_LIMITS.prompt.maxLength` is enforced and
reported.

The editor gained `maxLength` so swapping the control did not quietly drop the
hard cap — a field whose length the server refuses should not let somebody write
past it and find out on submit. `IngestionSettings` is embedded whole in the create
dialog, so one change fixes both.

## The four dialogs agree on a width scale, in one file

The same `IngestionSettings` was given 768px in two dialogs and **512px in the one
that also carries four fields above it**. That is the cramping — not a judgement
call but a disagreement: nobody had decided, each had picked.

`src/lib/dialog-widths.ts` holds the three sizes with the rule on each — a confirm,
a form, a form dense enough that 2xl crowds it — so the create dialog is no longer
the narrowest thing holding the widest form, and the fifth dialog has somewhere to
look.

## Each parser choice draws something

Three lines of text, where every other picker in the product (providers,
connectors, MCP servers) draws a mark. **Only one of the three is a product**, and
pretending otherwise is how a menu ends up with one logo and two blanks:

- **LlamaParse** takes LlamaIndex's own mark — a row in
  `scripts/gen-brand-icons.ts`, generated by `bun run gen:brand-icons`. Never a
  hand-authored path and never an icon-package import; #156 is what that rule cost.
- **PyMuPDF** and **liteparse** have no public mark and take a lucide icon, the way
  `brandMark(name)` already lets a table mix the two.

`Choice` grew an optional `Icon`, and the select item grew a `textValue` because
its child is no longer a bare string — without it a Radix item renders its own
markup into the trigger.

## Verified

Four new tests:

- the create dialog renders the editor's source/preview toggle, and
- still carries the length cap — the regression that matters, since the field is
  easy to swap back;
- the parser select draws an `svg` on every option, and LlamaParse's is the
  generated LlamaIndex glyph, read through `brandMarkIn` rather than by class name.

`make lint-frontend`, `make test-frontend-cov` (**5248 passed**, 100% / 97.67%) and
`make build-frontend` all clean. No backend path changed, so `test` is not expected
to run.

## Related

**#939** — the tabs on the same two pages. Different change, same screens; this one
landed first, so that one rebases.

Closes #940


DEENUU1:
@codex review

Issue #937, opened 2026-08-19
Sync source credentials are pasted into a JSONB column, encrypted by the one second mechanism the vault was meant to replace


DEENUU1:
Adding a Google Drive sync source asks you to paste a **Service Account JSON**
into a textarea. That credential then lives in `sync_sources.config`, a plain
JSONB column, encrypted by a module that exists solely because this one caller
was never converged onto the vault.

## What is actually there

`app/core/crypto.py` is honest about itself, and its docstring is the issue:

> **Not the vault, and not a second option.** [...] This module has no such
> property: one deployment-wide key encrypts everything, which is exactly the
> weakness the vault exists to remove.
>
> It survives for one caller - `app.services.sync_source` [...] and only because
> `sync_sources.organization_id` is nullable and the CLI creates rows without
> one. There is no owner to bind a ciphertext to, so converging it means making
> that column NOT NULL and giving the CLI an organization first.

So the blocker is named, and it is small. Meanwhile `CLAUDE.md` states the
boundary as absolute - "Every secret at rest goes through `app/core/vault.py`.
**There is no second mechanism**" - and a reader who believes that is wrong about
one table.

The consequences beyond the crypto:

- **The credential is per source, not per organization.** `google_drive.py`'s own
  docstring: "a source runs on the credential its own configuration carries or it
  does not run". Five knowledge bases fed from one Drive means the same JSON key
  pasted five times, rotated five times, and revoked in five places.
- **It is not in the Vault, so it is not in the Vault's story.** No "added by", no
  rotation, no `secrets:view` scoping, nothing on the Vault page that says this
  organization holds a Google credential at all.
- **The platform already has the right kind.** `SecretKind.GCP_SERVICE_ACCOUNT`
  exists, with a validator that refuses a document whose `type` is not
  `service_account` (`app/core/secret_kinds.py:162`) - written for Vertex AI, and
  exactly what a Drive connector wants.

## What it should be

A sync source references a vault secret by id, the way everything else does -
`ModelProfile.secret_id`, `CapabilityBindingSpec.secret_id` - and its `config`
holds no secret at all. The wizard's step becomes "pick a credential, or add one",
which is the `InlineSecret` control the model picker and the create-KB dialog
already use.

Order of work, because one step unblocks the rest:

1. `sync_sources.organization_id` NOT NULL, and the CLI given an organization
   (`rag-source-add` - see **#707**, which is about the same command writing an
   unvalidated, unowned collection name; the two want one branch).
[...]

PR #980, opened 2026-08-20
refactor(rag): type the sandbox and connector service contracts

### LINKED ISSUE #562: Type the dict[str, Any] service contracts in sandbox and connectors
`services/sandbox_connection.py:255,306,351` return service-constructed `dict[str, Any]` handed to routes; `sandbox_workspace.py:975`; `services/rag/connectors/__init__.py` (`CONFIG_SCHEMA: dict[str, dict[str, Any]]`, every `config: dict`), `connectors/s3.py:72` (`_get_s3_client(config: dict)`, no return type), `google_drive.py` same.

**Why it matters.** CLAUDE.md forbids `dict[str, Any]` where a model belongs. These are stable service-authored structures, so the service→route contract is untyped and `ty` cannot catch a renamed/dropped key. `validate_config` reads `field_spec.get("required")` from the untyped schema — a key typo (`"requird"`) silently disables validation for that field. (Honest exception: `sandbox_connection.py:495,521` proxy opaque `sandboxd` payloads — wrap or document, don't force a model.)

**Fix.** `*Read` schemas for the service-constructed returns; a `ConnectorFieldSpec` model for `CONFIG_SCHEMA`; add the missing return annotation.

**Done when.** The sandbox and connector contracts are typed; a schema-key typo is a type error.

---

DEENUU1:
Both halves already had the models; what was missing was the services
using them. Every route named a `response_model` and the service handed
back a `dict[str, Any]` for FastAPI to validate into it - so the
service→route contract was a mapping the type checker could not read, and
a renamed or dropped key was a runtime 500 rather than a red `ty` run.

`sandbox_connection.py` now returns the schema each route already
declares: `SandboxRuntimeOption`, `SandboxLocalServiceRead`,
`SandboxLocalCredentialRead`, `SandboxPolicyRead`, `SandboxSessionList`,
`SandboxEventList` and a new `SandboxSessionUsage`. Three consequences
worth naming:

- `usage_report.py` reads `sampled.memory_bytes` rather than
  `sampled.get("memory_bytes")`. Those two keys are the whole of what a
  usage footer shows for a container, and they were unchecked in the one
  place a rename shows up as a missing number instead of an error.
- `_attributed` builds the rows, so the daemon's `tenant` label is
  dropped where the filter reads it rather than by every caller
  remembering to. It is another organization's id when the session is
  theirs; `SandboxSessionList`'s docstring has always said it is absent,
  and now one place makes that true.
- `_read` keeps `dict[str, Any]` and is the only one left in the module,
  with a docstring saying why: it is `sandboxd`'s answer, and a model here
  would be a second declaration of somebody else's API.

`CONFIG_SCHEMA` is `dict[str, ConnectorConfigField]` - the model that
described it at the API edge is now its own type, so a misspelled key is
a type error where it is written. It was `dict[str, dict[str, Any]]` and
`validate_config` read `field_spec.get("required")` off it, so a
declaration that said `require` disabled that field's check in silence
and the wizard drew a required field as optional. `type` is a
`Literal` of the four widgets `SyncSourceConfigureStep` actually draws,
mirrored in `rag-api.ts`: its fall-through is a text input, so a
connector inventing a fifth got a field the form collects wrongly with
nothing reporting it.

Also `_get_s3_client` gained the return type it never had (`BaseClient`),
both `_list` closures theirs, and the four `sandbox_workspace.py` helpers
that take the state backend's stored document now say `FileData` - the
library's own type, which `StateBackend.__init__` has always been
annotated with.

## Docs

`docs/howto/add-sync-connector.md` was stale in a way this makes worse:
its worked example put a Notion `api_token` in `CONFIG_SCHEMA`, and
`list_files`/`_fetch` had no `credential` parameter at all - it described
the code as it was before #937, down to a tip reading "per-source
credentials go in `CONFIG_SCHEMA` and are stored in the database per sync
source". The typing change makes that example fail to type-check, and
fixing only the shape would have left a page teaching a credential into a
JSONB column, so the example, the CLI and API calls, the field-type table
and the tips are all corrected together. Pre-existing, found here.
`docs/patterns.md`'s snippet and two wrong `app/rag/connectors/` paths in
`docs/howto/configure-sync-sources.md` and `docs/architecture.md` went
the same way.

## Verified

`make lint`, `make docs-build`, `make test` - 6182 passed, the platform
layer at 100% including the new branches. No behaviour change intended:
validation of a daemon payload moves from the response model into the
service, which is the same requirement one step earlier.

One test fixture got more honest rather than less. `TestReadingTheSessions`
built sessions holding two keys, which the response model would already
have refused - `runtime`, `created_at`, `last_activity` and `idle_seconds`
are required and were never in the fixtures. There is now a `_session`
helper answering what `sandboxd` answers, and the assertions that used to
subscript a mapping read attributes.

Closes #562


DEENUU1:
Fixed the label and the CLI example in 94e584d.

`label` is required now — it defaulted to `""` and only `validate_config` fell back to the key, so a connector omitting it got an unlabelled input while the refusal still read sensibly. The base class's fallback is gone rather than left as a branch nothing reaches, and the field order in both declarations follows what the form does with them.

`rag-source-add` does take `--name` and `--type`; the example is corrected.

On the third finding: `docs/howto/add-rag-source.md` is **deleted**, not fixed. It was the same guide as `add-sync-connector.md` a second time, adjacent in the nav, stale in the same pre-#937 way (`api_key` inside `CONFIG_SCHEMA`, no `credential` parameter, a tip saying per-source credentials live in the database), and it told the reader to edit the generator's `post_gen_project.py`. Correcting it would have left two copies of one guide.

The normalisation shim for out-of-tree connectors is declined: there are no deployments carrying a custom connector, so it would be a branch for a caller that does not exist. The half of that finding which was true is the page above.

@codex review

DEENUU1:
Both fixed in 08303986.

`pids` is declared now. That one is a real regression this branch introduced: a model drops what it does not declare, so the field stopped reaching clients the moment the service built the model instead of passing the daemon's mapping on — and it is in the contract `sandbox-connections-api.ts` publishes even though nothing renders it yet. The test asserts every sampled field rather than the two `usage_report.py` uses, which is what would have caught it.

The tip was a contradiction this branch introduced on one page: the rewritten example's `validate_config` checks shape and explains that it cannot check connectivity, while the tips still said to test connectivity there — advice that requires the credential in `CONFIG_SCHEMA` or a deployment-wide fallback, both prohibited two sections earlier.

@codex review

Issue #988, opened 2026-08-20
Generalize S3Connector to an object store, before Azure Blob and GCS


DEENUU1:
`S3Connector` is an S3 connector. Azure Blob Storage and Google Cloud Storage are
the same connector with a different client — a bucket or container, a prefix, a
credential, a flat listing of objects with a size and a modified time. #938 put
them fifth on the list with a condition attached: *"if `S3Connector` is
refactored to an object-store shape, these are configuration rather than code."*

This is that refactor, and it is worth doing before writing either one rather
than after, because the alternative is three connectors sharing a shape by
resemblance.

## What it is

An intermediate class — `ObjectStoreConnector` — holding what all three do
identically: list objects under a prefix, skip keys ending in `/`, build a
`RemoteFile` per object with a `source_path` of `<scheme>://<container>/<key>`,
download one object to the path the base class chose. What a subclass supplies is
a client and the vocabulary: how its SDK paginates a listing, and which
`SecretKind` authenticates it.

`S3Connector` becomes that subclass and keeps its `CONNECTOR_TYPE`, its
`CONFIG_SCHEMA` and its behaviour exactly — this is a refactor, and a stored
source must sync the same way afterwards.

Azure needs a new `SecretKind` (a storage account plus a key, or a SAS token);
GCS reuses `SecretKind.GCP_SERVICE_ACCOUNT`, which the Drive connector already
has, so GCS is the cheaper of the two.

## The change signal

All three offer the same one: an object's `ETag` or its modified time. Whatever
the shared class does with it, it does once — which is half the point of having
it.

## Done when

`S3Connector` is a subclass with no behaviour change and its tests unchanged; the
shared class is what a new object store subclasses; and adding GCS is a client
and a `CONNECTOR_TYPE` rather than a second copy of the listing loop.

Part of #938. Do this **before** the Azure and GCS connectors, not alongside them.

Issue #987, opened 2026-08-20
Sync connector: import and synchronize Git documentation


DEENUU1:
Deliver the planned W8 source integration using the existing collection synchronization pipeline.

## Delivery scope

- Owner: DEENUU1.
- Milestone: W8 · Sep 21–25.
- Planning window: 2026-09-21 to 2026-09-25.
- Estimate: One week shared across WWW, Git, SharePoint and OneDrive; not one week per connector.

## Scope

- Keep the existing issue’s connector-specific design and safeguards; this section adds the agreed delivery scope.
- Expose source configuration, credential selection where required, manual/scheduled sync and visible status/errors in the current UI.
- Handle incremental changes, deletions and unchanged-content detection without duplicate ingestion or unnecessary embeddings.
- Keep credentials in the existing vault and enforce source/collection permissions; document the credential scope and who can read imported data.

## Acceptance criteria

- [ ] A configured source populates a collection and is searchable through existing retrieval.
- [ ] A second sync skips unchanged content, updates changed content and removes deleted content.
- [ ] Access failures, retries and partial sync failures are visible and tested; setup and troubleshooting docs are included.

## Design references

Source files in the planning bundle: `pasted-text.md — Kacper W8`.
The planning documents are in Polish; this issue defines the implementation scope in English. HTML demos are illustrative, use simulated data and do not prescribe the production styling.
Related workflow/table design and demo bundle: #56.

<details>
<summary>Previous issue context (historical; the delivery scope above takes precedence)</summary>

Markdown docs, ADRs and READMEs, with the commit as a change signal — the
cleanest incremental sync of anything on #938's list, and the one where "what
changed" is a solved problem rather than an approximation.

## What it is

`CONNECTOR_TYPE = "git"`, covering GitHub and GitLab, because what it needs is a
clone URL and a token rather than either platform's API. `SecretKind.API_KEY`
holds a personal access token; the clone runs over HTTPS with the token, into a
temporary directory the worker already makes for a sync.

`CONFIG_SCHEMA`: the repository URL, a branch, a path prefix within the repo, and
a glob of what to ingest — defaulting to documentation rather than to everything,
because a repository's source tree is not a corpus and ingesting it is how a
knowledge base fills with code nobody asked to search.

A file is one `RemoteFile` with `source_path = "git://<repo>/<path>"`.

## The change signal, which is why this one is worth building

The commit sha. A stored sha and `git diff --name-only <old>..<new>` is an exact
[...]

Issue #986, opened 2026-08-20
Sync connector: Confluence spaces


DEENUU1:
The wiki is where process documentation lives, and it is the corpus people most
often mean by "our knowledge". Third on #938's list.

## What it is

`CONNECTOR_TYPE = "confluence"`. Auth is an Atlassian API token plus the account
email and the site URL — which is `SecretKind.API_KEY` plus two config fields, or
its own kind if the review prefers the email bound to the token. Cloud first;
Server/Data Center is a different auth story and a separate issue if anybody asks.

`CONFIG_SCHEMA`: the site URL, the space key, optionally a CQL filter or an
ancestor page id to read a subtree rather than a whole space.

A page is one `RemoteFile` with `source_path = "confluence://<page-id>"`; storage
format converts to HTML, which the ingestion pipeline already parses. Atlassian
is already in `brand-glyphs.generated.ts`, so the mark is a row rather than
artwork.

## The change signal

`version.number` per page, and `lastModified` beside it — a page whose version
matches the stored one is skipped. That is a real signal, so `new_only` behaves
rather than re-embedding a wiki nightly
(`docs/file-processing.md#what-a-new-connector-owes`).

Attachments are a decision to make explicitly: a space's PDFs are often the
actual documents, and ignoring them silently is the wrong default while
ingesting them multiplies the file count. Whichever way it goes, it is a
`CONFIG_SCHEMA` field and not a surprise.

## Permissions

A Confluence token carries its user's own access, so a token belonging to
somebody senior reads more than the space it was pointed at — and everything it
reads becomes searchable by whoever can read the collection
(`docs/file-processing.md#who-ends-up-able-to-read-what-a-source-ingested`). The
mitigation is the same as everywhere on this list and belongs in the setup steps:
issue the token from an account with access to the one space.

## Done when

A space, or a subtree of one, can feed a collection; a page unchanged since the
last run is skipped on its version number; and whether attachments are ingested
is a field somebody chose rather than a default nobody saw.

Part of #938.

Issue #985, opened 2026-08-20
Sync connector: synchronize SharePoint and OneDrive


DEENUU1:
Deliver the planned W8 source integration using the existing collection synchronization pipeline.

## Delivery scope

- Owner: DEENUU1.
- Milestone: W8 · Sep 21–25.
- Planning window: 2026-09-21 to 2026-09-25.
- Estimate: One week shared across WWW, Git, SharePoint and OneDrive; not one week per connector.

## Scope

- Keep the existing issue’s connector-specific design and safeguards; this section adds the agreed delivery scope.
- Expose source configuration, credential selection where required, manual/scheduled sync and visible status/errors in the current UI.
- Handle incremental changes, deletions and unchanged-content detection without duplicate ingestion or unnecessary embeddings.
- Keep credentials in the existing vault and enforce source/collection permissions; document the credential scope and who can read imported data.

## Acceptance criteria

- [ ] A configured source populates a collection and is searchable through existing retrieval.
- [ ] A second sync skips unchanged content, updates changed content and removes deleted content.
- [ ] Access failures, retries and partial sync failures are visible and tested; setup and troubleshooting docs are included.

## Design references

Source files in the planning bundle: `pasted-text.md — Kacper W8`.
The planning documents are in Polish; this issue defines the implementation scope in English. HTML demos are illustrative, use simulated data and do not prescribe the production styling.
Related workflow/table design and demo bundle: #56.

<details>
<summary>Previous issue context (historical; the delivery scope above takes precedence)</summary>

The single largest gap on #938's list. A self-hosted platform sold to companies
meets more documents in SharePoint and OneDrive than in Google Drive, and today
there is no path at all — a Microsoft 365 company feeds a knowledge base by
uploading files by hand.

## What it is

`CONNECTOR_TYPE = "sharepoint"`, covering OneDrive as well: both are drives in
Microsoft Graph, so one connector with a drive id rather than two.

Auth is an Entra app registration — client id, tenant id, client secret. That is
a **new `SecretKind`**, and it is the first one this repository has added since
the vault work: `SecretKind.ENTRA_APP` (or a name the review prefers) with those
three fields, added to `app/core/secret_kinds.py` and `_KIND_MODELS`, so the
wizard offers it and the credential is one vault entry rather than a fourth
pasted blob. `docs/secrets.md` gets the row.

`CONFIG_SCHEMA`: the site or drive to read, a folder path within it, whether to
include subfolders. Never the credential (#937).

[...]

PR #989, opened 2026-08-20
docs(rag): decide which sync connectors follow Drive and S3

### LINKED ISSUE #938: Which sync connectors after Google Drive and S3 - decide the list and the order
`CONNECTOR_REGISTRY` holds two entries - `gdrive` and `s3`
(`app/services/rag/connectors/__init__.py:110`). A company whose documents live in
SharePoint, Confluence or a docs site has no way to feed a knowledge base except
by uploading files by hand.

Decide which connectors are worth having, in what order.

## First, the boundary that decides several of these

**A sync connector is not an MCP server.** MCP is how an agent reaches a SaaS
product *live*, mid-run, and `.claude/skills/mcp-connections` says to reach for it
before writing "a capability that would just be an API client". A sync source is
the other thing: a bulk pull with change detection, run on a schedule, whose
output is chunks in pgvector. Notion-as-a-tool is an MCP server; Notion-as-a-corpus
is a connector. Some candidates below are genuinely both, and the issue should say
which half each one is being built for.

## Candidates, ranked by what a company actually has

1. **SharePoint / OneDrive (Microsoft 365).** The single largest gap. A
   self-hosted platform sold to companies meets more documents here than in Drive,
   and there is no path at all today. Auth is an Entra app registration - a client
   id, tenant id and secret - which the vault work beside this makes clean rather
   than a fourth pasted blob.
2. **Confluence.** The wiki is where the process documentation is, and it is the
   corpus people most often mean by "our knowledge". Atlassian is already in the
   brand-mark table.
3. **A web crawler** - a sitemap or a root URL, depth-limited. The cheapest of the
   lot and the only one needing **no credential**, which makes it the natural
   second connector rather than the fifth: a product's own public docs site is how
   most agents get product knowledge. It needs the SSRF guard that
   `validate_webhook_url` and the MCP work already established (**#840**, **#860**)
   - a crawler is the most obviously remote-chosen URL in the product.
4. **A git repository** (GitHub / GitLab). Markdown docs, ADRs, READMEs, with the
   commit as a natural change signal - the cleanest incremental sync of anything
   on this list.
5. **Azure Blob and Google Cloud Storage.** Siblings of the S3 connector rather
   than new ideas; if `S3Connector` is refactored to an object-store shape, these
   are configuration rather than code.
6. **Notion.** Smaller organizations, and the API is pleasant. Below the four above
   because a Notion-first company is usually also small enough to upload files.

Deliberately not on the list: Slack and email archives. Both are conversations
rather than documents - retrieval over them answers badly, and the channel
integrations already put an agent *in* Slack, which is the question people
actually ask.

## What each one costs, so the order is honest

A connector is `list_files` + `download_file` + a `CONFIG_SCHEMA`, plus a wizard
step, plus the brand mark, plus the ingestion path already handling whatever the
file turns out to be. The expensive parts are not the API calls:

- **Incremental sync.** `sync_mode` is `new_only` by default; a connector with no
  change signal re-ingests, and re-ingesting a wiki nightly is a bill.
- **Permissions.** A Drive folder shared with a service account is a decision
  somebody made once. A SharePoint site or a Confluence space is usually not, and
  a connector that ingests everything the credential can see puts documents into a
  collection whose readers are a different set of people. That question wants
  answering **once, for all connectors**, not per connector.
- **These are template-inherited**, so they sit outside the 100% coverage gate.
  That is a reason to be deliberate about how many exist, not a licence.

## Done when

This issue is a decision, not an implementation: it closes when the list above is
cut to what is being built, in order, each with its own issue - and when the
permissions question in the paragraph above has an answer written down in
`docs/file-processing.md`.

## Related

- **#937** - sync source credentials belong in the vault. **This should follow it.**
  Five more connectors each pasting a credential into `sync_sources.config` is five
  times a problem that is currently one table with two entries.
- **#840** / **#860** (SSRF and remote-chosen URLs) are what a web crawler inherits.
- **#27** (open) - paginate `GET /rag/documents` and stop full-scanning on ingest.
  Any connector that brings thousands of files makes that one urgent rather than
  tidy.

DEENUU1:
#938 is a decision, not an implementation: it closes when the candidate
list is cut to what is being built, in order, each with its own issue,
and when the cross-connector permissions question has an answer written
down here.

**The permissions answer.** The collection is the boundary and a source's
credential is its reach: a source ingests into exactly one collection,
access is decided at the collection, there is no per-document isolation
inside one, so everything that credential can see becomes readable by
everyone who can read that collection. The platform makes that explicit
rather than clever - mirroring each source's ACLs and filtering at
retrieval is decided against, and the section says why so it is not
proposed again as an obvious win: there is no identity map between an
Entra or Atlassian principal and an `organization_members` row, a
permission changed in the source is invisible until the next sync so a
mirrored ACL is stale authorization, and a crawler has no ACL at all. The
rule that follows is one sentence - scope the credential, not the
collection - and it is the only knob that narrows what gets ingested.

Two things that rule owes and does not have, both filed rather than
described as though they existed: #982 states the consequence in the
wizard where the collection is chosen and re-asks when a source is
repointed, and #983 records creating, repointing and deleting a source in
the audit log. Today both are silent, which is the implicitness the
section names.

**The list, cut and ordered.** A web crawler first (#984) - the cheapest,
the only one needing no credential, and how most agents get product
knowledge; then SharePoint and OneDrive (#985), the largest gap, which
needs a new `SecretKind` for an Entra app; Confluence (#986); a git
repository's documentation (#987), where the commit sha makes incremental
sync exact rather than approximate; and Azure Blob and GCS only after
`S3Connector` is an object store rather than an S3 one (#988), which is
the condition #938 attached to them. Notion is decided against for now -
MCP covers Notion-as-a-tool - and Slack and email archives stay off,
because a conversation retrieves badly and the channel integrations
already put an agent in Slack.

**What a new connector owes** is written down alongside, because the API
calls are the cheap part: a change signal named in the docstring, a
credential scoped where the source is created, and a file count somebody
has thought about while reading a collection's listing is still a full
scan (#27). Each connector issue names its own signal - a Graph delta
token, a page version, a commit sha, an HTTP ETag.

Also fixed: `create_source`'s docstring still opened "Secret fields are
Fernet-encrypted before persisting", which has been untrue since #937
deleted `app/core/crypto.py`. It now says what the collection it names
decides, and points at the section above.

Verified with `make lint`, `make docs-build` and `scripts/docs_drift.py`.
No behaviour changed; this is a decision and a docstring.

Closes #938

### DIFF
--- a/backend/app/services/sync_source.py
+++ b/backend/app/services/sync_source.py
@@ -196,9 +196,11 @@ class SyncSourceService:
     async def create_source(self, data: SyncSourceCreate, *, ctx: AuthContext) -> SyncSourceRead:
         """Create a new sync source.
 
-        Secret fields are Fernet-encrypted before persisting.
         `collection_name` is optional - omit to create an org-level integration
-        not yet linked to a knowledge base.
+        not yet linked to a knowledge base. Whatever it names decides who can
+        read what this source ingests, because access is decided at the
+        collection and the credential's reach is the source's reach
+        (`docs/file-processing.md#who-ends-up-able-to-read-what-a-source-ingested`).
 
         The name's *shape* is judged here rather than only at the routes,
         because a sync writes into whatever collection this row names and the
--- a/docs/file-processing.md
+++ b/docs/file-processing.md
@@ -667,6 +667,15 @@ matches. A store that cannot answer the listing is treated as "no match" rather
 than as a match: a failed query is not evidence that a document is absent, but
 acting on it as though a document *were* present would delete one.
 
+**That is the local-directory sync. A connector sync implements none of it**
+([#990](https://github.com/vstorm-co/agenticos/issues/990)): `sync_source_flow`
+downloads and ingests every file the connector lists, `sync_mode` reaches only
+`ingest_file`'s `replace` argument, and `ingest_file` never skips — so on the
+default `new_only` the previous document is neither found nor deleted and a
+duplicate is inserted on every run. The `skipped` counter beside it is
+initialised and never incremented, which is what a sync log truthfully reporting
+`skipped=0` every night has been saying all along.
+
 One source's own history is `GET /kb/{kb_id}/sync-sources/{source_id}/logs`. The
 source is resolved against that knowledge base first, so a source belonging to
 another base answers **404** rather than an empty list — the two render the same
@@ -745,13 +754,23 @@ must not grow one.
 
 ### Who ends up able to read what a source ingested
 
-**The collection is the permission boundary, and a source's credential is its
-reach.** A sync source ingests into exactly one collection, access is decided at
-the collection (see [Who may reach a collection](#who-may-reach-a-collection)),
-and there is no per-document isolation inside one — so **everything that
-credential can see becomes readable by everyone who can read that collection.**
-A Confluence token good for the whole instance, pointed at an `org` collection,
-publishes the whole instance to every member holding `collections:view`.
+**The collection is the permission boundary, and a source's reach is its
+credential's permissions narrowed by its own configuration.** A sync source
+ingests into exactly one collection, access is decided at the collection (see
+[Who may reach a collection](#who-may-reach-a-collection)), and there is no
+per-document isolation inside one — so **everything that source reads becomes
+readable by everyone who can read that collection.**
+
+The two halves of that reach are not equally reliable, which is the part worth
+knowing. A Drive source is bounded by its `folder_id` and an S3 source by its
+`bucket` and `prefix`, so a broad credential pointed at one folder ingests one
+folder. But `config` is a field on the row, editable by anyone holding
+`collections:edit` on that collection — so **configuration narrows the reach and
+cannot be relied on to keep it narrow**, while the credential's own permissions
+are a ceiling nothing in this product can raise. A Confluence token good for the
+whole instance, on a source somebody later repoints at a wider space, publishes
+the whole instance to every member holding `collections:view`; the same token
+scoped to one space cannot, whatever the config says.
 
 That is a decision somebody has to make, and the platform's answer is to make it
 **explicit rather than clever**. The alternative — mirroring each source's own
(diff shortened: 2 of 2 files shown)


DEENUU1:
Both fixed in 02f7d54b, and the second one was worth more than it looked.

**Reach is the intersection.** Right — a Drive source is bounded by `folder_id`, an S3 source by `bucket`/`prefix`, so "the credential's reach is the source's reach" overstated it. The section now names both halves and why they differ: `config` is a field on the row that anyone with `collections:edit` can widen, so configuration narrows the reach but cannot be relied on to keep it narrow, while the credential's permissions are a ceiling nothing here can raise. The rule became "scope the credential, not just the config".

**The change signal.** Also right, and chasing it down found a defect: `sync_source_flow` lists, downloads and ingests every file unconditionally. `sync_mode` reaches exactly one argument — `ingest_file`'s `replace` — and `ingest_file` never skips. So on the default `new_only` the previous document is neither found nor deleted and a **second copy is inserted every run**. A nightly sync of 500 files inserts 500 documents a night, each fully re-embedded, until a search returns seven copies of the same passage. The `skipped` counter is initialised and never incremented, which is a sync log honestly reporting `skipped=0` every night.

Filed as #990 (severity:high, W4). `sync_local_flow` in the same module implements the modes properly, so there is a comparison step to copy rather than design.

That also made a third correction necessary on the same page: the `sync_mode` paragraph under Sync Operations described the local flow's hash comparison as though it were the rule for both flows. It now says which flow that is and cites #990.

@codex review

DEENUU1:
Merging. CI is green end to end and Codex made a full pass on this branch earlier — two findings, both fixed in 02f7d54, and the second one produced #990 (severity:high), which is the most useful thing this PR found. The re-review has answered "usage limits" on both connected instances for the last 40 minutes.

Reviewing my own diff turned up one more: the change-signal bullet said #990 comes before the connectors are worth having, while the paragraph naming the order still opened on #984. Fixed in a6267327.

Docs-only apart from one docstring.

Issue #992, opened 2026-08-20
Documents ingested by a connector sync are invisible to every per-document surface


DEENUU1:
Sync a Google Drive folder into a knowledge base. The sync log says
`ingested: 40`. The knowledge base's **Documents** tab stays empty.

The documents are real — they are in the vector store and an agent retrieves
them — but `GET /kb/{kb_id}/documents` reads `rag_documents`
(`RAGDocumentService.list_for_kb` → `rag_document_repo.get_for_kb`), and the
connector sync never creates a row there.

## Where it is

`_run_source_sync` in `backend/app/worker/tasks/rag_tasks.py:628-645` calls
`ingest_file` and moves to the next file. Compare `_run_sync`, the
local-directory flow in the same module at lines 425-447, which does the
bookkeeping:

```python
result = await ingestion_service.ingest_file(...)
if result.status.value == "done":
    doc = await RAGDocumentService(db).create_document(...)
    await RAGDocumentService(db).complete_ingestion(
        str(doc.id),
        vector_document_id=result.document_id,
        chunk_count=result.chunk_count,
        replaced_document_id=result.replaced_document_id,
    )
```

`grep -n "create_document" app/worker/tasks/rag_tasks.py` returns one hit, in
the local flow.

## What it costs

**Every per-document surface is blind to a synced document.** Not only the
listing:

- The Documents tab shows nothing, which is the "an empty page is ambiguous"
  failure exactly — "no documents yet" and "40 documents you cannot see" render
  the same pixels.
- `GET /rag/documents/{doc_id}/download` and `/parsed` have no row to serve, so
  a synced document cannot be opened or re-read.
- `DELETE /rag/documents/{doc_id}` cannot reach it. A document ingested from a
  Drive folder can be removed from the collection only by dropping the whole
  collection.
- `POST /rag/documents/{doc_id}/retry` cannot reach a failed one either — and a
  connector sync's failures are counted in the sync log and recorded nowhere
  per-document, so "which four of the forty failed, and why" has no answer.
- The collection stats a knowledge base shows (`document_count`,
  `indexed_count`) are computed from `rag_documents`, so a collection fed only by
  sync reads as empty while search over it works.

## Reproduce

1. Add a Drive or S3 integration to a knowledge base and sync it.
2. The sync log reports files ingested; `POST /rag/search` finds their content.
3. The Documents tab is empty, and the collection's stats read zero.

## Fix

The connector flow needs the same per-file bookkeeping the local flow has: a
[...]

PR #993, opened 2026-08-20
fix(rag): implement sync_mode in the connector sync

### LINKED ISSUE #990: A connector sync ignores sync_mode and duplicates every document on every run
`sync_mode` is implemented for the local-directory sync and **not at all** for
the connector sync — so a scheduled Google Drive or S3 source re-ingests every
file on every run, and because `new_only` does not look up the existing document,
each run *adds a second copy* rather than replacing it.

`new_only` is the default (`app/db/models/sync_source.py:62`,
`app/schemas/sync_source.py:82`), so this is what an unconfigured scheduled source
does.

## Where it is

`sync_source_flow` in `backend/app/worker/tasks/rag_tasks.py:628-645` lists,
downloads and ingests unconditionally:

```python
files = await connector.list_files(config, credential)
for remote_file in files:
    local_path = await connector.download_file(...)
    await ingestion_svc.ingest_file(
        filepath=local_path,
        collection_name=collection_name,
        replace=(sync_mode == "full"),
        source_path=remote_file.source_path,
    )
    ingested += 1
```

`sync_mode` reaches exactly one place — `replace=` — and `ingest_file`
(`app/services/rag/ingestion.py:109`) never skips anything: it parses, embeds and
inserts, and `replace` only decides whether the previously matching document is
deleted first. So:

| `sync_mode` | What happens on a connector source |
|---|---|
| `full` | Correct — the matching document is deleted, the new one inserted |
| `new_only` (**default**) | `replace=False`, so no lookup, so the old document stays **and** a new one is inserted — one duplicate per file per run |
| `update_only` | Identical to `new_only`. The mode's whole meaning ("skip a file that is new") is absent |

`skipped` is initialised at line 624 and never incremented, which is the tell:
the counter exists because the flow was meant to skip, and a sync log therefore
reports `skipped=0, ingested=<every file>` every night, truthfully.

Contrast `sync_local_flow` at line 407-423 in the same file, which implements the
modes fully — `existing_document`, a sha256 of the bytes, a hash comparison and
the skip counters. That is where the behaviour the documentation describes
actually lives.

## What it costs

A nightly sync of a 500-file Drive folder inserts 500 new documents a night,
every one of them fully re-embedded. Two consequences, and the money is the less
serious:

- **A provider bill for unchanged content**, recurring, on the organization's own
  embedding key.
- **Retrieval degrades.** After a week there are seven copies of every chunk, so
  a search returns the same passage seven times and the top-k a user sees is one
  document's worth of variety. Deleting the duplicates is not something the
  product offers.

## Reproduce

1. Create a Drive or S3 sync source on a collection, leaving `sync_mode` at its
   default, with a schedule.
2. Let it run twice (or hit `POST /rag/sync/sources/{id}/sync` twice).
3. `GET /rag/collections/{name}/documents` — every file appears twice, with
   different document ids and the same `source_path`.
4. The sync log for the second run says `skipped: 0`.

## Fix

The connector flow needs the comparison the local flow already has, before the
download rather than after it: `existing_document(collection_name,
remote_file.source_path)` answers the id and the stored `content_hash` in one
scan (#548), and `RemoteFile` already carries `modified_at` — which is a cheaper
first filter than downloading a file to hash it, and the only one available before
the bytes arrive.

Worth deciding in the fix: `new_only` on a *changed* file. The local flow
re-ingests it with `replace=True` semantics; the name says otherwise. Whatever it
does, the two flows should agree, since one column feeds both.

`docs/file-processing.md#sync-operations` currently states the local flow's
behaviour as though it were the rule for both, and is corrected in #989 to say
what happens today and cite this issue.

## Done when

A connector source run twice over unchanged files ingests them once, the second
run reports them as skipped, and `new_only` / `update_only` mean the same thing
on a connector source as on a local directory. With a regression test that fails
without the fix — two runs, one document.

Found reviewing #989 (the connector roadmap), whose "a connector owes a change
signal" rule cannot deliver anything until this exists. Pre-existing; not caused
by that branch. Blocks the usefulness of #984, #985, #986 and #987, each of which
names a change signal.

DEENUU1:
A scheduled Google Drive or S3 source inserted a second copy of every
document on every run, and re-embedded all of them to do it.

`sync_mode` reached exactly one argument in `_run_source_sync` -
`ingest_file`'s `replace` - and `ingest_file` never skips anything: it
parses, embeds and inserts, and `replace` only decides whether the
document it matched is deleted first. So `replace=(sync_mode == "full")`
meant the two commonest modes told it not to look. On the default
`new_only` the previous document was neither found nor deleted and a new
one was inserted beside it, so a week of nightly syncs was seven copies of
every chunk, ranked against each other in every search, each paid for in
embeddings on the organization's own key. `skipped` was initialised beside
the loop and never incremented, which is a sync log truthfully reporting
`skipped=0` every night.

`sync_local_flow` in the same module had implemented the modes properly
the whole time, so this is that logic rather than a new design - one
`sync_mode` column feeds both flows and a mode meaning one thing for a
directory and another for a Drive folder is the defect whatever either
does alone. `replace=True` unconditionally, as there: once the flow has
decided to ingest, whatever it matched has to go.

Where the decision is taken differs, because remote bytes cost something.
`update_only` needs none to skip a file it has never seen, so that answer
is given before the download - otherwise it is a transfer per new file in
the folder, every run. A hash needs them, so an unchanged file is
recognised after the download and before the embedding, which is the
expensive half. A stored document carrying no `content_hash` is
re-ingested rather than assumed current: skipping a file that may have
changed is the answer nothing later corrects, and re-embedding costs once.

## Verified

`tests/test_connector_sync_modes.py`, eleven tests, six of which fail
without this change - including both duplication cases and the
before-download skip. They leave `existing_document` real and control the
listing it reads, because that lookup and its precedence (#548) are the
thing being relied on. `make lint` and the backend suite: 6193 passed.

Not fixed here, and filed: #992, that a connector sync creates no
`rag_documents` row at all, so its documents are invisible to the
Documents tab, to download, delete and retry, and to a collection's own
stats. Same root cause - the connector flow does not do the per-file work
the local flow does - and a different symptom. It has an ordering
consequence worth knowing: with this change a second sync ingests nothing,
so #992 will need a backfill for documents already synced rather than only
a new call site.

`docs/file-processing.md` is updated: the `sync_mode` paragraph said the
connector flow implemented none of this, and "what a new connector owes"
now asks for a signal readable *before* the download, since a
`content_hash` comparison saves the embedding and not the transfer.

Closes #990

### DIFF
--- a/backend/app/services/rag/ingestion.py
+++ b/backend/app/services/rag/ingestion.py
@@ -36,6 +36,27 @@ def _stored(doc: DocumentInfo) -> StoredDocument:
     )
 
 
+def _unaddressed(doc: DocumentInfo) -> bool:
+    """Whether this document may be matched by its *name* alone.
+
+    Only one that does not already claim an address of its own. A file uploaded
+    through the browser stores its filename as its `source_path`, so the two
+    agree and it stays reachable by name - which is what the fallback is for: a
+    document uploaded once and later synced from the folder it came from should
+    be replaced rather than duplicated.
+
+    A document that names a *different* address is a different document, and
+    matching it by basename loses one of them. An S3 bucket holding
+    `a/readme.md` and `b/readme.md` is the case: the second key found the
+    first's document by name, so equal contents skipped it and unequal contents
+    replaced the first - either way a first sync could not keep both, silently
+    (#990). The same collision existed for two local files of the same name in
+    different directories.
+    """
+    stored_path = str((doc.additional_info or {}).get("source_path") or "")
+    return not stored_path or stored_path == doc.filename
+
+
 class IngestionService:
     """File → Parse/Chunk → Deduplicate → Embed/Store → Query-Ready."""
 
@@ -97,9 +118,10 @@ class IngestionService:
         by_hash: DocumentInfo | None = None
         for doc in docs:
             meta = doc.additional_info or {}
-            if source_path and meta.get("source_path") == source_path:
+            stored_path = str(meta.get("source_path") or "")
+            if source_path and stored_path == source_path:
                 return _stored(doc)
-            if by_filename is None and filename and doc.filename == filename:
+            if by_filename is None and filename and doc.filename == filename and _unaddressed(doc):
                 by_filename = doc
             if by_hash is None and content_hash and meta.get("content_hash") == content_hash:
                 by_hash = doc
@@ -142,14 +164,21 @@ class IngestionService:
                     )
                 ).document_id
 
-            if existing_id:
-                await self.store.delete_document(collection_name, existing_id)
-                logger.info("Replaced existing document %s for '%s'", existing_id, filepath.name)
-
+            # Inserted before the old one is removed, not after. `insert_document`
+            # is where the embeddings are computed, so a provider that refuses
+            # between the two statements used to leave the collection with
+            # *neither* document - permanently, since the failure is returned
+            # rather than raised and nothing retries it. In this order the worst
+            # case is both for as long as the insert takes, which a search
+            # survives and a reader can see (#990).
             await self.store.insert_document(
                 collection_name=collection_name,
                 document=document,
             )
+
+            if existing_id:
+                await self.store.delete_document(collection_name, existing_id)
+                logger.info("Replaced existing document %s for '%s'", existing_id, filepath.name)
 
             action = "replaced" if existing_id else "ingested"
             chunk_count = len(document.chunked_pages or [])
--- a/backend/app/services/rag/ingestion.py
+++ b/backend/app/services/rag/ingestion.py
@@ -118,8 +118,7 @@ class IngestionService:
         by_hash: DocumentInfo | None = None
         for doc in docs:
             meta = doc.additional_info or {}
-            stored_path = str(meta.get("source_path") or "")
-            if source_path and stored_path == source_path:
+            if source_path and meta.get("source_path") == source_path:
                 return _stored(doc)
             if by_filename is None and filename and doc.filename == filename and _unaddressed(doc):
                 by_filename = doc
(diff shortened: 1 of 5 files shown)


DEENUU1:
All three fixed in f4a257a. The two P1s were the interesting ones, and one was a regression this branch would have shipped.

**The basename collision** — right, and it is mine: `full` already had it via `ingest_file`'s own lookup, but `new_only` and `update_only` never reached `existing_document` before this branch. With the modes implemented, equal contents skipped the second key and unequal contents deleted the first, so a first sync could not keep both.

Narrowed rather than removed, because the fallback has a job: an upload stores its filename *as* its `source_path`, so the two agree and it stays reachable by name — that is how a document uploaded once and later synced from the folder it came from gets replaced instead of duplicated. A document naming a *different* address is a different document. One condition, in the one place both call sites read, which also fixes the same collision for two local files of one name in different directories.

**Insert before delete** — right. Pre-existing for `full`; this branch put it on the default path, which is why it is fixed here rather than filed. Both documents present for the length of an insert is a state a search survives; neither is not.

**The `updated` counter** — right, and I took your suggestion of `replaced_document_id` over `sync_local_flow`'s approach, which greps its own result message for the word "replaced".

Five new tests, each verified to fail without the change it covers. Making the counter testable also fixed something in the harness worth having: `ingest_file` was a bare `AsyncMock`, so `result.replaced_document_id` was a truthy mock.

@codex review

DEENUU1:
Merging. CI green end to end including `e2e`; Codex made one full pass on this branch and found three real things, all fixed in f4a257a with a test each that fails without it. The re-review has answered "usage limits" for the last hour and never reached 9730166.

Reading the diff myself found two more, both in 9730166: a comment claiming the reorder's worst case was "both for as long as the insert takes", when a delete that fails leaves both *permanently* — the honest claim is that this order fails the recoverable way — and a `str(... or "")` normalisation on the match line that changed nothing.

PR #995, opened 2026-08-20
fix(rag): record a document row for what a connector sync ingests

### LINKED ISSUE #992: Documents ingested by a connector sync are invisible to every per-document surface

DEENUU1:
A synced document was searchable and invisible. `GET /kb/{kb_id}/documents`
reads `rag_documents` through `get_for_kb`, and `_run_source_sync` created
no row there - so a Drive folder synced into a knowledge base reported
"ingested: 40" and left the Documents tab empty, the collection's own
`document_count` at zero, and the documents unreachable by delete: one
ingested from a folder could be removed only by dropping the whole
collection. A failure was a number in the sync log and a reason nowhere,
so "which four of the forty failed, and why" had no answer.

The bookkeeping is `sync_local_flow`'s, as the mode logic was: a row per
file, `complete_ingestion` with the chunk count and the id it replaced,
`fail_ingestion` with the reason on the other branch. `complete_ingestion`
already retires the superseded row, so the tab does not grow a row per
sync the way the store grew a document per sync before #990. A *skipped*
file creates nothing - it has a row from the run that ingested it.

The knowledge base id is resolved once, in the session that is already
open, rather than per file: `_config_for_collection` was already finding
that row to read its parser settings, so `_knowledge_base_for` now answers
both questions from one lookup and the config helper is written on top of
it.

One thing deliberately not done. A synced document keeps **no original**:
an upload stores a copy under `rag/{collection}` for its retry button, and
mirroring every synced file onto this deployment's disk is a cost per
corpus rather than per failure. `storage_path` is empty and `has_file` is
false, which is what a surface offering a download reads, and `retry`
refuses with the message it already had. Re-running the sync is the retry -
since #990 that skips everything unchanged and re-fetches exactly what has
no document, so four failures out of forty cost four transfers.

No backfill either, and this is the reason rather than an omission: no
deployment carries synced documents, so a migration for rows that do not
exist is a branch written for nobody. If one ever does, the store's own
listing is the source - `existing_document` already reads it, and each
entry carries the `source_path`, the hash and a filename.

## Verified

Eighteen tests in `tests/test_connector_sync_modes.py`, all eighteen of
which fail against the unfixed flow. Five are new and about this: a row is
created and completed, it names the knowledge base that owns the
collection, a replacement retires what it superseded, a skipped file
creates nothing, and a file that failed to parse keeps its own reason.
`make lint`, `make docs-build`, and the backend suite - 6205 passed.

`docs/file-processing.md` already claimed every ingest path writes a
tracking row, which was true of the upload, the CLI and the local sync
only. It now says what a synced document does and does not keep.

Closes #992

### DIFF
--- a/backend/app/api/routes/v1/knowledge_bases.py
+++ b/backend/app/api/routes/v1/knowledge_bases.py
@@ -14,6 +14,7 @@ from fastapi.responses import FileResponse
 from app.api.deps import (
     Auth,
     CollectionAccessSvc,
+    IngestionSvc,
     KnowledgeBaseSvc,
     RAGDocumentSvc,
     SyncSourceSvc,
@@ -257,6 +258,7 @@ async def delete_kb_document(
     doc_id: UUID,
     service: KnowledgeBaseSvc,
     rag_doc_service: RAGDocumentSvc,
+    ingestion_service: IngestionSvc,
     ctx: Auth,
 ) -> None:
     """Remove a document from the KB (cascades to vectors + file storage).
@@ -264,6 +266,13 @@ async def delete_kb_document(
     Verifies the doc actually belongs to this KB's collection - without that
     check a KB owner could pass any doc_id and remove docs from KBs they
     don't own.
+
+    The ingestion service is what removes the *vectors*, and this route did not
+    take one: it deleted the tracking row and left the content searchable, so a
+    collection held a document nobody could see, delete or re-ingest - the next
+    `new_only` sync matched its unchanged hash and skipped it (#992). The
+    argument has no default on the service any more, which is what stops a third
+    route repeating it.
     """
     kb = await service.get_for_write(kb_id, ctx=ctx)
     doc = await rag_doc_service.get_document(str(doc_id))
@@ -272,7 +281,7 @@ async def delete_kb_document(
             message="Document not found in this knowledge base",
             details={"kb_id": str(kb_id), "doc_id": str(doc_id)},
         )
-    await rag_doc_service.delete_document(str(doc_id))
+    await rag_doc_service.delete_document(str(doc_id), ingestion_service)
 
 
 @router.get("/{kb_id}/sync-sources", response_model=SyncSourceList)
--- a/backend/app/services/rag_document.py
+++ b/backend/app/services/rag_document.py
@@ -17,6 +17,7 @@ from app.db.models.knowledge_base import KnowledgeBase
 from app.db.models.rag_document import DocumentStatus, RAGDocument
 from app.services.rag.config import get_supported_formats
 from app.services.rag.documents import has_indexable_text
+from app.services.rag.ingestion import IngestionService
 from app.services.rag.vectorstore import BaseVectorStore
 from app.repositories import rag_document_repo
 from app.schemas.rag import (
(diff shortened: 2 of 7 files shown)


DEENUU1:
Three of the five fixed in c5d5062, one filed, one declined.

**Row before the ingest** — the sharpest of the five, and it's a trap this branch set for itself. Written afterwards and failing, the row left the vector document stored and untracked, and the next `new_only` run matched its hash and skipped the file before reaching the write — reintroducing precisely the state the branch removes, under one unlucky condition. Now the upload path's order; worst case is a row saying `processing` beside a finished document, which is visible and deletable.

**App-scoped collections** — right, and pre-existing for the config half: `_config_for_collection` had the same equality test, so a source on an `app` collection has always been parsed with deployment defaults rather than that collection's settings. Two passes now, because the caller's own row must win over a deployment-wide one of the same name; another tenant's still matches neither.

**The models** — right, both resolved once for the collection.

**The failed-row retirement (#996 filed).** Real, and it can't be fixed here: retirement matches `vector_document_id` and a failed parse writes none, so the succeeding run has nothing to name — and matching by filename instead would delete the other row when a bucket holds two keys of one basename, which is the collision #990 just removed on the vector side. The fix is a `source_path` column on `rag_documents`, which is worth having anyway since the Documents tab cannot currently say where a document came from.

**The backfill is declined**, for the reason in the branch's first commit: no deployment carries synced documents, so it is a migration for rows that do not exist. If one appears, the store's own listing is the source — `existing_document` already reads it.

Seven more tests; the three about this commit fail against the previous one.

@codex review

DEENUU1:
Both fixed in ab2481d, and the first was the sharpest finding on this PR.

**The delete didn't delete.** `delete_document` took `ingestion_service: Any = None` and removed vectors only when a caller passed one — `/rag/documents/{doc_id}` did, `/kb/{kb_id}/documents/{doc_id}` did not. So deleting from the Documents tab removed the row and left the content searchable, and for a synced document that was permanent: the next `new_only` run matched its unchanged hash and skipped it, so the row never came back. Searchable, invisible and undeletable again — through the button meant to remove it.

The parameter is required and typed now, which is the half that lasts: a third route cannot repeat it without `ty` saying so. Same trap `complete_ingestion` describes one method below — "an argument the caller may omit is an argument some caller will" — and the `Any` was the other half of how it hid. Pre-existing and reachable today for an *uploaded* document deleted from the tab; this branch made it reachable for synced ones, which is why it's fixed here rather than filed.

**The "on every path" claim** — right, and I took the second option you offered. `sync_local_flow` still writes its row after the ingest and only on success, so the claim is scoped to the upload and connector paths, the local gap is stated plainly in the page, and #997 is the reorder. Not folded in because `POST /rag/sync/local` is `is_app_admin` naming a server directory, and opening its row first changes what a failed local file leaves behind — an improvement, and not what #992 asked for.

Five new tests. Suite at 6217.

@codex review

PR #999, opened 2026-08-20
fix(rag): open a local sync's document row before indexing too

### LINKED ISSUE #997: The local-directory sync writes its document row after indexing, and only on success
`sync_local_flow` writes a document's tracking row **after** `ingest_file`
returns, and only when it returned `done`. Two consequences, both of which the
connector sync no longer has since
[#992](https://github.com/vstorm-co/agenticos/issues/992).

## Where it is

`backend/app/worker/tasks/rag_tasks.py:450-474`:

```python
with metered_by(ledger):
    result = await ingestion_service.ingest_file(...)
if result.status.value == "done":
    ...
    async with get_worker_db_context() as db:
        doc = await RAGDocumentService(db).create_document(...)
        await RAGDocumentService(db).complete_ingestion(...)
else:
    failed += 1
```

## What it costs

**A row written after the ingest can fail after it.** A database blip, or a
filename longer than the 255-character column, and the vector document is stored
with no row — searchable, invisible, and undeletable. The next `new_only` run
then matches its unchanged hash and *skips* the file before reaching the write,
so it stays that way permanently. This is exactly the state #992 removed for the
connector sync, and the same reasoning applies here.

**A file that failed to parse leaves no row and no reason.** `failed` is
incremented in the sync log and nothing anywhere says which file or why — the
same gap #992 closed for a connector sync with `fail_ingestion`.

## Fix

The two helpers already exist and are the shape to use:
`_open_document_row` before the ingest, `_settle_document_row` after. The local
flow supplies `organization_id=None` and `knowledge_base_id=None`, since it names
a directory on the server rather than a collection somebody's organization owns.

Worth doing at the same time, because it is the same call: the local flow creates
its rows with **no** `ingestion_config`, so `_tracked_item` reports `parser: null`
for every locally-synced document. The config is already resolved a few lines
above to build the ingestion service; keeping a reference to it is the whole
change.

## Done when

A locally-synced document's row exists before its vectors do, a file that failed
to parse carries its own reason, and the row says which parser read it. With a
test that the row is opened first, as
`tests/test_connector_sync_modes.py::TestWhatASyncedDocumentLeavesBehind::test_the_row_is_opened_before_the_file_is_indexed`
does for the connector path.

Found reviewing #995. Not folded into it because `POST /rag/sync/local` is a
narrower path — it names a directory on the server and is `is_app_admin` only —
and because opening the row first changes what a *failed* local file leaves
behind, which is an improvement but not what #992 asked for. `docs/file-processing.md`
says plainly that this path still has the exposure.

DEENUU1:
The last path still writing its row afterwards, and the one the previous
change had to leave a note about in the docs.

`sync_local_flow` created its row after `ingest_file` returned, and only
when it returned `done`. Two consequences, both of which the connector sync
stopped having in #992:

- **A row written after the ingest can fail after it.** A database blip, or
  a name longer than the 255-character column, and the vector document is
  stored with nothing tracking it - then the next `new_only` run matches its
  unchanged hash and skips the file before reaching the write, so it stays
  searchable, invisible and undeletable for good.
- **A file that failed to parse left no row and no reason.** `failed` was
  incremented in the sync log and nothing anywhere said which file or why.

Both fixed by the two helpers the connector path already uses -
`_open_document_row` before, `_settle_document_row` after - with
`organization_id` and `knowledge_base_id` at `None`, because a path on the
server belongs to no tenant and no knowledge base. That is also why
`POST /rag/sync/local` is the one route still carrying `is_app_admin`.

Two smaller things came with it, both in the same lines:

- **The rows now say which parser read the document.** They carried no
  `ingestion_config` at all, so `parser` read `null` for every
  locally-synced file while the setting that chose it sat a few lines above,
  already resolved to build the ingestion service.
- **`updated` is counted off `replaced_document_id`** rather than by
  searching the result's own message for the word "replaced" - the string
  dependency #990's review removed from the connector flow. Equivalent
  today, since `ingest_file` writes that word exactly when it replaced
  something; one of the two is a fact and the other is a sentence.

## Verified

Two tests, both failing without the change: a locally-synced file that fails
to parse keeps its own reason, and its row says which parser read it.
`make lint`, `make docs-build`, and the backend suite - 6219 passed.

`docs/file-processing.md` said the row-before-index order held for the
upload and the connector sync, and named this path as the exception. It now
says every path, which is true.

Closes #997

### DIFF
--- a/backend/app/worker/tasks/rag_tasks.py
+++ b/backend/app/worker/tasks/rag_tasks.py
@@ -619,12 +619,13 @@ async def _open_document_row(
     whole collection (#992). A failure was counted in the sync log and recorded
     nowhere per-file, so "which four of the forty failed, and why" had no answer.
 
-    No original is stored, unlike an upload: a synced file's bytes live in the
-    system it was synced from, and mirroring every one of them onto this
-    deployment's disk to make a retry button work is a cost per corpus rather
-    than per failure. `has_file` is false for these, and re-running the sync is
-    the retry - which since #990 skips everything unchanged and re-fetches
-    exactly what has no document.
+    No original is stored, unlike an upload, and that holds for both callers:
+    a connector's file lives in the system it was synced from and a local one is
+    already on this host's disk at the path the sync named, so keeping a second
+    copy to make a retry button work is a cost per corpus rather than per
+    failure. `has_file` is false for these, and **re-running the sync is the
+    retry** - which since #990 skips everything unchanged and re-fetches exactly
+    what has no document.
     """
     from app.services.rag_document import RAGDocumentService
 
--- a/backend/app/worker/tasks/rag_tasks.py
+++ b/backend/app/worker/tasks/rag_tasks.py
@@ -375,7 +375,6 @@ async def _run_ingestion(
 async def _run_sync(
     sync_log_id: str, source: str, collection_name: str, mode: str, path: str
 ) -> dict[str, Any]:
-    from app.services.rag_document import RAGDocumentService
     from app.services.rag_sync import RAGSyncService
 
     target_path = Path(path).resolve()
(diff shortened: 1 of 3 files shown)


DEENUU1:
Merging on CI (green end to end, including `e2e`) plus a read of my own diff. Codex has answered "usage limits" since before this PR opened and has not reviewed either commit.

The self-review found one thing, fixed in 42466b3: `_open_document_row`'s docstring justified "no original is stored" with a reason true only of the connector sync — that the bytes live in the system they were synced from. The local sync shares that helper now, and its file is already on this host at the path the sync named, which reaches the same conclusion by a different route.

Issue #983, opened 2026-08-20
A sync source is created, repointed and deleted with no audit entry


DEENUU1:
Nothing in the audit log records that a sync source was created, repointed or
deleted.

`grep -n record_audit app/services/sync_source.py` returns nothing. Every other
service that decides who can reach what writes an entry — `sandbox_connection`
does, `organization_secret` does — and this one binds a **credential** to a
**collection**, which is the platform's actual authorization decision for
ingested documents (`docs/file-processing.md#who-ends-up-able-to-read-what-a-source-ingested`).

So "who decided that this credential's reach becomes searchable by this
collection's readers" has no answer after the fact. Nor does "when did this
source start pointing at the org collection instead of my personal one".

## What breaks

1. Member A creates a sync source on an `org` collection with a broad
   credential; it syncs; the documents are searchable org-wide.
2. Somebody notices weeks later.
3. There is no record of who created it, when, which credential it named, or
   whether it was originally pointed somewhere narrower.

`backend/app/services/sync_source.py:196` (`create_source`), the update path, and
the delete path.

## What to build

Entries for `sync_source.created`, `sync_source.updated` and
`sync_source.deleted`, following the shape `sandbox_connection` already uses:
`actor_user_id`, `organization_id`, `target_type="sync_source"`, `target_id`.

`details` carries the decision and not the payload — the rule in
`.claude/rules/exceptions-security.md` applies here as much as to a refusal:
`{"connector_type", "collection_name", "secret_id", "fields": sorted(update_data)}`.
Never the config document, which is a place a credential has been posted before,
and never the row.

An update that changes `collection_name` is worth naming as such in the entry,
since that is the audience change rather than a rename.

## Done when

Creating, repointing and deleting a source each leave an entry naming the actor,
the collection and the secret id — and a test asserts no config value reaches
`details`.

Found deciding #938. Same reason as #982: the model there says the decision is
the operator's to make explicitly, and an explicit decision nobody recorded is
only half of that.

PR #1010, opened 2026-08-20
feat(rag): say who will be able to read what a source ingests

### LINKED ISSUE #982: A sync source's wizard never says who will be able to read what it ingests
The wizard never says who will be able to read what a source ingests.

`SyncSourceWizard` asks for a connector, a credential, a config and a schedule,
and the collection is chosen without a word about the consequence. But access is
decided at the **collection** and there is no per-document isolation inside one
(`docs/file-processing.md#who-may-reach-a-collection`), so everything that
credential can see becomes readable by everyone who can read that collection. A
Confluence token good for a whole instance, pointed at an `org` collection,
publishes the whole instance to every member holding `collections:view`.

The decision is the operator's to make — the platform deliberately does not
mirror source ACLs, and
`docs/file-processing.md#who-ends-up-able-to-read-what-a-source-ingested` says
why. What is wrong is that it is made **silently**.

## What breaks

1. Vault → store a Confluence token issued for the whole instance.
2. Knowledge bases → open an `org` collection → Integrations → add a Confluence
   source pointing at it.
3. Nothing on any step names the audience. The source syncs, and every member
   with `collections:view` can now search documents from spaces they were never
   given.

`frontend/src/components/rag/sync-source-wizard.tsx` — the step that carries
`collection_name`.

## What to build

- The step that chooses the collection states the consequence, naming the
  collection's scope: an `org` collection says "everyone in this organization
  with `collections:view`", a `personal` one says "only you". The scope is already
  on the `knowledge_bases` row the wizard has.
- The sentence names the credential too, because the pair is the decision:
  *"<credential> can read whatever it has been granted; all of it becomes
  searchable by <audience>."*
- **Repointing re-asks.** `PATCH` on a source's `collection_name` widens the
  audience without touching the credential, which is the same decision made
  again — and today it is one dropdown with no prompt at all.
- Copy through `next-intl`, and the step is gated on nothing new: whoever can
  create the source can read this.

## Done when

Creating a source, and repointing one, both state which people will be able to
search what that credential can reach — and a `personal` collection says so
differently from an `org` one.

Found deciding #938, which is where the model is written down. Blocks nothing;
every connector on that list inherits it, so it is worth having before the
third one rather than after.

DEENUU1:
The wizard asked for a connector, a credential, a config and a schedule, and the
collection was chosen without a word about the consequence. Access is decided at
the **collection** and there is no per-document isolation inside one, so
everything that credential can reach becomes readable by everyone who can read
that collection — a Confluence token issued for a whole instance, pointed at an
`org` collection, publishes the instance to every member holding
`collections:view`.

The decision is the operator's to make; the platform deliberately does not mirror
a source's own ACLs, because a mirrored ACL goes stale between syncs and stale
authorization is worse than none. What was wrong is that it was made silently.

## What it says

The last step — the one that decides the collection — names the credential and
the audience together, because the pair is the decision. The credential's own
permissions are a ceiling nothing in this product can raise; `config` narrows the
reach and cannot be relied on to keep it narrow, being a field anyone with
`collections:edit` can widen later.

One sentence per scope, because each names different people:

| Scope | Ends with |
|---|---|
| `personal` | by you alone |
| `org` | by everyone in this organization who can view that collection |
| `app` | by anybody in this deployment |
| no collection | nothing can search it yet; decided when it is cloned into a base |

## Where it appears — three things worth a reviewer's eye

- **Not conditional on the collection picker.** That control renders only where
  there is more than one collection to choose from, and the case the issue's repro
  walks is a knowledge base offering exactly one. The notice reads
  `defaultCollection` when there is nothing to pick — there is a test for exactly
  that.
- **The clone step carries it too**, which is the half of "repointing re-asks"
  this branch can honour: a clone references the same vault secret and names a
  different collection, so the audience changes while nothing about the credential
  does.
- **Repointing an existing source has no UI to ask in.** The issue says it "is one
  dropdown with no prompt at all" — there is no such dropdown.
  `PATCH /rag/sync/sources/{id}` has no frontend client, and there is no source
  editor anywhere in `src/components/rag`, so a repoint is reachable only through
  the API and the CLI. #983 (merged) is what records one. Say the word if you want
  the editor built and the prompt put on it; that is a feature, not this fix.

## Details

The credential's name comes from the same `/secrets` list the credential step
offers, so what is shown is what was chosen. A caller who cannot read the vault
gets the sentence *without* a name rather than one with an empty parameter in it —
an ICU `select`, not a ternary, per `.claude/rules/frontend.md`.

`collections` now carries `scope` alongside `name`; both call sites already had it
(`kb.scope` on the KB page, an empty list on the org-integration list, which is
what produces the "filed under no knowledge base" sentence).

No new page, tab or create control, so the onboarding walkthrough owes no stop —
the integrations stops in `tour.ts` already cover the surface this sits inside.

## How it was verified

- `sync-source-audience-notice.test.tsx` — 6 tests: one per scope, the unassigned
  case, the unnamed credential, and a name with no scope beside it.
- `sync-source-wizard.integration.test.tsx` — 5 more, including the issue's repro
  (a pinned `org` collection, no picker), the picker changing the sentence, and the
  clone path.
- `make lint-frontend`, `bun run check:i18n`, `make test-frontend-cov` — 5264
  passed, 100% lines/statements/functions — `make docs-build`, and `make check`
  green end to end.

Closes #982

### DIFF
--- a/docs/file-processing.md
+++ b/docs/file-processing.md
@@ -901,12 +901,23 @@ naming the row it came from: it points a credential somebody already scoped at a
 different collection, so its audience changes while nothing about the credential
 does (#983).
 
-What this rule still owes, and does not yet have, is the other half - saying the
-consequence *before* the fact rather than after it:
-[#982](https://github.com/vstorm-co/agenticos/issues/982) states it in the
-wizard where the collection is chosen, and re-asks when a source is repointed at
-a different one. Today that step is silent, which is exactly the implicitness
-this section exists to name.
+**And it is said before the fact, not only after it.** The wizard's last step -
+the one that decides the collection - names the credential and the audience
+together, because the pair is the decision: *"<credential> can read whatever it
+has been granted, and everything it ingests becomes searchable in <collection>
+by ..."*. Each scope ends that sentence differently - `personal` is its owner,
+`org` is everyone who can view the collection, `app` is anybody in the
+deployment - and an integration filed under no knowledge base says that nothing
+can search it yet. The sentence does not wait for the collection *picker*, which
+only appears where there is more than one collection to choose from: the case
+this was filed from is a knowledge base offering exactly one, where there is
+nothing to pick and the consequence is the same (#982).
+
+Cloning says it too, and for the reason above: it is the only way to change a
+source's audience from this product's own UI. Repointing an existing one is a
+`PATCH` on `collection_name`, which no screen sends today - there is no source
+editor - so it is reachable through the API and the CLI, where the audit entry
+above is what records it.
 
 ### What a new connector owes
 
--- a/docs/file-processing.md
+++ b/docs/file-processing.md
@@ -905,7 +905,9 @@ does (#983).
 the one that decides the collection - names the credential and the audience
 together, because the pair is the decision: *"<credential> can read whatever it
 has been granted, and everything it ingests becomes searchable in <collection>
-by ..."*. Each scope ends that sentence differently - `personal` is its owner,
+by ..."*. A connector that authenticates with nothing has no credential to name
+and the sentence does not invent one; nor does it name one whose reader holds no
+`secrets:view`. Each scope ends that sentence differently - `personal` is its owner,
 `org` is everyone who can view the collection, `app` is anybody in the
 deployment - and an integration filed under no knowledge base says that nothing
 can search it yet. The sentence does not wait for the collection *picker*, which
--- a/frontend/messages/en.json
+++ b/frontend/messages/en.json
@@ -3913,6 +3913,11 @@
   },
   "rag": {
     "addSyncSource": "Add sync source",
+    "audienceApp": "{named, select, no {The credential this source uses} other {{credential}}} can read whatever it has been granted, and everything it ingests becomes searchable in {collection} by anybody in this deployment.",
+    "audienceOrg": "{named, select, no {The credential this source uses} other {{credential}}} can read whatever it has been granted, and everything it ingests becomes searchable in {collection} by everyone in this organization who can view that collection.",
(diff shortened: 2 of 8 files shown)


DEENUU1:
Both findings were real and are fixed in 21da947.

**The closed wizard reading the vault** — right, and worse than it looks: `useSecrets()` was in `SyncSourceWizard`'s own body, the knowledge-base page mounts that wizard whether or not it is open, and a member without `secrets:view` therefore got two refusals plus a retry on every page load. The lookup moved into `SourceAudienceNotice`, which renders inside the dialog and only on the step that shows the sentence — so the query starts when somebody is reading it, and by then the credential step has usually filled the cache. There is now a test asserting what a *closed* wizard requests, which is the shape of the defect rather than a proxy for it.

**A credential-free connector** — also right. `secret_kind: "none"` is supported and the credential step says so in as many words, so the notice contradicted the step before it. The sentence selects on three states now rather than two: no credential to name, one whose name this caller cannot read, and one it can. The message keys carry all three in one ICU `select` per scope, so no locale can end up with the head of one and the tail of another.

`make test-frontend-cov` is green at 5266 tests, 100% lines/statements/functions.

DEENUU1:
Merging. Codex reviewed the first commit, found two things, both real, both fixed in 21da947 — and has not answered again since (it was back to "usage limits" before that review, so a second pass may not come). So the merge rests on CI, green end to end including `e2e`, plus the reply above and a read of my own diff.

What that read found and did not change: the notice renders inside the dialog, so its vault query shares the credential step's cache key rather than adding a request; and the three-state `select` lives in one message per scope, so a locale cannot end up with the head of one variant and the tail of another.

Issue #996, opened 2026-08-20
A synced file's error row is never retired by the sync that succeeds


DEENUU1:
A file that fails to parse on one sync and succeeds on the next leaves two
`rag_documents` rows: the error row from the first attempt and the completed row
from the second. Every repeated failure adds another, permanently, and each one
counts toward the collection's `document_count`.

## Why the usual mechanism does not catch it

`complete_ingestion` retires the row it superseded, but only by
`vector_document_id` — `_retire_superseded` matches on the vectors a replacement
just deleted. A failed parse writes **no vectors**, so there is nothing for the
successful retry to name: `result.replaced_document_id` is `None` and the error
row is invisible to the retirement.

## Why it cannot be fixed by filename

The obvious fix — retire prior untracked rows for the same `collection_name` and
`filename` before creating a new one — reintroduces the collision that
[#990](https://github.com/vstorm-co-agenticos/issues/990) removed on the vector
side: `rag_documents` has **no `source_path` column**
(`app/db/models/rag_document.py:39-49`), so `a/readme.md` and `b/readme.md` in
one bucket are indistinguishable by filename, and retiring by name would delete
the other file's row.

## Fix

Add `source_path` to `rag_documents`, set it on every ingest path that has one
(the connector sync, the local sync), and retire a prior row for the same
`(collection_name, source_path)` when opening a new one. That is what makes a
row identify the remote file it tracks rather than a name that repeats.

The column is worth having on its own account: the Documents tab currently
cannot say where a document came from, so an uploaded `handbook.md` and one
synced from Drive are the same row to a reader.

Migration and backfill: nothing needs backfilling — no deployment carries synced
documents, and a nullable column suits an uploaded document, which has no remote
address.

## Done when

A file that fails on one sync and succeeds on the next leaves **one** row, and
two objects with the same basename in one bucket keep two. With a test for each,
since the second is the trap the naive fix falls into.

Found reviewing #995, which is where the rows start being written at all. Not
caused by that branch — before it a connector sync wrote no rows, so there was
nothing to accumulate. Bounded in practice: it needs a file that keeps failing
to parse across several syncs, and each row is a genuine record of a genuine
failure. What it corrupts is the count, not the content.

PR #1018, opened 2026-08-20
refactor(rag): S3Connector is an object store's subclass

### LINKED ISSUE #988: Generalize S3Connector to an object store, before Azure Blob and GCS

DEENUU1:
`S3Connector` was an S3 connector. Azure Blob Storage and Google Cloud Storage are
the same connector with a different client — a container, a prefix, a credential,
and a flat listing of objects with a size and a modified time. #938 put them fifth
on the list with a condition attached: *"if `S3Connector` is refactored to an
object-store shape, these are configuration rather than code."*

This is that refactor, before either of them rather than after, because the
alternative is three connectors sharing a shape by resemblance — where a fix to
the listing loop lands in one of the three.

## What moved

`ObjectStoreConnector` holds what all three do identically:

- the listing, and the skip for a key ending in `/` — a console's "folder" is a
  zero-byte object, which ingests as a document with no bytes and no filename;
- `source_path` as `<scheme>://<container>/<key>`, which is the address the whole
  sync path matches a stored row on (#996);
- the container read back out of that address on download;
- the destination, which is `BaseSyncConnector`'s answer and not a connector's —
  the property a new store most easily loses.

A subclass supplies a client, a `SCHEME`, and which `CONFIG_SCHEMA` field names
the container: `bucket` for S3 and GCS, `container` for Azure, because the form
should say what the store's own console says.

Both hooks are deliberately **blocking**, run on a worker thread by the shared
class. boto3, `azure-storage-blob` and `google-cloud-storage` all ship synchronous
clients, so a hook declared `async` over one of those would either block the event
loop or be wrapped twice.

## What was deliberately not carried

- **No `etag` on `StoredObject`.** Every object store offers one and the sync path
  reads none — it compares a `content_hash` of the bytes it downloaded (#990) — so
  it would be a field nothing reads. The right place for a bytes-free change signal
  is with the caller that reads it, which `docs/file-processing.md` already asks a
  new connector for.
- **`S3Connector.validate_config` is gone**: an override that called `super()` and
  added nothing.

## No behaviour change

A stored source lists and downloads exactly what it did before. `Prefix` is still
omitted rather than sent empty (boto3 means the same by `Prefix=""`, but the call
stays the call it was), and the string-or-`datetime` `LastModified` is still parsed
in the client's own method, because which of the two arrives is the client's own
vocabulary.

## How it was verified

- `tests/test_object_store_connector.py` — 10 tests. Seven are written against a
  subclass that is **not** S3: a test using `S3Connector` for the shared half would
  prove the same code twice and say nothing about the next store. Three pin S3's own
  calls unchanged — the paginator it asks for, the arguments it passes with and
  without a prefix, and the bucket/key a download names.
- The existing connector tests are untouched and pass: 102 across
  `test_object_store_connector`, `test_untrusted_remote_names`,
  `test_sync_source_credentials` and `test_commands`.
- `make check` green — every CI job except e2e — and `make docs-build`.

## Docs

`docs/file-processing.md` said what a new connector owes, and now says that an
object store owes less than that, plus what its subclass supplies. The connector
roadmap paragraph no longer states the #988 condition as pending.

Adding GCS from here is a client and a `CONNECTOR_TYPE`; Azure is that plus a new
`SecretKind` for a storage account key or a SAS token. Neither is in this PR.

Closes #988

### DIFF
--- a/backend/app/services/rag/connectors/object_store.py
+++ b/backend/app/services/rag/connectors/object_store.py
@@ -1,0 +1,138 @@
+"""What every object store's connector does identically.
+
+S3, Azure Blob Storage and Google Cloud Storage are one connector with three
+clients: a container, a prefix, a credential, and a flat listing of objects with
+a size and a modified time. #938 put the other two fifth on the list with a
+condition attached - *"if `S3Connector` is refactored to an object-store shape,
+these are configuration rather than code"* - and this is that shape, written
+before either of them rather than after, because the alternative is three
+connectors sharing a form by resemblance (#988).
+
+What a subclass supplies is a client and its SDK's vocabulary: how a listing
+paginates, and how one object is written to a path. What it inherits is
+everything a reader of three connectors would otherwise have to check three
+times - that a key ending in `/` is a directory marker and not a document, that
+`source_path` is `<scheme>://<container>/<key>` and is the dedup key the whole
+sync path matches on, and that the destination is the base class's answer rather
+than the connector's (`BaseSyncConnector.download_file`, which is inherited for
+the reason its own docstring gives).
+"""
+
+import asyncio
+import logging
+from abc import abstractmethod
+from datetime import datetime
+from pathlib import Path
+from typing import ClassVar
+
+from pydantic import BaseModel
+
+from app.core.secret_kinds import StorableSecret
+from app.services.rag.connectors import BaseSyncConnector, ConnectorConfig, RemoteFile
+
+logger = logging.getLogger(__name__)
+
+
+class StoredObject(BaseModel):
+    """One object in a listing, in the terms all three stores describe one in.
+
+    Deliberately not an `etag`. Every object store offers one and the sync path
+    reads none: it compares a `content_hash` of the bytes it downloaded (#990),
+    so an `etag` carried here would be a field nothing reads. A connector that
+    can answer "changed?" *without* the bytes is worth having - it saves the
+    transfer, not just the embedding - and `docs/file-processing.md` says so under
+    what a new connector owes; the place to add it is there and then, with the
+    caller that reads it.
+
+    `modified_at` is carried because `RemoteFile` already has somewhere to put
+    it.
+    """
+
+    key: str
+    size: int | None = None
+    modified_at: datetime | None = None
+
+
+class ObjectStoreConnector(BaseSyncConnector):
+    """A connector for a store addressed by container and key.
+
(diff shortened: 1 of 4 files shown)


DEENUU1:
Real, and fixed in 701d8b6.

The refactor did hold two copies of the largest thing a sync ever holds: `_objects` built a complete `StoredObject` list and the comprehension then allocated the complete `RemoteFile` list beside it, where the old connector kept only the second.

`_objects` yields now, and the conversion moved into `_listing` — which runs **inside** the worker thread, so one `StoredObject` exists at a time beside the list being built. That placement is the part worth naming: converting in `list_files` instead would hand a generator back to the event loop and do the next page's network round trip on it, which is the reason these hooks are blocking to begin with.

The guard is `inspect.isgeneratorfunction` rather than a measurement, because a list is what somebody puts back without noticing.

PR #1025, opened 2026-08-20
fix(auth): rate-limit the auth surface and take bcrypt off the event loop

### LINKED ISSUE #947: Rate-limit the auth surface, and take bcrypt off the event loop
Two things that are each survivable and together are not: no route in `auth.py`
carries a rate limit, and `verify_password` runs bcrypt on the request event
loop.

**Severity: Critical** · AUD-001, audit of 2026-08-19

## Where

- `backend/app/api/routes/v1/auth.py:33` `/login`, `:55` `/register`, `:64`
  `/refresh`, `:112` `/password-reset/request`, `:149` `/magic-link/request`,
  `:175` `/magic-link/verify`
- `backend/app/core/security.py:89` `verify_password`, `:97` `get_password_hash`
- `backend/app/services/user.py:218` `authenticate`

`grep` for `limit`, `throttle`, `attempt` or `lockout` across `auth.py` and
`user.py` returns nothing. There is no per-IP counter, no per-account lockout and
no cooldown.

**The limiter is not missing from the codebase.** `app/services/rate_limit.py` is
Redis-backed, configured in the lifespan, and applied to five public surfaces plus
`limit_agent_run` through `api/deps.py`. The auth surface is the one it was never
wired to — which is why this is not a duplicate of #7, and why #7 wants
re-scoping to exactly this rather than closing.

## How it fails

Measured on an M-series laptop at the repository's default bcrypt cost (12
rounds, what `gensalt()` gives):

```
checkpw: 171 ms per attempt
=> ~6 attempts/second saturate one worker's event loop completely
```

**1. Unauthenticated denial of service.** POST `/login` for any address that has
an account — a company email pattern is enough to guess one — at 20 requests a
second. Each blocks the event loop for 171 ms with no suspension point. For that
time the worker serves nothing else: not the readiness probe the load balancer
reads, not the next frame of an in-flight agent WebSocket, not another person's
request. No credentials, no account, nothing to rate-limit it. On a
single-worker deployment, which is what `docker compose` gives you, the product
is down for as long as the attacker keeps typing.

**2. Unlimited credential brute force.** Same endpoint, no lockout and no
counter. The bcrypt cost is the only brake, and (1) is what that brake buys the
attacker.

**3. A user-enumeration timing oracle.** `authenticate` short-circuits:

```python
# app/services/user.py:218
if (
    not user
    or not user.hashed_password
    or not verify_password(password, user.hashed_password)   # only reached if the user exists
):
```

An address with no account never reaches bcrypt and is refused in a couple of
milliseconds; an address with one takes ~171 ms. Two orders of magnitude,
measurable over the internet. The surrounding code is otherwise careful about
enumeration — `/password-reset/request` and `/magic-link/request` both answer
symmetrically on purpose and say so in their docstrings — so this is an
inconsistency as much as a leak.

**4. An email amplifier.** `/password-reset/request` and `/magic-link/request`
issue a token and send mail on every call, for any address that has an account,
unlimited. One employee's address plus a loop is a mailbox nobody can use and an
SMTP reputation nobody can repair.

## Fix

Three edits, none structural.

```python
# 1. The limiter this repo already has. Per IP for the DoS, per submitted address
#    for the brute force - both, because they stop different attacks.
#    `rate_limit.consume(surface=..., caller=..., limit=...)` is the existing
#    shape; add `login_limit()` beside `run_limit()`.
@router.post("/login", response_model=Token, dependencies=[Depends(limit_login)])

# 2. bcrypt off the loop, in both directions.
async def authenticate(self, email: str, password: str) -> User:
    user = await user_repo.get_by_email(self.db, email)
    stored = user.hashed_password if user and user.hashed_password else _DUMMY_HASH
    ok = await asyncio.to_thread(verify_password, password, stored)
    if not user or not user.hashed_password or not ok:
        raise AuthenticationError(message="Invalid email or password")

# 3. `_DUMMY_HASH` is what closes the oracle: a real bcrypt hash of a random
#    string, computed once at import. An unknown address then costs the same
#    171 ms as a known one, in a thread, and the two are indistinguishable.
```

`get_password_hash` on `/register` and on the password-reset confirm path takes
the same `to_thread` — `hashpw` plus `gensalt` costs more than `checkpw`, not
less.

## Acceptance criteria

- [ ] Every route in `auth.py` is rate-limited, per IP and per submitted address
- [ ] No bcrypt call runs on the event loop
- [ ] An unknown address and a known one take indistinguishable time to refuse
- [ ] Regression test: a concurrent `asyncio.sleep(0)` task completes while a
      login attempt is in flight
- [ ] Regression test: the Nth request to `/login` inside the window is refused
      with 429

OchnikBartek:
## What breaks (Critical · AUD-001)

Two things, each survivable, together not: **no route in `auth.py` was rate-limited**, and **`verify_password` ran bcrypt on the request event loop** (~170ms, no suspension point).

1. **Unauthenticated DoS.** `/login` for any address that has an account, at 20 req/s: each blocks the loop 171ms with no `await`. The worker serves nothing else — not the readiness probe, not an in-flight agent socket. On a single-worker deploy the product is down for as long as the attacker types.
2. **Unlimited brute force** — the bcrypt cost was the only brake, and (1) is what that brake bought the attacker.
3. **A user-enumeration timing oracle** — `authenticate` skipped bcrypt for an unknown address (~ms) and ran it for a known one (~171ms).
4. **An email amplifier** — `/password-reset/request` and `/magic-link/request` sent mail on every call, unlimited.

## Fix — three parts, none structural

**Rate limit.** The limiter this repo already has (`services/rate_limit.py`, Redis-backed, shared across workers) was never wired to this surface. Added `auth_limit()` + `deps.enforce_auth_limit`, called at the top of **every** auth route — before any bcrypt/DB work. Counted **per IP** always and **per submitted address** where the body carries one (`RATE_LIMIT_AUTH_PER_MINUTE`, default 10). Called from the handler rather than as a `Depends` because the per-address half needs the parsed body.

**bcrypt off the loop.** Every bcrypt call — the verify in `authenticate` and the three `get_password_hash` sites — runs in a thread via `asyncio.to_thread`.

**Close the oracle.** An unknown address is verified against `_DUMMY_HASH` (a real hash computed once at import) rather than short-circuited, so known and unknown addresses refuse in the same time.

## Verification (acceptance criteria)

- ✅ Every `auth.py` route rate-limited, per IP and per submitted address
- ✅ No bcrypt call on the event loop — test holds bcrypt in its thread while the loop keeps turning
- ✅ Unknown and known addresses refuse indistinguishably — test asserts an unknown address is still put through bcrypt against the dummy hash
- ✅ The Nth `/login` in the window is refused 429; under the window it reaches the service
- Full backend suite 5668 passed; `rate_limit.py` at 100% (gated); `make lint` clean; `docs/configuration.md` documents the new setting.

Note: GET `/me` (authenticated, returns the cached user) is not rate-limited here — a per-IP limit on an authenticated no-op would harm an office behind one NAT, and per-user limiting of an already-authenticated cheap read is a separate call. Happy to add it if you'd prefer.

Closes #947

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -585,6 +585,30 @@ def _refuse_if_over(decision: rate_limit.Decision, message: str) -> None:
         )
 
 
+async def enforce_auth_limit(
+    request: Request, *, surface: str, identifier: str | None = None
+) -> None:
+    """Refuse an auth attempt from a caller who has made too many this minute.
+
+    Called at the top of an `auth.py` route rather than as a `Depends`, because
+    the per-address half needs the parsed body the route has and a dependency
+    does not. Both halves count against `auth_limit()`: the IP first, because it
+    cannot be varied for free and it is what bounds the unauthenticated bcrypt
+    DoS; then the submitted address, where the body carries one, which is what
+    bounds a brute force against a single account. Lower-casing the identifier so
+    two spellings of one address share a bucket.
+    """
+    limit = rate_limit.auth_limit()
+    decision = await rate_limit.consume(
+        surface=surface, caller=f"ip:{rate_limit.caller_ip(request)}", limit=limit
+    )
+    if decision.allowed and identifier:
+        decision = await rate_limit.consume(
+            surface=surface, caller=f"id:{identifier.strip().lower()}", limit=limit
+        )
+    _refuse_if_over(decision, "Too many attempts. Please wait and try again.")
+
+
 async def limit_embed_script(request: Request) -> None:
     """Refuse an address asking for a widget's script too often.
 
--- a/backend/app/api/routes/v1/auth.py
+++ b/backend/app/api/routes/v1/auth.py
@@ -6,7 +6,13 @@ from typing import Annotated, Any
 from fastapi import APIRouter, Depends, Request, status
 from fastapi.security import OAuth2PasswordRequestForm
 
-from app.api.deps import CurrentUser, DeploymentSettingsSvc, SessionSvc, UserSvc
+from app.api.deps import (
+    CurrentUser,
+    DeploymentSettingsSvc,
+    SessionSvc,
+    UserSvc,
+    enforce_auth_limit,
+)
 from app.core.config import settings
 from app.core.exceptions import AuthenticationError
 from app.core.security import (
@@ -38,6 +44,7 @@ async def login(
     session_service: SessionSvc,
 ) -> Any:
     """OAuth2 password login, returns access and refresh tokens."""
+    await enforce_auth_limit(request, surface="auth_login", identifier=form_data.username)
     user = await user_service.authenticate(form_data.username, form_data.password)
     access_token = create_access_token(subject=str(user.id))
     refresh_token = create_refresh_token(subject=str(user.id))
@@ -54,10 +61,12 @@ async def login(
 
 @router.post("/register", response_model=UserRead, status_code=status.HTTP_201_CREATED)
 async def register(
+    request: Request,
     user_in: UserCreate,
     user_service: UserSvc,
 ) -> Any:
     """Register a new user."""
+    await enforce_auth_limit(request, surface="auth_register", identifier=user_in.email)
     return await user_service.register(user_in)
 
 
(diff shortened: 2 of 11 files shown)


OchnikBartek:
@codex review

PR #1031, opened 2026-08-20
fix(permissions): offer only the roles the caller may actually assign

### LINKED ISSUE #1028: The role pickers offer roles the caller cannot assign, and the service refuses them after the fact
The invite dialogs and the members page offer every role in the catalog except
`owner`, whoever is asking. The service then refuses the ones the caller cannot
assign, which is the shape `.claude/rules/frontend.md` names: *"A control the
caller may not use is **not rendered** — not rendered and then 403."*

## Where

`frontend/src/hooks/use-permissions.ts:104` — `useAssignableRoles`:

```ts
catalog?.roles.map((role) => role.name).filter((name) => name !== "owner")
```

Read by `components/teams/invite-member-dialog.tsx`,
`components/teams/invite-link-dialog.tsx` and
`app/[locale]/(dashboard)/orgs/[id]/members/page.tsx`.

The backend answer is `assignable_roles(role)` in `app/core/permissions.py:288`:
every permission the offered role holds, held at least as widely by the assigner,
**and** something the assigner holds that it does not. So an Admin may assign
`builder`, `operator`, `member`, `viewer` — and not `admin`.

## What breaks

1. Sign in as an Admin, open Members → Invite.
2. The role select offers **Admin**.
3. Send it. `InvitationService.invite` refuses with 403, after the email address
   has been typed.

Pre-existing: the literal-`admin` ceiling refused this too. What changed is the
*set of callers* it can happen to — #1019 (closing #696) replaced that literal
with `assignable_roles`, so a Phase 2 custom role holding `members:manage` is now
capped by what it actually holds, and every one of those gets the same
offer-then-refuse. Same on the members page's role picker, where the refusal is
`change_role`'s (#672).

## Fixed when

The three call sites offer only what the caller may assign, derived from the same
relation the service enforces rather than from a second list — the catalog the
hook already reads carries each role's permissions, so the client can compute it
without a new endpoint. A test asserts an Admin is not offered `admin`.

## Related

Found reviewing #1019, which closes #696 — the backend half. #700 is the
neighbouring hole in `change_role` (an Admin demoting a peer Admin), and #672 is
where this relation came from. Not a duplicate of any of the three: all three are
about what the server allows, and this is about what the client offers.

DEENUU1:
`useAssignableRoles` answered "every role in the catalog except `owner`", whoever
was asking, and the three pickers built on it — both invite dialogs and the
members table — offered exactly that. The service then refused what the caller
could not assign, which is the shape `.claude/rules/frontend.md` names: a control
the caller may not use, rendered and then 403'd.

Sign in as an Admin, open Members → Invite: the role select offers **Admin**. Send
it, and the refusal arrives after the email address has been typed.

Pre-existing — the literal-`admin` ceiling refused this too — but #1019 widened
who it happens to. With the server's ceiling now derived from what a role holds,
every Phase 2 role composed with `members:manage` gets the same offer-then-refuse.

## The fix

`assignableRoles` in `src/lib/assignable-roles.ts` is the client's copy of
`app.core.permissions.assignable_roles`, and deliberately the same arithmetic
rather than a list: `GET /roles/catalog` already returns each role's permissions,
so the answer needs no endpoint of its own, and computing it from the same input
the server uses is what keeps the two from drifting.

| Caller | Offered |
|---|---|
| Owner | admin, builder, operator, member, viewer |
| Admin | builder, operator, member, viewer |
| Member | viewer |
| Viewer | — |
| A Phase 2 role holding only `members:manage` | — |

Three things came with it:

- **The members table draws a label where it used to draw a picker.** Radix shows
  the chosen item's text in the trigger, and a peer Admin's `admin` is not among
  an Admin's options — so that row rendered an *empty control*. It is a badge now,
  like the owner's row and one's own.
- **A picker seeded with a role it did not offer.** Both dialogs held
  `useState<OrgRole>("member")` and never reconciled it with the list.
  `defaultAssignable` keeps Member where it is on offer — which for every built-in
  role that may invite at all, it is — and otherwise takes the least privileged
  role that is.
- **Nothing is offered until both answers are in**, and submit is disabled until
  then. A picker that offers nothing for a beat is a control somebody waits for;
  one that offers too much is a refusal they walk into.

## The fixtures were part of the bug

Both dialogs' tests served a catalog whose roles had **empty** permission sets —
from when the picker was a list and the permissions were decoration. Empty sets
make every role reach every other, so none strictly outranks any and the picker
offers nothing at all: the three existing tests would have gone red for the right
reason. The fixture is now shaped like the built-ins and shared from
`src/test-utils/role-catalog.ts`, which sits outside the coverage `include` list
so a helper does not drag the 100% gate along.

## How it was verified

- `src/lib/assignable-roles.test.ts` — 11 tests on the relation: the pairs the
  backend's own tests assert, a Phase 2 role holding only `members:manage`, an
  unknown role, an unanswered catalog, and a scope compared **by rank** rather
  than as a string (`all < own` alphabetically, which is the mistake `Scope`
  refuses to make on the server).
- Four more across the two dialogs: an Admin offered four roles and not its own on
  each path, the Member default preserved, and nothing sendable before the answers
  land.
- A first integration test for the members page: an Admin offered four roles on a
  Builder's row, a **label** on a peer Admin's, and an Owner still offered the
  picker there.
- `make check` green, `make test-frontend-cov` at 5289 passed and 100%
  lines/statements/functions, `make docs-build`.

## One thing found and not fixed here

`OrgRole` is the closed union of the six built-ins, so a Phase 2 role name does
not type-check as one — the custom-role test needs a cast. The relation itself
takes any role the catalog names, so this is a type-level gap rather than a
behavioural one, and widening that union touches every consumer of it. Worth its
own issue rather than this branch; say the word and I will file it.

## Docs

`docs/permissions.md#who-may-hand-out-which-role` now says that the console
computes the same relation rather than being told it, and why the members table
will not draw a picker for a role the caller cannot assign.

Closes #1028


DEENUU1:
Three findings, all real. Two fixed in 49afbc0, the third filed as #1032 rather than half-fixed.

**A peer Admin's row lost a supported action** — right, and the sharper half of it: `change_role` judges the role being handed *out*, not the one being replaced, refusing only when the target is the Owner. So an Admin may demote a peer Admin to Builder and my badge removed the only way to. The picker is back with the row's current role in the list as a **disabled** item — the trigger needs it there (Radix draws the chosen item's text, so a value the list does not hold renders blank) and disabling it is what says "not this one" without taking the other four away.

**A failed catalog looked like a caller who may assign nothing** — also right, and it is the worse of the two states because it is permanent: two dialogs with a blank field and a submit disabled for good, plus a table of labels, which is exactly what a Viewer sees. `useRoleCatalog` returns its error now and all three surfaces say the list could not be read.

**The organization scoping is real and older than this branch** — `apiClient` puts `X-Organization-Id: activeOrgId` on every request, so `/me/permissions` answers for the active org while this page acts on the one in its URL, and the organizations list opens any org's members page through an overlay link while *switching* is a separate button that navigates to the dashboard. So the ordinary route to another org's members page is the one that leaves the active org behind.

I have not fixed it here, deliberately. `canManage` one line above reads the same wrong answer and has since long before this branch, so scoping the picker alone would leave the page deciding *which controls exist* from one organization and *what they offer* from another — a new inconsistency in place of an old one. #1032 has the repro in both directions, names the two candidate shapes (make the URL authoritative, or take an org id on `/me/permissions`), and asks for a test that opens org B's page with org A active.

`make lint-frontend` and `make test-frontend-cov` green — 5292 passed, 100% lines/statements/functions.

PR #1034, opened 2026-08-20
fix(api): pin the served content-type so a stored file cannot run as script

### LINKED ISSUE #702: Three proxy routes echo the backend's content type, and one of them can serve HTML
Split out of the review of #634. The fourth instance of this — `api/embed/[publicKey]/logo` — is fixed there, in `bd6dddb9`, along with the backend route feeding it. These three are pre-existing and were not touched by that branch.

## What breaks

`frontend/src/app/api/files/[id]/route.ts:28`, `frontend/src/app/api/users/avatar/[userId]/route.ts:35` and `frontend/src/app/api/orgs/[id]/avatar/route.ts:41` each pass on whatever `content-type` the backend answered with:

```ts
"Content-Type": response.headers.get("content-type") || "image/jpeg",
```

The backend decides that type from the *name on disk* (`FileResponse` with no `media_type` lets Starlette guess from the suffix), and the name on disk keeps the extension whoever uploaded the file chose — the upload paths validate the `Content-Type` header the client *declared*, never the bytes. So:

1. Upload `x.html` whose bytes are `&lt;script&gt;…&lt;/script&gt;`, declaring `Content-Type: image/png`. The avatar upload accepts it (`app/services/organization.py`, `app/services/agent.py` — same check `set_page_logo` had).
2. It is stored as `&lt;hex&gt;_x.html`.
3. The authenticated backend route serves it as `text/html` (verified: `mimetypes.guess_type` answers `text/html` for `.html` and `.htm`, `image/svg+xml` for `.svg`, `application/xhtml+xml` for `.xhtml`).
4. The proxy passes that through **from the frontend's own origin**, where `next.config.ts:17` allows `script-src 'self' 'unsafe-eval' 'unsafe-inline'`.

`X-Content-Type-Options: nosniff` does not help: the type is not being sniffed, it is being declared.

## How it differs from the one already fixed

These three require a session to read, so the audience is members of the same organization rather than anyone with a link — which is why they are an issue rather than part of that branch. It is still stored XSS on the app origin, and `api/files/[id]` is reachable for any file in a conversation the caller can open.

## The fix, and it is the same shape as #634's

Pin the type on the way out rather than echoing it. `bd6dddb9` is the worked example: an allowlist at the backend route (`media_type` passed explicitly to `FileResponse`, 404 for anything that is not an image) plus an allowlist in the proxy, and `X-Content-Type-Options: nosniff` on both. `api/files/[id]` cannot use an image allowlist — it serves PDFs and spreadsheets deliberately — so it needs `Content-Disposition: attachment` for anything outside a render-safe set instead.

Minting the stored filename from the validated MIME, the way `set_page_logo` now does, closes the other end for files uploaded after the fix; files already on disk need the serving-side pin.

`frontend/src/app/api/binary-routes.test.ts` has the four cases for the logo route already — `text/html`, `image/svg+xml`, `application/xhtml+xml` and an unnamed type all answer 502. The same table against each of these three, plus a backend test per route that a stored `.html` is refused rather than served.

Found reviewing #634. Related: #701 (the other finding from that review that needed its own decision).

OchnikBartek:
## What breaks (stored XSS, severity:high)

The bytes behind an **avatar** or a **chat attachment** are served with a type the backend takes from the *name on disk* — and the upload paths validate the `Content-Type` the client *declared*, never the bytes, keeping whatever extension the uploader chose. So:

1. Upload `x.html` whose bytes are `<script>…</script>`, declaring `Content-Type: image/png` — accepted.
2. Stored as `<hex>_x.html`; the backend serves it as `text/html` (Starlette/`mimetypes` guess), or the avatar route echoes it.
3. The frontend proxy passes that through **from the app's own origin**, where the CSP allows `'unsafe-inline'` script.

`X-Content-Type-Options: nosniff` doesn't help — the type is *declared*, not sniffed. Same shape #634 fixed for the hosted-page logo; these three require a session, so the audience is the organization (hence pre-existing, not part of that branch). `api/files/[id]` is reachable for any file in a conversation the caller can open.

## Fix — pinned at both ends (#634's shape)

**Frontend proxies** (`src/lib/proxy-content-type.ts` centralises the allowlists):
- Both avatar proxies refuse (**502**) anything not in the four image types, and drop the `image/jpeg` default over unknown bytes.
- The file proxy serves PDFs/spreadsheets on purpose, so it forces a **download** for anything outside the render-safe set (images + PDF) — `text/html`/SVG saved, not shown inline. `nosniff` on all three.

**Backend serving routes** (`image_media_type_for` in `file_storage`, beside `IMAGE_MIME_TYPES`):
- User and org avatar routes guess the type, **404** a non-image, pass it explicitly with `nosniff`.
- The chat-file route serves a render-safe type inline and forces everything else to download, `nosniff` both — the declared `mime_type` is not trusted to decide rendering.

## Verification

- Frontend: the table `text/html`, `image/svg+xml`, `application/xhtml+xml`, unnamed → **502** on each avatar route, **download** on the file route; images pinned + nosniff; PDF stays inline. (`binary-routes.test.ts`, 63→ more.)
- Backend: `image_media_type_for` refuses `.html/.svg/.xhtml`/no-suffix; each avatar route 404s a stored `.html`; a `text/html` chat attachment downloads, a PDF stays inline.
- Full backend suite 5693 passed (`_chat_file_bytes` is gated, 100%); frontend coverage gate green; `make lint` clean.

`docs/channels.md` already states this principle (and names "an avatar uploaded years ago through another route") — this makes that claim true for avatars, so no page change is owed.

Out of scope, filed separately: the **agent avatar** GET (`agents.py`) has the same guess-from-suffix serving with no allowlist/nosniff.

Closes #702


OchnikBartek:
@codex review

PR #1036, opened 2026-08-20
fix(permissions): the organization in the URL is the tenant

### LINKED ISSUE #1032: The members page judges permissions from the active organization while acting on the one in the URL
`/orgs/{id}/members` acts on the organization in the **path** and judges
permissions from the **active** one, and the two are routinely different.

## Why they diverge

`apiClient` puts `X-Organization-Id: activeOrgId` on every request
(`frontend/src/lib/api-client.ts:86`), so `GET /me/permissions` answers for the
active organization. The members page reads it through `usePermissions()` and
decides:

- `canManage` — whether any role control, invite button or spending-limit field
  renders at all (`page.tsx:91`);
- which roles a picker offers, since #1028 (`useAssignableRoles`).

Meanwhile `useMembers(id)`, `useInvitations(id)` and every write name the org from
the URL.

And nothing keeps them the same: the organizations list opens **any** org's
members page through an overlay link (`orgs/page.tsx:147`), while *switching* the
active org is a separate button that navigates to the dashboard
(`orgs/page.tsx:195`). So the ordinary way to reach another org's members page is
the way that leaves the active org behind.

## What breaks

1. Be an Owner of Acme and an Admin of Globex. Have **Globex** active.
2. Organizations → click the **Acme** card.
3. The page lists Acme's members, and every decision about them is Globex's: the
   picker offers what an Admin may assign, so Acme's Owner cannot offer Admin on
   their own organization.

The inverse is the one that matters more: Admin of Acme, **Owner** of Globex
active → the picker offers `admin`, `change_role` refuses it, and the
offer-then-refuse #1028 removed is back through another door.

Worse than the picker: with `members:manage` in one organization and not the
other, `canManage` shows or hides the whole set of controls against the wrong
answer. That half is older than #1028.

## Fixed when

The page's decisions and the page's writes name one organization. Two shapes, and
the choice is worth making explicitly:

- **Make the URL authoritative.** The page (or the `[id]` layout) switches the
  active organization to the one in the path, so the header and the path agree
  for every request the page makes. Cheap, and it makes the org switcher and the
  URL one thing rather than two.
- **Ask per organization.** `/me/permissions` takes an org id, and the page asks
  for the one it is showing. More surgery, and it leaves two notions of "current"
  in the app.

Either way a test opens org B's page with org A active and asserts the controls
describe B.

## Related

Found reviewing #1031 (which closes #1028) — the automated review raised the
picker half. Not a duplicate: #1028 was "the picker offers what the service
refuses", and this is "the page asks the wrong organization what the caller may
do". #696 and #700 are the server-side ceiling in the same area.

DEENUU1:
`/orgs/{id}/members` acts on the organization in its **path** and judged
permissions from the **active** one, and the two are routinely different.

`apiClient` stamps `X-Organization-Id: activeOrgId` on every request, so
`/me/permissions` answers for whatever is active. Meanwhile the organizations
list opens any org's members page through an overlay link, and *switching* is a
separate button that navigates to the dashboard — so the ordinary route to
another organization's members page is the one that leaves the active
organization behind. The page then judged Acme's members by the caller's role in
Globex.

Not only the role picker #1028 added: `canManage` decides whether any role
control, invite button or spending field renders at all, and it read the same
wrong answer long before that.

## Why the URL wins, rather than the permissions request taking an org id

The issue named two candidate shapes. The second one — ask `/me/permissions` for
the organization on screen — **cannot fix this page**, and that is what decided
it: not everything the page reads is path-scoped. The spending section renders
`/spend`, which has no organization in its path at all, so an org's cap would go
on sitting beside another org's month-to-date figure. One tenant per page is the
only answer that covers the page; two notions of "current" is the defect itself.

## Where it lives, and why there

`ActiveOrgGuard` — which the dashboard layout already renders **before**
`{children}`, and which already owns the tenant cache reset. Three things follow
from that placement, and each of them is the reason for it:

- **It is race-free.** A layout effect there runs before the page's own passive
  effects, which is where react-query starts a request and therefore when the
  header is read. So the selection is already the URL's by the time the page asks
  anything.
- **The cache reset is given the *effective* tenant** — the URL's organization,
  not the stored one — so a direct load records it as the tenant the cache
  belongs to, instead of dropping the page's first requests one commit later.
  That is the failure the hook's first-tenant skip exists for, and it would
  otherwise have been reintroduced from the other side.
- **A page does not do it for itself.** One place, as the recovery already is,
  for the reason its own docstring gives: `usePermissions` is rendered by a dozen
  components at once.

A refused organization is not adopted from a URL: opening its page would hand the
selection back to the one the recovery has just moved off, which is how a switch
loop starts.

`organizationInPath` takes a UUID and nothing else, so a later `/orgs/new` is not
adopted as a tenant id and refused on every request the page makes. It tolerates a
locale prefix because it reads `next/navigation`'s pathname, which keeps one —
reading, not navigating, so `.claude/rules/frontend.md`'s rule against that module
still holds.

## What this changes for a person

Opening another organization's members page **switches to that organization** —
the sidebar, the switcher and the dashboard follow. That is the intended
semantics rather than a side effect: "I am looking at Acme" and "Acme is my
organization" being two different things is the root cause here.

## How it was verified

- Four tests on the parser: a members page, a roles page, a locale prefix, `/orgs`,
  an unrelated path, and `/orgs/new`.
- Three on the adoption: the switch, a page that names no organization, and a
  refused one left alone. Plus one that the cache a directly loaded page filled
  survives.
- The one that matters most, on the page itself: with the permissions mock
  answering **per organization**, an Owner of the org in the URL is offered an
  Owner's five roles while an Admin is the active organization.
- All four of the behavioural ones fail with the adoption removed — I checked
  that rather than assuming it.
- `make check` green, `make test-frontend-cov` at 5301 passed and 100%
  lines/statements/functions, `make docs-build`.

## Docs

`docs/permissions.md` gains the rule under "Who may hand out which role", and
`.claude/rules/frontend.md` gains it beside the picker rule from #1028 — both say
where the adoption lives and that a page does not do it itself.

Closes #1032

### DIFF
--- a/.claude/rules/frontend.md
+++ b/.claude/rules/frontend.md
@@ -252,6 +252,20 @@ control the caller may not use is **not rendered** — not rendered and then 403
 Prove it with an integration test (`*.integration.test.tsx`, Testing Library against a
 mocked API), not with Playwright.
 
+**Which roles a picker offers is arithmetic, not a list.** `assignableRoles` in
+`lib/assignable-roles.ts` mirrors the server's `assignable_roles` over the catalog's
+permission sets: a role is offered only when the caller's own strictly outranks it.
+A list of "every role but owner" is what offered an Admin the Admin option and 403'd
+after the email address was typed (#1028).
+
+**And a page that names an organization in its URL *is* that organization.**
+`apiClient` stamps `X-Organization-Id: activeOrgId` on every request, so a page acting
+on the org in its path while reading permissions for the active one decides Acme's
+members by the caller's role in Globex — which is what `/orgs/{id}/members` did, because
+the organizations list opens it without switching (#1032). The adoption lives in
+`ActiveOrgGuard`, once, beside the tenant cache reset it has to happen before; a page
+does not do it for itself.
+
 ## An empty page is ambiguous
 
 Every dashboard page fans out to several queries and renders its empty state when one
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -147,6 +147,14 @@ Derived from the catalog rather than from a role name, so a custom role (Phase
 against the literal `"admin"` and could not see one at all - on the invitation
 paths as well as on `change_role`, which is what #696 closed.
 
+**A page's organization is the one in its URL.** `X-Organization-Id` travels on
+every request from the console and names the *active* organization, so a page
+that acts on an org from its path - `/orgs/{id}/members` - has two notions of
+"which tenant" and the organizations list opens that page without switching. They
+are one now: the dashboard's `ActiveOrgGuard` adopts the organization a path
+names, before the page asks anything, so what a caller may do there is what they
+may do *there* (#1032).
+
 **The console computes the same relation rather than being told it.** Every role
 picker - the two invite dialogs and the members table - offers what
 `assignableRoles` in `frontend/src/lib/assignable-roles.ts` answers, over the
--- a/frontend/src/app/[locale]/(dashboard)/orgs/[id]/members/members.integration.test.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/orgs/[id]/members/members.integration.test.tsx
@@ -5,8 +5,9 @@ import { Suspense, type ReactNode } from "react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import MembersPage from "./page";
+import { ActiveOrgGuard } from "@/components/layout/active-org-guard";
 import { apiClient } from "@/lib/api-client";
-import { useAuthStore } from "@/stores";
+import { useAuthStore, useOrgStore } from "@/stores";
 import { permissionsOf, ROLE_CATALOG } from "@/test-utils/role-catalog";
 
 /**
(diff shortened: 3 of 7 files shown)


DEENUU1:
Both real, both fixed in bb4802f — and the second one was mine.

**The organization switcher was inoperative on two pages.** Exactly as described: keyed on the selection, the adoption wrote back whatever else moved it, and `OrgSwitcher` sets the id without navigating — so on `/orgs/{B}/members` the store went to A and was snapped to B before the menu had closed. The product's primary switcher, silently dead on the two pages this branch is about.

Both halves you named are needed, and each is wrong alone:

- **Adoption happens once per path**, because that is what a *navigation* means. Keyed on the selection it fights every other writer.
- **A switch on a scoped route pushes the same page for the organization picked.** Adoption-on-path-change alone would leave the switch standing and the page still acting on the organization just left — the mismatch this branch exists to remove, re-entered by hand.

Together: pick Globex on Acme's members page → land on Globex's members page, nothing snapped back. Three tests on the switcher, which had none, and one on the hook for the deliberate switch standing.

**The casing** — right, and worse than a cosmetic mismatch: it is *stored*, so an upper-case spelling matches no organization in the list (`activeOrg` is `===`), the switcher falls back to showing the first one while every request carries another tenant, and `refusesOrganization` cannot recover it because it compares those same two strings. Lower-cased at the parse, with a test.

One thing I took while I was there, and flagging it rather than burying it: `OrgSwitcher` moves onto `@/lib/locale-navigation` instead of mixing two routers in one component. Its `usePathname` returns the path without the locale prefix, so the pattern does not have to know about one, and its router puts the prefix back — which the component's two existing `/orgs` pushes were losing, sending a Polish reader to the English page.

`make test-frontend-cov` at 5306 passed, 100% lines/statements/functions.

PR #1037, opened 2026-08-20
fix(vault): seal a multi-column row under one key version

### LINKED ISSUE #552: Vault 'seal N fields under one version' is hand-rolled at 4 models; agent_embed has no key-version column
The shape "a row has several ciphertext columns sharing one `key_version`" is hand-rolled four times, each differently: `channel_bot.py:224-235` (`_seal_at`, resets version — the AUD-008 bug), `mcp_connection.py:114-117` (`_seal_for`, honours version) mixed with bare `seal(...)` at `:295,337,516,626,682`, `agent_embed.py:318-321` (`_seal`, **discards `sealed.key_version`**, and `AgentEmbed` has no version column), `organization_secret.py:185-200` (clean). The vault (`core/vault.py`) offers only `seal`/`unseal`/`rewrap` — no multi-field primitive.

**How it fails.** All latent — `rewrap` (master-key rotation) has no production caller. The day it does: `agent_embed._verify_token` unseals at the implicit v1 (`:285`), so a rotated embed row can never be opened → every `jwt` widget fails `EmbedDenied("token rejected")`, with no column to bump.

**Fix.** Add `seal_fields(values, *, scope, key_version) -> (ciphertexts, version)` to the vault; convert all four sites; add a `key_version` column to `AgentEmbed` (migration + backfill to 1). The helper makes "no version column" and "reset to 1" unrepresentable.

**Done when.** One vault helper seals multi-field rows; `AgentEmbed` records/reads its version; a test rotates v1→v2 via `rewrap` and round-trips a widget token.

Extends AUD-008 ( from the 2026-08-01 audit — the channel_bot reset is one instance of this shape).

---

OchnikBartek:
## What breaks

"A row has several ciphertext columns sharing one `key_version`" was hand-rolled at several models, each differently — and the vault (`seal`/`unseal`/`rewrap`) offered no primitive for it. The failures are **latent** (`rewrap`, master-key rotation, has no production caller yet), but the day it runs:

- **agent_embed** has *no* `key_version` column and `_verify_token` unseals at an implicit v1 — so a rotated `jwt` widget can never be opened again (`EmbedDenied` for every visitor).
- **channel_bot**'s `update` re-sealed a changed token at the default v1 and reset the column, while siblings kept the rotated version — the row's version disagreeing with its envelopes (AUD-008).

## Fix

- **`vault.seal_fields(values, *, scope, key_version) -> (dict, version)`** — seals every field at one version and hands that version back to store. The one way to seal such a row, so "seal at v2 but record v1" and "no version column" cannot be written by hand.
- **agent_embed** gains a `secret_key_version` column (migration backfills existing rows to 1), seals through `seal_fields`, and unseals at the row's version. `docs/secrets.md` now lists the embed among the sealed rows.
- **channel_bot** `update` seals the token at the row's existing version, beside its siblings, and never resets the column.

**Left as-is, deliberately:** `mcp_connection` already seals one field per write at the row's `secret_key_version` (no reset, no bug), and `organization_secret` stores its `(ciphertext, hint, key_version)` triplet through the typed `secret_kinds` wrappers — both already record one version per row, so routing their single/typed fields through the multi-field helper would be churn, not a fix. Happy to converge them if you'd prefer.

## Verification

- `seal_fields` round-trips, skips empty fields, and a whole row rewraps together (`test_vault.py`).
- An embed secret sealed at v1, **rewrapped to v2** with the row's version bumped, still verifies a visitor token — the round trip that fails against the old implicit-v1 read.
- A channel-bot token update on a v2 bot keeps v2 and seals the new token there.
- Migration cycles **up → down → up** cleanly and `alembic check` reports no drift; full backend suite 5683 passed (`vault.py` gated, 100%); `make lint` clean.

Closes #552

### DIFF
--- a/backend/alembic/versions/0044_agent_embed_key_version.py
+++ b/backend/alembic/versions/0044_agent_embed_key_version.py
@@ -1,0 +1,40 @@
+"""An embed records which master-key version sealed its signing secret.
+
+`agent_embeds.jwt_secret_encrypted` was sealed by the vault but the row kept no
+`key_version`, so the verifier unsealed at an implicit v1. That is a latent bug:
+the day a master-key rotation runs `rewrap` over the vault, every `jwt` widget's
+secret moves to v2 and can never be opened again - `EmbedDenied("token
+rejected")` for every visitor, with no column to bump (#552).
+
+Every existing envelope was sealed at v1, so the column backfills to 1 with a
+server default; the default is then dropped, because the model carries no
+server default - a new row's version comes from the code that seals it.
+
+Revision ID: 0044_agent_embed_key_version
+Revises: 0043_rag_document_source_path
+Create Date: 2026-08-20
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "0044_agent_embed_key_version"
+down_revision: str | None = "0043_rag_document_source_path"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "agent_embeds",
+        sa.Column("secret_key_version", sa.Integer(), nullable=False, server_default="1"),
+    )
+    op.alter_column("agent_embeds", "secret_key_version", server_default=None)
+
+
+def downgrade() -> None:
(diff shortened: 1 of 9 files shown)


OchnikBartek:
@codex review

PR #1041, opened 2026-08-20
fix(chat): persist and replay an answered ask_user question

OchnikBartek:
Closes part of #502.

## What changed

A mid-turn `ask_user` question and the person's answer were written down
nowhere, so a reopened conversation showed neither the question the agent
put nor the answer it acted on. `ask_user` is a callback, not a tool, so
it never touched the turn timeline the transcript is replayed from.

- **Backend** — `MessagePart` gains an `ask_user` kind carrying the
  question and answer; `TurnTimeline.add_ask_user` records it; the session
  hands the running turn's timeline to `_ask_one`, which appends the pair
  once the answer is in hand, in the position it happened. A lone
  `ask_user` part is stored even as a turn's only part — unlike a tool
  call or a block of text it has no column to fall back to. No migration:
  the timeline is already a JSONB column and only the part union widens.
- **Frontend** — the raw/typed part shapes, the replay in
  `conversation-to-chat`, a `runsOf` run, and an `AskUserBlock` that draws
  the question and answer as a step in the turn rather than a chat bubble.

## Scope

Replay only: live, the question is the composer's own `ask_user` form and
the answer returns through it, so this is what a conversation reopened
from history was missing.

**Delegate attribution — the issue's "say which delegate asked" — is not
in this PR, on purpose.** `subagents-pydantic-ai` reaches the parent
through `ctx.deps.ask_user(question, [])` with no asker name, and
`SubAgentState` carries none, so naming the delegate needs an upstream
change to that package first. Filed as a follow-up; #502 stays open for it.

## Verified

- `TurnTimeline` records the pair and stores it even alone; `_ask_one`
  writes it onto the turn and is safe to reach between turns.
- The stored pair round-trips through `conversationMessageToChatMessage`,
  `runsOf` keeps a lone `ask_user` part as its own run, and the transcript
  renders the question and answer.
- Both 100%-gated backend modules stay at 100%; `make lint`,
  `make test-frontend-cov` green. Backend `test` coverage is at 100% for
  the touched modules; the only misses are integration-covered modules
  this branch does not touch, skipped locally for want of a DB (CI has
  one).

### DIFF
--- a/backend/app/schemas/conversation.py
+++ b/backend/app/schemas/conversation.py
@@ -33,6 +33,12 @@ class MessagePart(BaseSchema):
     `tool_call_id` refers to the `tool_calls` row carrying the arguments and the
     result. It is not repeated here - a tool call is written once, and a copy in
     JSONB is a copy that disagrees the first time one is re-run or redacted.
+
+    An `ask_user` entry is the exception that stores its own payload: a mid-turn
+    question and the person's answer have no column of their own - unlike text
+    (`content`), reasoning (`thinking`) or a tool call (its `tool_calls` row) - so
+    without this entry a reopened conversation shows neither the question the agent
+    put nor the answer it acted on (#502).
     """
 
     # The one schema in this module that must not strip its strings. `BaseSchema`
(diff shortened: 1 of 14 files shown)


OchnikBartek:
@codex review

Issue #943, opened 2026-08-19
Impersonation mints a bearer token onto the clipboard, and nothing it does is attributable to the admin


DEENUU1:
**Impersonate** on `/admin/users` mints a one-hour access token for another
person's account and copies it to the operating system clipboard. Nothing is
started, nothing can be ended, and every request made with it is attributed to
the target with no trace of who was actually acting.

## What happens

`POST /admin/users/{id}/impersonate`
(`backend/app/api/routes/v1/admin_users.py:90`):

```python
token = create_access_token(subject=str(target.id), expires_delta=timedelta(hours=1))
```

The payload is `{exp, sub, type}` - the platform's ordinary access token, whose
`sub` is the target. The frontend then writes it to the clipboard and says so:

```ts
const token = await onImpersonate(subject.id);
await navigator.clipboard.writeText(token);
toast.success(t("impersonationTokenCopiedValid"));
```
(`user-detail-drawer.tsx:94`)

## Three problems, in order of seriousness

**1. An impersonated action is indistinguishable from the person's own.** The
token carries no actor claim, so every row it writes, every run it starts and
every audit entry it triggers names the target. `record_audit` logs
`admin.user.impersonate` at issue time - that a token was *created* - and nothing
after it. An admin who reads a customer's private conversation and one who deletes
their agent leave the same trace: none. This is the one that matters, and it is
also the cheapest to fix: an `act` claim on the token, read wherever the audit
actor is resolved.

**2. It cannot be ended.** There is no revocation - the token is valid for its
full hour whatever anybody does, including the target changing their password or
the admin closing the tab. The product has session rows and a `DELETE /sessions`
that revokes them; an impersonation is the one credential that outlives every
control the platform has.

**3. It is a bearer token in the OS clipboard.** It survives the tab, is readable
by anything the operator pastes into, and is likely to end up in a paste buffer
history. The admin then has to do something manual with it - the product does not
say what - so the flow is "here is a credential, good luck".

## What it should be

An impersonated **session**, not a token on the clipboard:

- the admin's browser starts acting as the target, from the product, with a
  persistent banner naming who they are impersonating and an **End impersonation**
  button;
- the credential carries the admin's id as an actor claim, and everything written
  during it records both - "X, acting as Y";
- it is revocable, and it ends when the admin ends it or the window expires,
  whichever is first;
- `docs/deployment.md` (or `governance.md`, wherever the audit trail is described)
  says what impersonation records, because an operator answering a customer's "who
[...]

PR #1043, opened 2026-08-20
fix(audit): make an impersonated action name who was really acting

OchnikBartek:
Addresses the actor-attribution half of #943 — the issue's stated most
important and cheapest piece.

## The problem

`POST /admin/users/{id}/impersonate` mints an access token whose `sub` is
the **target** account. Every request made with it — every row written,
every audit entry triggered — was attributed to the target and nobody
else. An admin who read a customer's conversation and one who deleted
their agent left the same trace: the customer's own. "Who accessed my
account" had no answer.

## What changed

- **`act` claim.** `create_access_token(..., act=...)` carries the
  administrator behind the subject; the impersonate route sets it. Absent
  on every ordinary token, so those are byte-for-byte unchanged.
- **Audit context.** The auth dependency (HTTP and WebSocket) reads `act`
  onto a request-scoped context variable — the actor behind a request is a
  property of the request, not something to thread through every service
  that records an action.
- **`record_audit`** writes it as `impersonator_user_id` beside the actor;
  `0044_audit_impersonator` adds the nullable column and index; the audit
  read schema/service expose it. Null on an ordinary request; nothing
  backfilled (whether a past action was impersonated is unknowable after
  the fact).
- **`docs/governance.md`** says what an impersonated action records.

## Scope — deliberately not the whole issue

Done-when #2 (an impersonated request is identifiable; the audit says who
acted as whom) and #4 (a test asserts the claim survives into an audit
entry) are here. **#1 (no raw token on the clipboard) and #3 (revocable,
endable impersonation)** are one larger full-stack flow — an in-product
impersonation session with a banner and an End button, plus revocation —
filed as a follow-up. #943 stays open for it. The issue also flags a
policy question (whether the target is notified); that belongs with the
flow, not this change.

## Verified

- The claim survives from a minted token through the auth context into the
  entry; an ordinary action and an admin acting as themselves record no
  impersonator; the read service reports it.
- Migration up/down/up clean, `alembic check` finds no drift.
- Full backend suite green (5688 passed); `make lint-backend`,
  `make docs-build` green. Backend `test` coverage is 100% on the touched
  gated module (`services/audit.py`); the core auth/audit modules changed
  here are not in the gate.

### DIFF
--- a/backend/alembic/versions/0044_audit_impersonator.py
+++ b/backend/alembic/versions/0044_audit_impersonator.py
@@ -1,0 +1,51 @@
+"""An audit entry records who was acting behind the actor.
+
+An impersonated request is attributed to the account it acts *as* - its token's
+`sub` is the target - so every row it writes and every action it records named
+the target and nobody else. An administrator who read a customer's conversation
+and one who deleted their agent left the same trace: the customer's own.
+
+So the actor behind the actor is stored. `impersonator_user_id` is the
+administrator an `act` claim on the token names, written by `record_audit` from
+the request's audit context (#943). Null on an ordinary request, where nobody is
+acting as anybody else, and nothing is backfilled - there is no way to know after
+the fact whether a past action was impersonated, and guessing one would be a
+false accusation rather than a missing one.
+
+Revision ID: 0044_audit_impersonator
+Revises: 0043_rag_document_source_path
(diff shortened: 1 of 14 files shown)


OchnikBartek:
@codex review

PR #1081, opened 2026-08-24
fix(auth): deliver the OAuth sign-in via a single-use code

### LINKED ISSUE #14: Stop putting tokens in the OAuth redirect URL

OchnikBartek:
## What changed

The Google OAuth callback redirected the browser to the frontend with the
access and refresh tokens in the query string. That URL reaches the address
bar and session history, the frontend server's access log and any reverse
proxy in front of it, and the `Referer` of the next same-origin request the
callback page makes (`Referrer-Policy: strict-origin-when-cross-origin` sends
the full URL same-origin). The refresh token is valid for a week, so a reader
of any access log had a week of full account access.

The callback now hands out a **single-use, one-minute code** and keeps the
token pair in Redis. A new server-to-server endpoint, `POST /api/v1/oauth/exchange`,
redeems it with `GETDEL`, so a replayed, expired, or forged code all redeem to
nothing and answer 401. The frontend BFF swaps the code for the pair, verifies
the access token against `/auth/me`, and moves both into HttpOnly cookies — the
tokens never touch a URL. This also closes the session-fixation shape the issue
flagged, since the BFF no longer accepts a client-supplied token pair.

## Verification

- Backend: `tests/test_oauth_signin_exchange.py` — the redirect Location carries
  a `code` and neither token; a code redeems its pair exactly once; an unknown
  code is refused with 401.
- Frontend: `session-routes.test.ts` and `session-adoption.integration.test.tsx`
  updated to the code contract, including that the callback page posts `{ code }`
  rather than the token pair.
- `docs/configuration.md` now records the token-delivery decision under the
  OAuth section.

Closes #14

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -83,6 +83,7 @@ Redis = Annotated[RedisClient, Depends(get_redis)]
 
 from app.services.user import UserService
 from app.services.session import SessionService
+from app.services.oauth_exchange import OAuthExchangeService
 from app.services.conversation import ConversationService
 from app.services.sandbox_connection import SandboxConnectionService
 from app.services.sandbox_workspace import SandboxWorkspaceService
@@ -99,8 +100,14 @@ def get_session_service(db: DBSession) -> SessionService:
     return SessionService(db)
 
 
+def get_oauth_exchange_service(redis: Redis) -> OAuthExchangeService:
+    """Create OAuthExchangeService instance with the Redis client."""
+    return OAuthExchangeService(redis)
+
+
 UserSvc = Annotated[UserService, Depends(get_user_service)]
 SessionSvc = Annotated[SessionService, Depends(get_session_service)]
+OAuthExchangeSvc = Annotated[OAuthExchangeService, Depends(get_oauth_exchange_service)]
 
 
 def get_conversation_service(db: DBSession) -> ConversationService:
--- a/backend/app/api/routes/v1/oauth.py
+++ b/backend/app/api/routes/v1/oauth.py
@@ -1,15 +1,18 @@
 """OAuth2 authentication routes."""
 
 import logging
+from typing import Any
 from urllib.parse import urlencode
 
 from fastapi import APIRouter, Request
 from fastapi.responses import RedirectResponse
 
-from app.api.deps import UserSvc
+from app.api.deps import OAuthExchangeSvc, UserSvc
 from app.core.config import settings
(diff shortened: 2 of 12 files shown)


OchnikBartek:
@codex review

Issue #1080, opened 2026-08-23
Dev-stack watchdog kills a healthy reloaded server and leaves nothing serving


OchnikBartek:
## What breaks

On the dev stack (`agenticos_backend:dev`, uvicorn `--reload`), the event-loop watchdog killed a **healthy, serving** process twice in one session, and after one kill the reloader never spawned a replacement — the container stayed `Up (unhealthy)` serving nothing until a manual `docker restart`.

The smoking log — a 200 answered one second before the watchdog claims 24 seconds of silence:

```
INFO:     Application startup complete.
22:08:17  POST /api/v1/auth/login
22:08:19    COMMIT;
INFO:     192.168.65.1:26479 - "POST /api/v1/auth/login HTTP/1.1" 200 OK
ERROR:    Server process [8] has not run its event loop for 22s. Killing it and starting a replacement.
```

A loop that answers a login did run in the last 22 seconds, so the watchdog's beat is what went stale — the beat task appears not to be running in a server process created by a WatchFiles reload (both incidents followed `WatchFiles detected changes ... Reloading`). A third shape appeared on `docker restart`: `Server process [798] never ran its event loop and ignored SIGTERM for 1s. Killing it.` during ordinary cold start.

## Sequence

1. `docker compose up` the dev stack.
2. Edit any backend file so WatchFiles reloads the server.
3. Wait ~`EVENT_LOOP_WEDGED_AFTER` seconds after `Application startup complete`, keep requests flowing.
4. The watchdog kills the serving process; sometimes no replacement follows and the container serves nothing while `docker ps` says Up.

Workaround used: `EVENT_LOOP_WEDGED_AFTER=0` on the smoke stack.

## Where

- `backend/app/core/watchdog.py` — the beat and the verdict.
- `backend/app/main.py` — `watchdog.start()` before the lifespan yield; the reload path re-enters this in the new process.

## How you'd know it is fixed

A dev-stack server reloaded by WatchFiles survives past the wedge threshold under request load, and a killed process is always replaced (or the container exits so Docker's restart policy replaces it — a kill that leaves a running-but-dead container is the worst of both).

Pre-existing; found standing up #12's smoke stack. Related: #358 (the watchdog's own origin). Checked #168 — no existing issue covers a watchdog false positive; nearest is the reload-supervisor work the watchdog docstring cites.

Issue #1079, opened 2026-08-23
Knowledge singleton's pooled engine can cross worker event loops


OchnikBartek:
## What breaks (suspected — needs a worker repro)

The knowledge capability's retrieval singleton (`app/agents/capabilities/knowledge/_search.py`) caches a `PgVectorStore` for the life of the process, riding the process's pooled engine (`app/db/session.py`). Agents also run **inside the Prefect worker** — a trigger fire goes through `AgentTriggerService` → `AgentRunnerService.execute` — so an agent with the knowledge capability reaches this singleton from whatever event loop its flow runs in.

A pooled asyncpg connection is bound to the loop that created it. Wherever two flow runs share one worker process but not one event loop (e.g. `check_scheduled_syncs_flow` fans out sub-flows with `asyncio.create_task` in-process, while `serve()` normally gives each flow a subprocess), the second search can check out a connection made on the first loop and fail with `InterfaceError: another operation is in progress` / "attached to a different loop" — intermittent, unreproducible in any single-loop test.

`reset_retrieval_service()` is called only from the API lifespan; nothing resets the cache in the worker. The same class of hazard exists for `get_db_context()` used inside flows (`_keep_lease_alive`, the trigger recovery's corroborating read).

## Where

- `backend/app/agents/capabilities/knowledge/_search.py` — the singleton and its `process_vector_store` construction.
- `backend/app/db/session.py` — the module engine; `get_worker_db_context`'s docstring states the cross-loop rule the singleton sidesteps.

## How you'd know it is fixed

A worker that fires two knowledge-searching trigger runs on two different event loops in one process searches successfully both times. Either the singleton becomes per-loop / the worker path builds a per-flow store (the way `rag_tasks._ingestion_service` does), or the process is shown to guarantee one loop per flow run and the docstrings say so.

Pre-existing: the old private pooled engine in the singleton had identical exposure, so #12's engine injection neither introduced nor fixed it. Found reviewing #12. Related: #12, #537, #948. Checked #168 — nearest cluster is the #12/#948 connection-lifecycle work; no existing issue names the worker's cross-loop hazard.

DEENUU1:
Still open, and PR #1128's approach is superseded rather than nearly-done — the
detail is in [my comment there](https://github.com/vstorm-co/agenticos/pull/1128#issuecomment-placeholder).

Short version: #1128 keys the cached retrieval service on the running loop while
building a store with its own pool, and #948/#12 removed the private pool — the
[...]

Issue #948, opened 2026-08-19
Dispose the vector store's engine, or stop building one per ingested document


DEENUU1:
`PgVectorStore.__init__` creates its own pooled SQLAlchemy engine. The ingestion
worker builds one per document and disposes none.

**Severity: High** · AUD-002, audit of 2026-08-19

## Where

- `backend/app/worker/tasks/rag_tasks.py:127` — `_ingestion_service_for`, called
  from `_run_ingestion:277`, `_run_sync:349` and `_run_source_sync:557`
- `backend/app/services/rag/vectorstore.py:225` — `create_async_engine` in
  `__init__`; `:227` — `aclose`

`aclose()` has exactly one caller in the whole repository: `main.py:170`, on the
store the API's lifespan built. Every other construction site abandons its
engine.

The class docstring already says this is wrong:

> This class creates its own SQLAlchemy engine per instance. In production,
> prefer injecting a shared engine from app.db.session to avoid multiple
> connection pools. Call `await self.aclose()` on shutdown to release pool
> connections.

## How it fails

Ingest 200 documents, or sync a Drive folder holding 200 files:
`_ingestion_service_for` is called per document, so 200 pooled engines are
created and abandoned. Each holds its checked-in connections open until the
process exits — the default pool is 5 plus 10 overflow and connections are
created lazily, so the realistic cost is one to two live Postgres connections per
abandoned engine.

Somewhere between 50 and 100 documents the worker reaches `max_connections`, and
every query after that raises — including the ones that would mark a document
failed. The symptom is an upload stuck at `processing` with a connection error in
the worker log, which is indistinguishable from four other failure modes.

## The correct pattern is in the module next door

`app/db/session.py:95` — `get_worker_db_context` creates an engine per call *with
`NullPool`* and disposes it in a `finally`, and its docstring says why:

> Creates a fresh engine with NullPool on every call so there are no cross-fork /
> cross-event-loop connection issues. The engine is disposed automatically when
> the context manager exits.

Two engine-per-call sites in one worker process, one careful and one not.

## Fix

Either shape works; the first is smaller.

```python
# Option A - dispose what you built. `_ingestion_service_for` becomes a context
# manager and the three callers `async with` it.
@asynccontextmanager
async def _ingestion_service_for(...) -> AsyncIterator[IngestionService]:
    store = VectorStore(...)
    try:
        yield IngestionService(processor=..., vector_store=store)
    finally:
        await store.aclose()

[...]

PR #1084, opened 2026-08-24
fix(agents): stop holding a pooled DB connection across slow work

### LINKED ISSUE #3: Commit the run row so a failed run is still accounted
**Severity: High** · AUD-003

## What's wrong

`AgentRunnerService._run` records the run in a `finally` and then re-raises. `finish` flushes; nothing commits; `_managed_session` rolls back on the way out. The run row created at `agent_runner.py:340` goes with it.

`agent_chat.py` — the one surface that gets this right — documents the exact defect:

> Committed here rather than left to the session context: that exit rolls back on any exception, and cancellation never reaches it at all, since `CancelledError` is not an `Exception`. A run that failed, was stopped or ran out of budget still spent money, and a run missing from history is a run nobody is accountable for.

`grep '\.commit()' app/` returns exactly one non-infrastructure caller: that line.

## Where

- `backend/app/services/agent_runner.py:754-765`
- Contrast `backend/app/services/agent_chat.py:286-299`
- Mechanism at `backend/app/db/session.py:34-44`

## How it fails

POST `/api/v1/agents/{id}/run` against an agent whose provider 500s after three model steps. `finish()` flushes `cost_usd = 0.42`; the `RuntimeError` propagates; the session rolls back. There is no run in history, no cost, and `sum_cost_since` — the budget baseline for the *next* run — is unchanged. An agent failing in a retry loop spends unbounded money that no budget can see.

Affects the public API, all three channel surfaces (`mentions.py:198,257`), the embed widget (`embed_session.py:160`) and `resume`. Web chat alone is safe.

`asyncio.CancelledError` is worse: not an `Exception`, so `_run` records `FAILED` rather than `CANCELLED` and then unwinds past the session context entirely.

Verified against a real Postgres — a flushed-but-uncommitted row is gone after the body raises.

`docs/governance.md:41-46` states the opposite: *"Accounting happens in a `finally` block on every surface, and the commit is explicit rather than left to the session context."*

## A second, sharper consequence

A run parks on an approved `send_invoice`, someone approves, `resume` replays the call — the invoice is sent — and the *next* model request times out. The rollback discards `mark_running` too, so the run is `awaiting_approval` again with its original `paused_state` and an approval still marked `approved`. A second resume produces identical `DeferredToolResults` and **sends the invoice again**. `claim_parked_run`'s `FOR UPDATE` protects against simultaneous resumes; it protects against nothing here, because the state transition it guards is thrown away.

## Fix

Move the commit into `AgentRunnerService.finish` (or `_run`'s `finally`, after `finish`), give `_run` the `except asyncio.CancelledError: status = RunStatus.CANCELLED; raise` clause `agent_chat.py:268-272` already has, and delete the duplicate from `agent_chat.py`. Consider committing `mark_running` in its own transaction so the "out of the queue before anything is replayed" comment at `agent_runner.py:572-575` becomes true across a crash.

## Acceptance criteria

- [ ] Every surface's failed run appears in history with its cost
- [ ] A cancelled run records `CANCELLED`
- [ ] Regression test (integration, real Postgres): run an agent whose model raises, then in a **fresh session** assert the `agent_runs` row exists with `status='failed'` and the recorded cost. `tests/test_agent_runner.py:513` asserts `finish_run` was *called* on an `AsyncMock` and cannot observe the rollback
- [ ] Regression test: park → approve → resume with a crash after the deferred call → the run is `failed`, and a second resume raises

### LINKED ISSUE #12: Stop holding a pooled DB connection across slow work
**Severity: High** · AUD-012

Three separate paths, one consequence: `DB_POOL_SIZE=5` + `DB_MAX_OVERFLOW=10` means fifteen concurrent anything exhausts the pool, and the sixteenth request of *any* kind — a health check, a login — blocks for `DB_POOL_TIMEOUT=30`s and then raises.

## 12a. The run's transaction spans the entire LLM call

`agent_runner.py:340` inserts and flushes the run row; the commit only happens after `agent.iter(...)` completes (`agent_chat.py:299`). For a multi-turn tool-calling run — seconds to minutes — one connection sits `idle in transaction`, holding an uncommitted row lock and blocking `VACUUM` from reclaiming across the window.

The fix for the run-commit issue makes the commit correct but does not shorten this. Split the boundary: open the run row in its own short session and commit it, run the model with no session held, re-open in `finish`. `PreparedRun` would carry `run_id` rather than a live `AgentRun`.

## 12b. The embed widget pins a connection per open browser tab

`embed.py:113` wraps `while True: await websocket.receive_json()` in `async with get_db_context()`. One connection per visitor, from handshake to close, including all the idle time between questions. Fifteen simultaneous visitors on a marketing page take the whole API down, not just the widget.

`AgentSession` (`agent_session.py:202`) does this correctly — a session per turn. Move `get_db_context()` inside the frame loop and give `EmbedSession` a factory rather than a session.

## 12c. `PgVectorStore` builds a private engine per instance and ingestion never disposes it

`vectorstore.py:211-216` calls `create_async_engine` in `__init__`, giving each instance its own `QueuePool` (5 + 10). `rag_tasks.py:53` constructs one per document. `aclose()` exists, is documented (*"Call `await self.aclose()` on shutdown"*), and `main.py:134` is its only caller — so a worker that ingests a few dozen documents leaks a few dozen pools and walks into Postgres `max_connections`. SQLAlchemy async engines do not close asyncpg connections on GC; they warn.

Have `PgVectorStore` accept an engine rather than create one.

## Acceptance criteria

- [ ] The run row is committed before the model call and visible from another session mid-run
- [ ] `EmbedSession` takes a factory; no connection is held between frames
- [ ] Ingesting two documents creates one engine, not two

### LINKED ISSUE #1078: Reap runs a dead process left durably running
## What breaks

Since the run row is committed before the model call (#12), a process death mid-run — SIGKILL, OOM, a deploy that does not drain — leaves a durable `agent_runs` row in `running` that nothing will ever finish. Three consequences, in increasing severity:

1. **Activity lies.** The run shows as `running` forever. Reproduced on a local stack: a dev-server kill mid-wave left 18 rows `running` until the next demo overwrote the view.
2. **Its spend is unrecorded.** The terminal write carries the tokens and cost; a crashed run's money is invisible to every budget and report.
3. **A trigger's schedule can wedge on a crashed resume.** `claim_due`'s `last_run_id` join blocks while the linked run is non-terminal. A resume commits `mark_running` before the replay (that durability is deliberate — it is what stops a second resume re-sending an approved call), so a crash mid-resume leaves the linked run `running` with no resolver, and the schedule never fires again. The conversation-level guard was already narrowed to `awaiting_approval` for exactly this reason (see `claim_due`'s docstring); the join is the remaining edge.

## Sequence

1. Fire a run on any surface (or resume a parked run in a trigger conversation).
2. `kill -9` the API/worker process while the model call is in flight.
3. `select status from agent_runs where id = ...` → `running`, forever; for the resume case, `claim_due` skips the trigger on every tick.

## Where

- `backend/app/services/agent_runner.py::_run` — the opening commit that makes the row durable (`#12`).
- `backend/app/repositories/agent_trigger.py::claim_due` — the `last_run_id` join over `_NON_TERMINAL_STATUSES`.

## Fix shape

A periodic sweep (the existing worker heartbeat is a natural home) that marks `running` rows `failed` after a liveness bound — e.g. `started_at` older than a generous ceiling, or a `last_seen_at` the runner stamps if per-run liveness is wanted. Once stale runs are reaped, `claim_due`'s conversation guard could safely block on `running` again, which would also restore schedule-vs-`run_now` overlap protection.

## How you'd know it is fixed

Kill a worker mid-run; within one sweep interval the row reads `failed` with an error naming the crash, and a trigger whose resume died claims again on the next tick.

Pre-existing in shape (a crash always lost the run — before #12 the row rolled back entirely and the spend was equally lost); the durable `running` remainder is new. Found working #12, reviewing the commit-boundary change against the trigger scheduler. Related: #12, #3, #537 (heartbeat & triggers). Checked #168 — belongs beside the #12 cluster; nothing existing covers a run-liveness reaper.

OchnikBartek:
## Summary

Three paths held a pooled Postgres connection across work measured in
seconds to minutes, so fifteen concurrent anything exhausted the pool
(`DB_POOL_SIZE=5 + DB_MAX_OVERFLOW=10`) and the sixteenth request of any
kind blocked for `DB_POOL_TIMEOUT` and raised. This branch closes what was
still open of #12, and with it the crash-replay half of #3:

- **12a — the run's transaction spanned the model call.** `_run` and
  `ChatAgentRunner.run` now commit once *before* the model is asked
  anything (the terminal commit from #231 stays). The run row is visible
  from every other session mid-run, the connection goes back to the pool
  for the duration of the call, and a resumed run's `mark_running` is
  durable before the approved call is replayed — a crash mid-replay can no
  longer hand the same approval out twice (#3's sharper consequence). The
  budget capability's baseline read moved onto a session of its own; read
  on the run's shared session it silently re-opened the idle transaction.
- **12b — the embed widget** was already fixed on main (#39):
  `EmbedSession` takes a session factory and opens one per turn. Verified,
  no change needed.
- **12c — `PgVectorStore` built a private pool per instance.** The store
  now borrows an injected engine and `aclose()` is gone. The API, CLI and
  the knowledge capability share the process engine through one
  `process_vector_store()` factory (the #306-shaped four-site repetition
  collapsed into it); the worker builds one engine per flow in
  `_ingestion_service`, disposed on every path out.
- **Consequence fix:** the agent-triggers scheduler (#537) reasoned from
  "an executing run's row is invisible". `claim_due`'s conversation guard
  now blocks on `awaiting_approval` alone (a crash-orphaned `running` row
  would wedge the schedule forever), and the fire-recovery branch
  corroborates a `running` tail against a fresh session before settling it
  as its own orphan — a concurrent fire's live run is left alone.
- **#1078 — the crash-orphan remainder, closed here too.** An hourly
  stale-run sweep ends anything still `running` past
  `STALE_RUN_REAPED_AFTER_HOURS` (6h default, 0 disables) as `failed`,
  with the sweep's own sentence on the row. One conditional UPDATE whose
  status guard re-evaluates under the row lock; a live run flipped anyway
  is flipped back by its own terminal write. No spend is invented, nobody
  is mailed, `awaiting_approval` is never touched — each documented.

## Related Issue

Closes #3
Closes #12
Closes #1078
Refs #948, #537
Follow-ups filed: #1079 (knowledge singleton cross-loop engine,
pre-existing), #1080 (dev watchdog false positive, pre-existing).

## Testing

- Unit: commit-before-model ordering on the execute, resume and chat
  paths; the trigger claim SQL and both recovery corroboration outcomes;
  engine-counting across all three ingestion flows and their failure
  paths; the store factory's two invariants; the reaper's window, off
  switch and controlled error sentence.
- Integration (real pgvector Postgres): the run row is readable from a
  second session mid-run (fails without the fix); a failed and a
  cancelled run are accounted on a fresh connection; exactly one of four
  seeded rows falls to the reap — the seven-hour `running` orphan.
- `make check` green; backend suite 6900+ tests at the 100% platform gate.
- Live stack: 18 concurrent runs against a 25s stub model — all 18 ran in
  one wave (was: 15 + three waiting 26s on the pool), `pg_stat_activity`
  showed zero `idle in transaction` during model calls, health p50 158ms,
  worst bystander read 9.9s at the finish burst (was 29s). Web chat,
  hosted embed page and KB upload→ingestion clicked through end to end;
  each surface's run visible as `running` mid-turn.

## Notes for Reviewers

- The opening commit is deliberately in `_run` + `ChatAgentRunner.run`
  rather than centralized in `PreparedRun.execute/iterate`:
  `PreparedRun` does not carry the session, and both sites are tested and
  documented. Weighed and noted.
- The per-flow worker engine keeps SQLAlchemy's default pool knobs, as
  the old in-store engine did; tying it to `DB_POOL_SIZE` would couple
  worker sizing to API tuning silently.
- API-process vector work now shares the request pool; with connections
  no longer held across model calls, 15 is comfortable at the measured
  load, but heavy-RAG deployments may want a bigger `DB_POOL_SIZE` —
  called out rather than changed.


OchnikBartek:
@codex review

OchnikBartek:
@codex review

PR #1086, opened 2026-08-24
perf(frontend): ship only the identity-provider marks to the auth pages

### LINKED ISSUE #955: Stop shipping the whole brand-glyph table to the sign-in page
The sign-in and register pages ship all 89 brand marks — 104 KB of source — to
draw three of them.

**Severity: Medium** · AUD-009, audit of 2026-08-19

## Where

- `frontend/src/lib/brand-glyphs.generated.ts` — 104 KB, 89 marks
- `frontend/src/components/icons/brand-icon.tsx:22` —
  `<GlyphIcon glyph={BRAND_GLYPHS[name]} ... />`
- `frontend/src/components/auth/oauth-buttons.tsx:1` — `"use client"`, rendered by
  `login-form.tsx` and `register-form.tsx`

## How it fails

`BrandIcon` reads the table by dynamic key, which no bundler can tree-shake:
every entry in the record is reachable, so every entry is emitted.
`oauth-buttons.tsx` is a client component and imports `BrandIcon` to draw three
marks — Google, GitHub and Microsoft.

So the first pages any visitor loads, and the only ones an unauthenticated person
sees, carry all 89 marks of raw SVG path data. Minified and gzipped that is
roughly 25-30 KB of JavaScript on the critical path of the page whose whole job is
to be fast, and it grows every time a mark is added for a connector nobody signs
in with.

## Not a criticism of the generator

`frontend.md` is explicit that a mark is a row in `scripts/gen-brand-icons.ts` and
never a hand-authored path, and #156 is what that rule cost to establish. This is
about how the output is consumed, not how it is produced.

## Fix

```ts
// Option A - emit one module per mark and let the import be static.
//   brand-glyphs/google.ts, brand-glyphs/github.ts, ...
//   `BrandIcon` keeps its dynamic API for tables; a component that knows its
//   three marks imports them directly.
// Option B - keep one module, and have the generator also emit a small
//   `auth-glyphs.ts` holding just the identity providers, imported by
//   oauth-buttons.
```

Option B is smaller and enough: the auth pages are the only place where the
whole-table import sits on a critical path. Everything else drawing marks is
behind sign-in, where the table is already in the shared chunk.

## Acceptance criteria

- [ ] The sign-in page's JS no longer contains marks it does not draw — checked
      against the build output rather than asserted
- [ ] The generator remains the only source of glyph data

OchnikBartek:
## What changed

The sign-in and register pages drew three marks (Google, GitHub, Microsoft)
through `BrandIcon`, which reads the 89-mark `BRAND_GLYPHS` table by dynamic key.
A dynamic record access can't be tree-shaken, so the whole table — ~104 KB of
source, 25–30 KB gzipped — shipped on the critical path of the only pages an
unauthenticated visitor loads, and grew with every connector nobody signs in
with.

The generator now also emits `auth-glyphs.generated.ts` with just the three
identity-provider marks, **derived from the same fetched data** so the generator
stays the single source of glyph data (the acceptance criterion). `oauth-buttons`
imports `AUTH_GLYPHS` and draws it through `GlyphIcon` directly, so the full table
no longer reaches the auth bundle.

## Verification

- New `oauth-buttons.test.tsx`: renders a link and a mark per configured
  provider, carries the invitation token on sign-up, renders nothing when
  unconfigured; and guards the regression — the auth component must not import
  `BrandIcon`/`brand-glyphs.generated`, and `AUTH_GLYPHS` holds exactly the three.
- `.claude/rules/frontend.md` updated to record the second emitted module.
- The generated `auth-glyphs.generated.ts` was produced offline from the current
  brand table (no network re-fetch), byte-shaped as `gen:brand-icons` emits it,
  so a future regenerate is a no-op diff.

Closes #955


OchnikBartek:
@codex review

PR #1088, opened 2026-08-24
test(frontend): complete the local next-intl mocks with t.rich

### LINKED ISSUE #612: Twelve local next-intl mocks model t but not t.rich
Found while sweeping copy for #395, not caused by it.

Twelve test files mock `next-intl` locally to get key-as-value behaviour, and each models
`useTranslations` as a bare function:

```tsx
vi.mock("next-intl", () => ({
  useTranslations:
    (ns: string) =>
    (key: string): string =>
      `${ns}.${key}`,
}));
```

`t` is not a bare function. It carries `t.rich`, `t.markup` and `t.has`, and a component
that reads a message with a tag in it calls `t.rich`. So the first component under one of
these mocks to gain a rich message fails as:

```
TypeError: t.rich is not a function
 ❯ IngestionDialog src/components/kb/ingestion-dialog.tsx:93:16
```

— a render-time throw inside a component several files from the assertion, in a spec that
is about something else entirely. That happened once already in #610
(`kb-detail-sections.integration.test.tsx`, five tests, none of them about ingestion
settings), and it was fixed there by teaching that one mock `rich`.

## The twelve still one component away from it

```
src/app/[locale]/(dashboard)/dashboard/dashboard.integration.test.tsx
src/app/[locale]/(dashboard)/rag/copy.integration.test.tsx
src/app/[locale]/(dashboard)/rag/rag-page.integration.test.tsx
src/app/[locale]/auth/session-adoption.integration.test.tsx
src/components/chat/rating-buttons.test.tsx
src/components/layout/active-org-guard.integration.test.tsx
src/components/layout/app-sidebar.test.tsx
src/components/layout/command-palette.test.tsx
src/components/layout/mobile-tab-bar.test.tsx
src/components/layout/sidebar-search.test.tsx
src/components/layout/sidebar-user.test.tsx
src/components/runs/spend-by-person.integration.test.tsx
```

All twelve are green today, because none of the components they render reads a rich
message yet. That is the point: the suite is not wrong, it is one `t.rich` away from being
confusingly wrong — and the guard in #610 now refuses a sentence split across an element,
so `t.rich` is the *remedy* it pushes people towards. These will be hit.

## Fix

Either give each mock a `rich` (one line, `Object.assign` over the function), or - better -
put the key-as-value translator in one shared test helper, so there is one definition of
"a mock translator" to keep complete. `vitest.setup.ts` already carries the global mock
that uses the real `createTranslator`; the local ones exist only to return keys, and its
docstring explains why a file's own mock wins.

- A component under any of those twelve reading `t.rich` renders rather than throwing.
- A spec asserts it, so the next mock added by copy-paste cannot lose it silently.

## Where it sits

Tests/CI, adjacent to the i18n cluster on #168 (#199 → #246 → #249 → #314 → #141 → #395).
Not a duplicate - grepped the open list for `t.rich` and `next-intl` and found nothing.
Found reviewing #395; unblocked, and cheap.

OchnikBartek:
## What changed

Twelve specs mocked `next-intl` as a translator missing part of `t`: eleven as a
bare `(key) => key`, and `kb-detail-sections` as a hand-rolled cache that had
grown `t.rich` (via #610) but still lacked `t.markup`/`t.has`. `t` carries all of
`rich`, `markup` and `has`, and a component reading a message with a tag calls
`t.rich`. All are green today only because none of their components reads a rich
message yet — the #395 guard now steers copy toward `t.rich`, so they will, and
then one throws `t.rich is not a function` inside a component several files from
the assertion, exactly as #610 hit once.

A shared `keyTranslations(format)` helper (`src/test-utils/intl.ts`) is now the
one definition of a key-returning translator, complete with `rich`/`markup`/`has`,
and each local mock uses it — preserving its own key shape (and, for kb-detail,
its per-namespace cache) so no assertion moves. The mock factory is `async` and
imports the helper inside it, because `vi.mock` is hoisted above the static
imports (a direct import reference throws "Cannot access … before
initialization"). `copy.integration` is untouched — it already builds a real
translator through `createTranslator`.

## Verification

- `intl.test.ts` asserts the helper carries `rich`/`markup`/`has`, so the next
  mock added by copy-paste cannot lose them silently.
- All twelve migrated specs + the helper test are green (116 tests).
- `make lint-frontend` clean.

Closes #612

### DIFF
--- a/frontend/src/app/[locale]/(dashboard)/dashboard/dashboard.integration.test.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/dashboard/dashboard.integration.test.tsx
@@ -28,10 +28,11 @@ vi.mock("@/lib/api-client", async () => {
 vi.mock("sonner", () => ({ toast: { success: vi.fn(), error: vi.fn() } }));
 
 // Full-key echo, so "dashboard.errors.title" is assertable and unambiguous.
-vi.mock("next-intl", () => ({
+vi.mock("next-intl", async () => ({
   useLocale: () => "en",
-  useTranslations: (namespace?: string) => (key: string) =>
+  useTranslations: (await import("@/test-utils/intl")).keyTranslations((namespace, key) =>
     namespace ? `${namespace}.${key}` : key,
+  ),
 }));
 
 const auth = {
--- a/frontend/src/app/[locale]/(dashboard)/rag/[id]/kb-detail-sections.integration.test.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/rag/[id]/kb-detail-sections.integration.test.tsx
@@ -23,22 +23,26 @@ vi.mock("sonner", () => ({ toast: { success: vi.fn(), error: vi.fn() } }));
 // runs `useEffect(() => refresh(), [refresh])`, so a translator rebuilt per call made
 // that effect re-fire forever and every test here timed out (#446).
 //
-// `rich` hangs off the same function, because a component under this tree reads a
-// message with a tag in it. A mock modelling only half of `t` fails as
-// `t.rich is not a function` inside a render, several files from the assertion.
-const translators = new Map<string, (key: string) => string>();
-vi.mock("next-intl", () => ({
-  useLocale: () => "en",
-  useTranslations: (ns: string) => {
-    const cached = translators.get(ns);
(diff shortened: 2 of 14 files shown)


OchnikBartek:
@codex review

PR #1094, opened 2026-08-24
fix(audit): fail closed when an audit write cannot be recorded

### LINKED ISSUE #20: Make an audit write failure loud, and type record_audit
**Severity: Medium** · AUD-018

## What's wrong

```python
# backend/app/core/audit.py:12-37
async def record_audit(db, *, actor_user_id: UUID, action: str, ...) -> None:
    """Persist an audit log entry. Failures are logged but do not raise."""
    try:
        ...
        db.add(entry)
        await db.flush()
    except Exception:
        logger.exception("Failed to write audit log for action=%s actor=%s", action, actor_user_id)
```

Three problems.

1. The `db` parameter is **untyped**, against `.claude/rules/code-style.md`.
2. The swallow is fail-open on the audit trail, which `docs/governance.md:149` makes load-bearing for the app-admin bypass story. It is called from the platform-layer services (`organization_secret`, `model_profile`, `agent_embed`, `agent_environment`, `agent_registry`) on ~20 privileged actions.
3. **The swallow does not achieve what it claims.** `flush()` inside the `try` leaves the session needing a rollback, so `_managed_session`'s commit then raises `PendingRollbackError` and the request 500s anyway — with an opaque error that names the session, not the audit. Either outcome is bad, and neither is "the action succeeds quietly".

`app/core/audit.py` is also outside both the coverage gate and the ty overrides in `backend/pyproject.toml`, despite every caller being inside them.

## Where

`backend/app/core/audit.py:12-37`

## Fix

Type the parameter, let the exception propagate (or catch narrowly and re-raise as a domain error), and add the module to both lists in `pyproject.toml`.

## Acceptance criteria

- [ ] `db: AsyncSession`
- [ ] An audit write failure is not silently discarded
- [ ] `app/core/audit.py` added to `[tool.coverage.run] include` and `[[tool.ty.overrides]] include`, verbatim and in the same order

OchnikBartek:
## What & why

`record_audit` swallowed every exception, which was **fail-open on the audit trail** — the trail `docs/governance.md` makes load-bearing for the app-admin bypass story. Worse, the swallow never even bought silence: `flush()` inside the `try` left the session needing a rollback, so the request's `scope="function"` commit then raised `PendingRollbackError` and 500'd anyway, with an opaque error naming the session rather than the audit.

The write now shares the caller's transaction: a failure propagates and rolls the recorded action back rather than letting a privileged mutation land unaudited. `db` is typed `AsyncSession` (it never was), and `app/core/audit.py` is added to both the coverage `include` and the ty overrides — every service that records an action was already gated; the trail they write to was not.

## Test fixtures the swallow was hiding

The fix surfaced three unit-test fixtures (`sync_source`, `sandbox_connection`, `skill_proposal`) built with a bare `MagicMock()` db, where `await db.flush()` raised `TypeError` inside `record_audit`. The old swallow hid it, so the audit write was a **silent no-op the green suite never noticed**. They now build the db the way every other audit-calling test already does (`db.flush = AsyncMock()`).

## Accepted trade-off (not a defect)

A failure specific to the audit row (a bad `details`, an FK on `organization_id`) now rolls the *action* back rather than being dropped — the intended atomicity per governance.md. The recorded `details` were reviewed and are all serialisable.

## Verification

- New `tests/test_audit_record.py`: fail-closed propagation (a regression that passed silently before — confirmed failing against `main`'s `audit.py`), the defensive `target_id` stringify, and the impersonator context. The impersonator-onto-entry behaviour stays covered by the existing `test_audit_impersonation.py`.
- `make lint` and `make test` green; **100% coverage** including the newly-gated module.
- Reviewed by a peer session (blast radius across 60+ callers, #353 transaction ordering, two-lists rule, docs) — APPROVE.

## Noted, out of scope

The `CLAUDE.md` doc trigger-map has no explicit `app/core/audit.py → governance.md` row (the concept already maps there). Left for a follow-up so a future `audit.py` editor finds the page.

Closes #20

### DIFF
--- a/backend/app/core/audit.py
+++ b/backend/app/core/audit.py
@@ -5,6 +5,8 @@ from contextvars import ContextVar
 from typing import Any
 from uuid import UUID
 
+from sqlalchemy.ext.asyncio import AsyncSession
+
 from app.db.models.audit_log import AppAdminAuditLog
 
 logger = logging.getLogger(__name__)
@@ -36,7 +38,7 @@ def current_impersonator() -> UUID | None:
 
 
 async def record_audit(
-    db,
+    db: AsyncSession,
     *,
     actor_user_id: UUID | None,
     action: str,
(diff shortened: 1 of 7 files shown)


OchnikBartek:
@codex review

PR #1098, opened 2026-08-25
fix(vault): refuse a malformed sealed payload without naming it

### LINKED ISSUE #21: Refuse a malformed sealed payload without naming its contents
**Severity: Medium** · AUD-019

## What's wrong

`unseal_secret`'s docstring promises `BadRequestError` "if the envelope … holds something that is not a secret payload". It does not: `_STORABLE_ADAPTER.validate_json(...)` raises `pydantic_core.ValidationError`, which is not an `AppException`, so it reaches `unhandled_exception_handler` and `logger.exception`.

## Where

`backend/app/core/secret_kinds.py:252-270` (line 262), reaching `backend/app/api/exception_handlers.py:175`

## How it fails

Pydantic embeds the offending input in the message. Reproduced against this repo's own `_STORABLE_ADAPTER`:

```
Input tag 'bogus' found using 'kind' does not match any of the expected tags: …
[type=union_tag_invalid, input_value={'kind': 'bogus', 'api_ke...: 'sk-live-SUPERSECRET'}, input_type=dict]
```

The decrypted credential lands in the log line and, under `logfire.instrument_fastapi`, in the exception event on the span — breaking the guarantee `docs/secrets.md` states ("No log line or audit entry contains one") and `organization_secret.py:1-7` claims.

Reachable from `OrganizationSecretService.resolve_for_bindings` (`:296`), `ModelProfileService` (`model_profile.py:298`) and `_listing_key` (`model_providers.py:148`).

Trigger: any stored payload that no longer validates — a hand-edited row, a rollback to a build whose `SecretKind` enum lacks a kind a newer build wrote, or a future field tightening.

## Fix

Catch `PydanticValidationError` and re-raise `BadRequestError(message="Stored secret is not a usable payload", details={"recorded": kind.value})` with **`from None`** — chaining with `from exc` keeps the plaintext in the traceback.

## Acceptance criteria

- [ ] Regression test `test_a_payload_that_does_not_validate_is_refused_without_naming_the_value`: seal a JSON document with an unknown `kind`, call `unseal_secret`, assert `BadRequestError` and that the plaintext appears in neither `str(exc)` nor `exc.__cause__`

`tests/api/test_no_secret_escapes.py` sweeps the OpenAPI surface thoroughly and says nothing about log paths — worth extending in the same spirit.

OchnikBartek:
## What & why

`unseal_secret` promised `BadRequestError` for an envelope that holds something that is not a secret payload, but `_STORABLE_ADAPTER.validate_json(...)` raises `pydantic_core.ValidationError` — not an `AppException` — so it reached `unhandled_exception_handler` and `logger.exception`. A pydantic `ValidationError` **embeds the offending input in its message**, and here that input is the decrypted credential, so the plaintext landed in the log line and, under `logfire.instrument_fastapi`, on the exception span — breaking the guarantee `docs/secrets.md` states ("No log line or audit entry contains one").

Catch `ValidationError` and re-raise `BadRequestError(message="Stored secret is not a usable payload", details={"recorded": kind.value})` with `from None`.

- The **type change** is the primary guard: a 4xx `AppException` is logged at warning with no `exc_info`, so no traceback is formatted on the HTTP path, and it never reaches the `logger.exception` handler.
- **`from None`** is load-bearing for the non-HTTP readers whose traceback *is* rendered — a Prefect task failure, `doctor.py`'s `f"...: {exc}"` — where it sets `__suppress_context__` and keeps the chained `ValidationError` (which still holds the plaintext in `__context__`) out of the formatted traceback.

Reachable from `resolve_for_bindings`, `ModelProfileService` and the provider listing key whenever a stored payload no longer validates — a hand-edited row, a rollback to a build whose `SecretKind` enum lacks a kind a newer build wrote, or a future field tightening.

## Verification

- Regression test seals a JSON document with an unknown `kind`, calls `unseal_secret`, and asserts `BadRequestError` with the plaintext absent from `str(exc)` **and from the rendered traceback** — pinning `__suppress_context__`, the property that actually stops the leak (an adversarial review caught that `__cause__` is `None` with or without `from None`, so asserting it alone would let a future edit silently reintroduce the leak). Fails without the fix.
- `make lint` green. `secret_kinds.py` is at complete coverage locally; the full 100% gate needs the integration suite, which skips here for want of a local DB (the only uncovered lines are DB-layer repositories) — CI's environment has one.

Closes #21

### DIFF
--- a/backend/app/core/secret_kinds.py
+++ b/backend/app/core/secret_kinds.py
@@ -58,6 +58,7 @@ from pydantic import (
     PlainSerializer,
     SecretStr,
     TypeAdapter,
+    ValidationError,
     field_validator,
     model_validator,
 )
@@ -330,9 +331,19 @@ def unseal_secret(
             is not a secret payload, or holds a different kind than the column
             says - which would mean a row was edited underneath the schema.
     """
-    value = _STORABLE_ADAPTER.validate_json(
-        unseal(ciphertext, scope=scope, key_version=key_version)
-    )
+    try:
+        value = _STORABLE_ADAPTER.validate_json(
+            unseal(ciphertext, scope=scope, key_version=key_version)
+        )
+    except ValidationError:
+        # A `ValidationError` message embeds the offending input - here the
+        # decrypted credential - so it must reach neither a log line nor a span.
+        # `from None`, not `from exc`: chaining would keep the plaintext in the
+        # traceback (#21).
+        raise BadRequestError(
+            message="Stored secret is not a usable payload",
+            details={"recorded": kind.value},
+        ) from None
     if value.kind is not kind:
         raise BadRequestError(
             message="Stored secret does not match its recorded kind",
--- a/backend/tests/test_secrets.py
+++ b/backend/tests/test_secrets.py
@@ -7,6 +7,7 @@ worth testing here is that constraint holding.
 """
 
 import json
+import traceback
 import uuid
 from datetime import UTC, datetime
 from unittest.mock import AsyncMock, MagicMock, patch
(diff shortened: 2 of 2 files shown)


OchnikBartek:
@codex review

PR #1104, opened 2026-08-25
test(sanitize): cover the DNS-failure paths and gate the SSRF module

### LINKED ISSUE #28: Return the five repository-calling routes to the service layer
**Severity: Medium** · AUD-026

## What's wrong

CLAUDE.md and `.claude/rules/architecture.md` both state that routes never import or call repositories. Five handlers do.

## Where

- `backend/app/api/routes/v1/org_integrations.py:152` — `sync_log_repo.get_all`
- `backend/app/api/routes/v1/knowledge_bases.py:453` — `sync_log_repo.get_all`
- `backend/app/api/routes/v1/runs.py:41,52` — `agent_run_repo.list_runs`, `get_run`
- `backend/app/api/routes/v1/audit.py:32,35` — `audit_log_repo.list_for_org`, `count_for_org`
- `backend/app/api/routes/v1/model_providers.py:144` — `organization_secret_repo.list_secrets`

(`knowledge_bases.py:28` and `skills.py:21` import types only — `CollectionCounts`, `SkillSort` — which is fine.)

## To be fair

All five currently get tenant isolation right — each hand-rolls `organization_id=ctx.organization_id` into the repo call. The cost is not a live bug; it is that the rule "tenant isolation lives in a service" now has five exceptions, so the next reader copying `runs.py` gets a template with no service to put the check in, and the isolation is expressed in five places instead of one.

## One of the five is worse than a layering nit

`model_providers.py:135-156` defines `_listing_key` — a private helper in a **route file** that unseals vault secrets:

```python
async def _listing_key(db: Any, provider: str, *, organization_id: UUID) -> str | None:
    secrets = await organization_secret_repo.list_secrets(db, organization_id=organization_id, purposes=[provider])
    for secret in secrets:
        value = unseal_secret(...)
```

Three problems in one function:

1. Decryption in the HTTP layer.
2. `db: Any` — a typing escape the repo's own rules forbid.
3. `list_secrets` is called with its default `see_all=True`, so it walks **every** secret in the organization regardless of visibility or grants and unseals the first API key it finds. The route is gated only on `AGENTS_VIEW`, which a Viewer holds — so Member A's *private* OpenAI key is unsealed and spent to serve Member B's model picker. The plaintext never leaves the function, so this is unauthorized *use* rather than disclosure: quota, billing and provider-side request logs. It contradicts `permissions.py:45-50`, which says gating the whole vault was the defect `SECRETS_VIEW` was introduced to fix.

Move it into `ModelProfileService` or `OrganizationSecretService` and resolve visibility the way the secret listing does — `visible_resource_ids(..., Perm.SECRETS_VIEW)`, `user_id=ctx.subject_id`, `shared_ids`.

## The gate is the reason this matters

`model_providers.py` is in neither `[tool.coverage.run] include` nor `[[tool.ty.overrides]] include`. Neither are `app/core/audit.py`, `app/core/sanitize.py`, `app/services/agent_embed.py`, `app/services/embed_session.py`, or `app/services/channels/slack.py` / `telegram.py` / `mattermost.py` / `router.py` (only `mentions.py` is included).

Those modules hold the supervisor busy-loop, the Telegram fail-open, the audit swallow, the embed token expiry and half the concurrency races between them. The 100% gate is doing exactly what it was built to do inside its boundary; the boundary just does not include several modules implementing the refusals CLAUDE.md names as must-cover.

## Acceptance criteria

- [ ] The five handlers go through a service
- [ ] `_listing_key` moves into a service, is typed, and respects visibility
- [ ] Regression test `test_a_listing_does_not_reach_another_members_private_key`
- [ ] Incrementally add the un-gated modules to both lists in `backend/pyproject.toml`, highest-risk first

OchnikBartek:
## What & why

`app/core/sanitize.py` — the SSRF allow/deny every outbound URL a tenant chooses is checked against (a webhook target, an MCP server address) — was in neither `[tool.coverage.run] include` nor `[[tool.ty.overrides]] include`, the 100% gate #28 asks to extend to the modules implementing the refusals CLAUDE.md names as must-cover. It sat at 95%: the two DNS-failure exits in `resolve_pinned_url` (a lookup that raises `gaierror`; one that resolves to no address) had no test.

Cover both with a mocked `getaddrinfo`, then add the module to both lists (verbatim, same position), beside `pinned_http.py` — the validator paired with the client that dials what it validated. It types clean under the stricter checker and reaches 100% line+branch.

## Scope — the coverage-gate tail (this issue's remaining work)

Per Kacper's decision, this PR does the coverage-gate tail only; it does **not** implement the `_listing_key` visibility narrowing (AC 2/3), which the maintainer declined. The five-routes layering (AC 1) already landed via #232.

sanitize.py is the one tail module ready to gate right now:

| Module | Status |
|---|---|
| `app/core/audit.py` | gated by its own branch (#20) |
| `app/services/embed_session.py` | already in both lists |
| `app/services/agent_embed.py` | 85% — needs substantial tests first |
| `channels/slack.py`, `telegram.py`, `mattermost.py`, `router.py` | hold open bugs (supervisor loop, concurrency races); need to reach 100% and a real DB to measure |

#28 stays open for those.

## Verification

`test_coverage_gate.py` green (the two lists stay in sync), `sanitize.py` at 100% line+branch via the SSRF suite, `make lint` (ruff + ty + guards) clean. The two new tests fail without the DNS-failure handling they cover.

Closes #28

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -468,6 +468,10 @@ include = [
     "app/core/secret_kinds.py",
     "app/core/vault.py",
     "app/core/background.py",
+    # The SSRF allow/deny every outbound URL a tenant chooses is checked against -
+    # a webhook, an MCP server address. A hole here is a request to an internal
+    # address the operator cannot see, which is the pair to pinned_http below (#28).
+    "app/core/sanitize.py",
     # The client that dials the address it validated, for a URL the party being
     # validated chose. A miss here is an SSRF the operator cannot see.
     "app/core/pinned_http.py",
@@ -688,6 +692,10 @@ include = [
     "app/core/secret_kinds.py",
     "app/core/vault.py",
     "app/core/background.py",
+    # The SSRF allow/deny every outbound URL a tenant chooses is checked against -
+    # a webhook, an MCP server address. A hole here is a request to an internal
+    # address the operator cannot see, which is the pair to pinned_http below (#28).
+    "app/core/sanitize.py",
     # The client that dials the address it validated, for a URL the party being
     # validated chose. A miss here is an SSRF the operator cannot see.
     "app/core/pinned_http.py",
--- a/backend/tests/test_ssrf.py
+++ b/backend/tests/test_ssrf.py
@@ -3,6 +3,7 @@
 Covers validate_webhook_url() and _is_ip_blocked() from app.core.sanitize.
 """
 
+import socket
 from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


OchnikBartek:
@codex review

Issue #561, opened 2026-08-10
The observability secret is validated at run time; the factory docstring claims publish refuses it


DEENUU1:
`agents/factory.py:318-345` (`_instrument`) tolerates a missing token, docstring: "publishing is where a missing secret is refused, and a run is far too late." But `services/agent_registry.py:666-737` (`validate_spec`) never references `spec.observability` — only capability-binding secrets are validated — while `agent_runner.py:203-214` unseals it at run time.

**How it fails.** An author sets a wrong/missing/other-org observability secret id, publishes with no error, and every run executes untraced — the only signal a log line nobody reads. Benign (graceful fallback, org-scoped, no cross-tenant leak) but a documented guarantee that is false, and inconsistent with capability-secret validation.

**Fix.** Add the observability-secret check to `validate_spec` mirroring `_secret_problems`; or, if untraced fallback is intended, correct the `_instrument` docstring.

**Done when.** Publishing with a bad observability secret is refused, or the docstring no longer claims it is.

---
_Found in the 2026-08-10 duplication & code-quality audit. See the issue map #168._

OchnikBartek:
Duplicate of #33 — this is the second checkbox item there (`spec.observability.token_secret_id` is not validated at publish). Keeping it in the #33 audit batch so it isn't worked twice.

PR #1105, opened 2026-08-25
fix(channels): refuse a non-ASCII credential instead of crashing

OchnikBartek:
## What & why

One item of #33's audit-cleanup batch (plus one already fixed on `main`, ticked).

`secrets.compare_digest` — and `hmac.compare_digest`, the same function — raises `TypeError` on a non-ASCII `str` rather than returning `False`. Three checks compare `str` and take the left operand from the caller: `deps.verify_api_key` (the API-key header), the Mattermost webhook bearer check, and the Slack signature check. So `{"token": "é"}` to the **unauthenticated** Mattermost webhook, or a non-ASCII `X-Slack-Signature`, was a logged 500 with a traceback instead of a 403 — a free log-flooding primitive. Not an auth bypass (the comparison never matched anyway).

Compare `.encode()`d bytes at all three sites — constant-time, and non-ASCII now refuses cleanly. Three regression tests assert a non-ASCII token/signature/key is refused rather than raising; each fails against the pre-fix `str` comparison.

## #33 batch status

| Item | Status |
|---|---|
| **compare_digest on non-ASCII** (this PR) | fixed |
| **`get_worker_db_context` duplicates `_managed_session`** | already done on `main` — it wraps `_managed_session` (ticked) |
| `validate_webhook_url` has no callers | already ticked in the issue |
| `spec_version` write-only | remaining — touches the spec format/`from_yaml`, needs the agent-spec care |
| `observability.token_secret_id` unvalidated (= #561) | remaining |
| `hint` reveals a short secret | remaining — `min_length=8` on `SealedStr` breaks ~80 short-secret test fixtures; not sub-hour |
| Slack Socket Mode client never closed | remaining — async socket lifecycle |
| `McpOAuthPayload` plain-str credentials | remaining — `SecretStr` cascades through ~15 read sites whose truthiness checks (`not payload.access_token`) change semantics |
| `ModelProfile.allow_byo` dead | remaining — a stored column, so deletion needs a migration |

#33 stays open for those.

## Verification

`make lint` clean; full non-integration suite green (6283 passed, 0 failed). Self-reviewed given the change is three one-line `.encode()`s with textbook constant-time semantics.

Refs #33

### DIFF
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -898,7 +898,7 @@ async def verify_api_key(
     """
     if api_key is None:
         raise AuthenticationError(message="API Key header missing")
-    if not secrets.compare_digest(api_key, settings.API_KEY):
+    if not secrets.compare_digest(api_key.encode(), settings.API_KEY.encode()):
         raise AuthorizationError(message="Invalid API Key")
     return api_key
 
--- a/backend/app/api/deps.py
+++ b/backend/app/api/deps.py
@@ -267,7 +267,7 @@ from app.core.exceptions import (
 )
 from app.services import rate_limit
 from app.core.audit import set_impersonator
-from app.core.security import verify_token
+from app.core.security import encode_untrusted, verify_token
 from app.db.models.user import User
 
 oauth2_scheme = OAuth2PasswordBearer(tokenUrl=f"{settings.API_V1_STR}/auth/login")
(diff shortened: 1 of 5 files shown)


OchnikBartek:
@codex review

OchnikBartek:
@codex review

Issue #23, opened 2026-08-01
Require a freshness claim on an embed token


DEENUU1:
**Severity: Medium** · AUD-021

## What's wrong

The max-age check is guarded by `if isinstance(issued_at, int | float)`. A token carrying only `sub`, correctly signed, has no `iat` and no `exp` — PyJWT requires neither — so it is accepted forever.

## Where

`backend/app/services/agent_embed.py:284-288`

## How it fails

A customer's backend mints `{"sub": "user-42"}` — the minimal thing that satisfies the subject check at line 291. One token scraped from a browser's network tab keeps that widget answering on the organization's bill indefinitely.

This is the same failure the surrounding code twice calls out as the dangerous one: *"`jwt` mode silently degrading to `public`"*, *"a single leaked token becomes the whole widget's budget"*.

`test_a_stale_token_is_refused` (`tests/test_agent_embed.py:179-190`) passes only because its fixture supplies `iat`. `test_a_token_signed_with_the_wrong_secret_is_refused` two tests earlier demonstrates the no-`iat` shape is perfectly representable.

## Fix

Require a freshness claim rather than checking one opportunistically: `jwt.decode(..., options={"require": ["sub"]})`, then refuse when neither `exp` nor a within-window `iat` is present. Say so in `docs/channels.md` — customers minting tokens need to know the claim is mandatory.

## Acceptance criteria

- [ ] Regression test `test_a_token_with_no_issued_at_is_refused_rather_than_treated_as_fresh`
- [ ] `docs/channels.md` states the requirement

PR #1120, opened 2026-08-25
fix(channels): refuse a malformed embed-token claim instead of 500ing

### LINKED ISSUE #1107: A malformed exp/iat claim 500s the embed verifier instead of refusing
## What's wrong

`AgentEmbedService._verify_token` (`backend/app/services/agent_embed.py`) decodes a
visitor token with `jwt.decode(...)` inside a `try/except jwt.PyJWTError`. A token
whose `exp` or `iat` claim is `null`, `[]`, or `{}` makes PyJWT do `int(payload["exp"])`
internally, which raises a raw `TypeError` - **not** a `PyJWTError` - so it escapes the
handler and propagates as an uncaught 500 instead of a clean refusal.

## Where

`backend/app/services/agent_embed.py` - the `try: jwt.decode(...) except jwt.PyJWTError`
block in `_verify_token`.

## Severity

Low. Not attacker-exploitable: claim validation runs after signature verification, so
producing such a token needs the customer's own signing secret, and it fails closed
(a 500, never an admission). It is a robustness/clean-4xx gap, not a security hole.

## Fix

Broaden the catch to `except (jwt.PyJWTError, TypeError)` (or `ValueError` too), raising
`EmbedDenied("token rejected")`, so a customer's malformed-but-signed token yields a 4xx
rather than a 500 with a logged traceback.

## Acceptance criteria

- [ ] A signed token with `exp` (or `iat`) as `null`/`[]`/`{}` is refused as a clean 4xx, not a 500
- [ ] Regression test with such a token asserting `EmbedDenied` rather than a raised `TypeError`

## Notes

Pre-existing (the `jwt.decode` line predates it); found reviewing #23, which fixed the
freshness-claim leak and deliberately left this out of scope. Same embed-token area as #23.

OchnikBartek:
## What

`AgentEmbedService._verify_token` decoded a visitor token inside a
`try/except jwt.PyJWTError`. A signed token whose `exp` or `iat` is `null`, `[]`
or `{}` makes PyJWT's own validation run `int()` on it, which raises a raw
`TypeError` — not a `PyJWTError` — so it escaped the handler and became an
uncaught 500 with a logged traceback instead of a clean refusal.

Not attacker-exploitable: claim validation runs after signature verification, so
minting such a token needs the customer's own signing secret, and it fails
closed. A robustness gap — a customer's malformed-but-signed token should be a
4xx, not a 500.

The catch now also covers `TypeError`, raising `EmbedDenied`. `ValueError` was
left out deliberately: on the pinned PyJWT only these coercions raise a bare
`TypeError`; every malformed-segment/base64/JSON/non-numeric case is a
`PyJWTError` subclass, so a `ValueError` catch would be an unreachable branch.

## How verified

Regression test parametrized over `exp`/`iat` × `null`/`[]`/`{}` (6 cases) asserts
`EmbedDenied`; without the broadened catch each raises `TypeError` from PyJWT's
`int()` (confirmed: `jwt/api_jwt.py` `int(None)`). `make lint`, `ty`, `vulture`
and the full embed suite green.

Adversarially reviewed (a `@codex review` follows): no P1/P2; the reviewer's
points — drop the unreachable `ValueError`, parametrize `exp`/`iat` × the three
malformed shapes — are applied here.

Pre-existing, found reviewing #23 (which fixed the freshness-claim leak and left
this out of scope).

Closes #1107

### DIFF
--- a/backend/app/services/agent_embed.py
+++ b/backend/app/services/agent_embed.py
@@ -608,7 +608,11 @@ class AgentEmbedService:
         )
         try:
             claims = jwt.decode(token, secret, algorithms=["HS256"])
-        except jwt.PyJWTError as exc:
+        except (jwt.PyJWTError, TypeError) as exc:
+            # `exp`/`iat` as null, [] or {} makes PyJWT's own validation run
+            # `int()` on it - a raw `TypeError`, not a `PyJWTError` - so without
+            # this a malformed-but-signed token escaped as a 500 rather than a
+            # clean refusal (#1107).
             raise EmbedDenied("token rejected") from exc
 
         issued_at = claims.get("iat")
--- a/backend/tests/test_agent_embed.py
+++ b/backend/tests/test_agent_embed.py
@@ -237,6 +237,21 @@ class TestTokenMode:
 
         assert _service()._verify_token(embed, token) == "user-42"
 
+    @pytest.mark.parametrize("claim", ["exp", "iat"])
+    @pytest.mark.parametrize("value", [None, [], {}])
+    def test_a_signed_token_with_a_malformed_time_claim_is_refused_not_500ed(self, claim, value):
+        """`exp`/`iat` as null, [] or {} makes PyJWT's own validation run `int()`
+        on it - a raw `TypeError`, not a `PyJWTError` - so a malformed-but-signed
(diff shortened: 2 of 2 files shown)


OchnikBartek:
@codex review

PR #1121, opened 2026-08-25
fix(rag): give the document listings a stable pagination tiebreaker

### LINKED ISSUE #1103: Give the RAG document listings a stable pagination tiebreaker
## What's wrong

The two paginated RAG document listings order by `created_at DESC` with no unique
tiebreaker, so under `offset`/`limit` a client paging a collection where two rows
share the exact `created_at` (a bulk import lands many rows in the same
microsecond) can see a row twice or skip one across page boundaries.

## Where

- `backend/app/repositories/rag_document.py` — `get_all` (added in #27) and `get_for_kb`

The sibling `vectorstore.get_documents` already adds a secondary `id` sort for
exactly this reason (`vectorstore.py`, #548); these two do not.

## Fix

Append `RAGDocument.id` (or `created_at DESC, id DESC`) as a secondary sort on both
`get_all` and `get_for_kb`, so page boundaries are deterministic.

## Acceptance criteria

- [ ] Both listings order by a unique tiebreaker; paging over tied timestamps neither repeats nor skips a row

## Notes

Found reviewing #27 (parity issue pre-dating it — both siblings lack the tiebreaker).
Low severity: timestamptz microsecond precision makes ties rare. `get_for_kb` is
frontend-consumed, so the change is additive-only (a deterministic tiebreaker never
changes non-tied order). Same cluster as #27 (see #168).

OchnikBartek:
## What

`get_all` and `get_for_kb` (`app/repositories/rag_document.py`) ordered by
`created_at DESC` with no unique secondary key. A bulk import lands many
`rag_documents` in one microsecond, so over tied timestamps that is not a total
order: an `offset`/`limit` client paging such a collection can see a row twice or
skip one across page boundaries, because each page is a separate query whose sort
of the tied rows is undefined. The sibling `vectorstore.get_documents` already
adds a secondary `id` sort for this reason (#548); these two did not.

Both listings now order by `created_at DESC, id DESC`.

## Additive

A unique tiebreaker is only consulted when `created_at` ties, so rows with
distinct timestamps keep their exact prior order — the frontend-consumed
`get_for_kb` changes only among rows sharing a `created_at` to the microsecond,
order that was previously undefined and nothing could depend on.

## How verified

Two integration tests seed six documents at one identical `created_at` and assert
the exact `id DESC` order — through `get_all`, and paged across three `get_for_kb`
pages. Six random ids hitting that sequence by heap order is ~1/720, so both fail
without the tiebreaker (verified) and are deterministic with it. `make lint`, the
guard scripts and the new tests are green.

Adversarially reviewed (a `@codex review` follows): clean, no P1/P2.

Pre-existing parity gap, found reviewing #27.

Closes #1103


OchnikBartek:
@codex review

PR #1122, opened 2026-08-25
fix(rag): move the worker's blocking file walk, hash and stat off its loop

### LINKED ISSUE #1100: Move the worker's blocking file walk/hash off its event loop
## What's wrong

Blocking filesystem work runs synchronously inside the Prefect worker's async
tasks, the same defect #25 fixed on the request path. Lower severity because it
is the worker rather than an API request, but a large tree or file still stalls
the worker's event loop with no suspension point.

## Where

- `backend/app/worker/tasks/rag_tasks.py:256` — `rglob` over a whole tree
- `:289`
- `:302` — `hashlib.sha256(filepath.read_bytes())`
- `:323` — `filepath.stat()`

(Line numbers from the #25 audit; re-confirm against current `main`.)

## Fix

`await asyncio.to_thread(...)` around each blocking call, as #25 did for
`file_upload.py` / `file_storage.py`. `rglob` + the per-file hash is the one
worth batching into a single `to_thread` over the walk rather than a hop per file.

## Acceptance criteria

- [ ] The blocking walk/hash/stat calls run off the worker loop
- [ ] The tasks still produce identical manifests/hashes

## Notes

Split out of #25, which fixed the request-path half (23a/23b) and deliberately
left the worker paths as a lower-severity follow-up. Same cluster as #25; part
of the AUD-023 blocking-IO family (see #168).

OchnikBartek:
## What

The Prefect RAG sync flows (`app/worker/tasks/rag_tasks.py`) did their filesystem
work synchronously on the worker's event loop — the defect #25 fixed on the
request path, left as a lower-severity follow-up. A large tree or file has no
suspension point: `rglob` over a directory, `sha256(read_bytes())` per file,
`stat()` and a per-file `resolve()` each stall the loop.

Each **per-file** blocking call now goes through `asyncio.to_thread`:
- the tree walk (`rglob` + a `stat` per entry via `is_file`) is one hop over
  `_walk_files`, not a hop per file;
- the per-file hash is `_hash_file`, shared by the local and connector paths;
- `stat` and `resolve` are offloaded per file.

Values are unchanged — identical walks, hashes, sizes, resolved paths — only the
thread moves. The one-off resolve/`is_file` of the sync *target* path itself is
O(1) per flow and left on the loop.

## How verified

A flow-level regression test patches `_hash_file` with a thread probe and drives
the connector skip path, asserting the hash ran off the loop thread — it fails
against the pre-offload direct call (the probe lands on the loop thread). Plus
unit tests for `_hash_file` (== the file's sha256) and `_walk_files` (non-hidden
files, recursive, no directories). `make lint` green.

Adversarially reviewed (a `@codex review` follows): no P1/P2. Two P3s, both
addressed or noted: the per-file `resolve()` the first pass left behind is now
offloaded here; and `_run_sync` (the local-directory path) shares the proven
helpers but has no flow-level probe of its own — acceptable for this
template-inherited module (not in the coverage/ty gate).

Split out of #25 (AUD-023 blocking-IO family).

Closes #1100


OchnikBartek:
@codex review

Issue #1124, opened 2026-08-25
seed --clear / delete_non_admins still 500s on the personal-org RESTRICT FK


OchnikBartek:
## What's wrong

`user_repo.delete_non_admins` (`backend/app/repositories/user.py:119`) is a bulk
`DELETE FROM users WHERE is_app_admin = false`. It bypasses
`UserService._release_owned_rows` entirely — that reconciliation runs only from
the single-row `UserService.delete` (`app/services/user.py:317`), never from the
bulk path (`app/services/user.py:79` → repo).

Every user `seed` creates goes through `UserService.register` (`app/commands/seed.py:95`),
which calls `create_personal_org` — giving each user a personal org whose
`organizations.created_by_user_id` FK is `ondelete="RESTRICT"`, `nullable=False`
(`app/db/models/organization.py:51-56`; baseline `0001_baseline.py:202-207`).

So `agenticos cmd seed --clear` (`app/commands/seed.py:88` → `delete_non_admins`)
raises `ForeignKeyViolation` on `organizations_created_by_user_id_fkey` for every
seeded non-admin user, because each owns a personal org. **A second `seed` after
a `seed --clear` — or any `--clear` on a seeded DB — 500s**, and there is no
`IntegrityError` handler, so it surfaces as an unhandled error in the command.

## Why this is not #1110

#1110 removed the NO-ACTION *invite* FK bar, which fixed the single-row
`DELETE /users/{id}` path (it goes through `_release_owned_rows` plus the new
invite `SET NULL`s). The bulk `delete_non_admins` path bypasses that
reconciliation and hits a **different** wall — the personal-org `RESTRICT` — so
#1110's invite-FK change does not lift it. Found reviewing #1110.

## Where

- `backend/app/repositories/user.py:119` — `delete_non_admins`, the bulk delete.
- `backend/app/services/user.py:79` — the service passthrough (no reconciliation).
- `backend/app/services/user.py:321` — `_release_owned_rows`, the reconciliation the bulk path skips.
- `backend/app/db/models/organization.py:51-56` — the `RESTRICT` FK.

## Fix

Either delete non-admins one at a time through `UserService.delete` (so
`_release_owned_rows` runs per user — purges/reassigns each personal org), or
give `delete_non_admins` a bulk equivalent that releases owned rows for the whole
set before the delete. The sole-owner-of-a-shared-org case
(#1117) overlaps and should be reconciled the same way.

## How you'd know it's fixed

[...]

PR #1125, opened 2026-08-26
fix(organizations): SET NULL the three invite FKs into users

### LINKED ISSUE #1110: User deletion still 500s on three NO-ACTION invite FKs
**Found while fixing #9.** Same family as #941 (account deletion).

## What's wrong

#9 reconciled the three `SET NULL`/`RESTRICT` cascades that collide with a
CHECK constraint, so a *normally-registered* user (one personal org, no
invites) now deletes cleanly. But three foreign keys to `users.id` carry **no
`ondelete`** — PostgreSQL `NO ACTION` — so they still block deletion the same
way, and this is a separate wall (a missing `ondelete`, not a CHECK conflict),
needing a migration rather than a service step:

- `OrganizationMember.invited_by_user_id` — `backend/app/db/models/organization.py:113-117`
- `Invitation.invited_by_user_id` (NOT NULL) — `backend/app/db/models/organization.py:169-173`
- `Invitation.accepted_by_user_id` — `backend/app/db/models/organization.py:186-190`

### Trigger

`DELETE /api/v1/users/{id}` (`UserService.delete`) 500s for any user who has
**invited someone** — the invitee's `OrganizationMember.invited_by_user_id` and
the `Invitation.invited_by_user_id` row both still reference the deleted user,
and `NO ACTION` raises a `ForeignKeyViolationError` at the end of the statement.
There is no `IntegrityError` handler in `app/api/exception_handlers.py`, so it
becomes a 500. This is a common case in any real deployment.

`delete_non_admins` (`app/repositories/user.py:119`, used by `seed --clear`)
hits the same NO-ACTION wall in bulk — #9 left it untouched for this reason.

## Fix

Pick the right `ondelete` per column and add a migration:

- `invited_by_user_id` (both tables): `SET NULL` — "who invited them" is audit
  context that should outlive the inviter, matching how the secret/KB
  attribution FKs already `SET NULL`. `Invitation.invited_by_user_id` is
  currently `NOT NULL`, so it must be made nullable in the same migration.
- `Invitation.accepted_by_user_id`: `SET NULL` (already nullable).

Then handle `delete_non_admins`: with the FKs fixed it should no longer 500, but
add coverage that seeds an inviter + an invitation and asserts the bulk clear
succeeds.

## How I'd know it's fixed

- [ ] Integration test: a user who invited another member can be deleted; the
      invitee's membership survives with `invited_by_user_id` nulled.
- [ ] Integration test: `delete_non_admins` succeeds on a DB holding
      invitations authored by non-admins.
- [ ] `make test-migrations` green (forward + back).

## Minor, same area

`OrganizationService.purge` drops org-scoped collections' vector tables only
when a store is wired in (the org-delete route). A **personal** org deleted via
account teardown (`UserService.delete`) has no store, so its org-scoped
collections' `rag_<collection>` tables are orphaned (rows are still removed).
Minor dev-side leak; close it by threading a store into the user-delete route or
a background cleanup.

Part of the deletion cluster on #168.

OchnikBartek:
> **Stacked on #9** (`fix/9-delete-cascade-check-reconcile`). Based against that branch so the diff is only #1110's; the `DELETE /users/{id}` test relies on #9's `_release_owned_rows`. **Retarget to `main` once #9 (PR #1112) merges.**

## What

Three foreign keys into `users.id` — `OrganizationMember.invited_by_user_id`, `Invitation.invited_by_user_id`, `Invitation.accepted_by_user_id` — carried no `ondelete`, so PostgreSQL's NO ACTION made each an **absolute bar on deleting the referenced user**. #9 reconciled the CHECK-versus-cascade collisions, but a user who had ever invited another member still failed `DELETE /users/{id}` with a foreign-key violation surfaced as a 500 — common in any real deployment.

All three become `ON DELETE SET NULL`: who invited a member and who accepted an invitation are audit context that should outlive the user, the way the secret/KB attribution FKs already null. `Invitation.invited_by_user_id` was NOT NULL, so **migration 0056** makes it nullable in the same step (set on every create, null only once the inviter is gone).

## Tests

Three real-Postgres integration tests (`tests/integration/test_user_deletion_invites.py`):
- deleting an inviter leaves the invitee's membership and the invitation, both with the inviter nulled;
- deleting the accepter nulls `accepted_by_user_id`;
- the SET NULL fires on the bulk `delete_non_admins` path too.

Each **fails without the model change**. Migration applies forward+back+forward on a scratch DB; `alembic check` reports no drift. `make lint` + `make test` green (6912 passed, coverage 100%).

## Deliberately out of scope → #1124

`seed --clear`'s bulk `delete_non_admins` still 500s on `organizations.created_by_user_id` **RESTRICT** for a user who owns a personal org — that path bypasses `_release_owned_rows` and hits a different FK/mechanism than the invite bar. Filed as #1124 (found reviewing this).

Closes #1110
Refs #9, #1124

### DIFF
--- a/backend/alembic/versions/0056_invite_fk_ondelete.py
+++ b/backend/alembic/versions/0056_invite_fk_ondelete.py
@@ -1,0 +1,98 @@
+"""Let a deleted user keep their invite audit trail instead of blocking deletion.
+
+Three foreign keys into `users.id` - who invited a member, who authored an
+invitation, who accepted one - carried no `ondelete`, so PostgreSQL's NO ACTION
+made any of them an absolute bar on deleting the referenced user. #9 reconciled
+the CHECK-versus-cascade collisions for a normally-registered user, but a user
+who had ever invited another member still failed `DELETE /users/{id}` with a
+foreign-key violation surfaced as a 500 (#1110).
+
+`SET NULL` on all three: who invited a member and who accepted an invitation are
+audit context that should outlive the user, the way the secret and
+knowledge-base attribution FKs already null. `invitations.invited_by_user_id`
+was NOT NULL, so it is made nullable in the same step - it is set on every create
+and null only once the inviter is gone.
+
+The downgrade re-tightens that column to NOT NULL, which fails if the SET NULL
+schema has run long enough for an inviter to be deleted and left a null behind -
+the standard nullable->NOT NULL rollback risk, no data loss, and it aborts
+cleanly.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "0056_invite_fk_ondelete"
+down_revision: str | Sequence[str] | None = "0055_sandbox_operations"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.drop_constraint(
+        "organization_members_invited_by_user_id_fkey", "organization_members", type_="foreignkey"
+    )
+    op.create_foreign_key(
+        "organization_members_invited_by_user_id_fkey",
+        "organization_members",
+        "users",
+        ["invited_by_user_id"],
+        ["id"],
+        ondelete="SET NULL",
+    )
+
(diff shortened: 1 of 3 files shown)


OchnikBartek:
@codex review

PR #1127, opened 2026-08-26
fix(dev): judge the reload beat on wall clock, not monotonic

### LINKED ISSUE #1080: Dev-stack watchdog kills a healthy reloaded server and leaves nothing serving

OchnikBartek:
## What

The local dev stack runs `python -m cli.reload_supervisor`, whose `SupervisedReload` replaces a worker whose event loop has stopped turning. The worker reports liveness through uvicorn's `Config.callback_notify`, which `Server.on_tick` calls only once **`time.time() - last_notified > timeout_notify`** — so the beat **cadence is gated on the wall clock**. The beat stamped `time.monotonic()` and the supervisor judged with `time.monotonic() - beat` — a *different* clock from the one the cadence runs on.

On Docker Desktop's VM the wall clock can stall for tens of seconds relative to monotonic while the loop keeps serving. The cadence froze while the monotonic verdict climbed, so a **healthy, serving worker was read as wedged and killed** — and the replacement, beating once then killed again on the next poll, is the "container Up, nothing serving" the issue also reports. The module's own docstrings already named the mismatch.

## Fix

The beat now stamps `time.time()` and the verdict reads `time.time() - beat` — both on the clock uvicorn gates the cadence on, so they cannot diverge: a stalled wall clock stalls the cadence and the reading together.

`app/core/watchdog.py` (the in-process watchdog on the other two stacks) is **left untouched** — it beats via `loop.call_later` and judges with `time.monotonic()`, both the event loop's own clock, so it is internally consistent and never used uvicorn's notify hook. The issue pointed here, but the killer's log line is `SupervisedReload`'s.

## Trade, stated

A real wedge coinciding with a *backward* NTP step reads `silent_for` negative for the length of the step and defers the verdict — bounded by `POLLS_BEFORE_WEDGED`, self-healing once the wall clock passes `beat + WEDGED_AFTER`, and local-dev only. Since uvicorn gates the cadence on wall time, a wall-clock verdict is the only self-consistent choice.

## Tests

A new test pins both clocks divergent (monotonic ahead, wall stalled at the beat) and asserts no replacement past `POLLS_BEFORE_WEDGED`; it and the heartbeat-stamp test both **fail against the monotonic code** (verified by reverting). The ~18 existing beat-value assertions move to `time.time()`; the harness deadline loops stay monotonic. `make lint` + `make test` green (6901 passed, coverage 100%). Root cause verified against uvicorn 0.52.3's `on_tick` source.

Closes #1080

### DIFF
--- a/backend/cli/reload_supervisor.py
+++ b/backend/cli/reload_supervisor.py
@@ -40,10 +40,12 @@ while every request times out (#336).
 
 So the worker also says, from its own event loop, that the loop is turning.
 uvicorn already has the hook: `Config.callback_notify` is awaited by
-`Server.on_tick` every `timeout_notify` seconds, which is the integration point
-systemd's watchdog uses. The callback stamps `time.monotonic()` into a shared
-cell; the supervisor reads the cell on the same quiet poll as everything else,
-and replaces a worker whose loop has not turned for `WEDGED_AFTER` seconds.
+`Server.on_tick` once `time.time()` has advanced past `timeout_notify` since the
+last beat, which is the integration point systemd's watchdog uses. The callback
+stamps `time.time()` into a shared cell - the same wall clock the cadence is
+gated on, so the verdict cannot diverge from it (#1080); the supervisor reads
+the cell on the same quiet poll as everything else, and replaces a worker whose
+loop has not turned for `WEDGED_AFTER` seconds.
 
 Four properties of that choice are the reason for it:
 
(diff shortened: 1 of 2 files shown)


OchnikBartek:
@codex review

PR #1130, opened 2026-08-26
perf(organizations): batch the vault and org listing N+1 queries

### LINKED ISSUE #953: Two listings query per row where the code beside them batches
Two list endpoints run a query per row. In one of them the same function batches
two other lookups and documents doing so.

**Severity: Medium** · AUD-007, audit of 2026-08-19

## 7a. The vault listing

`backend/app/services/organization_secret.py:121`

Lines 99-119 batch the member identities and the grant counts, with a comment
saying why:

> How far each key reaches, in one grouped query rather than one per row. A
> listing that only says "private" cannot distinguish a key nobody else has from
> one five people were handed.

And then the loop at `:121` calls `organization_secret_repo.agents_using(...)`
per secret. An organization with thirty secrets loads its Vault page in
thirty-one queries. The rule was known and applied twice out of three times.

## 7b. The organization listing

`backend/app/services/organization.py:101` — `list_for_user` runs **two** queries
per organization, and one of them is redundant:

```python
for org in orgs:
    membership = await member_repo.get(self.db, organization_id=org.id, user_id=user_id)
    count = await organization_repo.count_members(self.db, org.id)
```

`organization_repo.list_for_user` (`repositories/organization.py:63`) already
joins `OrganizationMember` on that user to decide the row belongs in the result,
so the role it re-fetches was within reach of the first query. A user in eight
organizations costs 1 + 16 queries where 2 would do, on the endpoint the
organization switcher reads.

## How it fails

No failure — latency proportional to row count on two pages that should be flat.
It matters because both are read on navigation rather than on demand.

## Fix

The pattern is in the file above the bug.

```python
# vault: one grouped count keyed by secret id, like `count_for_resources` above it
usage = await organization_secret_repo.agents_using_many(
    self.db, organization_id=ctx.organization_id, secret_ids=[s.id for s in secrets]
)

# organizations: return the role from the join that already has it, and one
# grouped count for the sizes
select(Organization, OrganizationMember.role, func.count(...)).join(...).group_by(...)
```

## Acceptance criteria

- [ ] Both endpoints issue a constant number of queries regardless of row count
- [ ] Regression test: a query-count assertion on each, with two rows and with twenty

OchnikBartek:
## What

Two list endpoints ran a query per row where the code beside them already batched (#953).

**Vault listing.** It resolved emails and grant counts in one grouped query each, then asked which agents bind each secret **one secret at a time** — a thirty-secret vault loaded in thirty-one queries. `agents_using_for_secrets` answers the whole page in one: it unnests each agent's draft-spec capabilities and matches the bound `secret_id` against the requested set, grouped back per secret in Python. Semantically the same as the per-secret `@>` containment it replaces — a `CASE` guards `jsonb_array_elements` off a spec whose `capabilities` is not an array (the rows the old containment skipped) — and it keeps the tenant scope and name ordering.

**Organization listing.** `list_for_user` ran two queries per org: one to re-read the caller's membership for its role, one to count members. The role now rides the membership join the listing already makes, and `member_counts_for` counts a whole page in one grouped read. The per-row `member_repo.get` (redundant — the org came from that very join) and `count_members` are gone from the listing; `count_members` stays for the single-org read.

## Tests

Real-Postgres integration tests pin the batched SQL — grouping by secret, the non-array guard, dedup, tenant isolation, the role-carrying join, the grouped counts. A service test asserts the listing no longer makes a per-row `member_repo.get`/`count_members` call. `make lint` + `make test` green, coverage 100% including the gated `agents_using_for_secrets`.

Closes #953

### DIFF
--- a/backend/app/repositories/organization.py
+++ b/backend/app/repositories/organization.py
@@ -83,14 +83,37 @@ async def reassign_creator(
     return org
 
 
-async def list_for_user(db: AsyncSession, user_id: UUID) -> list[Organization]:
+async def list_for_user(db: AsyncSession, user_id: UUID) -> list[tuple[Organization, str]]:
+    """Each organization the user belongs to, with the role their membership carries.
+
+    The role comes off the same membership row the join already filters on, so a
+    caller needs no second per-organization lookup to learn it (#953).
+    """
     result = await db.execute(
-        select(Organization)
+        select(Organization, OrganizationMember.role)
         .join(OrganizationMember, OrganizationMember.organization_id == Organization.id)
         .where(OrganizationMember.user_id == user_id)
         .order_by(Organization.is_personal.desc(), Organization.created_at.asc())
     )
-    return list(result.scalars().all())
+    return [(org, role) for org, role in result.all()]
+
+
+async def member_counts_for(db: AsyncSession, org_ids: list[UUID]) -> dict[UUID, int]:
+    """How many members each of the named organizations has, in one grouped query.
+
+    One read for a whole page of organizations rather than a `count(*)` per row
+    (#953). An organization absent from the result has no members; every id the
+    caller passes for a listing has at least the caller, so in practice each is
+    present.
+    """
+    if not org_ids:
+        return {}
+    result = await db.execute(
+        select(OrganizationMember.organization_id, func.count(OrganizationMember.id))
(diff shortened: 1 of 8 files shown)


OchnikBartek:
@codex review

OchnikBartek:
@codex review

Issue #1134, opened 2026-08-26
Concurrent self-deletes of mutually co-owned shared orgs can deadlock


OchnikBartek:
**Found reviewing #1115** (the row locks it adds), which builds on #9.

## What's wrong

`UserService.delete` now takes `SELECT ... FOR UPDATE` on the user row, and
for a **shared** org the user solely created, `_release_owned_rows` hands the
`created_by_user_id` off to a co-owner via `reassign_creator`, which takes
`FOR KEY SHARE` on the *heir's* user row.

So `delete` acquires locks in the order (self, then heir). Two users who
co-own each other's shared orgs, deleting at the same moment, can deadlock:

- request A: `FOR UPDATE` on user A → wants `FOR KEY SHARE` on user B (its heir)
- request B: `FOR UPDATE` on user B → wants `FOR KEY SHARE` on user A (its heir)

Postgres detects the cycle and aborts one with `40P01`, so it **cannot hang** —
but the aborted request surfaces as a 500 rather than a clean result. The
pre-#1115 code would have 500'd on the integrity violation anyway, so this is a
new *shape* of an already-500 path, not a regression in behaviour a user can
reach without two simultaneous self-deletions of mutually-co-owned shared orgs.

The personal-org teardown path (user → own org) has no reverse locker and is
deadlock-free; only the shared-org reassignment introduces the second lock.

## Where

- `backend/app/services/user.py` — `UserService.delete` (FOR UPDATE on the user)
  → `_release_owned_rows` → `organization_repo.reassign_creator` (FK to the heir).

## How I'd know it's fixed

- [ ] Two concurrent `delete`s of users who each solely own a shared org the
      other co-owns either both resolve or one gets a clean domain refusal —
      never a `DeadlockDetected` 500. (Lock the two user rows in a stable order,
      e.g. by id, before reassigning, so the cycle cannot form.)

Rare, so `severity:low`. Pre-existing family: check-then-act / delete-governance
cluster (#17, #9, #1115, #941).

PR #1135, opened 2026-08-26
fix(concurrency): row-lock the reconcile before an org or user delete

### LINKED ISSUE #1115: Deletion reconcile-then-delete still 500s under a concurrent insert
**Found by codex review of #1112 (#9).** Same check-then-act class as #17.

## What's wrong

#9 reconciles the delete/CHECK cascades in the service before the row goes, but
the reconcile-then-delete is a check-then-act with no row lock, so a concurrent
insert reopens the exact 500 #9 closed:

- **Org delete** — `OrganizationService.purge` (`backend/app/services/organization.py:322`)
  lists org-scoped collections, then deletes the org. A member creating an
  org-scoped knowledge base after the list but before the `DELETE organizations`
  gets `ON DELETE SET NULL` on the new row → `ck_knowledge_bases_org_scope_has_org`
  violation → 500.
- **User delete** — `UserService._release_owned_rows` /`.delete`
  (`backend/app/services/user.py:311`) reads the user, promotes their private
  secrets and hands off their orgs, then deletes. A private secret or personal
  org created concurrently (referencing that user) after the reconcile lands a
  fresh CHECK/RESTRICT blocker → 500.

## Fix

Lock the row the reconcile is about before enumerating, so a concurrent insert
holding an FK key-share lock waits:

- `organization_repo.get_by_id_for_update` before listing collections in the
  teardown path.
- Add a `for_update` variant to `user_repo.get_by_id` (mirroring the ones #17
  added to `get_approval`/`get_by_token`) and take it at the top of
  `UserService.delete`.

## How I'd know it's fixed

- [ ] Integration test: a concurrent org-scoped KB insert during an org delete →
      the delete still succeeds, exactly one of the two wins.
- [ ] Integration test: a concurrent private-secret insert during a user delete →
      the delete still succeeds.

Part of the check-then-act cluster (#17). Refs #9.

OchnikBartek:
## Summary

#9 reconciles the FK/CHECK cascades before a delete, but reconcile-then-delete is check-then-act with no row lock, so a concurrent insert reopens the exact 500 #9 closed (#1115):

- an **org-scoped KB** inserted between `purge`'s collection list and its `DELETE organizations` is `SET NULL`-ed into a `ck_knowledge_bases_org_scope_has_org` violation;
- a **private secret / personal org** inserted between a user delete's reconcile and its `DELETE users` lands the same CHECK/RESTRICT blocker.

## Changed

- `OrganizationService.purge` and `UserService.delete` take `SELECT ... FOR UPDATE` on the row they are about **before** enumerating. A concurrent child insert takes `FOR KEY SHARE` on that parent through its FK, which `FOR UPDATE` conflicts with, so the insert waits and the reconcile sees every child (a fresh read under READ COMMITTED).
- Adds `user_repo.get_by_id_for_update`, mirroring `organization_repo.get_by_id_for_update`.

## Testing

- Two integration regression tests (`TestConcurrentInsertsDuringDeletion`) hold an uncommitted child insert open and race a delete against it — both **fail against the lock-free code** (the delete 500s on the CHECK once the insert commits) and pass with the lock.
- Affected delete unit tests retarget `get_by_id_for_update`.

## Notes for reviewers

Found and filed while reviewing: a rare deadlock when two users who co-own each other's shared orgs self-delete simultaneously → **#1134** (Postgres aborts one with a 500; pre-existing-shape, out of scope here).

Stacked on `fix/9-delete-cascade-check-reconcile`, because it builds on that reconcile. Retarget to `main` once #9 (#1112) merges.

Closes #1115

### DIFF
--- a/backend/app/repositories/user.py
+++ b/backend/app/repositories/user.py
@@ -17,6 +17,12 @@ async def get_by_id(db: AsyncSession, user_id: UUID) -> User | None:
     return await db.get(User, user_id)
 
 
+async def get_by_id_for_update(db: AsyncSession, user_id: UUID) -> User | None:
+    """Fetch a user row and acquire a SELECT FOR UPDATE lock."""
+    result = await db.execute(select(User).where(User.id == user_id).with_for_update())
+    return result.scalar_one_or_none()
+
+
 async def get_by_email(db: AsyncSession, email: str) -> User | None:
     """Get user by email."""
     result = await db.execute(select(User).where(User.email == email))
--- a/backend/app/services/organization.py
+++ b/backend/app/services/organization.py
@@ -318,7 +318,15 @@ class OrganizationService:
         `ON DELETE SET NULL`, and nulling an org-scoped row violates
         `ck_knowledge_bases_org_scope_has_org` (#9). Personal collections that
         merely carry this org's id are left to the `SET NULL`.
+
+        FOR UPDATE first: listing the collections and then deleting the org is
+        check-then-act, so an org-scoped collection inserted concurrently (it
+        takes FOR KEY SHARE on this row) would slip in between the list and the
+        DELETE, be nulled by the `SET NULL`, and violate the same CHECK. The lock
+        makes that insert wait, so the list sees every collection there is - a
+        fresh read under READ COMMITTED (#1115).
         """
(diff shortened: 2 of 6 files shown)


OchnikBartek:
@codex review

Issue #1137, opened 2026-08-26
Org teardown leaves post-commit and #913-rooted reachability residuals


OchnikBartek:
**Found reviewing #1116** (org/collection teardown), which builds on #9. #1116
closes the two concrete harms — a shared physical table dropped for a co-tenant,
and a deleted KB's document rows/files left reachable by `collection_name` — but
review named four residuals it cannot close on that branch. Filing so they are
tracked rather than implicit.

## Residuals

1. **External side effects commit outside the request transaction.**
   `OrganizationService.purge` drops the vector table (`PgVectorStore.delete_collection`
   commits its `DROP TABLE` on its own session) and unlinks stored uploads
   (`_purge_documents`), both before the request session commits the relational
   deletes. If that final commit fails, the org/KB/doc rows roll back and come
   back pointing at vectors and files that are already gone. A retry-safe
   post-commit cleanup (drop + unlink deferred to after commit, e.g. via
   `spawn_after_commit`) would close the window. Narrow today (#1115's lock makes
   a mid-teardown constraint failure unlikely), no regression in class (the old
   code also dropped on its own session), but undocumented until #1116.

2. **NULL-`knowledge_base_id` docs sharing the collection name still leak.**
   `rag_document_repo.delete_by_knowledge_base` keys on `knowledge_base_id`, so a
   `rag_documents` row with `knowledge_base_id IS NULL` but the same
   `collection_name` (the admin local-sync path creates these deliberately -
   `app/worker/tasks/rag_tasks.py`) survives with its name intact and is
   reachable by a later same-named KB. Widening the delete to `collection_name`
   is unsafe while the name is not tenant-unique (would hit a co-tenant's rows).

3. **A deleted tenant's vectors survive, searchable, in a shared/kept table.**
   When a table is shared and therefore kept, the deleted org's *rows* and files
   go but its *vectors* stay in the physical `rag_<name>` table, queryable by the
   surviving co-tenant. No per-document vector cleanup exists, and a shared table
   has no tenant column to scope one.

4. **TOCTOU between the reference check and the drop.**
   `_collection_still_referenced` reads under READ COMMITTED and the #1115 lock is
   on the org row, not the collection name; a second org concurrently creating a
   KB with the same name (and its `CREATE TABLE IF NOT EXISTS` backing the same
   physical table) between the check and the `DROP TABLE` could have its table
   dropped.

## Root

2, 3 and 4 are all rooted in `collection_name` not being tenant-unique - the
[...]

Issue #913, opened 2026-08-18
rag: per-collection resolution picks a KB by non-unique collection_name


OchnikBartek:
`collection_name` on `knowledge_bases` is indexed but **not unique** — the repo ships `list_by_collection_name` whose docstring says "two organizations can name a collection the same thing ... An authorization check built on [`get_by_collection_name`] is right in every single-tenant test and wrong in production".

Both per-collection resolvers rely on `get_by_collection_name(...).first()`:
- `app/services/embedding_resolution.py::embeddings_for_collection`
- `app/services/rerank_resolution.py::reranker_for_collection`

They reach the DB from contexts with no org in hand (an agent mid-run, a direct search), so when two orgs share a `collection_name` the `.first()` may return a different KB row than the one `CollectionAccessService` authorized — resolving another tenant's embedding/rerank config, unsealing that org's vault key (under its own scope) and billing that org. `PgVectorStore` is itself keyed by `collection_name`, so a cross-org collision is also a store-level isolation question, not only a resolver one.

`collection_name` is normally `<slug>_<rand6>`, so collisions require a caller to set `collection_name` explicitly on both rows — narrow, but supported.

Raised by the PR #911 reviewer (codex, P1) against `rerank_resolution.py:105`; applies identically to embeddings and predates the rerank work, so it is a resolution-layer fix rather than a rerank one.

**Options**:
- thread the authorized org (or the resolved KB identity) through the vector store's resolver interface so resolution is scoped; or
- a uniqueness constraint / disambiguation on `collection_name`.

**Fixed when**: a `collection_name` shared across two orgs resolves each org's own embedding + rerank config (integration test with two tenants).

Found reviewing #142 (PR #911). RAG cluster in #168 (#27, #142, #147, #148, #548).

OchnikBartek:
## Update — codex re-flagged the remaining gap on #911 (2026-08-21)

The org-scoping added on `feat/rag-reranker` (`get_for_collection`, resolvers
now taking `organization_id`) narrowed the exposure but did **not** close it.
A code review surfaced the precise residual:

**The access check and the resolution use different tie-breaks on a shared name.**

- Access (`CollectionAccessService._first_readable`) returns the **first
  readable** row for a `collection_name`.
- Resolution (`knowledge_base_repo.get_for_collection`) returns the caller's
  **own-org row**, unconditionally, regardless of readability.

When an `APP`-scoped KB and a **restricted** `ORG` KB in the caller's own org
[...]

PR #1138, opened 2026-08-26
fix(rag): purge documents and guard shared tables on org delete

### LINKED ISSUE #1116: Org/collection deletion vs collection_name reuse: shared-table drop, non-atomic drop, orphaned docs
**Found by codex review of #1112 (#9).** Rooted in `collection_name` not being
tenant-unique (`knowledge_base_repo.list_by_collection_name` documents that two
orgs can name a collection the same thing); the tenant-scoping work is #913.

Three related gaps in how a deleted org/collection cleans up its vector table and
documents. Pre-existing in shape — the single-collection `drop_collection` route
(`app/api/routes/v1/rag.py:200`) has the same characteristics — but #9's org
delete widens the surface, so worth doing together.

## 1. A shared physical table is dropped for both tenants

`OrganizationService.purge` (`app/services/organization.py:329`) drops
`rag_<collection_name>`. If two orgs hold KB rows with the same `collection_name`,
that is one physical table backing both, so deleting either org destroys the
survivor's vectors. Key backing tables by a tenant/KB-unique identifier, or check
for references outside the org before dropping.

## 2. The vector drop is not atomic with the row deletes

`PgVectorStore.delete_collection` commits its `DROP TABLE` on its own
`AsyncSession`, while the KB-row and org-row deletes are in the request session.
A later failure (a subsequent collection, the org delete, request cancellation,
the final commit) rolls back the relational side while the table stays dropped -
a live KB pointing at vectors that are gone. Defer the drop until the relational
transaction commits (a retry-safe post-commit cleanup), or accept and document it.

## 3. Tracked documents are orphaned and reachable by name

Deleting an org-scoped KB nulls `rag_documents.knowledge_base_id`, and the org
delete nulls `organization_id`, leaving the document rows and stored uploads
behind. Because document authorization and listing key on `collection_name`
alone, a later KB with the same permitted name can list and download the deleted
tenant's retained files. Purge the KB's document records and storage before
removing its identifiers.

## How I'd know it's fixed

- [ ] Deleting an org with a same-named collection in another tenant leaves the
      other tenant's vectors intact.
- [ ] A failed org delete leaves no KB pointing at a dropped table.
- [ ] Deleting an org removes its documents' rows and stored bytes; a later
      same-named collection cannot read them.

Refs #9. Related to the collection-name tenant-scoping in #913.

OchnikBartek:
## Summary

Tearing down an org's org-scoped collections (#9's `purge`) had three gaps, all raised in review (#1116):

1. **Shared physical table dropped for both tenants.** `collection_name` is not tenant-unique (#913), so two orgs can back onto one `rag_<name>` table; deleting either destroyed the other's vectors.
2. **Tracked documents orphaned and reachable by name.** `rag_documents` authorizes on `collection_name`, so a row outliving its KB was listable/downloadable by a later collection permitted the same name.
3. **The vector drop was interleaved with the row deletes** (not atomic; a later failure left a live KB pointing at a dropped table).

## Changed

- The physical table is dropped **only when no other KB still references the collection** (`_collection_still_referenced`), and **last** — after the relational deletes flush, so a failure among them aborts before any table is gone.
- Each KB's document rows and their stored uploads are deleted **before** its identifiers go, keyed on `knowledge_base_id` so a shared collection's co-tenant rows are untouched (`rag_document_repo.delete_by_knowledge_base`, RETURNING the storage paths to unlink).

## Residuals (documented on `purge`, tracked in #1137)

The drop and the file unlinks still commit outside the request transaction; and three reachability residuals remain — a NULL-`knowledge_base_id` doc sharing the name, the deleted tenant's vectors inside a *kept* shared table, and a TOCTOU on the reference check — **all rooted in the `collection_name` tenant-scoping of #913**. #1137 tracks them.

## Testing

- Integration: a deleted org's document rows and stored file are gone and unreachable by name (fails without the fix — the row survives).
- Integration: a table shared with a co-tenant survives the delete (fails without the fix — the old code drops it unconditionally).

## Notes for reviewers

Stacked on `fix/1115-delete-reconcile-row-locks` (which builds on #9). Retarget to `main` as the stack merges.

Closes #1116


OchnikBartek:
@codex review

Issue #1095, opened 2026-08-24
drain() on shutdown does not fully quiesce in-flight background work


OchnikBartek:
## What breaks

`background.drain()` and the shutdown ordering in `app/main.py`'s lifespan do
not fully quiesce in-flight background work before resources are torn down.
Three related gaps, all found reviewing #11 (which wired `drain()` into the
lifespan; these are pre-existing behaviours of `drain()` and the shutdown
sequence that the wiring now depends on):

1. **Tasks spawned during the drain are not awaited.** `drain()` snapshots
   `_running` once (`backend/app/core/background.py:163`) and waits only on that
   snapshot. A task in the snapshot can hand off new work while draining — e.g. a
   channel task finishing an agent run that requests approval or exceeds budget:
   `NotificationService._send()` spawns each email. The parent finishes, the
   snapshot completes, `drain()` returns, and the newly spawned email (documented
   as possibly taking ~10s) is cancelled at process exit.

2. **Cancelled tasks are not awaited to a terminal state.** After the timeout,
   `drain()` calls `task.cancel()` (`background.py:172-173`) and returns without
   awaiting the cancelled tasks. The lifespan then immediately disposes the vector
   store, Redis and the DB engine (`app/main.py`, after `await background.drain()`),
   while a cancelled task is still unwinding — e.g. a webhook task's cancellation
   handler awaits `release_delivery()` on the shared Redis client and then unwinds
   its DB context. Those cleanup awaits race the Redis/DB shutdown, leaving the
   delivery claim unreleased or the cleanup failing against a closed resource.

3. **Deferred stream opens can reopen intake during teardown.** A polling bot
   created/activated/updated just before shutdown leaves a committed, tracked
   `open_inbound_stream` task that may not have run yet. The lifespan stop loops
   (`app/main.py:181-186`) stop only the adapter tasks that already exist; then the
   drain waits for `open_inbound_stream()`, whose final step is
   `adapter.start_polling()` — creating a *new* adapter polling/socket task after
   the stop loops have passed, and adapter tasks are not tracked by `_running`.
   Intake is reopened during teardown (e.g. Slack acks a new event, then loses it
   at exit).

## Where

- `backend/app/core/background.py:154-173` (`drain()`)
- `backend/app/main.py` lifespan shutdown (stop loops `:181-186`, then `await background.drain()`, then resource disposal)

## How we'd know it's fixed

[...]

PR #1140, opened 2026-08-26
fix(channels): stop a deferred stream reopening intake at shutdown

### LINKED ISSUE #1119: Deferred open_inbound_stream can reopen channel intake during shutdown
**Split out of #1095 so it can close with #1097.** #1097 fixed the two
`drain()`-internal gaps (tasks spawned during drain; cancelled tasks awaited to a
terminal state). This is the third gap, which is a shutdown-**ordering** concern,
not a `drain()` one.

## What's wrong

A polling bot created/activated/updated just before shutdown leaves a committed,
tracked `open_inbound_stream` task that may not have run yet. The lifespan stop
loops (`backend/app/main.py:181-186`) stop only the adapter tasks that already
exist; then `drain()` waits for `open_inbound_stream()`, whose final step is
`adapter.start_polling()` - creating a **new** adapter polling/socket task after
the stop loops have passed. Adapter tasks are not tracked by `_running`, so
intake is reopened during teardown (e.g. Slack acks a new event, then loses it at
exit).

Must be fixed on the shutdown path only: rejecting the deferred open must not
touch the in-process `drain()` the RAG sync command (`app/commands/rag.py`) issues
while the process keeps running.

## Where

- `backend/app/main.py` lifespan shutdown (stop loops `:181-186`, then
  `await background.drain()`)
- the deferred `open_inbound_stream` handoff

## How I'd know it's fixed

- [ ] A bot activated moments before shutdown does not reopen a polling/socket
      task after the stop loops run.
- [ ] The in-process `drain()` path (rag command) is unaffected.

Refs #11, #1095. Was gap 3 of #1095.

OchnikBartek:
## Summary

A polling bot activated moments before shutdown leaves a committed, tracked `open_inbound_stream` task that may not have run yet. The lifespan stops the adapter tasks that exist, then `background.drain()` awaits that task — whose final step `adapter.start_polling(...)` creates a **new, untracked** adapter polling/socket task **after** the stop loops passed, reopening intake at exit (#1119). Was gap 3 of #1095.

## Changed

- The supervisor carries a `_shutting_down` flag. `open_inbound_stream` declines to open when it is set — re-checked after `stop_polling`, so a task suspended there when shutdown begins cannot slip a reopen through either.
- The lifespan sets it (`begin_shutdown()`) as the **first** shutdown statement, before the stop loops and `drain()`, and clears it (`allow_intake()`) at startup so a following lifespan (a test, a reload) serves again.
- The flag is the lifespan's alone: the in-process `drain()` the RAG sync command issues while the server keeps serving never touches it, so a legitimate reopen still opens (the "shutdown path only" constraint the issue requires).

## Testing

- A deferred open during shutdown does not reopen (fails without the guard); one suspended at `stop_polling` when shutdown begins does not either; a fresh startup re-permits.
- `conftest.py` resets the process-global flag after every test, so the real-lifespan drain test (`test_lifespan_drain.py`) cannot leak `_shutting_down=True` into a later test that opens a stream under the randomized parallel suite.

## Notes for reviewers

Stacked on `fix/11-drain-background-on-shutdown` (which adds the lifespan `drain()` this orders around). Retarget to `main` once #11 (#1089) merges. Conceptually closes with #1097 (#1095's drain-internal gaps).

Closes #1119


OchnikBartek:
@codex review

Issue #188, opened 2026-08-05
codespell fails over the repository, and nobody sees it because the hook only reads changed files


DEENUU1:
`uv run --project backend pre-commit run codespell --all-files` is red on `main`:

```
frontend/src/lib/tool-steps.ts:148: unparseable ==> unparsable
```

A docstring on `McpCall.domain`, from #119. Nothing runs codespell over the
whole tree — the hook is per-file at commit time and CI does not run pre-commit
at all — so this sits there until somebody edits `tool-steps.ts` for an
unrelated reason and their commit is refused by a word they did not write.

## The sequence

```bash
git switch main
uv run --project backend pre-commit run codespell --all-files   # red
```

## How you would know it was fixed

`pre-commit run --all-files` is green on `main`, and stays green — which
probably means CI running it, or at least a note that it is expected to pass
over the whole tree. Two separate things: the word, and the fact that no gate
covers the tree.

Pre-existing, found while reconciling `make check` with CI for #143. Not fixed
there: #143's diff is the Makefile and the workflow, and a one-word edit in a
frontend module does not belong in it.

PR #1144, opened 2026-08-26
ci: run yamlfmt, zizmor and the pre-commit basics over the tree

### LINKED ISSUE #203: yamlfmt, zizmor and the pre-commit basics read only changed files, and no gate reads the tree
**Pre-existing, found fixing #188.** That branch closes the gap for one hook and not for the class, deliberately: its diff is the Makefile, the workflow and one word.

#188 was "codespell is red over the tree and no gate reads the tree". The fix — `make lint-spelling`, in `make lint` and in CI's `lint` job — is specific to codespell. Three other hooks in `.pre-commit-config.yaml` have exactly the same property and no equivalent:

| Hook | Reads | Runs in CI |
|---|---|---|
| `yamlfmt` | the YAML files a commit touches | no |
| `zizmor` | the workflow files a commit touches | no |
| `pre-commit-hooks` basics (`end-of-file-fixer`, `trailing-whitespace`, `check-yaml`, `check-json`, `check-toml`, `check-merge-conflict`, `check-added-large-files`, `detect-private-key`) | the files a commit touches | no |

**All of them are green over the tree today** — `SKIP=no-commit-to-branch pre-commit run --all-files` passes every one of them as of `d0f200d`. So this is debt prevention, not a live failure, which is why it is `severity:low` and was not folded into #188.

## The sequence that makes it matter

The per-file scope is fine while the rules are fixed. It stops being fine when the rules move, and Dependabot moves them:

1. A `rev:` bump brings a new `zizmor` rule (or a `yamlfmt` formatting change).
2. Every workflow file in the tree now violates it. Nothing notices, because no commit has touched a workflow.
3. Weeks later somebody edits one line of `ci.yml` and their commit is refused by a finding that has nothing to do with their change — the shape of #188, one tool over.

`detect-private-key` has a second version of it: a file committed with `--no-verify`, or added through the GitHub web UI, is never seen by any hook at all.

`SKIP=no-commit-to-branch pre-commit run --all-files` is green in CI and stays green — one target, reachable from `make check`, so `backend/tests/test_ci_parity.py` holds it in place the way it now holds `lint-spelling`.

Worth deciding rather than assuming, when this is picked up:

- **What the target skips.** `ruff`, `ty`, `prettier`, `eslint` and `tsc` are already gated by `lint-backend` / `lint-frontend` and would double the lint job's time; `no-commit-to-branch` **fails by design** when CI checks out `main` on the push trigger, so it has to be skipped or the merge build breaks.
- **Whether the fixers belong in a gate at all.** `end-of-file-fixer` and `yamlfmt` rewrite files; in CI they can only report, so the target has to fail on a modification rather than commit one.
- A `--all-files` run in the `lint` job versus a scheduled weekly one. The scheduled version keeps the pull request fast and still catches a rev bump, but reports to nobody's pull request — `dependency-freshness.yml` already has this shape and opens an issue.

Cluster: **Tests, CI, dependencies** (#168). Nearest relative: **#188** — same defect, one hook; also #143, which is why the Makefile and the workflow are now one definition each.

Refs #188, #143, #168

OchnikBartek:
## What changed

`yamlfmt`, `zizmor` and the `pre-commit-hooks` basics (`end-of-file-fixer`, `trailing-whitespace`, `check-yaml/json/toml`, `check-merge-conflict`, `check-added-large-files`, `detect-private-key`) read only the files a commit touches, and no gate ever read the whole tree with them. That is fine while the rules are fixed and stops being fine when Dependabot bumps a `rev:`: a new `zizmor` rule makes every workflow in the tree violate it, nothing notices because no commit has touched a workflow, and weeks later an unrelated one-line edit to a workflow is refused by a finding that has nothing to do with it — the shape of #188, one tool over.

`make lint-spelling` already solved exactly this for codespell. This gives the rest the same treatment.

- **`make lint-precommit`** — `pre-commit run --all-files` with `SKIP` dropping the hooks another target already gates over the tree (codespell → `lint-spelling`; ruff/ty/vulture/guards/prettier/eslint/tsc/i18n → `lint-backend`/`lint-frontend`) plus `no-commit-to-branch`, which fails by design when CI checks out `main`. Neither doubles the gated hooks' time nor lets a fixer (`ruff --fix`, `prettier --write`) rewrite a file mid-check.
- A fixer that *does* run (`end-of-file-fixer`, `yamlfmt`) reports rather than commits: pre-commit exits non-zero on a modification, which is the failure CI needs.
- Wired as a dependency of `lint` (so `make check` reaches it) and as a step of CI's `lint` job. `test_ci_parity.py` holds both directions, the same way it holds `lint-spelling`.
- `docs/branching.md` and `docs/commands.md` updated to list the new tree gate.

## How it was verified

- `SKIP=… pre-commit run --all-files` green over the tree today; `make lint-precommit` exits 0 (yamlfmt + zizmor over the edited `ci.yml` included).
- `test_ci_parity.py` passes (19); `make lint-spelling` and `make docs-build --strict` clean.

Closes #203


OchnikBartek:
@codex review

Issue #162, opened 2026-08-04
e2e: the cost journey spec flakes on main after passing on the PR


DEENUU1:
`journey.spec.ts:51 › an agent goes from a stored key to a run with a cost` failed on `main` at `5780012`, then passed on a re-run of the same commit. 91 of 92 specs passed both times.

Not covered by #130, #132 or #154 — those name the vault rotation spec, the two seed/setup steps and the sharing/skills specs. This is a fourth spec with the same shape of failure.

## What happens

```
✘  42 [chromium] › e2e/journey.spec.ts:51:5 › an agent goes from a stored key to a run with a cost (15.8s)
   Error: expect(locator).toBeVisible() failed
   Error: element(s) not found
```

Same signature as the other three: `toBeVisible()` on something that should be there, `element(s) not found`, and no indication of *why* it was not.

## Why it looks like a flake rather than a regression

The commit was the TypeScript 5.9.3 → 6.0.3 bump (#118), and:

- its own PR run passed `e2e` on the identical tree
- the re-run of `5780012` passed
- `9087ac3`, one commit later, passed on the first attempt

Three passes and one failure across the same code. And a TypeScript bump cannot reach runtime behaviour: `tsc --noEmit` is type-checking only, the build output was byte-identical in behaviour, and nothing shipped differently.

## Why it is worth its own issue

`#132`'s diagnosis — "the step asserts a just-created row is on screen and says nothing useful when it is not" — probably applies here too, but this spec is doing something more: it walks a stored key through to a run with a **cost**, which crosses the model provider, the run row and the spend record. So the set of things that can be late is bigger, and "the refetch landed stale" is only one candidate.

Worth knowing which locator it was waiting on. The failure gives a screenshot and an `error-context.md` in `test-results/`, and neither is in the log — an artifact upload on failure would turn this from "it flaked again" into a diagnosis.

## How I would know it was fixed

The spec states what it is waiting for and why it is not there — a named locator and a reason, not `element(s) not found` — and it passes twenty consecutive runs on `main`.

## Note on the pattern

Four specs now flake this way. Individually each reads as bad luck; together it looks like every one of them is asserting on a screen that has not finished settling, and the suite has no shared "wait for this mutation to have landed" primitive. That may be the actual fix, rather than four separate hardenings.

Found while releasing 0.0.6. Pre-existing — this branch did not cause it.

OchnikBartek:
[...]

PR #1145, opened 2026-08-26
test(e2e): capture a trace on the one failure a flake gets

OchnikBartek:
## What changed

`playwright.config.ts` had `trace: "on-first-retry"` and `video: "on-first-retry"`, but `retries` is **0** — deliberately, so a real failure cannot go green on a second attempt. With no retry, `on-first-retry` never fires: on a flake the CI job kept a screenshot and nothing else. The **trace** is the one artifact that answers #162's question — *which locator was it waiting on, and why* (it carries the DOM timeline, the network log and the console) — and it was absent from exactly the runs that needed it.

- `trace` and `video` → **`retain-on-failure`**: captured on the first (only) failure, kept for the failed test alone.
- The e2e job already uploads `playwright-report/` on failure and the HTML report embeds the trace, so **no workflow change is needed** to carry it.

## Scope — this Refs, it does not Close

This does **not** fix the flake. Its root cause is unknown until a failure is actually captured, and this is what makes that capture possible. #162's own done-criteria — *the spec names the locator and reason, and passes twenty consecutive runs on `main`* — can only be met once a real trace has been read. So the issue stays open after this merges; it now produces a diagnosable artifact the next time `journey.spec.ts` (or any of the four specs flaking this way) fails.

## How it was verified

- `tsc --noEmit` clean — both literals are valid `TraceMode` / `VideoMode`, so the config still compiles.
- The comment guard passes. No product or spec behaviour changes; only which artifacts survive a failure.

Refs #162

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -91,14 +91,24 @@ export default defineConfig({
     /* Base URL to use in actions like `await page.goto('/')`. */
     baseURL: BASE_URL,
 
-    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: "on-first-retry",
+    // `retain-on-failure`, not `on-first-retry`, because `retries` above is 0: a
+    // test that flakes never gets a second attempt, so `on-first-retry` collected
+    // a trace on none of them and the one artifact that names which locator was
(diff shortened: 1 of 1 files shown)


OchnikBartek:
@codex review

PR #1173, opened 2026-08-26
refactor(channels): lift the transport boilerplate into shared helpers

### LINKED ISSUE #565: Lift the channel transport boilerplate into base.py helpers

OchnikBartek:
Four transport blocks were copied across the three channel adapters, so a change to any had to be made at 6–10 sites by hand and a fourth channel would copy them all again.

## What changed

- **`split_thread(id) -> (channel, thread)`** in `base.py` — replaces the `channel, _, root = id.partition(":")` unpack at 3 Mattermost + 3 Slack sites; `channel_key` now calls it too, so "which channel is this" has one implementation.
- **`SlackAdapter._web(token)`** — the lazy `AsyncWebClient` import + construction, previously repeated at 9 sites (including a dynamic `__import__` in the Socket Mode path). The import stays deferred so a deployment running no Slack bot does not pay for the SDK at import time.
- **`TelegramAdapter._bot(token, *, default=None)`** — an async context manager wrapping `Bot(...)` and the `try/finally: bot.session.close()` that **10** methods repeated, each of which leaked a TLS session if the close was forgotten. `default` carries the Markdown parse mode the streamed replies ask for; plain sends pass none.
- **`MattermostAdapter._client()` / `_headers(token)`** — the `httpx.AsyncClient(timeout=…)` and the bearer header, each repeated at 10 sites, so the timeout and the auth shape now have one definition.

## Behaviour-preserving

Every session still closes on every exit path (normal, early `return`, exception, the `except`-clause `return`s); the no-default `Bot` construction is byte-identical; `split_thread` drops only the separator the old unpack already discarded.

## How it was verified

- An adversarial diff review across all six axes (session lifecycle on every exit path, `default` mapping, `split_thread` semantics, post-block value reads, the lazy-import path, leftover/undefined names) found **no behaviour change**.
- **540 channel tests green**; `make lint-backend` clean (ruff, ruff-format, ty, vulture, deptry, guards); `ty` adds zero diagnostics over baseline. The adapters are template-inherited and not on the coverage gate.

Enables #547 (one normaliser per adapter).

Closes #565

### DIFF
--- a/backend/app/services/channels/base.py
+++ b/backend/app/services/channels/base.py
@@ -73,7 +73,21 @@ def channel_key(platform_chat_id: str) -> str:
     in with the import - and two implementations of "which channel is this" is
     how one of them ends up asking Mattermost about a thread id.
     """
-    return platform_chat_id.partition(":")[0]
+    return split_thread(platform_chat_id)[0]
+
+
+def split_thread(platform_chat_id: str) -> tuple[str, str]:
+    """The chat and the thread a message's id folds together.
+
+    Slack packs `channel:thread_ts` into `platform_chat_id` and Mattermost packs
+    `channel:root_id` the same way; every other platform's id is the chat with an
+    empty thread. `channel_key` keeps the channel half alone; this keeps both, for
+    a reply that has to be posted back into the thread it came from. The empty
+    string when there is no colon is what an unthreaded reply passes on, which each
+    adapter reads as "top of the channel".
+    """
+    channel, _, thread = platform_chat_id.partition(":")
+    return channel, thread
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


OchnikBartek:
@codex review

Issue #106, opened 2026-08-02
Uploads bypass the active organization: bare fetch omits X-Organization-Id


DEENUU1:
## What happens

`apiClient` attaches `X-Organization-Id` from `useOrgStore` on every request (`src/lib/api-client.ts:70,85`), and `platformProxy` forwards it to the backend (`src/lib/platform-proxy.ts:93,104`). A handful of calls use a bare `fetch` instead and therefore send no header at all, so the backend falls back to the caller's personal organization.

For an upload that is a **write across the tenant boundary**: an admin viewing organization B uploads into a collection whose name also exists in their personal organization A, the file is ingested into A, and the page reports success under B.

## Where

| Call | File |
|---|---|
| `ingestFile` — the RAG upload | `frontend/src/lib/rag-api.ts:171` |
| KB document download | `frontend/src/lib/rag-api.ts:114` |
| File upload | `frontend/src/lib/file-api.ts:19` |
| KB document download (again) | `frontend/src/components/kb/file-viewer.tsx:169` |
| File preview fetches ×3 | `frontend/src/components/chat/file-preview-card.tsx:230,262,394` |

The upload paths are the ones that write; the rest read, which is the same leak in the milder direction.

## Why a bare fetch at all

`FormData` and binary responses — `apiClient` sets `Content-Type: application/json`. That is a reason not to use `apiClient` as-is, not a reason to drop the header: it can be read from `useOrgStore.getState().activeOrgId` at the call site, or `apiClient` can grow a path that leaves the body alone.

## Not from #105

`rag-api.ts`, `file-api.ts` and `file-preview-card.tsx` predate that branch; it was found reviewing it. Filed rather than folded in, because a cross-tenant write deserves its own change and its own regression test.

## Done when

Every browser-side call that reaches an org-scoped endpoint carries the active organization, with a test that a request made while organization B is active does not arrive scoped to A.

PR #1176, opened 2026-08-26
refactor(api): forward the sessions BFF routes through platformProxy

OchnikBartek:
The first cluster of #564 — collapsing the hand-rolled `backendFetch` forwarders into `platformProxy`.

## What changed

`sessions/route.ts` and `sessions/[id]/route.ts` each hand-rolled the forwarder block #564 is collapsing (read cookie → 401, set bearer, map `BackendApiError`, `NextResponse.json`) — exactly what `platformProxy()` does, **plus** the active-organization header, byte-accurate bodies and the `no-store` those hand-rolled copies drop (#546, #553, #106).

- Both replaced with a **`sessions/[[...path]]`** mount, the same shape `kb`/`rag` use.
- The `skip`/`limit` query and the id-escaping the old routes did by hand are the proxy's path-and-search passthrough now — no endpoint behaviour changed.
- Added the mount to the `proxy-routes.test.ts` registry (auth-refusal + five verbs) and removed the sessions cases from `cookie-routes.test.ts`, whose behaviour `platform-proxy.test.ts` covers generically.

## Scope — sessions only, and why

This is `Part-of #564`, not the whole thing. The other clusters have blockers that want more than a mechanical swap, documented on the issue:

- **`orgs/**`** — the cluster keeps a binary `[id]/avatar` route, which cannot coexist with an `orgs/[[...path]]` optional catch-all (Next.js `[id]`-vs-catch-all conflict); and the org-header the proxy adds (the #106 fix) is a behaviour change that needs an e2e/backend to confirm.
- **`admin/**`** — carries a `requireAdmin` frontend gate rather than the plain cookie check.
- **`me/mcp-connections/oauth/callback`** — a redirecting route, not a plain forwarder.

## How it was verified

`make lint-frontend`, `make test-frontend-cov` (100% lines/functions, 97.67% branches) and `make build-frontend` all green; the route tree builds with the new mount and no conflict.

Part-of #564


OchnikBartek:
@codex review

PR #1178, opened 2026-08-26
test(rag): construct PgVectorStore with the engine it now requires

OchnikBartek:
## `main` is red — this unblocks it

The last **three** pushes to `main` failed CI in the `test` job's `coverage-all` step:

```
PgVectorStore.__init__() missing 1 required keyword-only argument: 'engine'
```

`TestDeletingAnOrg` (in `test_deletion_reconciliation.py`, added by #1112) built `PgVectorStore` with no `engine` and closed it with `store.aclose()`. The store was reshaped to **borrow** an injected engine rather than build and dispose its own (#1084): `engine` is now a required keyword-only argument and there is no `aclose`. The two pull requests were each green against a base without the other, so the incompatibility only surfaced once both were on `main` — and every backend branch now inherits the red on merge (it is why #1173's `test` job failed).

## Fix

- Pass the test's own `engine` fixture to `PgVectorStore(...)` — the same engine the `factory` already uses.
- Drop the `try/finally: await store.aclose()`; the store borrows the engine and the `engine` fixture disposes it.

No coverage or assertion changes — the reconciliation the two tests prove is unchanged.

## Verified

As far as a laptop without a database allows: `ruff`, `ruff format` and `ty` clean, and `ty` no longer reports the missing argument or the absent `aclose` (6 diagnostics on the file → 2, both pre-existing `MagicMock` resolver typing). The integration run itself needs the database CI has — please let CI confirm before merge.

### DIFF
--- a/backend/tests/integration/test_deletion_reconciliation.py
+++ b/backend/tests/integration/test_deletion_reconciliation.py
@@ -179,42 +179,38 @@ class TestDeletingAnOrg:
             settings=app_settings.rag,
             embedding_service=MagicMock(),
             resolver=MagicMock(),
+            engine=engine,
         )
         factory = async_sessionmaker(engine, expire_on_commit=False)
         collection = f"kbnine{uuid.uuid4().hex[:12]}"
         table = store._table(collection)
-        try:
-            async with factory() as s:
-                user = _user()
-                s.add(user)
-                await s.flush()
-                org = await _org(s, user)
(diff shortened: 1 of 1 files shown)


OchnikBartek:
@codex review

OchnikBartek:
Superseded — #1177 already built the deletion tests' store on the fixture's engine on main, so this is redundant. Closing.

Issue #954, opened 2026-08-19
Resolve a spec's references in one query, not one each - including on every run


DEENUU1:
Six loops resolve a list of ids by fetching one row at a time. They are one
finding because they are one habit, and one of them is on the hottest path in the
product.

**Severity: Medium** · AUD-008, audit of 2026-08-19

## Where

- `backend/app/services/agent_runner.py:1401` — `_collection_names`, **per run**
- `backend/app/services/agent_registry.py:1005` (MCP servers), `:1104`
  (collections), `:1360` (delegates)
- `backend/app/services/mcp_connection.py:795`
- `backend/app/services/agent_environment.py:323`

## How it fails

Five of the six are publish-time or admin-path, where N is small and the cost is
invisible. The sixth is not: `_collection_names` runs inside `prepare`, which runs
on **every agent run**. So every chat turn pays one round trip per bound
collection before the model is ever called — an agent bound to five knowledge
bases adds five serial round trips to the front of every turn, to fetch rows that
are very likely already in the session's identity map.

## Fix

One `get_by_ids(ids)` per loop, then the same per-id validation over the returned
map. The tenant check and the "collection deleted after publish degrades the
agent's reach rather than failing the run" branch stay exactly as they are — they
just read from a dict.

```python
async def _collection_names(self, spec: AgentSpec, ctx: AuthContext) -> list[str]:
    found = await knowledge_base_repo.get_by_ids(self.db, spec.collection_ids)
    names: list[str] = []
    for collection_id in spec.collection_ids:
        collection = found.get(collection_id)
        if collection is None or collection.organization_id != ctx.organization_id:
            logger.warning(...)      # unchanged
            continue
        names.append(collection.collection_name)
    return names
```

## Acceptance criteria

- [ ] Each of the six resolves its references in one query
- [ ] The refusal and degradation behaviour is unchanged, proved by the existing tests
- [ ] Regression test on the run path specifically: preparing a run with five
      collections issues one collection query

PR #1181, opened 2026-08-26
perf(agents): resolve a run's bound collections in one query, not N

OchnikBartek:
The hot-path loop of #954.

## What changed

`_collection_names` runs inside `prepare` — **on every agent run** — and fetched one collection row per bound id in a serial loop. An agent bound to five knowledge bases added five serial round trips to the front of every turn, before the model was called, to read rows very likely already in the session's identity map.

- `knowledge_base_repo.get_by_ids` reads them in one `WHERE id IN (...)`.
- `_collection_names` iterates `spec.collection_ids` over the returned map. The tenant check and the *"a collection gone or foreign narrows the agent's reach rather than failing the run"* degradation are unchanged — they read from a dict instead of awaiting a query each. An id with no row is absent from the map, the way `get_by_id` answered `None`.

## Scope — the hot loop, `Part-of #954`

The finding names six id-resolving loops as "one habit". This does the sixth — the only one on a per-run path. The other five are publish-time or admin-path where N is small and the cost is invisible (the finding says so), and their `2026-08-19` line references predate a since-moved file, so they want re-locating in a follow-up rather than a blind change.

## How it was verified

- Behaviour-identical: the degradation test (a live, a deleted and a foreign collection resolve to just the live name) passes with the mock moved to `get_by_ids`.
- 100 `test_agent_runner.py` tests green; `ty` adds zero diagnostics; `ruff` clean.

Part-of #954

### DIFF
--- a/backend/app/repositories/knowledge_base.py
+++ b/backend/app/repositories/knowledge_base.py
@@ -14,6 +14,19 @@ async def get_by_id(db: AsyncSession, kb_id: UUID) -> KnowledgeBase | None:
     return await db.get(KnowledgeBase, kb_id)
 
 
+async def get_by_ids(db: AsyncSession, ids: Sequence[UUID]) -> dict[UUID, KnowledgeBase]:
+    """The collections for a set of ids, keyed by id, in one query.
+
+    An id with no row is simply absent from the map, the same way `get_by_id`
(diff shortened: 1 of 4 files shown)


OchnikBartek:
@codex review

Issue #935, opened 2026-08-19
The approval email links to the agent's Builder page, which has no approvals queue


DEENUU1:
The email that says a run is parked sends you to the Builder page for the agent,
where there is nothing to approve.

## What it does

`app/services/notifications.py:144`:

```python
"approvals_url": f"{self._frontend}/agents/{agent.id}",
```

and the template puts that behind the call to action -
`emails/compiled/approval_requested.txt`:

```
Review it here:
[[approvals_url]]
```

with the HTML rendering it as a **Review the request** button
(`approval_requested.html:12`).

`/agents/{id}` is the Builder. Searching that page for "approval" finds one
sentence of prose about tool calls reaching the queue and no queue: there is no
approvals surface on it at all. The actual queue is `/runs` → the **Approvals**
tab, which the email cannot link to because the tab is not in the URL.

So the one email whose entire purpose is "somebody has to decide, now" lands a
click away from the decision, on a page whose own subject is editing the agent -
and the parked run is meanwhile burning whatever the expiry window is
(`ApprovalService.expire_stale` denies by timeout).

## Done when

- `approvals_url` addresses the queue - `{frontend}/runs?tab=approvals`, and
  better still the run itself, since `?run=` already deep-links the detail panel
  and the notification holds the `AgentRun`.
- A test asserts the URL, because this is the kind of string nothing else checks.
- `docs/governance.md` still describes what the alert does, if it says where it
  points.

## Blocked by

**#934** - `/runs` has no `?tab=`, so there is nothing to link to yet. That issue
adds it (controlling the tab and clearing the focused run on a switch); this one
is the reason it is worth adding rather than a tidy-up.

## Related

- Found reading `notifications.py` while scoping #934.
- The audience logic right above it is deliberate and correct - initiator *plus*
  administrators, because a scheduled or channel run has no initiator. Nothing
  wrong with who gets the mail; only with where it goes.

PR #1199, opened 2026-08-26
fix(runs): put the open tab in the URL, and close a run when it changes

### LINKED ISSUE #934: The run detail panel survives a tab switch on /runs, and below lg it hides the tab you switched to
Open a run from the **Runs** tab, then switch to **Approvals**: the run detail
panel is still there, beside a queue it has nothing to do with. Below `lg` it is
worse - the focused run *replaces* the list, so switching tabs appears to do
nothing at all.

## Mechanism

The panel is the page's state, the tab is not.

`src/app/[locale]/(dashboard)/runs/page.tsx:46` holds the focused run in the URL
(`const [focusedRunId, focusRun] = useUrlState("run")`) and renders the panel at
`:207`, outside every `TabsContent`, with the reasoning stated there:

> The panel belongs to the page rather than to one tab: a run row and an approval
> row are both doors to the same view.

That half is right - `ApprovalsTab` takes `onFocusRun` and an approval row does
open a run. What is missing is the other half: **a focused run belongs to the tab
that opened it.** `Tabs` is uncontrolled (`defaultValue="runs"`, `:121`), so the
page never learns the tab changed and cannot clear the focus when it does.

And on a narrow screen the two interact badly: the list column carries
`focusedRunId !== null && "hidden lg:flex"` (`:163`), so with a run focused the
tab strip is live but every tab's content is hidden. Clicking Approvals changes
nothing visible.

## The fix, and what it is not

**Not** "hide the panel on the Approvals tab" - that breaks the door an approval
row is supposed to be.

Control the tab, mirror it into `?tab=`, and clear `?run=` when it changes. One
consequence and one non-consequence:

- The tab becomes addressable, which nothing can do today - see the sibling issue
  about the approval email linking to `/agents/{id}` because there is no URL for
  the queue.
- **The reason `Tabs` is uncontrolled is already handled elsewhere.** The comment
  at `:106` warns that Radix captures `defaultValue` on first mount, so a strip
  mounted while `can()` still answers `false` would open on Runs and stay there
  once Approvals appeared. But the whole block is already gated on
  `permissionsLoading` (`:115`) - `Tabs` does not mount until the permission set
  has answered - so a controlled value cannot be captured too early. Keep the gate
  and the hazard stays closed.

## Done when

- Switching tabs with a run focused closes the panel, on every width.
- An approval row still opens the run detail, and `?run=` still deep-links to it
  with the correct tab.
- `?tab=` survives a reload and travels in a pasted link, like `?period=`,
  `?agent=` and `?sort=` already do on this page.
- An integration test switches tabs with a run focused and asserts the panel is
  gone and the new tab's content is visible - the narrow-width case specifically,
  since that is where it reads as a dead control.

## Related

- **#507** (closed - the approval panel followed you into another conversation,
  and still decided) is the same defect one surface over: a panel outliving the
  context that opened it.
- **#768** and **#766** (both closed) put this page's filters into the URL and
  gave the three tabs one control grammar. The tab itself was the one piece of
  state left out.

DEENUU1:
Which of Activity's three tabs is open was the one piece of this page's state not in the address bar. Both halves of #934 follow from that.

## What was wrong

`Tabs` was uncontrolled (`defaultValue="runs"`), so **there was no URL for the approvals queue**. That is why the approvals dashboard card's "See all" opens the run history — where nothing can be decided — and why the alert saying a run is parked links to the agent's Builder page (#935). Neither had an honest destination to point at.

And a focused run outlived the tab that opened it: the detail panel sat beside a queue it has nothing to do with, and below `lg` the focused run *replaces* the list — so the strip was live while every tab's content stayed hidden and clicking **Approvals** appeared to do nothing at all.

## What changed

- **`?tab=approvals` / `?tab=spend`.** `runs` is the default and is never written, like every other unset narrowing on this page. `parseRunsTab` joins `parseRunFilters` in `lib/runs/filter-params.ts`, and `runsHref` takes a `tab`.
- **A tab named by a link is resolved against what the reader may open**, not trusted. `approvals` is gated on `approvals:decide`, so a link carrying it that reaches somebody without the permission opens the run history rather than a strip whose selected value has no trigger and no content — which draws as a blank page under a live set of tabs. An unrecognised name falls back the same way.
- **The focused run is cleared on a switch**, as it already was (it landed incidentally in #537, untested) — now with tests, and `?run=` goes with it. Left behind, a reload would reopen a panel on a tab that never had one.
- **The approvals widget's `seeAll` points at the queue.** Same wrong destination as the email, enabled by the same missing parameter, one line — so it is fixed here rather than filed.

The comment warning that Radix captures `defaultValue` on first mount is gone with the hazard. The permission gate that made it moot stays, and its remaining reasons are written down: the strip's *shape* depends on that answer, and so does resolving `?tab=approvals`.

## How it was verified

`tab-url.integration.test.tsx` — six cases. Three of them fail against the page as it was:

```
× opens the queue for a link that names it
× writes the tab a reader switches to, so a reload comes back to it
× takes the run out of the URL with it
```

The other three guard behaviour that already held (the two fallbacks, the panel closing). The narrow-width case is asserted through the class that carries it — jsdom applies no stylesheet, so `hidden lg:block` is not observable as layout; what is observable is whether the page still asks for it.

`make lint-frontend` clean. `bun run test:coverage` green: 100% statements/lines/functions, 97.67% branches.

## Unblocks

**#935** — the approval email can now address the queue instead of the Builder page.

Closes #934

### DIFF
--- a/docs/governance.md
+++ b/docs/governance.md
@@ -599,6 +599,19 @@ it. They were local state until #768, so the p95 figure was the only number on
 the dashboard that could reach the runs behind it and three cards carried no link
 at all — there was nothing honest to point them at.
 
+**Including which tab is open.** `?tab=approvals` and `?tab=spend` open the queue
+and the cost screen; the run history is the default and is never written. That is
+the address a link to a *decision* needs, and the reason the parameter exists: the
+approvals card's "See all" and the alert that says a run is parked both had to
+point at the run history, where nothing can be decided (#934). A tab named by a
+link is resolved against what the reader may open — `approvals` is gated on
+`approvals:decide`, so a link carrying it that reaches somebody without the
+permission opens the run history rather than a strip whose selected tab has no
+content. Switching tabs closes an open run detail and takes `?run=` with it: a
+panel that outlives the tab that opened it sits beside a queue it has nothing to
+do with, and below `lg` it replaces the list, so the strip stayed live while every
+tab's content was hidden.
+
 **Duration is computed in SQL, over the whole narrowed set.** That is what gets
 from *"p95 is 14.8s"* on the dashboard to **those runs** — sorting one page of
 twenty-five sorts the wrong set, because the slowest run of a month is not in
--- a/frontend/src/app/[locale]/(dashboard)/runs/page.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/runs/page.tsx
@@ -14,7 +14,12 @@ import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui";
 import { useApprovals, usePermissions, useRuns, useSpend, useUrlState } from "@/hooks";
 import { periodEnd, periodStart } from "@/lib/dashboard/period";
 import { formatPeriodParam, parsePeriodParam, type Period } from "@/lib/dashboard/period";
-import { parseRunFilters, writeRunFilters, type RunFilters } from "@/lib/runs/filter-params";
+import {
+  parseRunFilters,
+  parseRunsTab,
+  writeRunFilters,
+  type RunFilters,
+} from "@/lib/runs/filter-params";
 import { cn, setUrlParam } from "@/lib/utils";
 import { Perm } from "@/types/permissions";
 import { useTranslations } from "next-intl";
@@ -88,6 +93,13 @@ export default function RunsPage() {
     setFilters(next);
     writeRunFilters(next);
   };
+  // Which tab is open, mirrored into `?tab=` like everything else the page
+  // narrows itself with. Addressable at all for the first time here: the alert
+  // that says a run is parked has to send somebody to the decision, and until
+  // the queue had a URL the only honest link was the agent's Builder page
+  // (#935). Resolved against the permission below rather than trusted, so a
+  // link naming a tab the reader may not open lands on one they may.
+  const [tabParam, changeTab] = useUrlState("tab");
   const { can, isLoading: permissionsLoading } = usePermissions();
   // Run history and spend take `runs:view` - `GET /runs` and `GET /spend` both
   // carry it - so a caller without it is not asked for: the figures and the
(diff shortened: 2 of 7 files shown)


DEENUU1:
Both findings addressed in 9e2d999b.

**`?tab=runs`** — right, and it contradicted the contract this PR states. `changeTab(next === "runs" ? null : next)`, with a test that fails against the old line.

**The focused run surviving a URL-driven tab change** — this one does not happen, and there are now two tests holding that shut. Radix indeed does not call `onValueChange` for a prop-driven change, but there is nothing left to clear by then: a navigation that changes `?tab=` replaces the whole query string, so `?run=` is gone from it and `useUrlState`'s render-time adjustment resets the value whose parameter changed under it — the panel closes on its own. The only navigation that keeps a run focused across a tab change is one that names `&run=` explicitly, and honouring that is the point of `?run=`:

```
✓ closes with a link that changes the tab and does not name the run
✓ honours a link that names both, because that link is asking for both
```

DEENUU1:
@codex review

DEENUU1:
Held out of the release batch (v0.0.248–v0.0.301 so far): the branch is checked out in a worktree at `agenticos-wt/i934`, so I have left it rather than pushing into work in progress.

One live thing to pick up before it merges — the Codex thread that is **not** outdated. A same-path navigation that changes `?tab=` while the page stays mounted updates the controlled `value`, but Radix does not call `onValueChange` for a prop-driven change, so `?run=` and its detail panel survive the switch. That is this PR's own second half arriving on the other path: below `lg` the focused run replaces the list, so the newly selected tab's content stays hidden — the same symptom as clicking Approvals and seeing nothing happen. Clearing the focused run on a change to the *resolved* tab, rather than only on a trigger click, covers both.

The other thread is outdated. #1202 depends on this one for `?tab=` to be honoured, so it is held too.

PR #1202, opened 2026-08-26
fix(approvals): send the parked-run alert to the queue, not the Builder

### LINKED ISSUE #935: The approval email links to the agent's Builder page, which has no approvals queue

DEENUU1:
The email that says a run is parked sent you to the Builder page for the agent, where there is nothing to approve.

`app/services/notifications.py`:

```python
"approvals_url": f"{self._frontend}/agents/{agent.id}",
```

`/agents/{id}` holds one sentence of prose about tool calls reaching a queue, and no queue — there is no approvals surface on it at all. So the one email whose whole purpose is *somebody has to decide, now* landed a search away from the decision, behind a button that says **Review the request**, while the parked run aged towards `ApprovalService.expire_stale`.

## What changed

`approvals_url` addresses `/runs?tab=approvals` — Activity's Approvals tab, the only surface carrying Approve and Reject. The tab had no URL until #934 put it in `?tab=`, which is why this was not simply a wrong string.

**Not `&run=`**, though the notification holds the `AgentRun` and `?run=` does deep-link the run detail panel, as the issue suggested. The decide controls are on the queue *row*, and below `lg` a focused run replaces the list — so naming the run would hide the buttons from the reader most likely to be reading this on a phone.

The templates need no change; they interpolate `[[approvals_url]]`.

## How it was verified

A new `TestWhereAnAlertSends` asserts the URL, because every one of these links is a hand-built f-string that no route table, no type and no other test looks at — the failure mode is a person following the link and not finding the control. It fails against the old string:

```
FAILED tests/test_notifications.py::TestWhereAnAlertSends::test_the_approval_alert_addresses_the_queue_not_the_builder
```

`make test` green against a real pgvector: **7129 passed, 100.00% coverage**. `make lint-backend` clean.

## Docs

`docs/governance.md` gains **An alert links to where the decision is** under Alerts: where approvals mail points and why, why it does not name the run, and that budget mail opening the agent is correct — the cap it reports is edited there.

## Depends on

**#934** (PR #1199) for `?tab=` to be honoured. Merged before it, this link opens Activity on the run history; the tab parameter is simply ignored.

Closes #935

### DIFF
--- a/backend/app/services/notifications.py
+++ b/backend/app/services/notifications.py
@@ -141,7 +141,14 @@ class NotificationService:
         context = {
             "agent_name": agent.name,
             "tools": ", ".join(tools) if tools else "a tool call",
-            "approvals_url": f"{self._frontend}/agents/{agent.id}",
+            # The queue, not the agent. This addressed `/agents/{id}` - the
+            # Builder - where there is nothing to approve, so the one email whose
+            # whole purpose is "somebody has to decide, now" landed a search away
+            # from the decision while the run aged towards `expire_stale` (#935).
+            # Not `&run=`: the Approve and Reject controls are on the queue row,
+            # and below `lg` a focused run replaces the list - which would hide
+            # them from the reader most likely to be on a phone.
+            "approvals_url": f"{self._frontend}/runs?tab=approvals",
             "app_name": settings.PROJECT_NAME,
         }
         self._send(EmailKey.APPROVAL_REQUESTED, recipients, context)
--- a/backend/tests/test_notifications.py
+++ b/backend/tests/test_notifications.py
@@ -705,6 +705,37 @@ class TestPreferences:
         roles.assert_not_awaited()
 
 
+class TestWhereAnAlertSends:
+    """The link is the whole point of an alert, and nothing else checks it.
+
+    An email that says a run is parked is read by somebody who has to decide, and
+    the only thing they can do with it is click. Every one of these URLs is a
+    hand-built f-string that no route table, no type and no other test looks at -
+    so a page that moves, or a surface built somewhere other than where the string
+    guessed, is found by a person following the link and not finding the control.
+    """
+
+    @pytest.mark.anyio
+    async def test_the_approval_alert_addresses_the_queue_not_the_builder(self, sent):
+        """The regression. `/agents/{id}` is the Builder: it holds one sentence of
+        prose about tool calls reaching a queue, and no queue. So the click landed
+        on a page whose own subject is editing the agent, while the parked run
+        aged towards `ApprovalService.expire_stale` (#935)."""
+        with (
+            patch(f"{MODULE}.member_repo.list_emails_by_role", new=_roles("ops@acme.test")),
+            patch(f"{MODULE}.member_repo.list_emails_for_members", new=_members()),
+        ):
+            await NotificationService(MagicMock()).approval_requested(
+                _run(), agent=_agent(), spec=_spec(), tools=["send_email"]
+            )
+
+        _, _, context = sent.calls[0]
+        assert context["approvals_url"].endswith("/runs?tab=approvals")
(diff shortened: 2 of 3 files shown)


DEENUU1:
Three findings, three different answers.

**P1 — `?tab=` is not honoured yet.** Correct on `main`, and it is the reason this PR names a dependency: **#934 / PR #1199** makes `Tabs` controlled and puts the tab in the URL. It is open and green. Merged in the other order this link simply opens Activity on the run history — the same page it opens today — so nothing regresses in the window between them; merged after, it lands on the queue. Splitting them was deliberate: one is a frontend change with an integration suite, the other is a string and a backend test.

**P2 — recipients who hold no `approvals:decide`.** Real, and worse than the comment says: `builder` and `member` both lack it, and a builder starting their own agent is the ordinary case, not a custom-spec edge. Filed as **#1203** rather than fixed here, because the honest answer is a copy and template decision — either the audience is filtered by permission (and then a parked run can tell nobody who is actually waiting on it) or the initiator gets a different email that does not carry a button they cannot use. #935 is explicit that the audience logic is deliberate; changing it belongs in its own change.

**P2 — no organization in the link.** Also real, and general: *none* of the four notification URLs carries one, and `activeOrgId` decides which organization the page acts on. `run.organization_id` is in scope at every one of those call sites and discarded. Filed as **#1204** — it needs a parameter decided once for all four emails plus an adoption in `ActiveOrgGuard`, which is the shape #1032 already established.

Neither is introduced by this change; both survive it.

DEENUU1:
Both findings on this diff were right, and both are fixed — on their own branches, because each is a decision this PR deliberately did not take:

- **the audience holding no `approvals:decide`** → #1203, PR #1280. The audience is split by the permission rather than trimmed to it: deciders get the request, everybody else gets a second email saying the run is held and that approving belongs to an owner, admin or operator. Dropping them would lose the one person definitely waiting on the run.
- **the link naming no organization** → #1204, PR #1289. Every alert URL carries `?org=<id>`, built in one place, and the console adopts it the way it adopts the id in `/orgs/{id}`. That covers this alert and the three others with the same defect.

Nothing on this branch changes as a result: it is about the destination (#935), and the destination it sets is what #1204's helper then parameterises.

@codex review

Issue #130, opened 2026-08-03
E2E: the vault rotation spec fails intermittently before the secret is created


DEENUU1:
`frontend/e2e/vault.spec.ts:125` ("rotating a secret changes its value and keeps its id") failed on `feat/sandbox` in run [30830855684](https://github.com/vstorm-co/agenticos/actions/runs/30830855684) and passed on a plain re-run of the same commit, so it is flaky rather than broken by that branch.

**What the artifact shows.** The failure is at line 147, waiting for the new row:

```
Locator: getByRole('row', { name: /e2e-rotate-msdfi15n/ })
Expected substring: "····AAAA"
Error: element(s) not found
```

In the captured page state the Add dialog is already closed, the table says "3 keys stored" and lists only `e2e-capability-token`, `e2e-vault-probe` and `openai (bootstrap)` - so `Store secret` did not create anything. There is an empty `- alert` at the root of the snapshot, which is where a refusal toast would be, but its text is not in the accessibility dump.

**Why it matters beyond one red run.** Every other test in that file passed in the same run, including one that stores a secret by the same "Something else" path (`e2e-capability-token`), so the create path works. Something makes this one create fail intermittently and the spec cannot say what: it asserts on the row that should have appeared, never on the refusal that did.

**How you would know it was fixed.** Two things:
1. The spec asserts the dialog closed *and* nothing is on the error toast before it waits for the row, so a failure names the refusal instead of timing out on a missing element.
2. The underlying create no longer fails - which needs the toast text to say what it was.

Pre-existing; found watching CI on the branch for #36, not caused by it.

OchnikBartek:
Two halves, and only one is fixable without a reproduction:

1. **Spec diagnostics (tractable).** The spec waits for the new row and, when `Store secret` silently did nothing, times out on a missing element instead of naming the refusal — the captured page had an empty `- alert` (the toast) whose text never reached the dump. Asserting the dialog closed *and* the error toast is empty before waiting for the row would make a failure say *what* was refused. This is a safe spec change.

2. **The underlying intermittent create failure (needs a reproduction).** Why `Store secret` intermittently creates nothing is unknown, and the spec as written cannot say — it needs the toast text a hardened spec would capture. It smells like the same post-write staleness family as #154/#230 (the dialog closed and the list refetched to "3 keys stored" without the new one), but that is a guess until the toast is seen on a red run.

[...]

PR #1260, opened 2026-08-27
fix(chat): cancel in-flight reads before a mutation invalidates them

### LINKED ISSUE #154: e2e: sharing and skills specs fail when a mutation's refetch lands stale

OchnikBartek:
## Summary

`sharing.spec.ts` and `skills.spec.ts` flaked because a mutation's cache
invalidation could be answered with pre-write data. A mutation's `onSuccess`
called `invalidateQueries` and relied on the refetch to show the new state —
but `invalidateQueries` **dedupes its refetch onto a fetch already in flight**.
Both the sharing panel and the skills gallery fan out several reads on mount, so
a read that began before the mutation committed resolved with the pre-write body,
marked the query fresh, and the panel kept the old value until a reload. It hit
the *second* mutation in a sequence, never the first — the exact shape of #154.

This is a client bug, not stale data on the wire: the read is ordered after the
commit and hits Postgres (no-store is already in effect since #405/#230), the
client just dropped the fresh answer.

## Changed

- Each hook funnels its mutations through one `invalidate` helper; that helper now
  **cancels the query before invalidating**, so the invalidation dispatches a
  genuinely new post-commit fetch rather than awaiting the stale in-flight one it
  meant to replace.
  - `use-sharing.ts` — covers `share` / `revoke` / `setVisibility`.
  - `use-skills.ts` — covers `create` / `update` / `remove` in `useSkills`, and
    `save` / `addResource` / `saveResource` / `removeResource` / `uploadResources`
    in `useSkill`.

## Testing

- New unit test per hook asserts `cancelQueries` runs (and before
  `invalidateQueries`) on a mutation success — the ordering the fix depends on.
- The existing `sharing.spec.ts` / `skills.spec.ts` panel-refresh assertions are
  what this makes reliable; they are left asserting the real re-render (no masking
  reload added). `make test-frontend-cov` green (100% stmts/funcs/lines, 97.7%
  branches).
- Reviewed the mechanism against TanStack Query v5 refetch dedup; `cancelQueries`
  before `invalidateQueries` is the standard idiom and behaviour-preserving on the
  happy path.

## Notes for reviewers

- The same in-flight-dedup shape exists in other hooks (e.g. `use-secrets`, the
  vault manifestation tracked as #130); this PR is scoped to the two hooks #154
  names. #130 is handled separately.
- `#230` is already closed (no-store, #405); the stale "nobody has closed it"
  notes in `.claude/rules/testing.md` / `docs/testing.md` are a separate
  documentation fix.

Closes #154

### DIFF
--- a/frontend/src/hooks/use-sharing.test.tsx
+++ b/frontend/src/hooks/use-sharing.test.tsx
@@ -135,6 +135,33 @@ describe("useSharing", () => {
     expect(toast.error).toHaveBeenCalledTimes(2);
   });
 
+  it("cancels the in-flight read before invalidating, so a stale refetch cannot win (#154)", async () => {
+    // invalidateQueries dedupes its refetch onto a fetch already running; a read
+    // that began before the mutation committed would otherwise resolve with the
+    // pre-write body and leave the panel stale. Cancelling first forces a new
+    // post-commit fetch.
+    vi.mocked(apiClient.put).mockResolvedValue({});
+    const client = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+    const cancelSpy = vi.spyOn(client, "cancelQueries");
+    const invalidateSpy = vi.spyOn(client, "invalidateQueries");
+    const scoped = ({ children }: { children: ReactNode }) => (
+      <QueryClientProvider client={client}>{children}</QueryClientProvider>
+    );
+
+    const { result } = renderHook(() => useSharing("agent", "a1"), { wrapper: scoped });
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    await result.current.share.mutateAsync({ subject_user_id: "u-sam", level: "read" });
+
+    const key = ["sharing", "agent", "a1"];
+    expect(cancelSpy).toHaveBeenCalledWith({ queryKey: key });
+    expect(invalidateSpy).toHaveBeenCalledWith({ queryKey: key });
+    const cancelOrder = cancelSpy.mock.invocationCallOrder[0] ?? 0;
+    const invalidateOrder = invalidateSpy.mock.invocationCallOrder[0] ?? 0;
+    expect(cancelOrder).toBeGreaterThan(0);
+    expect(cancelOrder).toBeLessThan(invalidateOrder);
+  });
+
   it("addresses a collection at the plural of its own noun", async () => {
     // Four routes exist per type and the paths are not derivable from the type
     // name; a wrong stem shares a resource that does not exist.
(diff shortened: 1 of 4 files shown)


OchnikBartek:
@codex review

PR #1261, opened 2026-08-27
fix(chat): stop the vault list refetch losing to an in-flight read

### LINKED ISSUE #130: E2E: the vault rotation spec fails intermittently before the secret is created

OchnikBartek:
## Summary

`vault.spec.ts` "rotating a secret changes its value and keeps its id" flaked
(~1 run in 8): the store/rotate committed and the list refetched, but the table
kept the pre-write list, so the new row never appeared and the spec timed out.
The create path itself worked — the artifact showed `3 keys stored` with an empty
error alert, i.e. the write succeeded and the render was stale.

Root cause is the client dedup race behind #154: a mutation's `onSuccess`
invalidates the list and relies on the refetch, but `invalidateQueries` dedupes
onto a read already in flight (the vault page issues its list read on load). A
read that began before the write committed resolves with the pre-write body,
marks the query fresh, and no further refetch runs.

## Changed

- `use-secrets.ts` — the one `invalidate` helper `create` / `rotate` / `remove`
  funnel through now **cancels the list query before invalidating**, so the
  invalidation dispatches a genuinely new post-commit fetch.
- `vault.spec.ts` — retires the three `page.reload()` workarounds this bug forced
  (the `#230`/`#130`-marked reloads): the row now appears on its own, which is the
  verification that it no longer flakes.

## On the issue's acceptance criteria

- *"the spec names the refusal instead of timing out"* — already provided by
  `submitDialog`, which asserts the write's response status and prints its body on
  a non-2xx, so a refused store fails at its source before the row is awaited. No
  separate error-toast assertion is needed; the response check is the stronger
  signal.
- *"the underlying create no longer fails"* — the create did not fail (it 201'd);
  the list render was stale. The `cancelQueries` fix removes that staleness.

## Testing

- `use-secrets.test.tsx` gains a regression test asserting `cancelQueries` runs
  before `invalidateQueries` on a rotate.
- `make test-frontend-cov` green (100% stmts/funcs/lines, 97.7% branches).
- The rotation and token-refresh specs now assert the row without a reload; CI's
  e2e job exercises them end to end.

## Notes for reviewers

- Same fix and shape as #154 (sharing/skills); this is the vault hook. Other
  surfaces with `#230`-marked reloads (surfaces/journey/seed) use different hooks
  and are out of scope here.
- `#230` itself is already closed (no-store, #405); its stale "nobody has closed
  it" notes in `.claude/rules/testing.md` / `docs/testing.md` are a separate docs
  fix.

Closes #130

### DIFF
--- a/frontend/e2e/vault.spec.ts
+++ b/frontend/e2e/vault.spec.ts
@@ -148,15 +148,10 @@ test.describe("Vault", () => {
       path: "/api/secrets",
     });
 
-    // Reloaded, and this is a concession to a live bug rather than a nicety.
-    // `submitDialog` has already proved the write was accepted and the dialog
-    // closed, so the list on screen *should* hold the row — and about once in
-    // eight runs it does not (#230: the row is committed, both server layers
-    // answer a list containing it, and the page keeps rendering the one without
-    // it). Asserting on it without a reload is asserting on that bug, which is
-    // how this spec came to flake as #130. What is under test here is rotation;
-    // the fresh load is how it gets a list it can trust to start from.
-    await page.reload();
+    // No reload: the store mutation cancels any in-flight secrets read before it
+    // invalidates, so the list refetch is a fresh post-write one rather than a
+    // pre-write fetch it deduped onto and rendered - the flake #130 was (once in
+    // eight runs the row was committed but the page kept the list without it).
     await expect(row(page, name)).toContainText("····AAAA");
     const before = await secretId(page, name);
 
@@ -176,7 +171,6 @@ test.describe("Vault", () => {
       method: "PATCH",
     });
 
-    await page.reload(); // #230, as above
     await expect(row(page, name)).toContainText("····BBBB");
     expect(await secretId(page, name), "rotation replaced the row instead of its value").toBe(
       before,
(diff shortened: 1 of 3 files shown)


OchnikBartek:
@codex review

PR #1276, opened 2026-08-27
fix(dashboard): the gate owns the clearance on its no-wrapper branch

### LINKED ISSUE #1241: The maintenance screen has no bottom clearance under the mobile tab bar
**Found reviewing #933** (PR #1207), which moved the dashboard's bottom clearance off `main` — where it painted nothing — onto `PageTransition`'s unconstrained branch.

## What's wrong

`DeploymentGate` has a branch that never reaches `PageTransition`: when maintenance mode is active for a non-admin it returns `MaintenanceScreen` directly. Before #933 that state inherited nothing either — `main`'s declaration was inert, which is the whole finding of #933 — so this is not a regression so much as the one page left behind when the clearance found a home that works.

On mobile the `MobileTabBar` is fixed and overlays the page. A long custom maintenance message therefore ends underneath it, and at maximum scroll the last 56px plus the safe-area inset stay covered — the operator's own text, on the one screen a visitor sees when nothing else is available.

## Where

`frontend/src/components/layout/deployment-gate.tsx` — the branch returning `MaintenanceScreen`, versus the one that renders `PageTransition`.

`page-clearance.test.ts` walks the pages rather than the wrapper, deliberately, so it does not see this branch. The check wants to be the same shape as the fix: either the gate's maintenance branch carries `pb-[calc(5rem+env(safe-area-inset-bottom))] lg:pb-16` itself, or it is wrapped by the same padded container the ordinary pages get — and a test asserts whichever it is, so a third branch added to the gate does not repeat this.

Worth deciding which, rather than pasting the class: a second copy of that calc is a second place for the inset to be forgotten, which is exactly how #933's four workarounds ended up at three different values.

DEENUU1:
`DeploymentGate` returns `MaintenanceScreen` **instead of** rendering `PageTransition`, which is where every other page takes its bottom clearance from. On mobile a long custom maintenance message therefore ended under the fixed `MobileTabBar` — the last 56px plus the safe-area inset stayed covered even at maximum scroll, on the one screen a visitor sees when nothing else is available.

## The decision the issue asks for

#1241 says to decide rather than paste the class: *"either the gate's maintenance branch carries it itself, or it is wrapped by the same padded container the ordinary pages get."*

The clearance moves **onto that branch**, and off `MaintenanceScreen`. The gate is what knows this *is* the whole page; on the screen it was a page property the component cannot know it has — it would inherit page padding anywhere else it were rendered — and the branch itself said nothing about needing any. Still one token, `PAGE_CLEARANCE`, so there is no second copy of the calc to forget `env(safe-area-inset-bottom)` in.

## The test it asks for

`page-clearance.test.ts` walks the *pages*, deliberately, so it cannot see this branch. The assertion is a render instead: the gate in maintenance, as a non-admin, and its root carries the token — spread as classes, so a token that loses the inset fails here too.

`make lint-frontend` clean · `bun run test:coverage` green (100% statements/lines/functions, 97.70% branches).

Closes #1241

### DIFF
--- a/frontend/src/components/branding/branding.integration.test.tsx
+++ b/frontend/src/components/branding/branding.integration.test.tsx
@@ -115,6 +115,10 @@ describe("a maintenance window", () => {
     );
 
     expect(container.firstElementChild).toHaveClass(...PAGE_CLEARANCE.split(" "));
+    // And the box has to grow with the message for that padding to land after
+    // it. `min-h-0` here capped it at the viewport, which left `main` unable to
+    // scroll to the end of a long message at all (Codex, P1 on this branch).
+    expect(container.firstElementChild).not.toHaveClass("min-h-0");
   });
 
   it("leaves it open for the administrator who has to end it", () => {
--- a/frontend/src/components/branding/branding.integration.test.tsx
+++ b/frontend/src/components/branding/branding.integration.test.tsx
@@ -11,6 +11,7 @@ import { DeploymentGate } from "./deployment-gate";
 import { MaintenanceScreen } from "./maintenance-screen";
 import { apiClient } from "@/lib/api-client";
 import { BUILT_IN_BRANDING, type Branding, type NoticeResponse } from "@/lib/branding";
+import { PAGE_CLEARANCE } from "@/lib/page-clearance";
 
 /**
  * What the deployment's own state does to the product.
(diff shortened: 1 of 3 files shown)


DEENUU1:
Fixed in f5fd033b — the finding was right, and the effect is worse than reported.

Measured in the real chain (`main` as the scroll container, the gate's branch inside it, 900px of message in a 400px viewport):

| gate branch | box height | `main.scrollHeight` | `scrollTop` at full scroll | gap under the message |
|---|---|---|---|---|
| no clearance (before #1241) | 900 | 916 | 516 | **0px** |
| `min-h-0 flex-1` + clearance | 384 | **400** | **0** | — |
| `flex-1` + clearance | 980 | 996 | 596 | **80px** |

Identical in Chromium 151 and WebKit 26.5. With `min-h-0` the message is squeezed to the viewport and `main` will not scroll at all, so the end of it is unreachable rather than merely tucked under `MobileTabBar`.

Dropping the floor is what `PageTransition`'s unconstrained branch already does for exactly this reason; both comments now say so, and `branding.integration.test.tsx" asserts the absence so it does not come back as a tidy-up.

@codex review

Issue #923, opened 2026-08-19
Six answers to "what models and providers exist", and none of them is the authority


DEENUU1:
"What models and providers exist" is answered in six places, each with a
different list, and none of them is the authority at run time. Map them, decide
which one is the source of truth, and collapse the rest.

## The six answers

| Where | Holds | Count |
|---|---|---|
| `backend/app/agents/model_resolver.py:105` — `PROVIDERS` | Which providers a profile may point at, and which credential shape each wants | 27 |
| `backend/app/core/catalog/model_listings.json` | How to read one provider's live `/models` endpoint (`ListingSpec`) | 20 |
| `backend/app/core/catalog/model_fallbacks.json` | The curated list used when there is no listing, no key, or the call fails | 7 |
| `genai_prices` snapshot, via `model_catalog.priced_model` | Price and context window, keyed on its own provider spelling (`_PRICE_PROVIDER_ALIASES` bridges three) | its own |
| `backend/app/core/catalog/image_models.json` + `services/image_models.py` | Image models, with their own `provider` / `prefix` / `caller` notion | 3 providers |
| `docs/models.md:72-131, 210-218` | Hand-written tables of every provider, its credential, its custom-URL param and its curated ids | 27 + 7 |

And the actual authority is a seventh: `pydantic_ai.providers.infer_provider_class`
and `pydantic_ai.models.infer_model`, which is what constructs the client. The
platform's job is only the part inference cannot know — the credential shape — and
`PROVIDERS` says so in its own docstring. Everything else in the table above is a
copy that can drift from it, and one already has.

## What that costs today, concretely

- **Seven providers have an empty model picker.** `models_for`
  (`services/model_catalog.py:250`) returns the curated list when there is no
  `ListingSpec`; `github`, `heroku`, `ollama`, `litellm`, `azure`, `bedrock` and
  `google_cloud` are in `PROVIDERS`, in neither JSON file, so they return
  `([], "curated")` — a dropdown that says the provider offers nothing. For
  `ollama` and `litellm` that is the worst case: both are keyless, self-hosted,
  and both publish an OpenAI-shaped `/v1/models` at the endpoint the profile
  already stores, so the one thing that would work is the one thing not wired.
  The same sentence one element over is **#863** (closed) — the picker claiming
  the organization has no models when the *profiles* read failed.
- **`model_fallbacks.json` is not about fallbacks.** A profile's fallback chain
[...]

PR #1277, opened 2026-08-27
test(models): bring the last three providers inside the drift guard

### LINKED ISSUE #1252: The model-catalog drift test compares less than it claims
**Found reviewing #923** (PR #1230), which chose a drift test over generating the docs tables — *"a drift test is the cheaper equivalent and is what the issue's done-when asks for"*. That is the right call, but the test as written compares less than the claim needs, in four ways. Each one lets the page and `PROVIDERS` disagree while the suite stays green, which is the single failure the test exists to prevent.

## The four

1. **The page is searched for a token, not compared to a table.** Removing a provider from `PROVIDERS` leaves its row in `docs/models.md` and the test still passes, because the token is still on the page — so the check holds in one direction only. Comparing the parsed table rows to the set is what makes it a drift test rather than a spell-check.

2. **Only the id is compared, though `PROVIDERS` is authoritative for the credential shape.** That is the one thing the PR names as the reason `PROVIDERS` is the source of truth, and it is exactly what the documentation rows also carry — so the columns most worth holding in step are the ones nothing checks.

3. **`_documented_rows()` excludes the Azure, Bedrock and Vertex table**, and none of those three is checked anywhere else. The three providers with the most involved credential shapes are the three outside the guard.

4. **The image catalog is checked on `provider` but not on `prefix`**, though it carries the pair. A future edit pairing `google` with an OpenAI prefix passes.

## Where

`backend/tests/test_model_catalog.py` — `TestOneAnswerPerQuestion` and its `_documented_rows()` helper.

The test that proves a drift test works is the one that fails when you introduce drift. #1230 already did this for the additive direction — a twenty-eighth provider was added and the suite went red. The same demonstration in the other three directions is the acceptance: remove a provider and watch the page's stale row fail; change a documented credential column and watch it fail; repoint an image entry's prefix and watch it fail.

## Also raised, and smaller

`_fallback(curated)` answers `([], "unlisted")` for a provider with no curated entry whose live request fails, but the warning logged immediately before it still says a curated shortlist is being used. The log and the return disagree about what just happened — worth a sentence that names the real outcome, since that line is what an operator reads when a picker is empty.

DEENUU1:
`_documented_rows` reads the two four-column tables, so **Azure, Bedrock and Vertex AI** — the three providers with the most involved credential shapes — were in no assertion but the id one. That is item 3 of #1252; items 1, 2 and 4 and the log-line note landed with #923.

## What those three rows can be checked against

Their credential is prose, so no field maps to it. But *which of the three tables a provider sits in* is itself a claim about its credential — the heading says "Credential is not an API key" — and that is comparable:

- **The three tables partition `PROVIDERS`.** A provider documented twice, or in none of them, fails.
- **Which table a row is in matches `secret_kind`.** Moving a row between tables without changing the spec, or the reverse, fails.

## The four directions, demonstrated rather than asserted

#1252's acceptance is that the drift test fails when drift is introduced. Each one, run:

| Drift | |
|---|---|
| Remove `cohere` from `PROVIDERS`, leave its docs row | `5 failed, 5 passed` |
| Spell Ollama's credential as `API key` | `FAILED …::test_the_page_says_which_providers_need_no_key` |
| Move Cohere's row into the not-an-API-key table | `FAILED …::test_which_table_a_provider_is_in_matches_its_credential` |
| Repoint the image catalog's `google` entry at `openai-responses` | `FAILED …::test_every_image_prefix_belongs_to_the_provider_it_is_paired_with` |

Clean tree: `10 passed`.

`make test` **100%** (7204 passed) · `make lint-backend` clean.

Closes #1252

### DIFF
--- a/backend/tests/test_model_catalog.py
+++ b/backend/tests/test_model_catalog.py
@@ -293,15 +293,20 @@ class TestCuratedFallbacksAreData:
         assert LISTINGS["together"].array_path == ""
 
 
-def _documented_providers() -> set[str]:
-    """The ids in the `| Provider | id | Credential | Custom URL |` tables.
+def _documented_ids() -> list[str]:
+    """Every id in the three `| Provider | id | ...` tables, in page order.
 
     Read off the second cell of every row under one of those headers, so a row
     added or removed is the thing being compared - and prose elsewhere on the
     page is not.
+
+    A **list**, not a set: a provider documented twice is a row somebody added
+    without noticing the one already there, and the tables' own claim is one row
+    each. Collapsing to a set answers "is it mentioned" and lets the duplicate
+    through every comparison below.
     """
     page = (Path(__file__).resolve().parents[2] / "docs" / "models.md").read_text()
-    ids: set[str] = set()
+    ids: list[str] = []
     in_table = False
     for line in page.splitlines():
         if line.startswith("| Provider | id |"):
@@ -313,7 +318,7 @@ def _documented_providers() -> set[str]:
             continue
         cells = [cell.strip() for cell in line.strip("|").split("|")]
         if len(cells) > 1 and cells[1].startswith("`"):
-            ids.add(cells[1].strip("`"))
+            ids.append(cells[1].strip("`"))
     return ids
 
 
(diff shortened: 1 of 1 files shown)


DEENUU1:
Fixed in c4b1d69c. The finding was right: both parse helpers keyed on the provider id, so a duplicated row collapsed before any assertion ran and the docstring's "documented exactly once" was not being checked.

`_documented_providers` is now `_documented_ids`, returning every id in page order, and the comparison is `sorted(_documented_ids()) == sorted(PROVIDERS)`; the partition test adds `len(four_column) + len(three_column) == len(_documented_ids())`, which is what catches a duplicate inside one table. Duplicating the first hosted row fails both tests and passed both before.

@codex review

Issue #17, opened 2026-08-01
Add row locks to four check-then-act races


DEENUU1:
**Severity: Medium** · AUD-015

Four check-then-act races. The codebase already has the right pattern in `claim_parked_run` (`repositories/agent_run.py:91-105`, `.with_for_update()`, with a comment explaining why). These four do not use it.

## 15a. `ApprovalService.decide`

`services/approvals.py:86-107`, `repositories/agent_run.py:263-272` (`get_approval` is a plain `select`).

Two approvers open the queue. A rejects, B approves, both POST within the same read-committed window. Both read `status == 'pending'`, both pass the guard, both write. The row ends at whichever committed last, and `record_audit` emits **two contradictory entries** — `approval.rejected` and `approval.approved` — for one decision. The service docstring says deciding twice *"would make the audit trail ambiguous about who authorised the action"*; that is the state produced.

Fix: a `with_for_update` variant of `get_approval`, used in `decide`.

## 15b. Concurrent runs share one budget baseline

`agents/capabilities/budget/_capability.py:289-309`, `repositories/agent_run.py:137-158` (`cost_usd` is `0` until `finish_run`).

An organization at $99 of a $100 cap. Fifty API runs arrive together; each opens a row with `cost_usd = 0`, each reads a baseline of $99, each finds `99 < 100`, all fifty proceed. The cap overshoots by up to N × (cost of one run), while `docs/governance.md:36-39` presents the check as a hard stop.

Fix: reserve before the request, or write the ledger total onto the run row periodically so in-flight spend is visible to the baseline — or state the limit on the governance page. Known design class, hence Medium rather than High.

## 15c. Channel identity get-or-create

`services/channels/router.py:318-342` (`_resolve_identity`), `:344-369` (`_resolve_session`), lock at `:31-40`.

Both resolvers are check-then-act with no `IntegrityError` handling, against tables that carry unique constraints. The in-process lock is keyed `(bot_id, platform_chat_id)` but the identity is keyed `(platform, platform_user_id)` — so **one Slack user messaging two channels at once takes two different locks**, both `create`, and the second flush violates `uq_channel_identity_platform_user`. `_managed_session` rolls back and re-raises, `process_channel_event` logs `channel_event_processing_failed`, and the user gets **no reply at all**.

The lock is per-process anyway, so with two uvicorn workers the session race and the rate-limit bucket are both wrong — `rate_limit_rpm` is N× looser than configured.

[...]

PR #1278, opened 2026-08-27
fix(chat): answer with the reader's star, and star idempotently

### LINKED ISSUE #1254: Conversation favourites: a permitted caller is refused, and the flag is false on most reads
**Found reviewing #929** (PR #1238), which added per-reader conversation favourites. Five things the feature does not yet hold. Two are ordinary races; the third is a wrong answer on the wire; the fifth refuses a caller who is allowed.

## 1. Starring a run log is refused for a caller who may read it

`set_favourite` calls `get_conversation` without the `AuthContext`, and `_may_read_trigger_log` returns false when `ctx` is `None` — so a caller who may read an ownerless trigger run log through `runs:view` plus agent access gets **404 from both favourite endpoints** on a conversation they can open. The existing conversation read routes take `Auth` and thread it through; these two do not.

This is the sharpest of the five: it is a refusal to somebody who is permitted, on a surface the run-log work deliberately opened up.

## 2. `is_favourite` is `false` on every read but two

Only `list_conversations` and `set_favourite` pass their rows through `_attach_favourites`. `GET /conversations/{id}`, the PATCH and archive responses, and `/shared-with-me` serialize ORM objects that never were — so the schema default emits `false` for a caller who really did star the thread. A newly advertised caller-specific field that is wrong on most of the responses carrying it.

## 3. Concurrent starring is not idempotent

Two overlapping POSTs for the same `(user_id, conversation_id)` can both read `existing is None`, and the primary key then makes one `flush()` raise — an integrity error where the endpoint promises idempotent success. A retried request does it too. `INSERT ... ON CONFLICT DO NOTHING` is the shape the codebase already uses for this (see `channel_identity_repo.get_or_create`, #17).

## 4. A double-click can leave it starred

Click the star twice before the first POST commits and the DELETE may check for the row first, succeed without deleting, and let the POST commit after it — so the thread stays starred though the reader's last action was to unstar it. Serializing the pending mutation per conversation, or disabling the control while one is in flight, closes it.

## 5. The optimistic rollback crosses an account change

If a request started by account A fails after account B has signed in and loaded the same shared thread, the unconditional rollback patches B's cached row and shows A's error in B's session. The surrounding hook already captures the initiating user for its other requests that can outlive their account; this one does not.

1 and 2 are assertions: a run-log conversation starred by a `runs:view` caller, and `is_favourite` true on a `GET` after a star. 3 and 4 want the interleaving orchestrated explicitly rather than raced through `gather` — the same shape #17's tests use, and for the same reason: the pair serialises and the first commits before the second reads. 5 is a hook test that changes the account between the request and its failure.

DEENUU1:
Three of the five things #1254 found in the conversation-favourites feature. Items 1 and 5 — the missing `AuthContext` on both favourite routes, and the optimistic rollback crossing an account change — are already on `main`.

## `is_favourite` was false on six responses out of eight

Only `list_conversations` and `set_favourite` passed rows through `_attach_favourites`, so `GET /conversations/{id}`, the PATCH, the archive response and `/shared-with-me` serialized ORM objects that never carried the flag — the schema default answered `false` to a caller who really had starred the thread, and the sidebar un-starred it on the next render.

Stamped in `get_conversation` instead: the one read every reader-scoped one goes through, so a route cannot forget. `list_shared_with_me` has its own repository call and gets its own stamp. A read with no reader — the admin listing, the run path resolving a thread — still asks for nobody's stars and pays no query to say so.

## Starring twice at once raised

`set_favourite` read the row and inserted when it saw none, so two overlapping POSTs for the same `(user_id, conversation_id)` both saw nothing and the second `flush()` violated the primary key: a 500 on an endpoint that promises idempotent success, and a retried request did it too. Now `INSERT ... ON CONFLICT DO NOTHING` — the shape `channel_identity_repo.get_or_create` already uses (#17) — and the unstar is an unconditional `DELETE` rather than a read followed by one.

## A double click could leave it starred

The POST and the DELETE were separate requests with nothing making the second wait, so the DELETE could be answered first and the POST commit after it, leaving the thread starred with nothing on screen saying so and no further click to reconcile them. One promise chain per conversation now, so the requests land in click order. The optimistic patch still happens at once — only the request is queued, so the star is as instant as it was — and a refusal rolls the row back only if its click is still the newest, which star/unstar/star makes observable.

## How it was verified

Both races are pinned deterministically rather than raced through `gather`, which the issue asked for and which does not reproduce either of them:

- one session's insert held open — uncommitted, holding the unique-index lock — while a second session stars the same thread, the shape `test_channel_identity_race.py` uses;
- star, unstar, star with the first POST refused and the queued DELETE never answering, so what is asserted is the patched cache rather than a fresh page.

Both fail without the fix, as do the two read tests (`is_favourite` true on a `GET` after a star, and a run-log conversation starred by a `runs:view` caller).

Nine tests in `test_services_conversation.py` gained a `favourite_ids` stub: they patch `conversation_repo` wholesale and now go through a read that asks the question.

`make check` green. `make test` at 100% on the platform layer, `tests/integration/test_conversation_favourites.py` green against pgvector on 5433 (9 passed), `make test-frontend-cov` green with the new guard's branch covered. `docs/architecture.md#a-favourite-belongs-to-the-reader-not-to-the-thread` updated: the second bullet said "on a listed row", which is what this change makes untrue.

Closes #1254

### DIFF
--- a/frontend/src/hooks/use-conversations.ts
+++ b/frontend/src/hooks/use-conversations.ts
@@ -431,14 +431,18 @@ export function useConversations(query: Partial<ConversationQuery> = {}) {
       const startedAs = useAuthStore.getState().user?.id;
       patchFavourite(id, favourite);
       const chains = favouriteChains.current;
+      let sent = false;
       const send = async (): Promise<void> => {
+        // Checked here as well as after, and this is the half that matters: a
+        // click queued behind a slow request is sent with whatever cookies the
+        // browser holds when its turn comes, so waiting out a sign-out would
+        // have A's queued star land as B's on a thread they both can read. The
+        // checks after the call only protect the cache.
+        if (!stillSameAccount(startedAs)) return;
         try {
           if (favourite) await apiClient.post(`/conversations/${id}/favourite`, {});
           else await apiClient.delete(`/conversations/${id}/favourite`);
-          if (!stillSameAccount(startedAs)) return;
-          // The band is an ordering the server applies, so the list is refetched
-          // to move the row - the star itself is already right on screen.
-          await invalidateLists();
+          sent = true;
         } catch (err) {
           if (!stillSameAccount(startedAs)) return;
           // Only the newest click owns the row's displayed state; rolling back
@@ -453,6 +457,13 @@ export function useConversations(query: Partial<ConversationQuery> = {}) {
       chains.set(id, mine);
       await mine;
       if (chains.get(id) === mine) chains.delete(id);
+      // Outside the chain: what has to be serialized is the write, and holding
+      // the next click behind a refetch of the lists means a slow GET can leave
+      // an unstar optimistic while the server keeps the star indefinitely.
+      if (!sent || !stillSameAccount(startedAs)) return;
+      // The band is an ordering the server applies, so the list is refetched to
+      // move the row - the star itself is already right on screen.
+      await invalidateLists();
     },
     [patchFavourite, invalidateLists, setError, t, tErrors],
   );
--- a/frontend/src/hooks/use-conversations.ts
+++ b/frontend/src/hooks/use-conversations.ts
@@ -16,6 +16,25 @@ import type {
   ConversationMessage,
 } from "@/types";
 
+/**
+ * One chain of pending stars per conversation, so two clicks on the same star
+ * cannot land out of order.
+ *
+ * The POST and the DELETE are separate requests and nothing made the second wait
+ * for the first, so a double click could have the DELETE answered first and the
+ * POST commit after it - leaving the thread starred while the screen said
+ * otherwise, with no further click to reconcile them (#1254). The optimistic
+ * patch still happens at once; only the request is queued, so the star is as
+ * instant as it was.
+ *
+ * Module scope rather than a ref, because what is being ordered is requests in
+ * flight in this tab and a request does not stop when the sidebar unmounts. Two
+ * components reading this hook are two refs, and navigating away from the chat
+ * and back is a third - each of which would send its click without waiting for
+ * a pending one it cannot see. An entry is removed when its chain drains.
+ */
+const favouriteChains = new Map<string, Promise<void>>();
+
 interface CreateConversationResponse {
   id: string;
   title?: string;
@@ -413,15 +432,6 @@ export function useConversations(query: Partial<ConversationQuery> = {}) {
     [queryClient],
   );
 
-  // One chain per conversation, so two clicks on the same star cannot land
-  // out of order. The POST and the DELETE are separate requests and nothing
-  // made the second wait for the first, so a double click could have the
-  // DELETE answered first and the POST commit after it - leaving the thread
-  // starred while the screen said otherwise, with no further click to
-  // reconcile them (#1254). The optimistic patch still happens at once; only
-  // the request is queued, so the star is as instant as it was.
-  const favouriteChains = useRef(new Map<string, Promise<void>>());
-
   const setFavourite = useCallback(
     async (id: string, favourite: boolean) => {
       // The account this started as, for the same reason every other request
@@ -430,7 +440,7 @@ export function useConversations(query: Partial<ConversationQuery> = {}) {
       // account's error.
       const startedAs = useAuthStore.getState().user?.id;
       patchFavourite(id, favourite);
-      const chains = favouriteChains.current;
+      const chains = favouriteChains;
       let sent = false;
       const send = async (): Promise<void> => {
         // Checked here as well as after, and this is the half that matters: a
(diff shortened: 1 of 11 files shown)


DEENUU1:
Both fixed in 8e2f6d49 — the P1 was right and the P2 with it.

**P1.** The account is now checked at the top of `send`, before the request goes. Every `stillSameAccount` call was after the await, which the queue makes worse rather than better: a click waiting its turn is sent with whatever cookies the browser holds when it gets one. Pinned by a test that queues a second click, signs a different account in, releases the first request, and asserts no DELETE is sent.

**P2.** `invalidateLists` is out of the chained function, after the chain entry is released, and runs only when the request it follows actually went. Pinned by a test where the list GET never answers and the next click's DELETE still fires.

The two `py/ineffectual-statement` threads from `github-code-quality` are the bare `await <task>` shape adjudicated in `docs/code-review.md` — the cancellation idiom in the `finally`, and running the task to completion so the assertion after it is about a finished task. Not changing those.

@codex review

frontend/src/hooks/use-conversations.ts
```diff
@@ -413,6 +413,15 @@ export function useConversations(query: Partial<ConversationQuery> = {}) {
...
   );
 
+  // One chain per conversation, so two clicks on the same star cannot land
+  // out of order. The POST and the DELETE are separate requests and nothing
+  // made the second wait for the first, so a double click could have the
+  // DELETE answered first and the POST commit after it - leaving the thread
+  // starred while the screen said otherwise, with no further click to
+  // reconcile them (#1254). The optimistic patch still happens at once; only
+  // the request is queued, so the star is as instant as it was.
+  const favouriteChains = useRef(new Map<string, Promise<void>>());
```
Labeeb2339: [P2] Preserve the mutation chain across hook remounts. If a star request is pending when this hook unmounts, this new `useRef` starts with an empty map on the next mount. An immediate unstar can therefore send DELETE without waiting for the old POST; DELETE may commit first and the late POST leaves the server starred while the UI shows unstarred after refetch. Keep the per-conversation queue in state that outlives the hook (or otherwise coordinate outstanding writes).

Labeeb2339 (commented):
Found one correctness issue in the favourite-write ordering.

DEENUU1:
Fixed in 45441700. Right on both counts — the ref is discarded when the sidebar unmounts while the request it was ordering carries on, and two components reading the hook are two refs with the same hole.

`favouriteChains` is a module constant now: what is being ordered is requests in flight in this tab, and their lifetime is the tab's rather than any one render's. Pinned by a spec that stars a thread, unmounts the hook, remounts it and unstars — the DELETE is not sent until the POST answers, and it *is* sent when it does. It fails against the ref.

One consequence worth naming: a held request now outlives its test, so the two specs that hold one open release it at the end. Without those lines eleven tests failed, all of them waiting behind a star left pending in an earlier one. Said once in the describe block, because it is the trap for the next spec written there.

@codex review

DEENUU1:
Both fixed in ecb19b5a.

**The organization.** Right — `apiClient` reads the store when the request is sent, so a click queued across a switch went out under the new tenant and was refused for a conversation in the old one, leaving the server on the previous click's answer. The account was already captured for exactly this; the organization is the same fact one layer down, and both are now one `stillHere()`.

**The superseded refetch.** Also right. Only the chain's newest entry invalidates now; a superseded one would answer with the state its own request left behind and patch that over the newer optimistic row, or leave the row visibly starred against the reader's last click when the DELETE is slow. The click that supersedes it refetches when its own write lands.

Both pinned by specs that fail without their half.

@codex review

DEENUU1:
Both fixed in 4c60b271.

**The chain key.** Right, and the mechanism is that a hung request never drains its chain — so a shared key leaves the next reader appended to a queue that will not move. The key is now `account:organization:conversation`: ordering is only owed within one person's view of one thread. Pinned by a spec where account A's star hangs and account B's unstar on the same conversation still reaches the server.

**The redundant queries.** Also right — `GET /conversations/{id}/messages` resolves the conversation through both `list_messages` and `conversation_cost`, and neither serializes the star. `include_favourite` defaults to on, which is what keeps a route from forgetting (the whole of #1254), and is off at the five call sites that authorize and discard: those two, `add_message`, `delete_conversation`, and `set_favourite`, which overwrites the flag anyway. A test asserts `favourite_ids` is not awaited on such a read, and `docs/architecture.md` says which way the default points and why.

@codex review

DEENUU1:
Fixed in e59d8524. Right on all four, and `_resolve_in_org` is the one that mattered: every turn of an existing chat goes through it, so a message was paying for a star nothing serialized. The three workspace guards authorize and discard the row the same way.

`include_favourite=False` at each. The default stays on — that is what keeps a route which really does serialize a conversation from forgetting, which is the whole of #1254 — and the opt-out is now stated at every call site whose result is discarded or is not a conversation. `docs/architecture.md` lists them rather than naming two examples.

@codex review

PR #1279, opened 2026-08-27
fix(api): read last-seen from every session, count only usable ones

### LINKED ISSUE #1256: The admin drawer reports "Never signed in" for anyone who has signed out
**Found reviewing #942** (PR #1239), and it lands on the drawer's own headline claim.

## What's wrong

The detail read takes `active_only=True` for both figures. That is right for *how many sessions are open*, and wrong for *when were they last here*:

- **A user who has signed out, or had their sessions revoked, has no active session rows at all** — so `last_seen_at` comes back null and the drawer reports **"Never signed in."** That is most accounts most of the time, and it says the opposite of the truth.
- The reverse on the other figure: an expired session row stays `is_active` until it is next reused, so it is still counted as open.

The PR argues for this field precisely because *created and never used* and *dormant since March* are different decisions an admin makes differently. As written, the two are rendered identically for anybody who logged out — the one case the distinction exists to separate.

## Where

`backend/app/services/admin.py` (or wherever the detail read assembles the two figures) — the single `active_only=True` serving both.

## The fix shape

Two scopes rather than one: derive last-seen from **all** sessions, and count open ones with `is_active` **and** an unexpired `expires_at`.

An integration case with a user whose sessions are all inactive: `last_seen_at` is their most recent session's timestamp, not null, and the open count is zero. And one with an `is_active` row already past `expires_at`: it does not count as open. Both fail against the current read.

## Also on the same review, smaller

- When the detail request fails, the value branch does not inspect the error and renders **"Never signed in"** while the memberships and sessions blocks below it report the failure — the drawer contradicting itself on one screen. The PR already gives each new block the three states; this is the fourth field not getting them.
- A membership row links to `/orgs/{id}`, whose read requires the *current* user's membership — so most of the newly linked rows 404 for the admin, including the target's personal organization in the normal case.
- The Activity link carries a person filter but not the conversation's tenant, while `/runs` is scoped to the active organization — so it commonly lands on no runs, or on unrelated runs for the same person in the current tenant.

These three are marked outdated against the head; the last-seen scope is not.

DEENUU1:
The admin drawer took both of its session figures off the same read — the user's *active* sessions — and both were wrong for it.

## "Never signed in" for anybody who had signed out

A user who signs out, or has their sessions revoked, has no active session row at all. `last_seen_at` came back null and the drawer rendered **"Never signed in."** That is most accounts most of the time, and it is the opposite of the truth — on the one field the drawer exists to answer, and in exactly the case the field is there to separate from *created and never used*.

Where somebody was last seen is a fact about every session they have ever had, so the read takes the whole history (`open_only=False`) and the head of it — most-recently-used first — is the answer.

## An expired session counted as open

Nothing sweeps a session that simply lapses: the row stays `is_active` until the next refresh finds it past `expires_at` and declines it. So a session nobody can use was reported as open.

"Open" now means `is_active AND expires_at > now()`, and it lives in `app/repositories/session.py` rather than at one call site — which is why the flag is `open_only` and not `active_only`: the old name described the column, not the question. The user's own devices list goes through the same two functions, so it stops offering an expired row to revoke as well; that surface had the identical defect and it is one predicate, so it is fixed here rather than filed.

`newest_session_at` stays scoped to the open ones — it is read beside their count, and "newest session August" under "0 open sessions" is a sentence about nothing.

## How it was verified

`backend/tests/integration/test_admin_user_last_seen.py`, against a real Postgres because both halves are a `WHERE` clause:

- every session inactive → `last_seen_at` is their most recent session's timestamp, open count 0, `newest_session_at` null;
- no session ever → `last_seen_at` still null (the distinction the field exists for);
- an `is_active` row past `expires_at` → not open, and still counted for last-seen;
- an unexpired row beside an inactive one → open count 1, and both dates are the open one's.

The first and third fail against the old read; the other two pin what was already right.

`make lint` and `make test` (100% on the platform layer) green. `tests/test_sessions_pagination.py` pinned the old kwarg name and now pins the new one.

The three smaller items on the issue were already fixed on `main` in #1239: the drawer's last-seen field renders `-` rather than "Never signed in" when the detail request fails, and both the membership rows and the Activity link are plain text rather than destinations that 404.

Closes #1256

### DIFF
--- a/backend/alembic/versions/0061_session_last_used_index.py
+++ b/backend/alembic/versions/0061_session_last_used_index.py
@@ -1,0 +1,45 @@
+"""The index every read of a user's sessions actually wants (#1256).
+
+All three of them ask the same question - this user's sessions, most recently
+used first: the devices list, its page count, and the admin drawer's last-seen.
+The table had an index on `user_id` alone, so answering meant fetching a user's
+rows and sorting them.
+
+That is fine for a week-old account and not for a year-old one. Nothing prunes
+`sessions`: a refresh deactivates the row it used and inserts another, so the
+history grows for as long as somebody keeps signing in, and the admin drawer got
+slower for exactly the accounts an administrator is most likely to open.
+
+`id` is in the index because `last_used_at` ties on two sign-ins in the same
+moment and the page order has to be total - the same reason the query orders on
+it.
+
+The single-column index goes: this one leads on `user_id`, so Postgres would
+never choose it and every insert would still maintain it.
+"""
+
+from collections.abc import Sequence
+
+from sqlalchemy import text
+
+from alembic import op
+
+revision: str = "0061_session_last_used_index"
+down_revision: str | None = "0060_conversation_favourites"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+_COMPOSITE = "sessions_user_id_last_used_at_idx"
+_SINGLE = "sessions_user_id_idx"
+
+
+def upgrade() -> None:
+    # `text(...)` for the descending column: a plain string would be quoted as
+    # one identifier named "last_used_at DESC".
+    op.create_index(_COMPOSITE, "sessions", ["user_id", text("last_used_at DESC"), "id"])
+    op.drop_index(_SINGLE, table_name="sessions")
+
+
+def downgrade() -> None:
+    op.create_index(_SINGLE, "sessions", ["user_id"], unique=False)
+    op.drop_index(_COMPOSITE, table_name="sessions")
--- a/backend/app/db/models/session.py
+++ b/backend/app/db/models/session.py
@@ -3,7 +3,8 @@
 import uuid
 from datetime import UTC, datetime
 
-from sqlalchemy import Boolean, DateTime, ForeignKey, String, Text
+from sqlalchemy import Boolean, DateTime, ForeignKey, Index, String, Text
+from sqlalchemy import text as sa_text
 from sqlalchemy.dialects.postgresql import UUID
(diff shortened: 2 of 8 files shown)


DEENUU1:
Fixed in 4780808e — right, and it lands hardest on exactly the accounts an administrator opens. Nothing prunes `sessions` (a refresh deactivates the row it used and inserts another), so the history read grew without bound.

`limit=1`, which the existing ordering already makes correct: most-recently-used first, so the head is the whole answer. The open-sessions read stays a list because its length is one of the two figures the drawer shows, and it is bounded by what is still usable.

An API test pins both scopes and the bound in one place — the calls are `open_only=False, limit=1` then `open_only=True`, in that order.

@codex review

DEENUU1:
Both fixed in ac835594.

**The index.** Right that `LIMIT 1` bounds the returned rows and not the scan. `0061` adds `(user_id, last_used_at DESC, id)` and drops `sessions_user_id_idx`, which the new one leads on — Postgres would never choose it and every insert would still maintain it. `id` is in the index for the same reason the query orders on it: `last_used_at` ties on two sign-ins in the same moment and the page order has to be total. Applied and downgraded against pgvector, `alembic check` clean against the models, and the built database reports `btree (user_id, last_used_at DESC, id)`.

**The cutoff.** Also right, and it breaks without a concurrent write: `list_sessions` asks for a page and then a total, and each `_open` took its own clock. `_open` now takes the caller's `now` and the service computes one per response; the test asserts the two calls carry the same value.

One thing for whoever merges: `feat/rag-reranker` (#911) carries a different `0061`, so whichever lands second renumbers.

@codex review

Issue #1204, opened 2026-08-26
Alert emails name no organization, so they open whichever one the reader last used


DEENUU1:
Every link in every notification email is organization-agnostic, and the page it
opens acts on whichever organization the reader last used.

`apiClient` stamps `X-Organization-Id: activeOrgId` on every request, and
`activeOrgId` is persisted per browser. None of the alert URLs carries an
organization:

| Email | Link |
|---|---|
| Approval requested | `/runs?tab=approvals` |
| Budget exceeded | `/agents/{agent_id}` |
| Usage report (org) | `/agents` |
| Usage report (agent) | `/agents/{agent_id}` |

`run.organization_id` is in scope at every one of those call sites and is
discarded.

So a person in two organizations who was last working in Globex opens the
approval alert for a run in Acme and reads **Globex's** queue — which is very
likely empty, and reads as *nothing is waiting* about a run that is parked and
ageing towards `ApprovalService.expire_stale`. The agent links are worse in a
quieter way: `/agents/{id}` under the wrong organization is a 404 or a refusal
for an agent the reader can genuinely see, one switch away.

This is the same class as **#1032** — a page that names an organization in its URL
*is* that organization, adopted once in `ActiveOrgGuard` — except that here the
URL names none, so there is nothing to adopt.

## Sequence

1. Be a member of two organizations. Work in the second one, so `activeOrgId` is
   its id.
2. Park a run in the first: a tool call needing approval.
3. Open the alert. Activity loads the *second* organization's queue. Nothing is
   waiting there.

`backend/app/services/notifications.py:115,151,187,252` build the URLs;
`frontend/src/lib/api-client.ts` stamps the header;
`frontend/src/components/layout/active-org-guard.tsx` is where an adoption would
live, beside the tenant cache reset it has to happen before (#1032).

## Done when

- Every alert link carries the organization the alert is about — one parameter
  name, decided once, not four call sites inventing one.
- The dashboard adopts it in `ActiveOrgGuard`, before the tenant cache reset,
  and a reader who lands there is switched rather than shown somebody else's
  page.
- A reader who is no longer a member of that organization is told so, rather
  than being silently left where they were — the refusal has to say which
  organization it was about.
- A test per email kind asserts the id is in the URL, beside the destination
  assertions `TestWhereAnAlertSends` now holds.

## Related

Found by the automated reviewer on **PR #1202** (#935). Sibling: **#1203** — the
[...]

PR #1280, opened 2026-08-27
fix(approvals): tell a non-decider their run is held, not to approve it

### LINKED ISSUE #1203: The approval alert is mailed to people who may not decide, and the page just has no queue for them
The default audience for a parked tool call is **the run's initiator plus the
administrators**, and the initiator is very often somebody who cannot decide.

`approvals:decide` is held by `owner`, `admin` and `operator`. A `builder` and a
`member` do not hold it — and a builder starting their own agent from the chat is
the ordinary case, not an edge one:

```
$ uv run python -c "from app.core.permissions import ROLE_PERMS, Perm; ..."
owner   -> approvals:decide
admin   -> approvals:decide
builder -> -
operator-> approvals:decide
member  -> -
viewer  -> -
```

So that recipient gets **"[Agent] is waiting on your approval"** with a **Review
the request** button, follows it, and Activity draws no Approvals tab at all —
`page.tsx` gates the tab on `canDecide`, correctly, because `GET /approvals`
carries the same permission. The mail is addressed to somebody the platform will
refuse, and the refusal is silent: not an explanation, just an absent tab.

This is not the destination bug (#935, fixed) under another name. That one sent
everybody to a page with no queue. This one is about *who is told*, and it
survives the fix.

## Two shapes of answer, and they are different products

- **Filter the audience by permission.** `_audience` drops a recipient who does
  not hold `approvals:decide`, so the mail goes only to people who can act. Risk:
  a run whose initiator is a builder and whose organization has no operator or
  admin online then tells nobody who cares that it is parked — and the initiator
  is the one person who is definitely waiting on it.
- **Tell them something true.** The initiator is told their run is held and by
  what, without a **Review the request** button they cannot use; the deciders get
  the button. One email key, two contexts — or two keys.

The second is probably right: the initiator has a legitimate interest in knowing
their run stopped. But it is a copy and template decision, not a one-line fix,
which is why this is filed rather than folded into #935.

## Sequence

1. Sign in as a `builder`, start a run of an agent whose spec parks a tool call.
2. Read the mail: subject "… is waiting on your approval", button "Review the
   request", link `/runs?tab=approvals`.
3. Follow it. Activity opens on Runs; there is no Approvals tab, no queue and no
   Approve or Reject control.

`backend/app/services/notifications.py:121` (`approval_requested`, the audience),
`backend/app/core/permissions.py` (`ROLE_PERMS`),
`frontend/src/app/[locale]/(dashboard)/runs/page.tsx` (`canDecide` gates the tab).

## How you know it is fixed

A recipient who does not hold `approvals:decide` either is not mailed, or is
mailed something whose call to action they can actually complete. A test asserts
which, per role, at the audience — the wiring the rest of
`tests/test_notifications.py` already pins.

## Related

Found by the automated reviewer on **PR #1202** (#935). Sibling: **#1204** — the
same email opening the wrong organization's queue. Cluster: the approvals chain
around **#925**, **#128**.

DEENUU1:
`approvals:decide` belongs to `owner`, `admin` and `operator`. The default audience for a parked tool call is the run's initiator plus the administrators — and a builder starting their own agent from the chat is the ordinary initiator, not an edge case. So the alert routinely reached somebody the platform will refuse: **"[Agent] is waiting on your approval"**, a **Review the request** button, and then an Activity page with no Approvals tab at all. The refusal arrived as an absent tab rather than a sentence.

## Split by the permission, not trimmed to it

| Recipient holds | Gets |
|---|---|
| `approvals:decide` | the request, with the link to the queue |
| anything less | a new `approval_pending` mail: the run is held not failed, the people who can decide have been told, nothing is asked of them — linking to the **agent**, which every role holds `agents:view` for |

Trimming instead — dropping a recipient who cannot decide — was the other shape the issue offered, and it loses the one person definitely waiting on the run. That is the whole reason `initiator` is in the default audience.

The deciding roles come off `ROLE_PERMS` rather than a list beside it:

```python
_DECIDING_ROLES = [str(role) for role, perms in ROLE_PERMS.items() if Perm.APPROVALS_DECIDE in perms]
```

A role gaining or losing the permission must not leave this routing behind — that is the same defect one level up — and a test pins the derivation, including that `builder` and `member` are not in it. App admins count as deciders: they hold no membership row and `AuthContext.permissions` gives them everything, so a request is a mail they can act on.

## Cost

One extra query per approval alert (`list_emails_by_role` for the deciding roles, plus the app-admin read). It is the same pair `_administrators` already makes, and it is what routing needs to know: the audience arrives as addresses, and whether an address may decide is not derivable from how it entered.

The new mail deliberately does **not** name the organization, which would have meant a third read on this path. Naming the tenant in alert links is #1204's subject and it will want the id in the URL anyway.

## Verified

`tests/test_notifications.py::TestApprovalRequested` now asserts each side separately: a decider gets only the request; an initiator who cannot decide gets only the pending mail, with `agent_url` and no `approvals_url`; an audience holding both gets one of each; an app admin is a decider. The spec-narrowed case still asserts that an audience nobody named costs no query — `list_emails_by_role` is awaited once, and only about the deciding roles.

`make test` at 100% on the platform layer, `make lint` and `make docs-build` green. `docs/governance.md` gains "The approval alert is two emails" under Alerts.

Note the link in the decider's mail is `/agents/{id}` on this branch, unchanged: #935 / PR #1202 is what moves it to the queue, and this branch deliberately does not touch that line.

Closes #1203

### DIFF
--- a/backend/app/services/email/service.py
+++ b/backend/app/services/email/service.py
@@ -21,6 +21,7 @@ class EmailKey(enum.StrEnum):
     NEWSLETTER_WELCOME = "newsletter_welcome"
     BUDGET_EXCEEDED = "budget_exceeded"
     APPROVAL_REQUESTED = "approval_requested"
+    APPROVAL_PENDING = "approval_pending"
     USAGE_REPORT = "usage_report"
 
 
@@ -42,6 +43,7 @@ _CATEGORIES: dict[EmailKey, EmailCategory] = {
     # columns), consulted where recipients are resolved, in NotificationService.
     EmailKey.BUDGET_EXCEEDED: EmailCategory.LIFECYCLE,
     EmailKey.APPROVAL_REQUESTED: EmailCategory.LIFECYCLE,
+    EmailKey.APPROVAL_PENDING: EmailCategory.LIFECYCLE,
     EmailKey.USAGE_REPORT: EmailCategory.LIFECYCLE,
 }
 
--- a/backend/app/services/notifications.py
+++ b/backend/app/services/notifications.py
@@ -170,16 +170,17 @@ class NotificationService:
             self._send(EmailKey.APPROVAL_REQUESTED, approvers, context)
         if not onlookers:
             return
+        # No link at all, and that is the point of it: `agents:view` being a role
+        # permission does not make one agent reachable, because agent access is
+        # resolved per resource - a `chosen` recipient with no grant to a private
+        # agent would get a second call to action the platform refuses. Nothing is
+        # asked of this reader, so nothing is offered.
         self._send(
             EmailKey.APPROVAL_PENDING,
             onlookers,
             {
                 "agent_name": context["agent_name"],
                 "tools": context["tools"],
-                # The agent, not the queue: `agents:view` is held by every role,
-                # so this is a destination the whole audience can open - which is
-                # the property the Review button did not have.
-                "agent_url": f"{self._frontend}/agents/{agent.id}",
                 "app_name": context["app_name"],
             },
         )
--- a/backend/app/services/notifications.py
+++ b/backend/app/services/notifications.py
@@ -48,7 +48,7 @@ from app.agents.capabilities.budget import BudgetScope
 from app.agents.spec import AgentSpec, AlertAudience, AlertSpec
 from app.core.background import spawn
 from app.core.config import settings
-from app.core.permissions import OrgRoleName
+from app.core.permissions import ROLE_PERMS, OrgRoleName, Perm
 from app.db.models.agent import Agent
 from app.db.models.agent_run import AgentRun
 from app.db.models.user import NotificationPreference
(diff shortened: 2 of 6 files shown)


DEENUU1:
Both fixed in 86e20d70.

**The false claim.** Right: an audience of one non-decider — an agent whose approvals go only to the initiator, or a deployment where every decider opted out — means nobody who can act was mailed, so "the people who can have been told" left the reader waiting on somebody who never heard. The copy now states what is true of the platform rather than of this send: approving belongs to an owner, an admin or an operator, and nothing is asked of them.

**The link.** Also right, and it is the same defect this branch is about one place along: `agents:view` being a role permission does not make one agent reachable, because agent access is resolved per resource, so a `chosen` Member with no grant to a private agent would get a second refused call to action. The mail carries no link at all — nothing is asked of that reader, so nothing is offered — and the test asserts the absence of any `*_url` in the context rather than a destination.

@codex review

DEENUU1:
Fixed in ed932be2. Right — a `chosen` recipient is an arbitrary organization member who neither started the run nor owns the agent, so the footer's "a run of yours is waiting" assigned it to them. It reads "an agent run is waiting" now. The body was already neutral: it explains why a run went quiet rather than claiming whose it is.

@codex review

DEENUU1:
Fixed in ed932be2. Right — a `chosen` recipient is an arbitrary organization member who neither started the run nor owns the agent, so the footer's "a run of yours is waiting" assigned it to them. It reads "an agent run is waiting" now. The body was already neutral: it explains why a run went quiet rather than claiming whose it is.

@codex review

DEENUU1:
Head is ed932be2 now — the neutral footer wording — which is not the commit the last review names.

@codex review

PR #1283, opened 2026-08-27
test(chat): cover the two-digit ordered-list indent deterministically

### LINKED ISSUE #1264: Frontend coverage gate flakes at 99.98% on markdown-content.impl.tsx:37
**Found running `make test-frontend-cov` while working #154 and #130.** The
frontend coverage gate intermittently reports **99.98% statements (one
uncovered)** and fails the 100% threshold, then passes on a plain re-run of the
same tree.

## What it is

The single missed statement is `src/components/chat/markdown-content.impl.tsx:37`
(confirmed via `coverage/coverage-final.json`: exactly one `s` entry at 0 hits,
that line). Lines and functions stay at 100%; only that one statement flips. Two
separate full runs on unrelated branches (#154, #130) each hit it once and each
passed on re-run — so it is a flaky-coverage statement, not a real gap, and it is
unrelated to the branch under test.

Likely a test whose render of `markdown-content.impl.tsx` does not deterministically
reach line 37 under the parallel/xdist-style worker scheduling vitest uses (the
same class of flake as a test that occasionally times out under load, per
`docs/testing.md`'s note on loaded-machine deadlines).

## Why it matters

A required gate that fails ~1 run in N on a clean tree costs a re-run per red CI
and trains people to ignore a red coverage job.

## How I'd know it's fixed

- [ ] `bun run test:coverage` reports 100% statements on ten consecutive runs of a
      clean `main`.
- [ ] `markdown-content.impl.tsx:37` is deterministically covered (a test that
      always exercises that branch), or the line is refactored so it cannot be
      conditionally skipped.

Pre-existing, surfaced by the coverage runs for #154/#130, not caused by them.
Checked #168 — no existing issue covers this coverage flake; nearest neighbours
are the e2e flakes #130/#154 (a different layer).

OchnikBartek:
## Summary

`make test-frontend-cov` intermittently failed the 100% statement gate at 99.98%,
the one miss being `markdown-content.impl.tsx:37` — the `pl-8` return in
`orderedIndent`, the indent band for a 10-99 item ordered list. Nothing in the
markdown-content suite renders a list that size; the statement was covered only
when some *other* suite happened to render one, and under parallel scheduling that
render is not guaranteed, so the statement flipped uncovered and the gate went red
on a clean tree (#1264).

## Changed

- The ordered-list indent test already pins the 1-9 (`pl-6`), 100+ (`pl-10`) and
  1000+ (`pl-12`) bands. Added the missing two-digit case — a 15-item list
  asserting `pl-8` — so the band is covered here, deterministically, rather than
  by accident.

## Testing

- `bunx vitest run markdown-content.impl.test.tsx` green (23 tests).
- `bun run test:coverage` green: 100% statements/lines/functions, 97.83% branches.

Test-only; the branch is live (a 10-99 item list is reachable), so it is covered,
not removed.

Closes #1264

### DIFF
--- a/frontend/src/components/chat/markdown-content.impl.test.tsx
+++ b/frontend/src/components/chat/markdown-content.impl.test.tsx
@@ -130,6 +130,12 @@ describe("rendering an answer", () => {
     expect(short.container.querySelector("ol")).toHaveClass("pl-6");
     short.unmount();
 
+    // The two-digit band: covered here deterministically so it does not depend on
+    // some other suite happening to render a 10-99 item list (#1264).
+    const tens = markdown(Array.from({ length: 15 }, (_, i) => `${i + 1}. item`).join("\n"));
+    expect(tens.container.querySelector("ol")).toHaveClass("pl-8");
+    tens.unmount();
+
     const long = markdown(Array.from({ length: 120 }, (_, i) => `${i + 1}. item`).join("\n"));
     expect(long.container.querySelector("ol")).toHaveClass("pl-10");
     long.unmount();


OchnikBartek:
@codex review

Issue #1077, opened 2026-08-22
A plan does not survive to the conversation's next turn, and the agent denies it exists


DEENUU1:
## What breaks

An agent with the `planning` capability writes a plan in one chat turn and denies
it in the next: *"Nie ma aktywnego planu — nigdy go nie stworzyłem."* `read_plan`
answers `No plan yet.` about a plan the same conversation created two messages
earlier, and the transcript above it still shows the three steps.

The plan's lifetime is one **run**, and in chat every user message is a new run.
`new_plan_store()` hands the capability a fresh `InMemoryPlanStore`
(`backend/app/agents/capabilities/planning/_capability.py:269`), and the only
thing that carries a checklist across a boundary is an approval park:
`open_plan_store` re-seeds from `PausedRunState.plan`, `dump_plan` writes it back
(`:280`, `:295`). Nothing seeds a store from the *conversation*, so turn N+1
starts empty.

That was a deliberate choice - the docstring says so, "not a record kept across
conversations" - and it is the wrong one for the surface this platform leads with.
A person watching a chat sees one thread; the agent sees a plan that evaporates
between their messages, then contradicts what is on screen.

## Sequence

1. Bind `planning` to an agent, open a chat.
2. `use planning and make 3 example tasks` → the agent calls `write_plan`, three
   steps land.
3. Next message: `ok work on the first one` → the agent works from the transcript,
   not the store, and any `update_task_status` fails with
   `Step with id '…' not found.`
4. Next message: `ok finish all of them` → `read_plan` answers `No plan yet.` and
   the agent says it never made one.

Reproduced on `feat/tool-standards` at conversation
`b9693bec-076d-4294-93d9-5249742c02b2`; the frontend plan strip added on that
branch is what made the contradiction visible on one screen.

## What "fixed" looks like

A plan written in a conversation is still there on the conversation's next turn,
and `update_task_status` on a step from an earlier turn works. Concretely:
the runner seeds the store from the conversation's stored plan the way it already
seeds from `paused_state`, and writes it back when the run ends rather than only
when it parks. Where "the conversation's stored plan" lives is the decision -
a column beside the conversation, reached through a repository, keeps the raw SQL
out of the library's hands (`pydantic_ai_harness.planning` also ships
`PostgresPlanStore` / `RedisPlanStore`, which own their own queries and would sit
outside the layering rule).

Two questions this needs an answer to before code:

- **Scope.** Per conversation is the obvious one for chat. A channel thread and an
[...]

Issue #788, opened 2026-08-16
Add a Memory capability: persist and recall facts across runs


DEENUU1:
> **Design first.** Still to be planned — a written plan goes to @DEENUU1 before any
> code. `effort:l` is a placeholder until the plan settles it.

## The brief

A **memory** capability: a tool an agent can call to persist facts and recall them on a
later run, so it stops re-learning the same thing every conversation. Write a memory,
read the relevant ones back on the next request.

## Read this before planning

The whole value of this platform is in what it refuses, and memory is a refusal surface
before it is a feature:

- **What is the scope of a memory?** Per-agent, per-organization, or per-end-user? A
  memory bleeding across those boundaries is a cross-tenant read (#1's class of bug) or
  one user reading another's facts. Decide the scope in the plan, and it is probably more
  than one axis (this agent, for this user, in this org).
- **Where does it live?** A new table, or does this ride on the RAG store? Semantic
  recall wants embeddings and pgvector; exact recall wants a key. Say which, and whether a
  memory is even the same thing as a knowledge-base document (it is not — a KB doc is
  authored content, a memory is something the agent decided to keep).
- **Retrieval strategy.** Injected on every request (context cost, budget-metered) vs. a
  tool the model calls when it wants to remember. The second is cheaper and auditable;
  the first is more like how a person remembers. Probably both, gated by config.
- **A memory has a lifetime.** Can it be listed, edited, deleted by the operator? A store
  the agent can write to but nobody can inspect or clear is a liability.

## Nearest issue

The context/harness roadmap family — adjacent to **#48/#49** (Add Context / context
manager) and distinct from **#57** (tool-output overflow): this is durable cross-run
memory, not in-window context handling. Storage may overlap the RAG store (**#36**,
pydantic-ai-backend / filesystem) — check before inventing a second store.

## The capability contract (abbreviated)

- **The registry is the only door.** No `@agent.tool`. Register with `@register(...)`,
  `tools=[...]` declared, add to `load_builtins()`. `app/agents/capabilities/clock/` is the
  shape; a stateful one closer to `code_execution/`.
- `app/agents/**` is already in the 100% coverage + strict-typing globs — no config edit.
- The capability `id` is permanent; published specs reference it.
- A write/read that spends tokens (embedding a memory) goes through `metered_by` /
  `record_ambient_usage`, or the budget guard cannot see it (#16).
- Docs: `app/agents/capabilities/**` → `docs/reference/capabilities.md`, from docstrings.

[...]

PR #1287, opened 2026-08-27
fix(agents): stop a new turn starting from a finished checklist

### LINKED ISSUE #1221: A completed plan comes back: the conversation seeds a checklist nobody is working to
A plan now belongs to the conversation rather than to one turn (#1077), which is
right — and it means a **finished** checklist comes back too.

`keep_plan` stores whatever the store held when the run ended, `completed` steps
included, and the next turn seeds from it. So a thread whose three steps were all
ticked off in August is reopened in November with those three steps still in the
agent's working state, and `read_plan` answers with them. The agent is then
working to a plan about a task nobody is doing.

#1077 named this as one of its two open questions and answered the other one
(scope: per conversation, and a run with no conversation persists nothing — the
whole block is guarded on `conversation_id is not None`). This is the half left.

## The decision

Not obviously "clear it". A checklist somebody can still see in the transcript,
answered with *No plan yet*, is the contradiction #1077 was filed about — and
clearing on completion recreates a smaller version of it: the last message shows
three ticked steps and the next turn denies them.

Three shapes, and the third is probably right:

- **Clear when every step is `completed` or `cancelled`**, at the end of the turn
  that finished it. Cheapest. Costs the agent any memory of what it just did,
  one message later.
- **Clear on age.** A plan older than N days is not seeded. Arbitrary, and the
  number would be invented rather than measured.
- **Seed it, and say what it is.** The store carries the finished checklist and
  the tail reminder says it is finished, so `write_plan` starting a new one is
  the obvious next move rather than an interruption of the old. No state is
  dropped and nothing contradicts the screen.

## Done when

- Reopening a thread whose plan was completed does not leave the agent working to
  it, and does not make it deny work it did.
- Whichever shape is chosen is written down in
  `app/agents/capabilities/planning/_capability.py`'s docstring beside the
  seeding rule, because the next reader will ask exactly this.
- A test per case: a completed plan, a half-finished one, an empty one.

## Related

The half of **#1077** its implementation did not answer; found verifying that
issue. **#788** (a Memory capability) is the same question about state that
outlives a run, one size up.

DEENUU1:
The half of #1077 its implementation left open. `keep_plan` records whatever the store held when the run ended — completed steps included — and the next turn seeds from it. So a thread whose three steps were all ticked off in August opens in November with the tail reminder calling them **"your current plan"** and `read_plan` answering with them. The agent works to a checklist about a task nobody is doing.

## The shape chosen

The issue offered three and expected the third. This is the **first, moved**: the filter is at the *seed* rather than at the moment the last step is ticked — which is what answers the issue's objection to clearing ("the last message shows three ticked steps and the next turn denies them").

- Within the turn that finishes a plan the store still holds it. `read_plan`, the tail reminder and the transcript all agree, and the agent can summarise what it just did.
- It is the **next** question that starts clean. The ticked checklist is still in the messages above it, where it reads as what *was* done rather than as what is being done — so nothing is denied, and there is no contradiction on screen.

The third shape ("seed it, and say it is finished") needs the library's tail reminder to say so: `_reminder_text` in `pydantic_ai_harness.planning` opens with "Your current plan (keep it updated with the planning tools)" unconditionally, so seeding a finished plan and labelling it correctly is a change in that package rather than here.

Nothing is deleted. The row keeps the finished plan; `still_open` decides only what a fresh turn is seeded with. A resume seeds from `PausedRunState.plan` and does not come through it, being mid-plan by construction.

**Finished** means at least one step and every step `completed` or `cancelled`. `blocked` is work outstanding — something still has to unblock it — which is the one status that could read either way, so it has its own test.

## Verified

Six cases on the helper: every step ticked, a cancelled step counting as finished, one step left seeding the whole plan, a `blocked` step keeping it, `None` and `[]` surviving as themselves (they mean different things to `keep_plan`), and the store `open_plan_store` actually opens. One on the runner — a conversation whose stored plan is all `completed`/`cancelled` prepares with an empty store — which fails without the change.

`make test` at 100% on the platform layer, `make lint` and `make docs-build` green.

The rule is written down in both places the issue asked for: `planning/_capability.py`'s docstring beside the seeding rule, and `docs/reference/capabilities.md`, whose paragraph said the opposite ("A finished checklist is kept rather than cleared").

Closes #1221

### DIFF
--- a/backend/app/agents/capabilities/planning/__init__.py
+++ b/backend/app/agents/capabilities/planning/__init__.py
@@ -31,6 +31,7 @@ from app.agents.capabilities.planning._capability import (
     dump_plan,
     new_plan_store,
     open_plan_store,
+    still_open,
 )
 
 __all__ = [
@@ -41,6 +42,7 @@ __all__ = [
     "dump_plan",
     "new_plan_store",
     "open_plan_store",
+    "still_open",
 ]
 
 
--- a/backend/app/agents/capabilities/planning/_capability.py
+++ b/backend/app/agents/capabilities/planning/_capability.py
@@ -50,6 +50,10 @@ from app.agents.capabilities._tool_text import ToolText
 if TYPE_CHECKING:
     from pydantic_ai_harness.planning import PlanStore
 
+# What "this step needs no more work" looks like in a stored `PlanItem` dump. A
+# `blocked` step is not one of them: something still has to unblock it.
+_FINISHED_STATUSES = frozenset({"completed", "cancelled"})
+
 PLANNING_STORE_RESOURCE = "planning_store"
 """Resource key under which the runner injects the run's :class:`PlanStore`.
 
@@ -61,6 +65,10 @@ run (its own default), which is the honest behaviour for a build with no run to 
 The store the runner injects is seeded from the conversation's `plan_items`, or from
 `PausedRunState.plan` on a resume, and written back to the conversation when the run
 ends - so the plan belongs to the thread rather than to one turn of it (#1077).
+
+A **finished** checklist is not seeded: see :func:`still_open`. The row keeps it;
+a new turn starts without it, because a plan whose every step is ticked is history
+rather than working state (#1221).
 """
 
 WRITE_PLAN_TEXT = ToolText(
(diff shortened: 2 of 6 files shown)


DEENUU1:
Fixed in 998a1648 — and it undid the guarantee the commit before it made, so thank you for catching it. A turn writes its store back when it ends, so the first ordinary turn after a finished plan (opened with an empty store by design) dumped `[]` over the row and deleted exactly what "the row keeps it" claimed.

`PreparedRun.finished_plan_withheld` records that the store was opened empty *over* a finished plan, and `finish` skips the write while the store is still empty: that turn did nothing to the checklist, so there is nothing to record. An agent that starts new work dumps a non-empty plan and replaces the row as before — that is the second test.

`docs/reference/capabilities.md` now says the pair: keeping the row takes the write rule as well as the seed rule.

@codex review

DEENUU1:
Correct as an observation, and I do not think it is worth a mechanism — here is the reasoning, and what would change my mind.

The case is: a turn opens over a withheld finished plan, writes a new plan, then empties it (`write_plan([])`). `finished_plan_withheld` is still true and the dump is empty, so `keep_plan` is skipped and the row keeps the *old* finished checklist rather than recording that this turn's plan was emptied.

**Nothing observes the difference.** `plan_items` has exactly one reader — the seed — and the seed withholds a finished plan. So both outcomes produce the same next turn (an empty store), and the row differs only in holding a finished list nobody will ever be seeded with instead of holding nothing. The persistence contract the comment states exists so a *seeded* plan the agent emptied does not come back; here the seed was empty, so nothing comes back either way.

**Tracking it exactly needs a wrapper.** `write_plan` goes through `store.set_items`, which the library documents as event-silent (`_toolset.py:404`), so the `PlanEventEmitter` hook cannot see it — dirtiness would mean a delegating `PlanStore` around `InMemoryPlanStore` whose only job is to set a boolean. Comparing values cannot distinguish it either: seeded-empty and mutated-to-empty are both `[]`.

So: about 25 lines of new machinery around a library store, to make an unread row exact. I would rather not, and the docstring now says which of the two facts it tracks.

If you want it exact anyway — say a future reader of `plan_items` (an audit view, #788's memory capability) would make the stale row visible — the wrapper is the way and I will add it.

@codex review

PR #1289, opened 2026-08-27
fix(api): name the organization in every alert link, and adopt it

### LINKED ISSUE #1204: Alert emails name no organization, so they open whichever one the reader last used

DEENUU1:
Every notification link was organization-agnostic, and the page it opens acts on whichever organization the reader last used — `apiClient` stamps `X-Organization-Id` from a selection persisted per browser. So somebody in two organizations who was last working in Globex opened the approval alert for a run in Acme and read **Globex's** queue: very likely empty, and reading as *nothing is waiting* about a run that is parked and ageing towards `ApprovalService.expire_stale`. The agent links were wrong more quietly — `/agents/{id}` under the wrong organization is a refusal for an agent the reader can genuinely see, one switch away.

`run.organization_id` and `agent.organization_id` were in scope at all four call sites and discarded.

## One parameter, decided once

```python
def _link(self, path: str, organization_id: UUID) -> str:
    return f"{self._frontend}{path}?org={organization_id}"
```

Four call sites inventing a parameter name is the same defect one level up, so there is one helper and a test that pins it.

| Email | Link | Organization |
|---|---|---|
| Approval requested | `/agents/{id}?org=…` | the **run's** |
| Budget exceeded | `/agents/{id}?org=…` | the run's |
| Usage report (org) | `/agents?org=…` | the one it reports on |
| Usage report (agent) | `/agents/{id}?org=…` | the agent's |

## The console adopts it, the way it adopts `/orgs/{id}`

`organizationInQuery` reads it under the same two rules as the path's reader, for the same reasons #1032 gives: a UUID only (so a future `?org=new` is not adopted as a tenant id and refused on every request) and lower-cased (the value is stored and found by `===` against ids the server serialises in canonical lower case). The adoption is the existing layout effect in the recovery hook — before the tenant cache reset and before the page's own queries, so the first request the page makes already carries the right tenant.

Two rules follow from what the parameter *is*:

- **The path outranks it.** `/orgs/{id}` *is* that organization; `?org=` only says which one an alert was about, so it cannot move somebody off the page they are standing on.
- **Adoption is keyed on the path and the adopted id together.** `?org=` does not change the path, so two alerts about two organizations arrive at the same page — keyed on the path alone, the second would read the first one's tenant.

A reader who has since left that organization is told the link is the reason, rather than being moved in silence and reading another organization's page as the answer to the alert. It cannot name the organization: they are not a member, so it is not in their list.

## Verified

Backend: one case per email kind asserting the id is in the URL, plus one that the parameter is built in a single place. Client: three on `organizationInQuery` (present, absent, UUID-only and lower-cased) and four on the hook — adoption from a link, the path winning over one, a second link to the same page, and the refusal naming the link.

`make test` at 100% on the platform layer, `make lint`, `make test-frontend-cov` and `make docs-build` green. `docs/governance.md` gains "Every link says which organization it is about" under Alerts.

One note: `/runs?tab=approvals` is #935's destination and is not on `main` yet, so the approval link here is still `/agents/{id}`. It gains the parameter either way — `_link` takes the path.

Closes #1204

### DIFF
--- a/backend/app/services/notifications.py
+++ b/backend/app/services/notifications.py
@@ -112,7 +112,7 @@ class NotificationService:
             "org_name": organization.name if organization else "your organization",
             "reason": reason,
             "spent": f"{run.cost_usd:.2f}" if run.cost_usd is not None else "0.00",
-            "run_url": f"{self._frontend}/agents/{agent.id}",
+            "run_url": self._link(f"/agents/{agent.id}", run.organization_id),
             "app_name": settings.PROJECT_NAME,
         }
         self._send(EmailKey.BUDGET_EXCEEDED, recipients, context)
@@ -141,7 +141,7 @@ class NotificationService:
         context = {
             "agent_name": agent.name,
             "tools": ", ".join(tools) if tools else "a tool call",
-            "approvals_url": f"{self._frontend}/agents/{agent.id}",
+            "approvals_url": self._link(f"/agents/{agent.id}", run.organization_id),
             "app_name": settings.PROJECT_NAME,
         }
         self._send(EmailKey.APPROVAL_REQUESTED, recipients, context)
@@ -184,7 +184,7 @@ class NotificationService:
             "total": f"{total:.2f}",
             "runs": str(sum(row[3] for row in rows)),
             "agents": str(len({row[0] for row in rows})),
-            "dashboard_url": f"{self._frontend}/agents",
+            "dashboard_url": self._link("/agents", organization_id),
             "app_name": settings.PROJECT_NAME,
         }
         self._send(EmailKey.USAGE_REPORT, recipients, context)
@@ -249,7 +249,7 @@ class NotificationService:
             "total": f"{sum((row[2] for row in mine), Decimal(0)):.2f}",
             "runs": str(sum(row[3] for row in mine)),
             "agents": agent.name,
-            "dashboard_url": f"{self._frontend}/agents/{agent.id}",
+            "dashboard_url": self._link(f"/agents/{agent.id}", agent.organization_id),
             "app_name": settings.PROJECT_NAME,
         }
         self._send(EmailKey.USAGE_REPORT, recipients, context)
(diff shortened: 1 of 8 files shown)


DEENUU1:
Fixed in 049cd501. Right — the dashboard layout outlives every navigation inside it, so the marker that keeps the adoption from fighting the organization switcher outlived the link too.

A URL that names no organization now clears the marker. The invariant it protects is untouched: on `/orgs/{id}` and on an alert URL `adopted` stays non-null, so a switch made without navigating is still left standing — which is what #1032's own test pins.

Pinned by the three-step sequence you described (open the alert, switch, navigate away, return), which fails without the reset.

@codex review

Issue #930, opened 2026-08-19
Sharing a conversation outside the organization is accepted and unreadable


DEENUU1:
`POST /conversations/{id}/share` resolves the target user **deployment-wide** and
never checks that they are in the conversation's organization. The row is created,
the dialog lists them as having access, and they can never open it: the read path
refuses on the tenant before it ever looks at the share.

## The two halves

**Share accepts anyone on the deployment.**
`ConversationShareService.share_conversation`
(`backend/app/services/conversation_share.py:20`) resolves by
`user_repo.get_by_email(...)` or `user_repo.get_by_id(...)`, checks that the user
exists, that it is not the sharer, and that the share does not already exist. It
never compares the target's membership against `conv.organization_id`. Any address
belonging to any account on the deployment is accepted.

**Read refuses on the tenant first.** `ConversationService.get_conversation`
(`app/services/conversation.py:165`) compares
`conversation.organization_id` against the caller's `organization_id` and raises
`NotFoundError` before `_may_read` is consulted. The caller's organization comes
from their membership, so somebody outside the tenant cannot produce it - the
share row `_may_read` would honour is unreachable.

**Not a leak - a lie.** The tenant gate holds, which is why this is `medium` and
not `high`. What breaks is the promise the dialog makes: the owner sees
`bob@other-company.com` in "Shared with", believes the conversation was shared,
and Bob gets a 404 with no explanation on either side. A refusal that belongs at
share time happens invisibly at read time.

## What it should do

Refuse at share time, with the reason: the target must be a member of the
conversation's organization. `AuthorizationError`, or a `NotFoundError` matching
how the rest of this service reports a target it will not name - whichever, it must
not be a silent success.

Worth deciding in the same change: **existing rows.** Any share already written to
somebody outside the tenant is dead weight that the dialog still lists. Either the
listing filters them out or a migration removes them; the first is cheaper and
honest, the second is tidier. Say which in the PR.

## Done when

- Sharing with a user who is not a member of the conversation's organization is
  refused, by id and by email alike, with a message naming why.
- A test covers exactly that - the shape `.claude/rules/testing.md` calls covering
  the refusal, and one that would fail today.
- Rows already in that state no longer appear as access somebody has.

## Related

- Found reading the share path while scoping the Share dialog redesign (the issue
  above this one). The picker there fixes the client half by construction - you can
  only pick members - but the API is the boundary and has to refuse it on its own.
- **#641** and **#701** are the neighbouring conversation-access decisions
[...]

PR #1291, opened 2026-08-27
feat(chat): pick who to share a conversation with, and read the list as people

### LINKED ISSUE #931: The Share conversation dialog: a member picker instead of an email box, room to breathe, and icons on view/edit
The Share conversation dialog is a 448px box with three controls crammed into one
row, an email text field where the product already has a member picker, and two
permission levels rendered as bare words. It is the oldest surface in the chat and
it shows.

## What is wrong, in order

**1. It asks for an email.** `<Input type="email" placeholder={memberEmail}>`
(`share-dialog.tsx:148`) with a hand-rolled suggestion list below it
(`matchingMembers`, `:42`) that only appears once you have typed something -
`if (!needle) return []`. So the default state of the control is a blank box and
you have to already know the address.

The product has the answer: **`MemberPicker`**
(`src/components/orgs/member-picker.tsx`) - a `Popover` plus a `cmdk` list, open it
and the members are there, type to narrow, each row a face and a name over the
address (`MemberIdentity`). Its own docstring says why it is not a
`DropdownMenu`: Radix's menu typeahead fights a search field, so typing moves
focus instead of filtering. The share dialog's hand-rolled list is a worse version
of a component two directories away, and `matchingMembers` can go with it.

**And the API already takes what the picker holds.** `share_conversation` accepts
`shared_with` (a user id) as well as `shared_with_email`, with a comment saying the
id path is kept "for callers that hold one"
(`app/services/conversation_share.py:56`). The picker holds one. So this is a
client change, not a contract change - and it removes the whole class of "No user
with that email" 404s that the free-text field invites.

**2. It is too small for what it holds.** `sm:max-w-md` = 448px, and inside it: an
email field, a `w-24` permission select and an icon button on one row; a
"generate link" row; the link itself; then the list of who already has access,
each row with a badge and a revoke button. At 448px the person's address in that
list truncates before the badge. `sm:max-w-lg` or `xl`, and the invite row split
from the access list with a separator - the shape `SharingPanel`
(`src/components/sharing/sharing-panel.tsx`) uses for agents, skills and
collections.

**3. View and Edit are bare words.** Both in the select and in the badges on the
access list. They want icons - `Eye` and `Pencil` - and a sentence saying what
each means, because "edit" on a conversation is not obvious: it carries renaming,
archiving, deleting the thread and appending turns
(`ConversationService._may_write`). Somebody granting it should be able to tell.

**4. The badge renders the backend's word.**
`<Badge variant="secondary">{share.permission}</Badge>` (`:246`) puts the raw
`"view"` / `"edit"` from the API on screen, so the access list is English in every
locale while the select one row above it is translated. The guard does not catch it
because it is an expression, not a literal.

## What it should be

- `MemberPicker` for who, sending `shared_with` - no email field, no
  `matchingMembers`.
- A wider dialog, invite row separated from the access list.
- `Eye` / `Pencil` on both levels, in the picker and on each row, with one line
  saying what edit permits.
- The access list rendered as people (`MemberIdentity`), with the avatar the rest
  of the product draws, and the level translated.
- The link half left where it is: a share token is a different thing from a person
  and should not sit in the same row.

## Done when

The dialog shares with a picked member and never with a typed address; the access
list reads as people with an icon per level; nothing on it renders an untranslated
server value; and `matchingMembers` is deleted along with its test rather than left
behind unused (`make lint` runs `knip` and `vulture` for exactly that).

## Related

- **#930** - sharing with somebody outside the organization is accepted by the API
  and unreadable by them, found reading this path. The picker fixes the client half
  by construction; the server refusal is its own bug.
- **#68** (Organizations redesign) and **#67** (Chat redesign) are the wider
  passes; this is one dialog and does not need either.

DEENUU1:
The Share conversation dialog asked for an **email address**, with a hand-rolled suggestion list that appeared only once something had been typed (`if (!needle) return []`). So the default state of the control was a blank box you had to already know the answer to fill, and every mistyped address was a 404. The product already had the answer two directories away.

## Who is picked, never typed

`MemberPicker` — a popover over a `cmdk` list; open it and the organization is there, each row a face and a name over the address. `matchingMembers` and its four tests are deleted with the field.

The API takes what the picker holds: `shared_with` has always been accepted beside `shared_with_email`, so **this is a client change, not a contract change** — and sharing outside the organization becomes impossible by construction, which is #930's client half.

Somebody who already has access is not offered again: a row that answers "already shared" after the click is worse than one that never appears.

## View and Edit carry icons and a sentence

`Eye` and `Pencil`, in the select and on every row of the access list, with one line saying what the level permits — because "edit" on a conversation is not obvious: it is rename, archive, delete and append turns, which `ConversationService._may_write` decides and nothing on this dialog used to say.

## The access list reads as people

`MemberIdentity`, the same row the members table and the alerts picker draw, resolved against the organization's members — with a fallback to whatever the share itself holds, because a share whose member is gone still has to be revocable.

The level is now the catalog's word. The badge printed the API's raw `"view"` / `"edit"`, so that one row was English in every locale while the select above it was translated; the i18n guard could not see it because it is an expression rather than a literal.

## Room to breathe

`DIALOG_FORM` (672px) instead of `DIALOG_CONFIRM` (512px), and three sections separated rather than three controls on one row: invite, who has access, and the link. A share token is not a person — no face, no level to read against a name — so it keeps its own row, which is the shape `SharingPanel` uses.

## Verified

28 cases in `share-dialog.test.tsx`, rewritten around the picker:

- there is **no text field at all** on the dialog;
- the picker offers the organization before anything is typed, and hides somebody who already has access;
- the share carries `shared_with: "u2"`, not an address;
- the Share button is disabled until somebody is picked — the disabled state *is* the guard, so the handler holds no branch nothing can reach;
- the sentence changes with the level;
- the list renders a member as a person (name *and* address), a link as a link, an unresolvable share by id, and the level as `Edit` rather than `edit`;
- `toPermission` refuses a level the catalog does not know rather than defaulting, and `sharedPerson` is covered in all four of its shapes.

`make lint`, `make test-frontend-cov` (100% statements/lines/functions) and `make build-frontend` green. `memberEmail` and `memberSuggestions` are removed from both catalogs — an unread key fails the build.

Closes #931


DEENUU1:
All five fixed in b73172cd, and the P1 about `_may_write` is the most valuable thing this review found — thank you.

**View access was not enforced (P1).** Right, and it is pre-existing: `_may_write` accepted any share whatever level it held, so the two levels this dialog offers meant the same thing — a conversation shared to *view* could be renamed, archived, deleted, or given a `role: "assistant"` turn the model is handed back as its own words. Found because this branch writes a sentence promising otherwise. A share now carries the write only at `edit`; two service tests pin it, and `docs/architecture.md`'s IDOR section states the rule.

**Member 51 was unreachable (P1).** Also right. `useMembers` asked for no page, so the route's default of fifty was every caller's whole list — a members table quietly missing rows, and now a colleague who could not be shared with at all. It pages through, stopping on a **short page** rather than on `total`: a loop that trusts a count asks forever if the count is ever wrong.

**The reader was offered themselves (P2).** Filtered out — every option the picker offers can now succeed.

**A nameless member drew as the local part (P2).** Right, and `MemberIdentity` omits the address beneath a nameless account precisely because the line above it is supposed to *be* the address. `displayName` keeps the whole address; three suites that asserted the shortened form now assert what is drawn, on the dashboard's people card and the alerts picker as well as here.

**A pick the refetch took away (P2).** The button is enabled on the chosen *member* rather than on the id, so there is nothing left to submit once they leave the list.

@codex review

DEENUU1:
Both fixed in 7d33cddf.

**The page order.** Right, and it is the same defect the sessions listing carries its own tie-breaker for: `list_for_org` ordered by `joined_at` alone, two people invited in one request share it, and a page boundary inside a tie shows one member twice and another never. Reading every page to fill the picker turns that into a colleague who cannot be shared with. `id` breaks it, with an integration test that twenty-five members sharing one `joined_at` come back once each across five pages, and that one page asked twice is one page.

**The silent failure.** Also right, and worse than before the change: the email field would at least have taken a typed address, where an empty disabled picker explains nothing. The dialog reads the hook's error now and says so, with a retry.

@codex review

DEENUU1:
Fixed in 65efaf03. Right, and it is the new copy that made it a lie: the level select drove both halves, so choosing Edit and then minting a link promised renaming, archiving and appending that a token cannot do — there is exactly one token route, `GET /conversations/shared/{token}`.

Link generation now sends `permission: "view"` whatever the select says, and the link section says so *before* one is minted rather than after. The test that asserted the old behaviour is inverted, with the reason in it.

@codex review

Issue #1039, opened 2026-08-20
Sandboxes, runtimes and workspaces: an attachment that never reaches the workspace, a CSP that blocks every PDF, and seven things nothing explains


DEENUU1:
Nine things about the Docker (`sandboxd`) backend and the surfaces around it,
asked in one sitting. `daytona` and `state` are out of scope except where a fix
touches both.

The full analysis — what was checked, how, and what each answer rests on — is in
`docs/plans/sandbox-runtimes-and-workspaces.md` on the branch. This issue is the
list.

## The one that is a bug

**An attachment never reaches the workspace.** `UPLOAD_DIR = "/uploads"` is an
absolute path, and the sandbox resolves it as one. Probed against the running
service:

```
write /uploads/probe.txt   -> /uploads/probe.txt          (nowhere on the host)
write uploads/probe2.txt   -> /workspace/uploads/probe2.txt
write probe3.txt           -> /workspace/probe3.txt
```

So a file an agent is given is written outside the bind-mounted work directory:
`ls` does not see it, `/workspaces` cannot list it, and it dies with the
container. Every workspace on this host is empty on disk. The agent that
summarised a PDF and then said `Directory '.' is empty` was reading the head
sample in its prompt, not the file.

- [ ] Uploads go inside the work directory (`uploads/`, relative)
- [ ] Every reader of that path updated: the model's reference, the browser's
      addressing, the extracted-text sibling, the tests asserting `/uploads/`
- [ ] A test that an attachment lands where a shell's `ls` finds it
- [ ] The capability's prompt says where attachments are

## The one that is a viewer

**No `frame-src`, so `default-src 'self'` blocks the blob iframe** — which is
every PDF and HTML preview. `img-src` already allows `blob:`, which is why images
work.

- [ ] `frame-src 'self' blob:` in the CSP
- [ ] A test on the header
- [ ] The other iframe viewers checked against it

## Runtimes: two lists, and where they come from

The connection dialog offers fifteen (the library's `BUILTIN_RUNTIMES` recipes);
the agent Builder offers three (what *this service allows* — `DEFAULT_RUNTIMES`,
because the deployment names none). Both are correct; the deployment is the
problem, because twelve of the fifteen cannot run here and nothing says so until
`Test` is pressed.

Runtimes are defined on the service, via `SANDBOXD_RUNTIMES=alias=image` /
`alias=@recipe;mem_limit=4g`. A UI for it is refused upstream on purpose —
[...]

Issue #131, opened 2026-08-03
Workspaces: a user-scoped workspace from Slack or Telegram is invisible to the linked account


DEENUU1:
Found while building the per-reader workspace listing (#36, item 1).

`agent_workspaces.owner_ref` is a **string, not a foreign key** — deliberately, because not every surface's caller is a row in `users`: a Slack member id and a Telegram number are both legitimate answers. The listing added in `list_for_reader` matches `owner_ref == str(user_id)`, which is the web account's own id.

**What that means.** Somebody who talks to a `user`-scoped agent in Slack gets a workspace keyed to their Slack member id. When the same person signs into the web app — even with a `ChannelIdentity` linking the two — the Workspaces page does not list those files. They are not lost and not leaked; they are simply unreachable from the surface where a person would look for them, and the chat panel only shows them from inside a conversation with that agent.

**Where it is.** `backend/app/repositories/agent_workspace.py:list_for_reader`, the `owner_ref` predicate. `channel` scope has the same shape and is deliberately operator-only for the same reason — its people are identified by the platform, not by a row in `users`.

**Two ways out** (the second is probably right):
1. Resolve the caller's linked identities and match `owner_ref` against all of them. No migration; one extra query, and it makes the predicate depend on `channel_identities` staying correct.
2. Add a nullable `owner_user_id` beside `owner_ref` for the workspaces whose owner *is* an account, set where the identity is already known. A migration, but then "mine" is a join rather than a string comparison, and the string stays for the identities that are not accounts.

**How you would know it was fixed.** A `user`-scoped workspace created by a Slack run appears on Workspaces for the linked account, and does not appear for anybody else — including another member of the same Slack workspace.

Pre-existing shape, surfaced by #36 rather than caused by it: before that branch the listing was operator-only, so the question of "mine" never arose.

DEENUU1:
This is a channel issue as much as a workspaces one, and had no link to that cluster.

**#10** is the one that matters: option 1 above ("resolve the caller's linked identities") depends on `ChannelIdentity` rows that actually carry a `user_id` — and `link_code` is minted by no code in the repo, so today there is no linked identity to resolve. Option 1 is unbuildable until #10 lands; option 2 may be right anyway, but the choice should be made knowing that.

Also **#41** / **#42** / **#43** (a Slack- or Telegram-sourced `owner_ref` is whatever those settle for identity) and **#137**, which already cites this issue from the UI side — that column has to render an unlinked label for exactly the case described here.

[...]

PR #1292, opened 2026-08-27
feat(sandbox): name the workspace's owner, and share one folder tree

### LINKED ISSUE #137: Workspaces: an owner column, a link to the conversation, and Open into a skills-style browser
`/workspaces` answers "what are the agents holding" with a table, and `/workspaces/{id}` answers "what is in this one" with a folder browser. Both are close, and both stop one step short of the thing somebody came to do.

## The list is missing the column that says whose it is

`components/sandboxes/workspace-browser.tsx:130-141` has: Agent · Conversation · Who can see it · Backend · Size · Last used · Files.

**There is no owner.** "Who can see it" is `access_label`, which describes the *scope* — "everybody who talks to this agent", "one person". It does not name the person. On an `agent`-scoped workspace shared by six people that is exactly the question an operator has, and the row cannot answer it.

The data is there: `agent_workspaces.owner_ref` plus `owner_label`, which the chat panel already renders (`chat/workspace-files.tsx:122`). Add **Owner** beside **Who can see it** — the first says who, the second says who else. Note #131: `owner_ref` is a string and a Slack-sourced workspace's owner is a platform id rather than an account, so the column has to render an unlinked label for those rather than assume a user row.

## The conversation cell is a title with no way to reach the conversation

`workspace-browser.tsx:151` renders `workspace.conversation_title`, as plain text. The thing anybody wants after reading a conversation title is to open that conversation, and there is nothing to click. Make it a link to `/chat?conversation={id}` when there is exactly one, and leave `N conversations` as it is when there is not — that is a real distinction and it should stay legible.

`WorkspaceRead` will need `conversation_id` alongside `conversation_title`; the repository already resolves the title, so it is one more column on the same query.

## Open lands on a folder browser, and it should land on what /skills has

`Open` goes to `/workspaces/{id}` → `WorkspaceExplorer` (259 lines): breadcrumbs, a search box, and one level of a directory at a time, with a click opening a modal over the top.

`/skills` does the better thing for the same shape of data — `components/skills/skill-files.tsx`: the whole tree on the left, collapsible, with the file open in a pane beside it. You can see where you are and what else is there while you read, and moving between two files is one click rather than close-navigate-open.

Rebuild the explorer on that layout. `FileTree` and `TreeLevel` in `skill-files.tsx` are already generic over a `TreeNode` from `lib/file-tree.ts`, and `levelAt` in `workspace-explorer.tsx` derives its structure from flat paths the same way — so this is mostly extracting the tree from `skills/` and building the workspace's `TreeNode[]` for it, not writing a tree.

Keep the search box: `WorkspaceExplorer` has one and the skills tree does not, and a workspace can hold considerably more files than a skill.

## Watch out

- Once the reading pane is inline, `WorkspaceFileViewer` is still the right thing for the *chat* panel and the flat file list, where there is no room for a tree. See the file-viewer standardization issue — the pane and the dialog should be the same renderer in two frames, not two renderers.
- `/workspaces` is deliberately **not** gated on `connections:manage` (`workspaces/page.tsx` says why). Adding an Owner column must not turn it into an operator screen: the backend already decides which rows a reader sees, and the column shows what the row already carries.
- Sorting and filtering on this table is the cross-cutting issue, not this one. Add the columns here; the header controls arrive with the shared table.

## Done when

- [ ] An **Owner** column, rendering a platform-sourced owner as a plain label rather than a broken link
- [ ] A single-conversation workspace links to that conversation in `/chat`; a shared one still reads `N conversations`
- [ ] `/workspaces/{id}` is a persistent tree beside a reading pane, with the search box kept
- [ ] The tree is one component, shared with `/skills`, not a second copy
- [ ] An integration test that the conversation link carries the right id, and that a workspace with no conversation renders no link

DEENUU1:
Two of #137's four bullets were already true on `main` — the redesign in #1039 rebuilt the explorer as a persistent tree beside a reading pane (search kept) and made the conversation reachable from the table. This finishes the other two.

## An Owner column

The table said who *else* could see a workspace and never who it belongs to. `access_label` describes the **scope** — "everybody who talks to this agent", "one person" — which is a different fact, and on an agent-scoped workspace shared by six people it is not the one an operator is asking.

`owner_label` was already on the row and already rendered by the chat panel; here it was only a fallback heading, used when a workspace had no conversation to be named after. It is a column now, beside Who can see it: the first says who, the second says who else.

Plain text and never a link — `owner_ref` is a string and a Slack-sourced workspace's owner is a platform id rather than an account (#131), so a linked cell would be broken on half the rows, and there is no user page to link to in any case. Sortable, because grouping a deployment by holder is what somebody opens this to do.

## One folder tree, not two

`/skills` and `/workspaces/{id}` had written the same tree twice: the same recursion, the same expand-collapse set keyed on a folder's path, the same chevron and two folder icons, the same `role="tree"` with `aria-expanded` — over two node shapes and **two polarities of open state**, one holding what was collapsed and the other what was open.

`PathTree` in `components/files` is now the mechanics and the semantics: indentation by depth, the roles, one selected file, the open set. What a row *says* stays with the caller, because a skill's file is a name and a workspace's is a name, a size and a download.

Two render props rather than one:

| Prop | Where it renders | Why |
|---|---|---|
| `renderFile` | inside the button that opens the file | the icon and the name — which is then all a screen reader announces |
| `renderFileMeta` | beside that button | a control belongs outside it: the workspace's download must not need the file opened first, and a button inside a button is invalid |

The empty-tree guard is the shared component's now, singular. A surface that wants to say something instead — the explorer distinguishes an empty folder from a host that would not answer — checks before it gets there.

**One deliberate visual change:** a skill's *file* rows were indented twelve pixels further than its folders and the workspace's were not, so the two trees disagreed about the same question. They now indent alike.

Net: `workspace-explorer.tsx` −110 lines, `skill-files.tsx` −80, and one 203-line component holding what both had.

## Verified

Both suites are unchanged in intent and green — 36 explorer cases and 31 skills-tree cases, including the download reached without opening a file (the property the split render props exist for) and a stale `openId` selecting nothing. Two new cases on the table: the owner named beside the access label, a platform-sourced owner drawn as words rather than a destination, and the owner sort.

`make lint-frontend`, `make test-frontend-cov` (100% statements/lines/functions) and `make build-frontend` green.

## Not in this PR

The issue's remaining "done when" — an integration test that the conversation link carries the right id — is already covered on `main` by `workspace-browser.test.tsx`, which asserts the `?id=` link is drawn only for the reader's own conversation and that a shared workspace reads `N conversations` instead.

Closes #137


DEENUU1:
All three fixed in c564f9be, and the third one is the finding that mattered — the Owner column named nobody.

**`owner_label` is the scope, not a person.** You are right and I should have read the helper rather than the issue's summary of it: it answers "This conversation", "Your files for this agent", "Shared by everyone who uses this agent", so the column repeated the access column and my tests passed only because they injected values the API cannot produce.

`WorkspaceOverview.owner_name` resolves the row's `owner_ref` — the only field that records an owner, and only for `user` scope. A UUID becomes the account's email through one grouped `get_emails_for_users`; a platform id is the answer verbatim (#131); an id nobody in the organization holds resolves to nothing, because naming a stranger's email off an id is the one read this listing must not do. Three of the four scopes have no owner, so the cell is an em dash and the ordinary listing pays no query at all. Four service cases pin those.

**The open state.** Also right on both halves. Held inside `PathTree` it was mount-local and frozen after the first toggle: the explorer swaps the tree for a flat list while searching, so clearing the box reset every fold; and a skill needs openness *derived* ("every folder except the collapsed ones") or a folder uploaded after a collapse stays shut. `openPaths` and `onToggleFolder` are props now — the explorer keeps its set above the search, the skill tree keeps what is collapsed — and the component holds no state. Both behaviours have a test.

**The duplicate key.** Right: `a` and `a/b.md` make sibling nodes whose path is both. The key carries the kind, with a test that the two draw as two rows.

@codex review

Issue #417, opened 2026-08-06
Ingestion and sync flows are spawned before the transaction that wrote the row they read


DEENUU1:
**Severity: medium** · pre-existing, found implementing #353

## What breaks

Three services start a background flow **inside the endpoint**, before the
request's transaction has committed, and the flow's first act is to read by id
the row the request has just written on a session nobody else can see:

| Spawn | Reads |
|---|---|
| `backend/app/services/rag_document.py:245` | `ingest_document_flow(rag_document_id=...)` |
| `backend/app/services/rag_sync.py:105` | the sync log by `sync_log_id` |
| `backend/app/services/sync_source.py:287` | the sync log by `sync_log.id` |

The flow opens a session of its own — correctly, it is a different unit of work
— so under `READ COMMITTED` it cannot see the uncommitted row. If it gets there
first it finds nothing and the work is lost: an upload that answers
`{"status": "processing"}` and stays that way, which is the shape
`app/core/background.py:66` already describes as "a document stuck in
`processing` forever".

Whether it gets there first is a scheduling question. `spawn` is
`asyncio.create_task`, so the flow runs at the first suspension point after it —
and the endpoint has several left before its commit.

## What #353 does and does not do about it

Nothing. #353 moves the commit earlier relative to the *response*; the spawn is
earlier still, so the window between spawn and commit is the same width before
and after. This is filed separately because it is a separate mechanism — the
loser there is another process, not the client — and because the fix is a
different change.

## The fix, and why it is not in #353's branch

The work has to be handed over after the commit, not before it. FastAPI's
`BackgroundTasks` runs after the response, which is now after the commit, and is
the obvious mechanism — but it means threading a `BackgroundTasks` through
`RAGDocumentService`, `RAGSyncService` and `SyncSourceService` and their
callers, which is a reviewable change of its own and not one to bury in a branch
about response ordering.

Worth deciding at the same time: `spawn` currently keeps in-process work alive
on the API worker, and `drain` waits for it at shutdown. Whether these three
should be Prefect deployments rather than in-process tasks is the same question
one layer up.

## How you would know it was fixed

An integration test that uploads a document and asserts the flow reads a
committed row — or, more cheaply, that the flow is not created until the request
has committed. Today the assertion has to be about ordering, because the race is
narrow enough that sampling proves nothing.

## Where this sits

**Runtime, lifecycle, background work** in #168 — the cluster with #2, #11, #12,
[...]

PR #1296, opened 2026-08-27
docs: a readability pass over the whole site, and the four pages that were wrong

### LINKED ISSUE #784: Docs: a readability and visual pass, modelled on pydantic-ai / FastAPI
## What

A readability and visual pass over the docs site. The **content is already good**; it is
hard to read and unpleasant to look at. Bring it up to the bar of the docs it should
stand next to — **pydantic-ai** and **FastAPI**: clear information hierarchy, generous
whitespace, admonitions/callouts, diagrams where prose is doing a diagram's job, tabbed
code samples, consistent page structure, and a landing page that orients a first reader.

## Why

Docs are how a self-hosted, open-source platform gets adopted without us in the room.
Good content that reads as a wall of text loses the reader before the value lands. This
is a presentation pass, not a rewrite — the words mostly stay, the shape and styling
change.

## Nearest issue

**Documentation** cluster of #168. Distinct from the *content-correctness* issues there
(#29, #664, #212 fix claims that are wrong); this one is purely readability and design
of pages whose content is already correct.

## Done when

- MkDocs Material features are used where they earn it (admonitions, content tabs,
  code annotations, a proper landing page).
- Concept pages follow one consistent structure and altitude.
- A side-by-side with a pydantic-ai / FastAPI page no longer looks a tier below.
- `make docs-build` still passes `--strict`.

DEENUU1:
## What changed

A presentation and correctness pass over **every page of the site** — 25 concept
and reference pages, the 8 Guides, and the 3 reference stubs.

The site was accurate and almost unreadable: 27 pages of unbroken prose, **one**
mermaid diagram between them, **no** content tabs anywhere, ~20 pages with no
callouts at all, and three flows drawn in ASCII that only line up in a monospace
font. The words are mostly unchanged. What changes is what a reader sees before
they start reading.

### Presentation

- **17 mermaid diagrams** where prose or ASCII described a flow: the request path
  and the transaction's ordering, both ingestion pipelines, the three permission
  layers and `resolve_access` as a decision tree, park → decide → resume, the
  sandbox's three processes, a webhook delivery's refusals, the embed handshake,
  choosing a surface, envelope encryption, progressive disclosure, MCP's OAuth
  2.1 handshake, `ai-review`'s three jobs, a sync's six stages, the local dev
  stack.
- **~130 callouts**, each promoting a rule the page already stated in bold prose
  — the ones whose violation costs something: a 2xx means the write is readable,
  a budget is checked before the request, an empty origin list allows nothing,
  the sandbox token is root-equivalent, a signature is the only thing between a
  stranger and your budget, losing every vault key is unrecoverable.
- **Content tabs** where alternatives were stacked vertically: the four platform
  quickstarts in `deploy.md`, curl vs Python in `triggers.md`.
- Prose restructured where it was a table or a list in disguise: the five run
  statuses (`concepts.md`), the production checklist (`configuration.md`), the
  rollback options (`deploy.md`), `first-agent.md`'s six-paragraph "guided way".
- Nine `---` rules that separated nothing, and `deploy.md`'s generator flag list.

### Correctness — what the pass turned up

Four pages were teaching things that are not true here:

- **`patterns.md`** — all three worked examples had drifted off the code. The DI
  example injected `Depends(get_db)` and `Depends(get_current_user)` (neither
  exists; the aliases do, and `DBSession`'s `scope="function"` is load-bearing,
  #353); the repository example was a `ConversationRepository` class, which is
  the one shape `.claude/rules/architecture.md` rules out; the service example
  held `self.repo`, which no service in the codebase does. Replaced with real
  signatures.
- **`howto/customize-agent-prompt.md`** — taught editing `app/agents/prompts.py`
  and overriding `DEFAULT_SYSTEM_PROMPT`, with `get_system_prompt_with_rag()` to
  customise and `AI_TEMPERATURE` in `.env` to tune. None of those exist, and the
  page contradicted the sentence CLAUDE.md calls the whole design. Rewritten
  around the spec and `default_instructions.py`.
- **`howto/add-background-task.md`** — step 2 was
  `asyncio.create_task(send_notification_flow(...))`, which is the shape #417
  was: the task starts before the request commits, so a flow reading its own row
  finds nothing — and it drops the exception too. Now `spawn_after_commit` /
  `spawn`.
- **`howto/add-api-endpoint.md`** was a second, already-diverging copy of
  `adding_features.md`'s walkthrough, and both taught the `self.repo` shape. It
  is now the single copy (`BaseSchema`, `TimestampMixin`, `__repr__`, `-> Any`,
  the `require`/`resolve_access` split); `adding_features.md` points at it.

Plus:

- `configure-sync-sources.md` named `app/rag/connectors/` twice (the package is
  `app/services/rag/connectors/`) and had a dangling "Scheduled syncs require a
  running background task system:" with nothing after the colon.
- `ROADMAP.md` was dated 2026-07-27 and contradicted itself — R12 said the 100%
  gate was not enforced while the Testing section said CI enforces it, and R8/R9
  described features that have shipped. Re-verified each (`fail_under = 100`,
  `org_mcp_connections.py` gated on `connections:manage`, `/e/{publicKey}`), all
  three moved to "Closed since that revision". R10 and R11 re-checked and still
  open.
- `index.md` promised "four nouns" where `concepts.md` has five.
- A dead anchor in `configuration.md` — mkdocs reports those at INFO, so
  `--strict` never caught it.

### The self-inflicted one

Promoting a bold sentence into a callout and leaving the sentence below it makes
a page say everything twice. A detector over every `!!!` block and the four lines
under it found **25** such repeats in my own diff; all are gone — the callout
keeps the claim, the prose keeps only what the callout does not carry, and where
it carried nothing else the prose is deleted.

## How it was verified

`make docs-build` (mkdocs `--strict`) after **every page**, and again at the end:
exit 0, and the one INFO-level dead anchor now fixed. Every symbol, signature,
flag and file path I asserted was checked against `backend/app/` — which is how
the four wrong pages were found in the first place.

Docs-only, so CI runs `lint`, `docs`, `changes` and the security jobs; `test`,
`test-frontend` and `e2e` skip by path scope, which is a pass.

Closes #784

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -4,9 +4,7 @@ The state of the platform and what is left, written so someone picking this up
 cold knows both what exists and *why it was built that way*. Sections match the
 board in the working plan (`VstormOS — MVP`).
 
-Updated: 2026-07-27.
-
----
+Updated: 2026-08-27.
 
 ## Where the platform stands
 
@@ -19,14 +17,12 @@ Updated: 2026-07-27.
 | Agent registry | Done. Draft → validate → publish → frozen version |
 | Runs, budgets, approvals | Done, resume included |
 | Skills and collections | Done, backend and UI |
-| MCP | Catalog, per-user connections, and agent bindings resolved at run time. Org-scoped connection *management* missing (R8) |
-| Surfaces | Web chat, Playground, API, Slack and Telegram `@mention` — one run path, one set of books |
+| MCP | Catalog, per-user **and** org-scoped connections, agent bindings resolved at run time |
+| Surfaces | Web chat, Playground, API, an embeddable widget, a raw WebSocket, a hosted page, Slack, Telegram and Mattermost — one run path, one set of books |
 | Ingestion | Per-collection parser, chunking, OCR and image-description model; overridable per upload. Embedding model fixed at creation |
 | Frontend | Agents, Builder, Skills, Sharing, Providers, MCP catalog, Activity, Roles, ingestion settings |
-| Tests | 1,644 backend, 564 frontend, E2E present. **Coverage gate is red at 97.51%** — see R12 |
+| Tests | Four layers plus Playwright. The platform layer is at **100%, enforced in CI**; the frontend gate is 100% lines/statements/functions and 97.5% branches |
 | CI | Lint, types, tests, migrations forwards and back, frontend, E2E, security |
-
----
 
 ## Architecture, in the order it matters
 
@@ -47,8 +43,6 @@ someone is looking at a form, not at 3am in a customer conversation.
 
 **Every surface goes through one runner.** If each surface assembled its own
 agent, budgets and history would hold whatever each one remembered to record.
-
----
 
 ## Delivered since the last revision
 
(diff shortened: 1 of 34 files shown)


DEENUU1:
All five P2s were right, and four of them were this branch's own — a readability pass shipping inaccuracies is the joke writing itself. Fixed in `c083234f`:

- **patterns.md, the repository example** — the worst of the five, because that page exists to be copied. It took `user_id` and filtered only on the tenant. It narrows on the argument now, with a note that the real module widens the predicate to vetted channel participants through `_reachable_by`.
- **adding_features.md, `db-check`** — you are right on both halves. The target prints a warning and exits 0 when nothing is listening on `CHECK_DB_PORT` (Makefile:563-569), so a model change with no migration *does* pass local `make check`; and CI runs it in the **`test`** job, which has Postgres, not in `lint`. Both stated now, plus `make docker-db` for the local answer.
- **add-api-endpoint.md, `total=len(items)`** — mine, and wrong in the direction that hides rows. The example grows `count_unread` and the service returns both.
- **deployment.md, the recovery command** — `email` is a required `click.argument`, so the documented command exits on Click's error. Now `create-app-admin <email>`.
- **skills.md, "the twenty-first costs nothing either"** — the page's own claim, but my diagram made it louder by putting "nothing paid for" on the discovery branch. `list_skills` returns every attached skill's name and description into the next model request. Corrected in both places, with why the numbers still work: a line per skill against a body per skill.

`make docs-build` green, and the `db-check` and `create_app_admin` claims were read out of the Makefile and the command module rather than from memory.

@codex review

DEENUU1:
All four right again, and all four checked against the source rather than from memory. Fixed in `f10c31e7`:

- **triggers.md** — `trigger_webhooks.py:47-73` submits the flow and answers `202` immediately; my diagram put the run before it, which reads as an acknowledgement that waits for the model. The run is a dashed edge now, plus a callout saying `202` means accepted rather than finished — the thing an integrator sizing a delivery timeout actually needs.
- **governance.md** — you are right that only rejection is replayed. `expire_stale` ends the run `cancelled` and makes no further model request, and its docstring says why: spending an organization's keys on a schedule, for a run nobody is waiting on, is not a thing to do unasked. Two branches now instead of one.
- **add-background-task.md** — the sharpest of the four. The flow took `(user_id, message)` and printed them while the call passed `notification.id`, so the example mislabelled the id *and* demonstrated nothing: a flow that reads none of its own rows works with `spawn` too. It now takes an id and loads it in its own session, which is the shape that makes `spawn_after_commit` necessary rather than advisable.
- **code-review.md** — `context` uploads only `pr-meta/` (title and body); `Assemble the diff` is a step of `review`, after its checkout. Relabelled, and the prose now says explicitly that the head enters inside the key-holding job — because that is the fact somebody editing the workflow has to hold, not a caption detail.

`make docs-build` green.

@codex review

Issue #120, opened 2026-08-03
Mobile View


DEENUU1:
This issue had `describe it later` in every field. Here is the audit, measured
rather than described, so the remaining work has a size.

## How it was measured

Chromium against the running app at **390x780** (iPhone 15) and **768x1024**
(iPad portrait), signed in, fourteen dashboard pages, each scrolled to the end.
For every element inside `main` the probe recorded: does it extend past the
viewport, is an interactive element under the fixed tab bar at maximum scroll, is
a tap target smaller than 40px, is text below 12px.

## The good news first

**The document never scrolls horizontally**, at either width, on any of the
fourteen pages. Whatever else is wrong, the page itself is not sideways.

## Fixed (PR below)

| | Before | After |
|---|---|---|
| Agent / skill / context / MCP / KB card grids at 390px | card **391px** in a 324px grid — clipped by 34px | fits |
| Chat control bar at 390px | ran **27px** past the composer | fits |
| A widget's info button | **14x14** target | 44x44 hit area, same glyph |

The grid one is worth stating because the cause is not where it looks. `grid
gap-3 md:grid-cols-2` declares no column count below `md`, so the implicit track
is `auto` and sizes to content: `grid-template-columns` computed **391.094px**
inside a 324px box. `min-width: 0` at every level of the ancestor chain changed
nothing, because the track is what is too wide. `grid-cols-1` is the fix.

### 1. Data tables are reachable but not usable on a phone

Three pages put a table in an `overflow-x-auto` scroller, which is the correct
pattern and means no column is lost. It also means this, at 390px:

| Page | Table width | Column width | Ratio |
|---|---|---|---|
| Activity | 1155px | 364px | **3.2x** |
| Sandboxes | 914px | 364px | 2.5x |
| Channels | 638px | 364px | 1.8x |

Reading one Activity row means three sideways scrolls. **This is the bulk of the
xl effort**: below `md` these want to be cards — the two or three fields that
matter per row, stacked, with the rest behind a tap. Not a CSS fix.

### 2. Tap targets

Both mobile platforms ask for 44px (Apple) or 48px (Material). At 390px:

| Size | Count | What |
|---|---|---|
| `324x32` | 9 | full-width list rows |
| `28x28` | 9 | `icon-sm` buttons — the repo's default icon size |
| `324x28` | 6 | list rows |
| `211x20` | 6 | inline links |
| `36x36` | 6 | `icon` buttons |
| `93x24` | 4 | chips and badges |
[...]

PR #1297, opened 2026-08-27
fix(runs): give Activity its room without clipping the sticky panel

### LINKED ISSUE #1206: Activity has no room under its last row: /runs is registered as a full-height route and is not one
Activity's last run row sits on the bottom edge of the viewport. `main`'s
`pb-16` is declared and painted on every other list page; on `/runs` it is not,
and the reason is one entry in a regex.

## Mechanism, measured

`PageTransition` puts `min-h-0` on its wrapper for **full-height routes**
(`src/components/layout/page-transition.tsx:8`):

```ts
const FULL_HEIGHT_ROUTES = /\/(?:chat|runs)(?:\/|$)/;
```

`min-h-0` is what makes the wrapper stop growing with its content, so the
content overflows *it* rather than extending it — and `main`'s bottom padding
is then measured from the wrapper's own box, above the end of the content, not
after it. The file's own comment says this ("with it, a long page overflows this
box and parks `main`'s bottom padding mid-content instead of after it"); what it
does not say is that `/runs` is in the list.

Measured in Chromium 151 against this app's stylesheet, on a transcription of
Activity's own markup — a 25-row list beside the sticky run detail panel, 1440 x
800:

| wrapper | room below the last row |
|---|---|
| `min-h-0` (today, `/runs`) | **0px** |
| no `min-h-0` (every other list page) | **64px** |

## Why `/runs` should not be in that list

Because it is not a full-height route any more. The page's own root is
`<div className="flex flex-col">` and its comment says so in as many words:

> An ordinary page that scrolls, not a full-height column with panes that scroll
> apart. That arrangement bought sticky column headers and cost the table its
> rows […] a second scrollbar to reach what the page had room for all along.

The run detail is `sticky` inside the page's own scroll, not a pane with a
scrollbar of its own. `page-transition.test.tsx` still asserts the opposite —
*"constrains Activity too, where the list and the run detail scroll apart"* —
which was true before the page was rebuilt and is not now.

## The consequence, which is why this is not a one-character fix

The panel's height and sticky offset were tuned for a scrollport with **no**
bottom padding: `h-[calc(100dvh-1rem)]` with `sm:top-[calc(0.5rem-2rem)]`, so it
pins 8px from the window top and ends 8px from the bottom. Restore the 64px and
the sticky element's containing block ends 64px above the viewport at maximum
scroll, so the panel's top goes to **-48px** — its own header, the agent, the
status and the cost, cut off by 56px at the end of the list. Measured in the same
run.

So this takes a decision about the two-column row: either the room under the page
belongs to the list column alone (the sticky column keeping the full height), or
the panel's height gives the padding back. Whichever, the panel's header must
still be readable at maximum scroll — that is what
`src/app/[locale]/(dashboard)/runs/page.tsx:231-260` is written to protect.

## Done when

- `/runs` clears the bottom of the viewport by the same room every other list
  page gets, scrolled to the end, at `lg` and below it.
- The run detail panel's header is still fully visible at maximum scroll.
- `/chat` is unchanged: its composer belongs on the bottom edge, which is what
  the constrained branch is for.
- `page-transition.test.tsx`'s Activity case says what is now true.

## Related

Found while verifying **#933**, whose stated mechanism — end padding on a
flex-column scroll container not being painted — **does not reproduce**: measured
at 80px / 64px in Chromium 151, WebKit 26.5 and Firefox 153, both hand-transcribed
and against the app's own CSS. `min-h-0` on the wrapper is the whole of it, and
`/runs` is the only page that has it and does not want it. #933 keeps the half
that is real (four ad-hoc `pb-*` on top of the working declaration) — PR below.

**#120** (Mobile View) touches the same shell and the same `pb-20`; whichever
lands second should not undo the other.

DEENUU1:
## The defect

`/runs` had **0px** under its last run row where every other list page gets 64px,
because it was in `FULL_HEIGHT_ROUTES` — so `PageTransition` gave it `min-h-0`
and withheld `PAGE_CLEARANCE`.

It stopped being a full-height route in #914. The page's root is
`<div className="flex flex-col">`, an ordinary scrolling page, and the run detail
is `sticky` inside the page's own scroll rather than a pane with a scrollbar of
its own. One regex was answering two different questions on its behalf.

## The fix, and why it is not a one-character one

Splitting the two questions:

- **`OWN_SCROLL_PANE`** — `/chat` alone. It needs the constrained chain so the
  transcript scrolls instead of the page.
- **`OWN_BOTTOM_ROOM`** — `/chat` and `/runs`. Both have something that has to
  reach the bottom edge of the viewport.

Activity then declares the same `PAGE_CLEARANCE` one level in, on its **list
column**, and that placement is the whole point: padding on the box *around* the
two-column row shortens the containing block the sticky panel is clamped to, so
the obvious fix trades one defect for a worse one.

Measured at 1440x800 in Chromium, against a transcription of the page's own chain
(`main` / `DeploymentGate` / `PageTransition` / the `items-start` row), scrolled
to the end:

| arrangement | room below last row | panel top | panel header |
|---|---|---|---|
| today (`/runs` constrained) | **0px** | 8px | visible |
| room on the outer box | 64px | **-48px** | cut off by 56px |
| room on the list column — **this PR** | **64px** | **8px** | **visible** |

The -48px reproduces the figure in the issue exactly. Inside the column the
padding lands under the last row while the row itself still ends at the viewport,
so the panel keeps its full height and pins 8px from the window top.

## Also fixed, found while measuring

Below `lg` the list column is `hidden` and the panel is the only column, so its
flat `h-[calc(100dvh-1rem)]` put its **last 56px behind the mobile tab bar** —
`fixed bottom-0`, `min-h-[56px]` plus the safe-area inset. Measured at 390x780:
56px hidden before, 8px clear after, which matches the 8px it already keeps at
the top. The full height stays above `lg`, where the bar is `lg:hidden`.

Same element, two lines, measured — so it is here rather than in a second PR.

## The test that was asserting the old reason

`page-transition.test.tsx` said *"constrains Activity too, where the list and the
run detail scroll apart"* — true before the page was rebuilt, false since. It now
asserts both halves of what is true: no `min-h-0`, and no `pb-` either. The
prefix-match case also checks `/runsomething` still **gets** the room.

## Verified

- `bunx vitest run` over the layout and runs specs — 36 files, 309 tests
- `make lint-frontend`
- `make test-frontend-cov` — 100% statements/lines/functions, 97.7% branches
- `make build-frontend`

## Note for #120

`PAGE_CLEARANCE` and the mobile bar's geometry are both touched here. #120 works
the same shell; whichever lands second should read this diff first.

Closes #1206

### DIFF
--- a/frontend/src/app/[locale]/(dashboard)/runs/page.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/runs/page.tsx
@@ -14,6 +14,7 @@ import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui";
 import { useApprovals, usePermissions, useRuns, useSpend, useUrlState } from "@/hooks";
 import { periodEnd, periodStart } from "@/lib/dashboard/period";
 import { formatPeriodParam, parsePeriodParam, type Period } from "@/lib/dashboard/period";
+import { PAGE_CLEARANCE } from "@/lib/page-clearance";
 import { parseRunFilters, writeRunFilters, type RunFilters } from "@/lib/runs/filter-params";
 import { cn, setUrlParam } from "@/lib/utils";
 import { Perm } from "@/types/permissions";
@@ -192,6 +193,15 @@ export default function RunsPage() {
                   // underneath the detail panel rather than scrolling inside its
                   // own column.
                   "min-w-0 flex-1 overflow-hidden",
+                  // The room under this page, declared here rather than by
+                  // `PageTransition` - which is where every other page gets it.
+                  // Padding on the box *around* this row shortens what the
+                  // sticky panel beside it may pin in, because a sticky box is
+                  // clamped to its containing block: 64px of it put the panel's
+                  // top at -48px at maximum scroll, cutting its own header off
+                  // by 56px. Inside the column it lands under the last row and
+                  // the row still ends at the viewport (#1206).
+                  PAGE_CLEARANCE,
                   focusedRunId !== null && "hidden lg:block",
                 )}
               >
@@ -257,7 +267,14 @@ export default function RunsPage() {
                 // pixels of nothing to the right of the panel while the table ran
                 // flush to the left edge. Sixteen of those are given back, which
                 // matches the eight above and below.
-                <div className="sticky top-[calc(0.5rem-1rem)] h-[calc(100dvh-1rem)] self-start sm:top-[calc(0.5rem-2rem)] lg:-mr-4">
+                //
+                // The height counts the mobile tab bar below `lg`, where the
+                // list column is hidden and this is the only column: the bar is
+                // `fixed bottom-0`, `min-h-[56px]` plus the safe-area inset, so
+                // a flat `100dvh-1rem` put the panel's last 56px behind it -
+                // measured at 390x780. Above `lg` the bar is `lg:hidden` and the
+                // full height is right.
+                <div className="sticky top-[calc(0.5rem-1rem)] h-[calc(100dvh-1rem-3.5rem-env(safe-area-inset-bottom))] self-start sm:top-[calc(0.5rem-2rem)] lg:-mr-4 lg:h-[calc(100dvh-1rem)]">
                   <RunDetailPanel runId={focusedRunId} onFocusRun={focusRun} />
                 </div>
               )}
--- a/frontend/src/app/[locale]/(dashboard)/runs/page.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/runs/page.tsx
@@ -268,13 +268,18 @@ export default function RunsPage() {
                 // flush to the left edge. Sixteen of those are given back, which
                 // matches the eight above and below.
                 //
-                // The height counts the mobile tab bar below `lg`, where the
-                // list column is hidden and this is the only column: the bar is
-                // `fixed bottom-0`, `min-h-[56px]` plus the safe-area inset, so
-                // a flat `100dvh-1rem` put the panel's last 56px behind it -
-                // measured at 390x780. Above `lg` the bar is `lg:hidden` and the
-                // full height is right.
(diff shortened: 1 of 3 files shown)


DEENUU1:
Right, and the fault was in my measurement rather than in the reasoning: the mobile probe transcribed the tab bar and left `MobileHeader` out. It is `h-14`, `md:hidden`, and in **normal flow** above the row that holds `main` (`layout.tsx:30`), so below `md` the scrollport starts 56px down — and subtracting only the bar left exactly the 56px the commit claimed to have fixed.

Re-measured at 390x780 with the header in the transcription:

| height | hidden behind the bar |
|---|---|
| `100dvh - 1rem` | 56px |
| `- 3.5rem` (bar only) | **56px** — what I had |
| `- 3.5rem - 3.5rem` (bar + header) | 8px |

Three steps now, since the two come off at different breakpoints: bar and header below `md`, the bar alone from `md` where the header is hidden and the bar is not, neither from `lg`. Verified at 390x780 and 800x900 — 8px clear at both, matching the 8px it keeps at the top. Fixed in `ce8a8b9c`.

@codex review

DEENUU1:
Right again, and this one changed my mind about the approach rather than just the numbers.

Measured at 390x780 after the last commit: the panel's last **8px** was still behind the bar. Chasing it meant the height had to account for the bar, the header, `main`'s padding *and* the row's `mt-2` — four numbers owned by other files. And your other point is what settles it: the element carrying `PAGE_CLEARANCE` is the list column, which is `display:none` in exactly the state where the panel is alone, so nothing was reserving room under it at all.

So the panel is now sticky **from `lg` only** — which is where there is a second column for it to stay beside. Below `lg` it is an ordinary block that scrolls with the page and carries the clearance itself. No `dvh` arithmetic on mobile and nothing to keep in step with the shell.

Measured with the list hidden, as it is when a run is focused:

| | result |
|---|---|
| 390x780 | panel content ends **25px above** the tab bar |
| 1440x800 | panel pins 8px from the top, ends **9px above** the viewport bottom |

309 tests, lint, and the coverage gate all green. `ed86392e`.

@codex review

PR #1299, opened 2026-08-27
fix(frontend): stop the card grids overflowing a phone viewport

DEENUU1:
## The audit first

#120 had `describe it later` in every field, so **the issue body is now the
audit** — measured at 390x780 and 768x1024 in Chromium against the running app,
fourteen pages, each scrolled to the end, recording overflow, tap-target size,
text size and anything intersecting the fixed tab bar. Read that for the full
picture and the suggested split; this PR is the three clipped-layout bugs it
found.

Good news up front: **the document never scrolls horizontally**, at either width,
on any of the fourteen pages.

## What this PR fixes

| | Before | After |
|---|---|---|
| Agent / skill / context / MCP / KB card grids at 390px | card **391px** in a 324px grid — clipped by 34px | fits |
| Chat control bar at 390px | ran **27px** past the composer | fits |
| A dashboard widget's info button | **14x14** target | 44x44 hit area, same glyph |

### The grid one is not where it looks

`grid gap-3 md:grid-cols-2 xl:grid-cols-3` declares no column count **below**
`md`, so the single implicit track is `auto` and sizes to its items' content. The
grid box measured 324px while its own `grid-template-columns` computed
**391.094px**.

I tried the obvious fix first: `min-width: 0` injected at every level of the
ancestor chain, one level at a time. The card stayed 391px at all seven, because
the track is what is too wide, not the item. `grid-cols-1` —
`repeat(1, minmax(0, 1fr))` — makes the track the container's 324px, and the
card's own `truncate` finally has something to truncate against.

Applied to **fourteen grids across nine files**: the ones whose cards carry
user-supplied unbreakable text (a slug, an email, a URL, an id), because those
are the ones whose min-content is unbounded. The sweep found **71** grids with
the same shape; the other 57 are deliberately left alone — the pattern is only a
defect when something inside cannot be broken, and a no-op class on 57 files is a
diff nobody can review.

### The chat bar

Three controls, 358px of them, in a `justify-between` row with the connection
pill. The pill is `shrink-0` now (two words, nothing to give) and the control
group `min-w-0` — which `AgentPicker`'s trigger needed too: its
`max-w-[160px]` on the name is a cap, not permission for a flex item to shrink.

### The 14px button

A bare `<button>` around a `size-3.5` icon, one per dashboard card — a third of
the 44px both mobile platforms ask for. `-m-2 p-2` moves the hit area into the
padding without moving anything on screen.

## What is deliberately left, and why

- **The three data tables.** Each sits in its own `overflow-x-auto` scroller, so
  no column is lost — but Activity is **1155px in a 364px column**, a 3.2x
  sideways scroll to read one row. Below `md` those want to be cards. That is the
  bulk of the `effort:xl` label and it is a design change, not a CSS fix; scoped
  on the issue.
- **45 sub-40px tap targets on the dashboard**, most of them the repo's own
  `icon-sm` (`h-7 w-7` = 28x28) in `button.tsx`. The fix is one hit-area token
  below `md`, not bigger buttons — making them bigger changes every dense layout.
  Scoped on the issue.
- **The 10–11px mono label register.** Fifteen strings. Two are sentences (the
  composer's `AI can make mistakes…` at 10px) and the rest are deliberate
  micro-labels. Whether that register moves on a phone is a design decision, so
  this PR does not take it.

## Verified

- Same probe before and after: **zero clipped elements at 390px** on every page,
  down from three pages with clipping
- 94 spec files / 1671 tests over the touched areas
- `make lint-frontend`, `make test-frontend-cov` (100% statements/lines/functions,
  97.7% branches)

## Note on #1206

That PR touches `PAGE_CLEARANCE` and the same mobile chrome, and its review turned
up a fact this audit needed: `MobileHeader` is `h-14` and **in normal flow**, so
below `md` the scrollport starts 56px down the viewport. Anything computing
`100dvh` has to subtract it. Whichever lands second, no conflict — different files.

Refs #120

### DIFF
--- a/frontend/src/app/[locale]/(dashboard)/agents/page.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/agents/page.tsx
@@ -145,7 +145,7 @@ export default function AgentsPage() {
           render inside the same frame. */}
       {isLoading ? (
         <AgentsCard visible={null} total={0} controls={galleryControls}>
-          <div className="grid gap-3 md:grid-cols-2 xl:grid-cols-3">
+          <div className="grid grid-cols-1 gap-3 md:grid-cols-2 xl:grid-cols-3">
             {[0, 1, 2].map((row) => (
               <div key={row} className="border-border rounded-xl border p-4">
                 <div className="flex items-start gap-3">
@@ -186,7 +186,7 @@ export default function AgentsPage() {
               }
             />
           ) : (
-            <div className="grid gap-3 md:grid-cols-2 xl:grid-cols-3">
+            <div className="grid grid-cols-1 gap-3 md:grid-cols-2 xl:grid-cols-3">
               {visible.map((agent) => (
                 <AgentCard
                   key={agent.id}
--- a/frontend/src/app/[locale]/(dashboard)/context/page.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/context/page.tsx
@@ -289,7 +289,7 @@ export default function ContextPage() {
             />
           ) : (
             <>
-              <div className="grid gap-3 md:grid-cols-2 xl:grid-cols-3">
+              <div className="grid grid-cols-1 gap-3 md:grid-cols-2 xl:grid-cols-3">
                 {files.map((entry) => (
                   <ContextCard
                     key={entry.id}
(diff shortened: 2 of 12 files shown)


DEENUU1:
Right, and the arithmetic was mine to get wrong: `p-2` around a 14px glyph is 30x30, and `-m-2` moves the box in the layout without adding a pixel to the clickable area. The commit claimed 44 and delivered 30.

Padding cannot get there without moving the title — it would need 15px a side — so the hit area is now an absolutely positioned `::before` with `-inset-[15px]`: 14 + 15 + 15 = 44, taking pointer events for the button and occupying no layout.

Verified by **clicking** rather than measuring: a transcription of the card header, clicked 20px above and 20px left of the glyph's centre — outside the 14px icon, inside the 44px box — registers on the button both times (2/2).

`4b564dac`. Dashboard specs and lint green.

@codex review

Issue #1347, opened 2026-08-31
rag: delete paths drop vector tables and remove vectors before the request commits


OchnikBartek:
### What breaks

#1293 deferred the *file* unlinks in the RAG delete paths to after the request
commits, so a rolled-back delete no longer strands a restored row on a missing
file. The same paths still do their **vector-store** side effects before the
commit, on the store's own engine (a separate transaction), so the same
rollback bug remains for those:

- `RAGDocumentService.delete_document` calls `ingestion_service.remove_document`
  before the row delete commits. A rollback restores the row pointing at vectors
  already gone.
- `KnowledgeBaseService.delete` calls `vector_store.delete_collection` (a
  `DROP TABLE`) before the commit. A rollback restores the `rag_documents` rows
  and the KB row pointing at a `rag_<collection>` table that no longer exists.

The table-drop case is the sharper one: it is a whole collection, and a restored
row describing a dropped table is a correctness fault, not a recoverable leak.

### Where

- `backend/app/services/rag_document.py` — `delete_document`
- `backend/app/services/knowledge_base.py` — `delete`

### How to fix

Defer the vector-store ops to `spawn_after_commit`, the way #1293 deferred the
files and #1274 deferred the org purge's table drops to a durable flow. Mind two
things the file case did not have: the store/ingestion-service must be held by
the deferred coroutine (both ride the process's vector engine, not the request
session, so this is safe), and the `list_by_collection_name` reference check that
gates the drop has to move into (or be re-read by) the deferred step, so a name a
second org reclaims between the commit and the drop keeps its table (#913) - the
same re-check `OrganizationService.purge`'s deferred cleanup already does.

### How you'd know it's fixed

A KB delete (or document delete) whose transaction is forced to roll back after
the row delete leaves the vector table (or the document's vectors) intact, so no
restored row describes a table or vectors that are gone.

Found finishing #1293 (deferred the file half); sibling of #1274 (org-teardown
durability). Cluster: RAG teardown / delete governance in #168.

Issue #1293, opened 2026-08-27
rag: teardown unlinks stored files before the request commits


OchnikBartek:
### What breaks

The RAG teardown paths unlink stored uploads inside the request's
transaction, before it commits. `RAGDocumentService.delete_by_collection`
and `KnowledgeBaseService.delete` both do the SQL row delete (via
`db.flush()`, not commit) and then loop over the returned storage paths
calling `get_file_storage().delete(path)`. The request session commits
once, after the route returns.

If that final commit fails — a dropped connection, a serialization
failure — the SQL deletion rolls back and the `rag_documents` rows are
restored, but the files have already been unlinked. The restored rows then
reference uploads that no longer exist: downloads and re-ingestion break,
and the originals that could have rebuilt the already-dropped vectors are
gone. The same before-commit ordering is used by `delete_document` and by
`complete_ingestion`'s replacement path, so a fix should cover the pattern,
not one method.

### Where

- `backend/app/services/rag_document.py` — `delete_by_collection`, `delete_document`, `complete_ingestion`
- `backend/app/services/knowledge_base.py` — `delete`

### How to fix

Defer the irreversible unlinks until after the commit succeeds, using the
codebase's `spawn_after_commit` idiom (`app/core/background.py`): capture
the paths from the DELETE ... RETURNING, then schedule the unlinks so a
rollback discards them. The tests that currently assert `storage.delete`
is awaited during the service call move to asserting the deferred work.

### How you'd know it's fixed

A teardown whose transaction is forced to roll back after the row delete
leaves the stored files on disk (regression test), and no path unlinks a
file before its owning transaction commits.

Found reviewing #1284 (PR for #1265). Related teardown work: #1265, #1266.
Belongs to the "Ingestion and retrieval" cluster in #168.

OchnikBartek:
Queued (W4 close-out) but blocked on merge order. Two of the four call sites this must fix — `RAGDocumentService.delete_by_collection` and `KnowledgeBaseService.delete` — live on the still-open #1284 (#1265) and #1286 (#1266), not on main. Fixing only the on-main sites (`delete_document`, `complete_ingestion`) would leave the issue half-done and would conflict with #1284's edits to `rag_document.py` when it merges. So this waits until #1284 and #1286 land, then fixes all four uniformly in one PR.

[...]

Issue #1274, opened 2026-08-27
Org-teardown cleanup is lost if the process dies after commit (spawn_after_commit is not durable)


OchnikBartek:
**Found in review of #1259** (#1137). The org-teardown cleanup — dropping the
vector table and unlinking the stored uploads — was moved to a post-commit
`spawn_after_commit` task to fix residual 1 (a failed final commit leaving rows
pointing at gone vectors). That closed the commit-failure window, but
`spawn_after_commit` is explicitly **not durable** (`app/core/background.py`:
"Work that must survive a restart belongs in Prefect"). If the process crashes
or is restarted after the relational commit but before the in-process cleanup
runs, the drop/unlink is lost — the org/KB/doc rows are already committed-gone,
so nothing carries the paths and collection names needed to retry, and the
vector table and files are retained indefinitely.

This is a narrower and more recoverable failure than the bug #1137 fixed (a
leaked, unreferenced table/file is reclaimable by a sweep; rows pointing at gone
vectors are a correctness fault), and the #1137 issue explicitly specified
`spawn_after_commit` as the mechanism — so it is out of scope there. Tracking the
durable version separately.

## Fix direction

Commit a durable teardown record with the relational delete (an outbox row, or a
Prefect flow handed over with `spawn_after_commit`) and have a worker drop the
table / unlink the files and clear the record, retrying until done. Same pattern
would serve the sequential-cleanup window on very large purges.

## How I'd know it's fixed

- [ ] A teardown whose process dies after commit but before cleanup still drops
      the vector table and unlinks the files on a later run (no orphaned
      `rag_<collection>` table or file left behind).

Cluster: RAG teardown / delete governance (#9, #1116, #1131, #1137, #913, #1265,
#1266). Checked #168.

OchnikBartek:
Investigated for the W4 follow-up batch. This one needs a **design decision**, not a local fix, so I'm documenting the options rather than forcing one — flagging it project-blocked per the batch's own "describe options and skip" rule.

## The gap (restated precisely)

[...]

PR #1356, opened 2026-09-01
fix(rag): make bulk RAG teardown cleanup durable across a restart

### LINKED ISSUE #1349: rag: deferred teardown file unlinks are in-process, not durable across a restart
### What breaks

#1293 moved the RAG teardown file unlinks to `spawn_after_commit`, which fixed
the ordering bug (a rollback no longer strands a restored row on a missing file).
But `spawn_after_commit` is an **in-process** handoff, not durable
(`app/core/background.py`: "Work that must survive a restart belongs in Prefect").

So when a collection drop or KB delete leaves enough files that the deferred
unlink is still queued or running during a worker restart or shutdown, the
transaction has already removed every `rag_documents` row while the paths live
only in process memory - and the unlink is lost. The files are orphaned on disk
with no row left to identify them.

This is a **recoverable leak** (unreferenced bytes a sweep could reclaim), not the
correctness fault #1293 removed (a live row pointing at a gone file) - which is why
it is out of #1293's ordering scope and lower severity.

### Where

- `backend/app/services/rag_document.py` — `delete_by_collection`, `delete_document`, `_retire_superseded`
- `backend/app/services/knowledge_base.py` — `delete`
- all via `app/services/file_storage.py::delete_files_best_effort`

### How to fix

The same durable pattern #1274 gave the org purge: hand the unlink to a Prefect
deployment run after the commit (the paths are the run's parameters, persisted on
the server and retried by a worker), rather than an in-process task. Once #1347
lands, the deferred vector ops on these paths want the same treatment, so a single
durable cleanup dispatch could carry both.

### How you'd know it's fixed

A collection drop whose worker dies mid-unlink still unlinks the remaining files
on a later run - no orphaned upload left under `MEDIA_DIR` for rows that are gone.

Found finishing #1293 (codex review). Sibling of #1274 (org-teardown durability)
and #1347 (the vector-op half). Cluster: RAG teardown / delete governance in #168.

OchnikBartek:
Stacked on #1354 (`fix/1347-defer-vector-ops`), itself on #1348. The diff here
is the single `#1349` commit. GitHub retargets this to `main` as the parents
merge; `Closes #1349` takes effect once the base is `main`.

## What changed

#1293 and #1347 deferred the RAG teardown side effects past the request commit
with `spawn_after_commit` — fixing the *ordering* (a rollback no longer strands a
restored row on a gone file or table) but not the *durability*: an in-process task
dies with the worker that queued it, orphaning the files and tables it had left to
clean (#1349).

The bulk teardown now goes through the durable Prefect flow #1274 gave the org
purge, generalized to serve both callers:

- **`teardown_tasks.py`** — `org_purge_cleanup` → **`external_state_cleanup`**
  (function, dispatch, flow, and the deployment name). Same `(storage_paths,
  collections)` signature; org purge and the RAG paths now share one durable
  cleanup, and its `#913` reference re-check runs on the worker's own session.
- **`rag_document.py`** — `_retire_superseded` and `delete_by_collection` dispatch
  the durable cleanup instead of an in-process `delete_files_best_effort`.
- **`knowledge_base.delete`** — dispatches the files *and* the collection drop in
  one durable run and no longer takes a `vector_store` (the flow builds its own
  engine). Sync-source deactivation stays inline. The `delete_knowledge_base` route
  drops the param too.

## Scope decision (please sanity-check)

`delete_document` (a **single** document) is left **in-process** on purpose. The
leak a lost single-document task would strand is one file and one document's
vectors; a durable Prefect run per document delete is disproportionate to that, and
making it durable would mean removing its `ingestion_service` param and undoing
#992's structural guarantee. Documented in its docstring. If you'd rather cover it
too, say so and I'll extend (accepting the #992 rework).

## Deliberately unchanged

The commit-to-dispatch window #1274 documented stays open: a crash after the commit
but before `run_deployment` fires still loses the cleanup — only an outbox closes
that.

## Verification

- Renamed the org-purge cleanup tests to `test_external_state_cleanup.py` and
  converted the RAG teardown tests to assert the durable dispatch `(files,
  collections)` rather than the in-process unlink; the `#913` re-check test moved to
  the flow's own tests.
- Fixed two integration call sites (`test_platform_flows.py`) that still passed the
  removed `vector_store` — found by a review pass, DB-less-skipped locally.
- `make lint-backend` and the affected unit suites pass. No local database, so the
  integration reconciliation tests are left to CI.

Closes #1349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,26 @@ Two things are versioned separately from this file and worth knowing about:
 
 ## [Unreleased]
 
+## [0.0.344] - 2026-09-01
+
+### Fixed
+
+- **The RAG delete paths unlinked stored uploads inside the request transaction.**
+  `rag_document.delete_by_collection`, `delete_document`, `complete_ingestion`'s
+  `_retire_superseded` and `knowledge_base.delete` all removed the file before the
+  commit, so a commit that then failed rolled the `rag_documents` rows back and left
+  the files gone - a restored row pointing at a missing upload, its download and
+  re-ingestion broken, and the original that could have rebuilt the vectors lost. All
+  four now hand the unlink to `spawn_after_commit` after deleting the rows, through
+  one shared `delete_files_best_effort(paths)` in `app/services/file_storage.py`
+  which holds only ids, resolves the storage backend when it runs, and suppresses a
+  file already gone. The table drop and the sync-source deactivation stay in the
+  transaction. (#1293)
+- The vector-store half of those same paths - `delete_document`'s `remove_document`
+  and `knowledge_base.delete`'s `DROP TABLE` - still runs before the commit, which is
+  the same rollback class on the store's own engine and needs the #913 reference
+  re-check moved with it. Filed as #1347 rather than widened into this change. (#1293)
+
 ## [0.0.343] - 2026-08-28
 
 ### Changed
--- a/backend/app/api/routes/v1/knowledge_bases.py
+++ b/backend/app/api/routes/v1/knowledge_bases.py
@@ -127,15 +127,15 @@ async def update_knowledge_base(
 async def delete_knowledge_base(
     kb_id: UUID,
     service: KnowledgeBaseSvc,
-    vector_store: VectorStoreSvc,
     ctx: Auth,
 ) -> None:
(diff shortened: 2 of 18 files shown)


OchnikBartek:
@codex review

Issue #1359, opened 2026-09-01
rag: two in-request collection-drop paths bypass the durable locked teardown


OchnikBartek:
### What breaks

Two collection-drop paths remove a `rag_<collection>` vector table **in the request
process**, before the request commits, and outside the durable teardown the rest of
the delete surfaces now route through (`cleanup_external_state`, #1349). Neither
takes the `COLLECTION_TEARDOWN` advisory lock #1355 added, so both can still lose the
race #1355 closes for the durable path — and one skips the `#913` reference check
entirely:

- `app/api/routes/v1/rag.py::drop_collection` (`DELETE /rag/collections/{name}`,
  line ~208-210) resolves one writable KB and calls `vector_store.delete_collection`
  directly, with **no `list_by_collection_name` re-check at all** and no lock. So it
  (a) drops the shared table while another organization's still-committed KB row
  references the same non-tenant-unique name (#913), and (b) races a concurrent
  `claim` of the name (the #1355 TOCTOU).
- `app/services/user.py::_purge_personal_collections` (reached from
  `UserService.delete` → `_release_owned_rows`, #1131, line ~550-557) re-reads
  `list_by_collection_name` then drops with no lock between the two, so a concurrent
  `claim` can create-and-commit a KB row and its table in the window and this drop
  then removes it.

Both are also **in-request** drops, so a rolled-back delete strands the restored rows
on a dropped table — the ordering bug #1347/#1349 fixed for the KB-delete and
org-purge paths but never reached these two.

### Where

- `backend/app/api/routes/v1/rag.py` — `drop_collection`
- `backend/app/services/user.py` — `_purge_personal_collections`

### How to fix

Bring both under the same durable, locked, re-checked teardown
`KnowledgeBaseService.delete` uses: delete the document rows and the KB row in the
request, then hand the file unlinks and the collection drop to
`dispatch_external_state_cleanup(storage_paths, collections_to_drop)`, which takes the
`COLLECTION_TEARDOWN` lock, re-reads the reference check on its own session, and drops
only what no base still references. That closes all four gaps at once (the missing
#913 re-check on the route, the #1355 lock, the #1347 rollback ordering, and the
#1349 restart-durability). `drop_collection` moving its drop into the flow lets it
drop its `vector_store` dependency, as `delete_knowledge_base` did.

### How you'd know it's fixed

A `DELETE /rag/collections/{name}` for a name a second base still holds keeps that
[...]

PR #1360, opened 2026-09-01
fix(rag): serialize a collection's claim against its vector-table drop

### LINKED ISSUE #1355: rag/org-teardown: collection drop's reference re-check is not serialized with collection claims (TOCTOU)
### What breaks

Tearing down a collection's vector table is gated by a `list_by_collection_name`
reference check, then the table is dropped — but the check and the drop are **not
serialized with collection claims**, so a name reclaimed in the window between
them loses its freshly-created table.

Concretely, against `POST /rag/collections/{name}`
(`app/api/routes/v1/rag.py::create_collection`), which runs
`claim → create_collection (CREATE TABLE) → create KB row (commit)` and takes no
shared lock:

1. A KB delete (or an org purge) commits, removing the last row for the name; the
   deferred/durable drop is queued.
2. The drop's `list_by_collection_name` re-check runs and returns empty.
3. A concurrent `POST /rag/collections/{name}` for the **same name** claims it,
   `CREATE TABLE`s the vector table, and commits a KB row pointing at it.
4. The drop's `delete_collection` (`DROP TABLE`) then executes and drops that live
   table.

Result: a committed KB pointing at a vector table that is gone — a correctness
fault (search/ingestion against a missing `rag_<collection>` table), not a
recoverable leak. Narrow window and specific preconditions (same name reclaimed
between the re-check and the drop), so it is most reachable for **explicit-name**
collections; derived names carry a random suffix and rarely collide.

### Where

The same non-atomic re-check-then-drop appears in all three places, so a fix has
to cover them together:

- `backend/app/services/knowledge_base.py` — `_drop_collection_if_unreferenced`
  (the deferred KB-delete drop, added in #1347).
- `backend/app/worker/tasks/teardown_tasks.py` — `purge_org_external_state` (the
  durable org-purge drop, #1274) — the identical `list_by_collection_name`
  re-check then `delete_collection`.
- `backend/app/api/routes/v1/rag.py` — `create_collection`, and
  `app/services/collection_access.py::CollectionAccessService.claim` /
  `vector_store.create_collection` — the claim/create side the drop races.

### How to fix

Serialize the check-and-drop with collection claims rather than relying on a
non-atomic re-read. A Postgres advisory lock keyed on the collection name
(`pg_advisory_xact_lock(hashtext(name))`), taken by **both** the create/claim path
and each drop path, makes "is this name still referenced?" and "drop it" atomic
against a concurrent claim. Apply it once, to the shared teardown so both the rag
KB-delete drop and the org-purge drop inherit it, and to the create path so the
two contend on the same lock space.

### How you'd know it's fixed

Two requests racing — one dropping a collection whose last KB was just deleted,
one re-creating a collection with the same explicit name — never leave a committed
KB pointing at a dropped table: either the create waits and re-creates the table,
or the drop sees the new reference and skips.

Found reviewing #1347 (codex P1). #1347 fixed the *ordering* bug (deferring the
drop past the commit) and re-reads the reference check exactly as its spec asked;
this is the deeper *concurrency* gap that re-read does not close, pre-existing in
#1274's merged org-purge cleanup too. Sibling of #1347 / #1274 / #913. Cluster:
RAG teardown / delete governance in #168.

OchnikBartek:
Stacked on #1356 (`fix/1349-durable-rag-teardown`), itself on #1354/#1348. The diff
here is the single `#1355` commit; GitHub retargets this to `main` as the parents
merge, and `Closes #1355` takes effect once the base is `main`.

## What changed

The teardown re-reads `list_by_collection_name` before dropping a collection's
vector table (#913), but the re-check and the drop are two statements and the claim
path is two more: under READ COMMITTED a claim that reads "this name is free" and a
drop that reads "no base holds this name" can both act, and the drop then removes
the table the claim just created and committed a KB row against (#1355). `POST
/rag/collections/{name}` creates the table before committing its row, so the window
is real.

A transaction-scoped advisory lock, keyed on the collection name, at both ends:

- `app/db/locks.py` gains **`hold_name`** (the string-subject sibling of
  `hold_subject`, keyed on `hashtext(name)`) and a **`COLLECTION_TEARDOWN`** scope.
- **`CollectionAccessService.claim`** takes it after the identifier rule and before
  the taken-check, held until the request commits the row — past `create_collection`,
  which builds the table it protects.
- **`cleanup_external_state`** takes it before each collection's re-check, held
  through the `delete_collection`.

Either the claim commits first (the drop's re-check sees the row and skips) or the
drop commits first (the claim re-creates the table its row points at).

## Scope — two paths deferred to #1359

This covers the claim path and the durable teardown every properly-deferred drop
routes through. **Two pre-existing in-request drops still bypass that teardown** —
`DELETE /rag/collections/{name}` (no `#913` re-check *and* no lock) and
`_purge_personal_collections` (re-check, no lock) — so the invariant remains
reachable-to-violate through them. Bringing both under the same durable, locked,
re-checked cleanup is **#1359** (filed, triaged), noted here because #1355 does not
close it.

## Verification

- A real-database integration test (`tests/integration/test_collection_teardown_lock.py`)
  proves two holders of one name serialize and two holders of different names do not.
- Unit tests assert both paths take the `COLLECTION_TEARDOWN` lock on the name; the
  mocked sessions the claim and teardown suites use grew an async `execute` because
  the lock issues one.
- `make lint-backend` and the full non-integration suite (6619 tests) pass; the
  serialization itself is exercised only in CI, which has a database.

Closes #1355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,45 @@ Two things are versioned separately from this file and worth knowing about:
 
 ## [Unreleased]
 
+## [0.0.345] - 2026-09-01
+
+### Fixed
+
+- **The RAG delete paths did their vector-store work inside the request
+  transaction.** #1293 deferred the file unlinks; the store side effects stayed
+  where they were, so the same rollback left the vectors gone.
+  `RAGDocumentService.delete_document`'s `remove_document` and
+  `KnowledgeBaseService.delete`'s `delete_collection` `DROP TABLE` now go over with
+  `spawn_after_commit` too. The collection drop is the sharper of the two: a
+  rollback used to restore the knowledge-base and `rag_documents` rows pointing at
+  a table that no longer existed, which is a correctness fault rather than a
+  recoverable leak. The deferred drop re-reads its `list_by_collection_name`
+  reference check on a session of its own, because the name is not tenant-unique
+  and a second organization can reclaim it in the window between the commit and
+  the drop (#913) - the same re-check the org purge's durable cleanup makes. (#1347)
+- Still deferred in-process, so a crash between the commit and the dispatch loses
+  the cleanup; that durability gap is #1349. (#1347)
+
+## [0.0.344] - 2026-09-01
+
+### Fixed
+
+- **The RAG delete paths unlinked stored uploads inside the request transaction.**
+  `rag_document.delete_by_collection`, `delete_document`, `complete_ingestion`'s
+  `_retire_superseded` and `knowledge_base.delete` all removed the file before the
+  commit, so a commit that then failed rolled the `rag_documents` rows back and left
+  the files gone - a restored row pointing at a missing upload, its download and
+  re-ingestion broken, and the original that could have rebuilt the vectors lost. All
+  four now hand the unlink to `spawn_after_commit` after deleting the rows, through
+  one shared `delete_files_best_effort(paths)` in `app/services/file_storage.py`
+  which holds only ids, resolves the storage backend when it runs, and suppresses a
+  file already gone. The table drop and the sync-source deactivation stay in the
+  transaction. (#1293)
+- The vector-store half of those same paths - `delete_document`'s `remove_document`
+  and `knowledge_base.delete`'s `DROP TABLE` - still runs before the commit, which is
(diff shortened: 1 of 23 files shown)


OchnikBartek:
@codex review

PR #1388, opened 2026-09-04
feat(memory): promote a fact from agent to operator

### LINKED ISSUE #1384: Add promote for facts (agent→operator), matching files
## What

Files have a **promote** (agent→operator) — a deliberate act that turns an agent-authored note into a trusted, injectable one. Facts do not. Introduced with the fact `origin` trust tier in #788: the standing brief injects only the *injectable* set (a person's own facts + operator-authored shared ones) and never an agent-authored **shared** fact.

So an agent-authored shared fact an operator wants in the brief has no one-click path — it must be **deleted and recreated** as an operator seed. The `origin` migration made this concrete: existing facts backfilled to `agent` (the conservative default), so any operator-seeded shared fact created before the column (e.g. a "company facts" seed) dropped out of the shared brief until reseeded. `promote_fact` is the clean fix — promote it in place instead of delete+recreate, which also removes the need for the one-time reseed.

## Where

- `backend/app/services/memory/facade.py` — files promote via a service method that sets `origin=operator` (see the file `promote` path and `test_promotion_sets_origin_operator_and_is_audited`); facts have no equivalent.
- `backend/app/repositories/memory.py` — a fact update-origin write would mirror the file one.
- Frontend `memory-facts-pane.tsx` — a "promote" control next to the origin badge, gated on edit.

## Done when

An operator can promote an `agent` shared fact to `operator`; it then appears in the standing brief (`list_brief_facts`), the action is audited, and editing a fact never launders its origin (promotion stays deliberate, as for files).

Part-of #788 (found while implementing the fact `origin` trust tier).

OchnikBartek:
## What

Files could be **promoted** from agent-authored to operator-trusted; facts could
not. So an operator who wanted an agent-learned fact in the standing brief had no
one-click path — the only route was to delete the fact and reseed it as an operator
one. This adds the fact counterpart, mirroring the file promote (#1384).

## Changes

- **`set_fact_origin`** (repository) writes the one field a fact can change — its
  content is replaced, never amended — and leaves the stored embedding untouched.
- **`MemoryService.promote_fact`** sets `origin=operator`, audits it
  (`memory.fact.promoted`), and refuses a mem0-backed agent like the other
  fact-management paths. It requires **edit** on the parent agent, not view:
  promotion reaches the shared brief every end-user reads, never just one person's
  own store.
- **`POST /memory/facts/{id}/promote`** exposes it, mirroring the file route.
- **Console**: an inline promote control next to each agent-authored fact (gated on
  edit), and a `promote` mutation on `useMemoryFacts`. No open detail to reconcile
  as there is for a file, so a plain list invalidation is enough. Reuses the file
  promote's copy, so no new strings.
- **Docs**: `capabilities.md` now notes a fact can be promoted, like a file.

Also corrects two stale notes left where they can no longer be amended: the
fact-seed schema docstring still said the embedding was "unmetered" (the metering
landed in #1383, approved and pending merge), and the repo test fixture named
migration `0064`, renumbered to `0072` on this branch.

## Verification

- **Backend**: `test_memory_service.py` (promote sets origin + audits; 404;
  view-only refused; mem0 refused), `test_memory_routes.py` (200 + trusted row),
  and an integration test proving the consequence end to end — an agent-authored
  shared fact is **absent** from `list_brief_facts` before the promote and
  **present** after.
- **Frontend**: `use-memory.test.tsx` (promote + failure) and the facts-pane
  integration test (promote offered only on an agent-authored fact; a viewer sees
  none).
- `make lint` (ruff / ty / guards), both type checks, and the full frontend
  coverage gate (404 files, 6159 tests, exit 0) pass locally.

## Stacking

Stacked on `feat/agent-memory` (#1383) so it need not wait for that merge —
**retarget to `main` once #1383 lands**. Do not merge before #1383.

Closes #1384

### DIFF
--- a/backend/app/api/routes/v1/memory.py
+++ b/backend/app/api/routes/v1/memory.py
@@ -124,6 +124,13 @@ async def get_memory_fact(fact_id: UUID, service: MemorySvc, ctx: Auth) -> Any:
     return await service.get_fact(ctx, fact_id)
 
 
+@router.post("/facts/{fact_id}/promote", response_model=AgentMemoryFactRead)
+async def promote_memory_fact(fact_id: UUID, service: MemorySvc, ctx: Auth) -> Any:
+    """Mark an agent-authored fact operator-trusted, so it may enter the standing
+    brief. Deliberate, like a file's promote; native-only."""
+    return await service.promote_fact(ctx, fact_id)
+
+
 @router.delete("/facts/{fact_id}", status_code=status.HTTP_204_NO_CONTENT, response_model=None)
 async def delete_memory_fact(fact_id: UUID, service: MemorySvc, ctx: Auth) -> None:
     """Forget a fact. Operators do not create or edit facts - only the agent does,
--- a/backend/app/repositories/memory.py
+++ b/backend/app/repositories/memory.py
@@ -471,6 +471,23 @@ async def list_facts(
     return list(items.scalars().all()), total or 0
 
 
+async def set_fact_origin(
+    db: AsyncSession, *, fact: AgentMemoryFact, origin: str
+) -> AgentMemoryFact:
+    """Change a fact's trust tier - the promote write - and return the refreshed row.
+
+    A fact's only mutation: its content is never amended, only replaced, so unlike a
+    file's `update` this takes just the one field. `embedding` is not a mapped column
+    (see `AgentMemoryFact`), so a plain attribute set, flush and refresh leaves the
+    stored vector untouched.
+    """
+    fact.origin = origin
+    db.add(fact)
+    await db.flush()
+    await db.refresh(fact)
+    return fact
+
+
 async def delete_fact(db: AsyncSession, fact: AgentMemoryFact) -> None:
     await db.delete(fact)
     await db.flush()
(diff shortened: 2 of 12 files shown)


OchnikBartek:
@codex review

OchnikBartek:
Closing per an updated workflow: for these memory follow-ups the push and PR are the maintainer's to make, not the worker session's. The change is committed locally on `feat/memory-fact-promote` (stacked on `feat/agent-memory`, 2 commits) and will be pushed and opened as a PR by the maintainer. For the record, CI here was green — test, test-frontend, e2e, lint, docs all passed. Superseded, not abandoned.

DEENUU1:
Closing without merging: everything this builds on was removed from #1383 before it landed.

`promote_fact` moves a memory row from `origin: agent` to `origin: operator`. All three parts of that sentence are gone:

- **the native fact store** — semantic memory is now its own capability (`memory_mem0`) against mem0, and nothing is stored here to promote;
- **the `origin` trust tier** — every note is the agent's own now, so there is no second tier to promote into. It existed to keep an agent from laundering its own content into the injected set, and the injection rule is decided by the run's audience instead;
- **the operator console** — there is no Memory tab. An operator paging through what an agent wrote about a named colleague is a surveillance affordance rather than a feature, and standing knowledge somebody *wants* an agent to have belongs in `context`.

All twelve files it touches are either deleted or rewritten by #1383 — `memory-facts-pane.tsx` among them — so there is no merge to attempt and nothing to salvage. #1384, the issue behind it, is already closed as superseded for the same reason.

If the want behind it returns — an operator vouching for something an agent wrote — it needs a fresh design against the shape that shipped, not this diff.

Issue #1389, opened 2026-09-04
rag: teardown lock-ordering residuals (gap-collection D↔K, user-delete U↔T) need a serialization point


OchnikBartek:
### What breaks

#1387 gives every teardown path one lock order - the `COLLECTION_TEARDOWN`
advisory lock before any `knowledge_bases`/`organizations` row lock - which
removes the `claim` vs `purge` ABBA. Two deadlock residuals survive it, both
needing a serialization point (an outbox, or a generation table) that this
lock-ordering pass cannot express. Both are narrow, and #1387 made both
*strictly narrower* than they were - but neither is closed.

**1. Gap-collection D↔K (a purge residual, narrowed by #1387).**
`OrganizationService.purge` takes teardown locks for a *snapshot* of the org's
collections read without the org row lock (`organization.py:384`), then the org
`FOR UPDATE` (`:389`), then the authoritative re-list and delete (`:392-395`,
docs `D` before the KB row `K`, as `ON DELETE SET NULL` + #1116 require). A
collection committed in the window between the snapshot and the org lock is in
the authoritative list but not in `held`, so purge deletes its `D` then `K`
under the org lock **without ever holding its teardown lock**.

- A concurrent `KnowledgeBaseService.delete` / `DELETE /rag/collections/{name}`
  of that same collection takes `T → K → D`; purge takes `D → K` on the same
  rows. Both hold one of the collection's row locks and wait on the other →
  `40P01`.
- The same gap collection is added to `to_drop` but **not** tombstone-reserved
  (`organization.py:409-414`) - taking its teardown lock after the org lock
  would reinvert #1387 against `claim`'s `T → O`. So the durable cleanup's
  reference re-check is the only guard, and a `claim` of the name in the
  commit-to-drop window is not refused (`is_reserved` is false), the #1362
  window reopened for this subset.

Reachable only for a collection *created* in the microsecond snapshot→lock
window inside an org being deleted, then *immediately* deleted-again or
reclaimed - and self-resolving (Postgres aborts one side; the caller retries).
Before #1387 the teardown lock was taken late everywhere, so **every**
org-scoped collection had this exposure; now only gap collections do.

**2. User-delete U↔T (pre-existing, untouched by #1387).**
`UserService.delete` takes the subject's `users` row `FOR UPDATE` first
(`user.py:397` → `_lock_for_delete` `:402`, for #1115/#1134), then
`_purge_personal_collections` takes each personal collection's teardown lock
[...]

Issue #1382, opened 2026-09-02
rag: serialise teardown against concurrent writes (upload, sync) end to end


OchnikBartek:
### What breaks

The collection-teardown reservation (`collection_teardowns`, #1362) is consulted by
`CollectionAccessService.claim` (create) and, best-effort, by
`RAGDocumentService.dispatch_upload` (#1364). Two write paths remain unserialised
against a teardown, both raised in the #1378 review:

1. **Worker ingestion does not check the reservation.** A sync
   (`_run_source_sync` → `_open_document_row` → `insert_document`) can create a
   tracking row and insert vectors into a name that is reserved and about to be
   dropped, then record the document `done` while its vectors are dropped by the
   durable cleanup. Reachable for a **default** collection cleared with an active
   sync source (its KB row and `sync_sources` entry survive the clear, #1364).
   Latent today because no code path sets `is_default=True`, but the mechanism is
   real, and `dispatch_upload`'s guard does not cover the worker paths.

2. **The reservation is created after the row delete.** `DELETE
   /rag/collections/{name}` runs `delete_by_collection` (deletes document rows)
   before `delete_for_rag_collection` reserves the name, so a concurrent upload can
   read no reservation, commit a document row, and lose its vectors to the drop.
   The `dispatch_upload` guard is best-effort for this narrow commit-to-reserve
   window.

3. **One lock order.** The `dispatch_upload` guard deliberately takes **no**
   teardown lock: taking `COLLECTION_TEARDOWN` before the insert's foreign-key lock
   on the `knowledge_bases` row inverts `KnowledgeBaseService.delete`'s order (the
   row `FOR UPDATE` first, then the teardown lock) and can deadlock. A full
   serialisation has to pick one order — the teardown lock before the KB-row lock
   in every path.

### The comprehensive fix

- Reserve the name **before** any document rows are deleted, under the teardown lock.
- Check the reservation at the **lowest write level** (the worker `insert_document`
  / the ingestion `still_wanted` gate), so upload, sync and retry are all covered
  uniformly rather than one entry point at a time.
- Take the teardown lock **before** the KB-row lock in every teardown path, so the
  upload guard can hold it again without inverting `delete`'s order.

### Where

- `app/services/rag_document.py` — `dispatch_upload` (guard), the insert paths.
- `app/services/knowledge_base.py` — `delete`, `delete_for_rag_collection`.
[...]

Issue #1364, opened 2026-09-01
rag: teardown reservations need a stuck-tombstone sweep, and the default-clear has an ingest window


OchnikBartek:
### What breaks

Two low-severity refinements the #1362 security review surfaced, both fail-safe and
strictly better than what they replace, so out of #1362's scope but worth tracking.

**1. A stuck teardown reservation blocks a name forever.** #1362 reserves a
collection name in `collection_teardowns` until the durable cleanup drops its table
and releases it. If the drop permanently fails (Prefect retries exhausted, or the
commit-to-dispatch gap loses the run entirely), the row lingers and
`CollectionAccessService.claim` refuses that name indefinitely. There is no reaper.
This is the safe direction - a blocked name beats an adoptable populated table, and
a lost dispatch now strands a *blocking* tombstone rather than an adoptable table -
but a name should not be unusable for good because one worker run failed.

Fix: an age-based sweep (a Prefect flow, like the other `*-sweep` deployments) that,
for a tombstone older than some threshold, re-attempts the drop and clears the row;
and/or an admin CLI to release one by hand.

**2. The default-KB clear has a same-tenant write-vs-drop window.** #1362 fixed
#1361's P1 so `DELETE /rag/collections/{name}` on a *default* base now drops (clears)
its table while keeping the row. A concurrent ingest into that default during the
commit-to-drop window would `_ensure_collection`-adopt the about-to-be-dropped table
and lose those chunks to the drop. It is **same-tenant and transient** (re-ingesting
recovers) - writes don't pass through `claim`, so the tombstone deliberately does not
block them - and strictly better than the stale-searchable-chunks bug it replaces.
Worth closing if the default-clear path stays.

### Where

- `backend/app/repositories/collection_teardown.py`, `app/worker/tasks/teardown_tasks.py` (sweep)
- `backend/app/services/knowledge_base.py::delete_for_rag_collection` (default-clear window)

### How you'd know it's fixed

A collection name whose drop failed becomes claimable again within the sweep window
rather than never; and a default collection cleared while an ingest races it does not
silently drop the freshly-ingested chunks.

Found reviewing #1362 (the teardown tombstone). Both are follow-ups to the RAG
teardown / delete governance cluster in #168; siblings of #1347/#1349/#1355/#1359/#1362.

OchnikBartek:
[...]

Issue #1273, opened 2026-08-27
The personal-collection purge runs before the refusal that can abort the delete


DEENUU1:
**Found reviewing PR #1267**, which tears down a deleted user's personal collections. The gap it closes is real — a personal-scoped knowledge base did survive its owner, with its documents, files and vector table retained and unreachable. Three things about *when* the teardown happens.

## 1. The purge runs before the refusal that can abort the delete

`_purge_personal_collections` deletes files and commits the vector-table drop through a separate session **before** the sole-owner-of-a-shared-organization checks below it. When either check raises `BadRequestError`, the request session rolls the relational deletions back — so the user and knowledge-base rows come back intact, pointing at files and a table that are already gone.

This is the simplest of the three and the most likely to be hit: it needs no concurrency at all, just a user who solely owns a shared organization. Moving the purge after the ownership checks is the whole fix.

## 2. An active ingestion recreates the table it just dropped

A worker that has already read a personal collection's tracking row and is still parsing is not cancelled by removing the relational rows. After the drop commits, its `insert_document` calls `_ensure_collection` and recreates the released `rag_<collection>` table — so the name is taken again, by a table holding one document from a deleted account.

## 3. Two concurrent deletes sharing a collection name drop nothing

`collection_name` is not tenant-unique (#913). When two users whose personal collections share one are deleted at the same time, each transaction removes its own row and still sees the other's pre-delete row under READ COMMITTED — so both take the "still referenced, skip the drop" branch. Both commit, no references remain, and neither dropped the table.

A leak rather than a correctness problem, but it is the exact inverse of the risk in #1269: there, a stale reference check drops a table it should not; here, two fresh ones each decline to drop a table nobody owns.

## Where

`backend/app/services/user.py` — `_release_owned_rows`, `_purge_personal_collections`, and their order relative to the ownership checks.

## How you would know it was fixed

1 is an integration case: delete a user who solely owns a shared organization, assert the refusal **and** that the personal collection's table and files are still there. It fails against the current order.

2 and 3 want the interleaving orchestrated rather than raced — the shape #17's tests use.

## Worth reading together

[...]

PR #1390, opened 2026-09-04
fix(rag): take the teardown lock before row locks in every path

### LINKED ISSUE #1387: rag: claim and org purge take the teardown and organizations locks in opposite order (ABBA deadlock)
### What breaks

`CollectionAccessService.claim` and `OrganizationService.purge` acquire the same two
locks in opposite order, so a knowledge-base create that names an existing collection,
racing a purge of that collection's organization, can deadlock.

- **claim / create** takes the `COLLECTION_TEARDOWN` advisory lock, then its caller
  inserts a `knowledge_bases` row whose `organization_id` FK takes `FOR KEY SHARE` on
  the `organizations` row. Order: **teardown lock → organizations row**.
- **purge** takes the `organizations` row `FOR UPDATE` first, then the per-collection
  `COLLECTION_TEARDOWN` lock. Order: **organizations row → teardown lock**.

`FOR UPDATE` conflicts with `FOR KEY SHARE`, so the two cross (ABBA):

- T1 = `purge(org X)`: holds `organizations[X]` FOR UPDATE, blocks on `teardown[foo]`.
- T2 = `POST /kb` (or `POST /rag/collections/foo`) naming existing collection `foo` in
  org X: holds `teardown[foo]`, its create INSERT blocks on `organizations[X]` FOR KEY
  SHARE.
- → PostgreSQL aborts one as a deadlock (40P01): a 500 on the create or the purge.

Reachable via org deletion and account deletion (`UserService.delete` → `purge` of the
personal org). Rare - it needs a create naming an *existing* collection concurrent with
that org's teardown - and self-resolving (the aborted side can retry), but it is a real
deadlock. **Pre-existing**: introduced when #1362 added the teardown lock to `claim`; it
predates #1378 and is not caused by it.

### Where

- `backend/app/services/collection_access.py:231` — `claim` takes `hold_name(...,
  COLLECTION_TEARDOWN, name)`; the create INSERT that FK-locks the org row is
  `backend/app/services/knowledge_base.py:185` (`create_for_rag_collection`) and `:356`
  (`create`), reached from `app/api/routes/v1/rag.py:168` and `knowledge_base.py:336`.
- `backend/app/services/organization.py:374` — `get_by_id_for_update` (`organizations`
  FOR UPDATE), then `:393` `hold_name(..., COLLECTION_TEARDOWN, collection)`.

### How you'd know it's fixed

One consistent lock order across every path that takes both locks - the teardown lock
acquired before any `organizations` (and `knowledge_bases`) row lock - so a concurrent
`claim`/create and `purge` of the same org cannot deadlock. The constraint that makes
this non-trivial: `purge` must hold the `organizations` row lock across its knowledge-
base deletes, because both `organization_id` FKs are `ON DELETE SET NULL` (not cascade),
so a naive reorder that drops the early org lock leaves a concurrently created org-scoped
base with a null org, violating its check constraint. The fix therefore needs a design
change - e.g. `purge` pre-scanning collections and taking their teardown locks before the
org lock, or the write path not holding a teardown lock across the org-FK insert (an
outbox/generation approach). See #1382 for the same lock-ordering analysis on the write
side.

Split from #1382, found reviewing #1378. Cluster: RAG teardown / "Ingestion and retrieval"
in the issue map #168 (siblings #1355, #1362, #1364, #1382).

OchnikBartek:
## What this fixes

`CollectionAccessService.claim` takes a collection's `COLLECTION_TEARDOWN`
advisory lock and then, on INSERT, a `FOR KEY SHARE` on the `organizations`
row. `OrganizationService.purge` took the `organizations` row `FOR UPDATE`
first and the teardown lock second. Opposite orders on the same two locks are
an **ABBA deadlock** (#1387): a KB create naming an existing collection, racing
a purge of that collection's org, aborts one side with `40P01`. Reachable via
org deletion and account deletion.

## The fix — one lock order everywhere

Every path that takes both now takes **the teardown lock before any
`knowledge_bases`/`organizations` row lock**:

- **`purge`** takes each doomed collection's teardown lock from a *snapshot*
  read (no row lock), **sorted**, before the org `FOR UPDATE`. The `FOR UPDATE`
  still precedes the authoritative list-and-delete, so #1115 (SET NULL + CHECK)
  holds unchanged — the snapshot only chooses which locks to take.
- **`delete`**, **`delete_for_rag_collection`**, **`_purge_personal_collections`**
  move `hold_name` above the KB row lock/delete. `_purge_personal` sorts its
  names so it and `purge` acquire any shared name (#913) in one order — no
  teardown-vs-teardown ABBA.
- **`DELETE /rag/collections/{name}`** now deletes the document rows *after*
  `delete_for_rag_collection` has taken the lock, so the drop no longer holds
  those rows while waiting on the teardown lock (this ordering is load-bearing —
  a comment says so).

## Verification

- Two **deterministic** integration tests against a real database: `purge`
  blocks on the teardown lock **before** taking the org row (`FOR UPDATE NOWAIT`
  probe of the org row succeeds while purge waits); the collection drop blocks on
  the teardown lock **before** deleting documents (`FOR UPDATE NOWAIT` probe of
  the doc row succeeds). Both would fail on the old order.
- An **adversarial lock-graph review** of every path pair (claim, purge, KB
  delete, rag-collection drop, personal-collection purge, two concurrent purges).
  It confirmed the #1387 deadlock and both round-1 regressions it caught in an
  earlier revision are closed, and #1115/#1116/#1266/#1362 all still hold.
- `make lint-backend` green; 149 unit tests green.

#1387's own note that this "needs a design change" was pessimistic — the pre-scan
is a contained fix, taken here.

## What is *not* closed (filed as #1389)

Two narrow, self-resolving residuals need a serialisation point (outbox), which
#1382 already weighed as disproportionate for this severity, so they are tracked
rather than fixed here:

- A collection created in the **snapshot→org-lock gap** is deleted under the org
  lock but not vector-reserved (its lock can't be taken after the org lock
  without reinverting the order); it can `D↔K`-deadlock with a concurrent
  delete/drop of that same just-created collection, and its commit-to-drop window
  is unreserved. Strictly **narrower** than before this change — which serialised
  no collections at teardown time.
- The pre-existing **users-dimension `U↔T`** inversion in `UserService.delete`
  (untouched here; the `users` FOR UPDATE predates #1364).

## Rebased onto `main` after #1378 merged

The branch was stacked on `fix/1364-teardown-sweep` (#1378) for the
default-KB clear in `delete_for_rag_collection`; that merged, so this now
carries the one commit on top of `main`. The only overlap was in
`KnowledgeBaseService.delete`, where `main` (#1273) had meanwhile also
placed `hold_name` before the KB row lock but kept the old second call
after the row delete. Both intents agree, so the merged version keeps the
#1273 explanation with the lock-order sentence appended, and this branch
removes the now-redundant second `hold_name`. Re-verified on the rebased
tree: both ordering tests pass against a real Postgres, alongside the
#1268 and #1273 tests `main` added to the same file.

Closes #1387
Refs #1382, #1389


OchnikBartek:
@codex review

DEENUU1:
Addressed in `63945d70`.

**Reserve collections discovered after the purge snapshot** — real, and a cross-tenant read rather than a storage residual, so not something to leave to #1389.

A collection created between the snapshot and the `FOR UPDATE` is found by the authoritative scan and deleted with the rest, but no teardown lock was ever taken for it, so it was dropped **unreserved**. Another organization claims the name in the commit-to-drop window; the deferred cleanup re-reads the reference check, sees the new row and *preserves* the table — and the claiming organization inherits a vector table still holding the deleted organization's vectors.

Locking it where it is found means taking a lock after the org row, which is exactly the ABBA deadlock the snapshot-first order exists to avoid. So `try_hold_name` (new, `pg_try_advisory_xact_lock`) takes it without ever waiting:

- **free** → held for the rest of the transaction, and the name reserves like any other;
- **held** → a claim for that very name is in flight, and nothing this transaction can still do makes the drop safe. It refuses with `ConcurrentChangeError` (409, retryable) and writes nothing, rather than dropping unreserved.

Six unit tests over the two readings: the late name is locked without waiting and reserved, a name the snapshot saw is not locked twice, a held lock refuses, nothing is dispatched when it refuses, a name another organization still references is left alone, and no store wired locks nothing. `make lint` clean, 7031 backend tests pass.

PR #1436, opened 2026-09-05
fix(channels): refuse a deactivated member's channel turn

### LINKED ISSUE #1427: Refuse a deactivated member's channel turn instead of running it at their old role
**Severity: High** · AUD-CHAN-01

## What's wrong

A linked channel sender whose account has been **deactivated** still runs agent turns at their full old authority — their role (up to Owner), their budget, and in a 1:1 DM their own MCP account.

## Where

- `backend/app/services/channels/mentions.py:559-569` — `_membership_context` derives the run's role from `member_repo.get` (the plain read), for every linked sender on the mention **and** default paths, on every platform and transport.
- `backend/app/repositories/member.py:37-49` — `get_active`'s docstring: *"anything reading a role off `get` alone answers with the authority of an account that is refused everywhere a person signs in … see `access.publisher_context`, which is **the one caller** and the reason this exists."*
- `backend/app/services/access.py:260-273` — `publisher_context` uses `get_active`, and names this exact case: *"a deactivated Owner's widget, hosted page **and channel binding** running turns at their full authority."*
- `backend/app/services/channels/router.py:259` — the sender gate checks only `identity.user_id is None`, never `User.is_active` (`bot.is_active` is checked at `:222`; the sender's is not).

## How it fails

Deactivating a user revokes their web sessions but does not clear `channel_identities.user_id` or the membership row — which is precisely why `get_active` exists. The binding path (`publisher_context`) was hardened to `get_active`; `_membership_context` is a **second** "nobody-signed-in" caller and was not. So an offboarded user (`is_active = False`), blocked everywhere they sign in, sends a message to a Telegram/Slack/Mattermost bot and the turn runs as them. Offboarding is a security boundary; this is the platform's own stated invariant (*"refused everywhere a person signs in"*) violated on the one surface that re-derives authority from a row that outlives deactivation.

**Not fixed by #1392** — its diff adds or removes no `member_repo.get(`/`get_active(` line in `mentions.py` (it renames `ctx→sender` and mocks `get_active` in tests only for `publisher_context`). It does rewrite the surrounding code, so land this **after #1392 merges**.

## Fix

`_membership_context` calls `member_repo.get_active`, so a deactivated linked member falls through to `admit_unlinked` — refused in a DM, run under the binding's publisher (floored at `VIEWER`) in a room — exactly as the widget and hosted-page paths already do. One line, with a sibling to copy.

## Acceptance criteria

- [ ] A deactivated user with a linked `channel_identity` sending a DM is refused (the link-first message), not run
- [ ] The same user in a room is admitted under the binding publisher at `VIEWER`, never at their old role
- [ ] Test: `test_a_deactivated_linked_member_does_not_run_a_channel_turn_at_their_old_role`

---
Cluster **Channels and bots** in #168. Nearest: #1392 (rewrites `mentions.py`; this must follow it), #639 (made the room/DM admit rule this fix falls back to). Pre-existing, found in the 2026-09-03 channels audit.

OchnikBartek:
## What changed

`ChannelAgentRouter._membership_context` now reads a linked sender's role off `member_repo.get_active` - the joined read that answers only while the account can still sign in - instead of the plain `member_repo.get`.

A linked channel sender whose account had been **deactivated** kept running agent turns at their old role (up to Owner), on every platform and on both the `@handle` and the default path. Deactivation revokes web sessions but leaves the membership row and the `channel_identities` link in place. The sibling **binding** path (`access.publisher_context`) was already moved to `get_active` for exactly this reason; the **sender** path was not.

A deactivated or departed linked sender now falls through to the unlinked rule: refused with the link-first message in a direct message, admitted under the binding's publisher in a room - never at their own old role.

Also: `get_active`'s docstring no longer claims `publisher_context` is its one caller; `docs/channels.md` and `docs/permissions.md` say what a deactivated linked sender is now.

## How it was verified

Three regression tests in `tests/test_channel_mentions.py` that **fail against the previous `mentions.py`** (checked by swapping the file back) and pass now:

- `test_a_deactivated_linked_member_does_not_run_a_channel_turn_at_their_old_role` - a DM mention from a sender whose `get` still answers Owner and whose `get_active` answers `None` is refused before the handle is looked up; the runner is never called.
- `test_a_deactivated_member_in_a_room_runs_under_the_binding_not_at_their_old_role` (x2) - the same sender in a room runs under the binding creator at the creator's role, and at `viewer` when the creator is gone too.

The remaining mention tests stub `get_active` for the sender, and the room fixture now raises if anything on the path reads the plain row.

- `uv run pytest tests/test_channel_mentions.py tests/test_channel_charts.py tests/test_surface_transcripts.py` - 106 passed
- `make lint-backend` clean (`ty` prints 60 pre-existing warnings in unrelated modules, exit 0)

## Notes

- **Sequencing with #1392:** it rewrites `mentions.py` (renames `ctx→sender`, removes `one_to_one`). This PR touches one production line there plus a docstring, so whichever lands second carries a one-line conflict. Not blocked on it.
- The refusal a deactivated sender sees is the same link-first message a stranger sees. It does not say "your account was deactivated" - a refusal must not tell a sender more about their standing than a stranger learns.

Closes #1427

### DIFF
--- a/backend/app/repositories/member.py
+++ b/backend/app/repositories/member.py
@@ -45,8 +45,9 @@ async def get_active(
     Deactivating a user leaves their membership row exactly where it was, so
     anything reading a role off `get` alone answers with the authority of an account
     that is refused everywhere a person signs in. That is only a difference on the
-    paths where nobody is signed in - see `access.publisher_context`, which is the
-    one caller and the reason this exists.
+    paths where nobody is signed in: `access.publisher_context`, the reason this
+    exists, and the channel router's `_membership_context`, which reads a linked
+    sender's own role off a chat account that stays linked after the deactivation.
 
     Joined rather than a second read: it is answered on every turn a public surface
     takes, and two round trips for one decision is one of them that can be true
--- a/backend/app/services/channels/mentions.py
+++ b/backend/app/services/channels/mentions.py
@@ -554,10 +554,14 @@ class ChannelAgentRouter:
         A linked account that is no longer a member takes the same path as an
         unlinked one rather than a third: there is no membership to read a role
         from, and in a room a former member is no more entitled than the stranger
-        beside them.
+        beside them. A deactivated account is one of those, which is why this is
+        the joined read: the membership row outlives a deactivation, and the link
+        in `channel_identities` outlives it too, so the plain read ran an
+        offboarded Owner's turns at their full authority from a chat account
+        nobody had unlinked - refused everywhere they sign in, except here.
         """
         if user_id is not None:
-            membership = await member_repo.get(
+            membership = await member_repo.get_active(
                 self.db, organization_id=organization_id, user_id=user_id
             )
             if membership is not None:
--- a/backend/tests/test_channel_charts.py
+++ b/backend/tests/test_channel_charts.py
@@ -227,7 +227,7 @@ class TestTheReplyAChannelTurnBuilds:
             patch("app.services.channels.mentions.agent_exposure_repo") as exposures,
             patch("app.services.channels.mentions.AgentRunnerService") as runner_cls,
         ):
-            members.get = AsyncMock(return_value=MagicMock(role=OrgRoleName.MEMBER))
+            members.get_active = AsyncMock(return_value=MagicMock(role=OrgRoleName.MEMBER))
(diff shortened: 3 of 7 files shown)


OchnikBartek:
@codex review

PR #1447, opened 2026-09-05
fix(channels): render a channel chart off the event loop

### LINKED ISSUE #1429: Render a channel chart off the event loop
**Severity: Medium** · AUD-CHAN-04

## What's wrong

The PNG chart a channel reply carries is rasterized with Pillow **synchronously on the event loop**, stalling every other concurrent channel turn on that worker for the duration of each chart.

## Where

- `backend/app/services/channels/mentions.py:373` and `:471` — `AnsweredTurn(..., image_png=drawn_chart(called), ...)` built synchronously inside the `async def` turn methods, right after `await self.runner.execute`
- `backend/app/services/channels/mentions.py:93-112` — `drawn_chart` calls `render_chart_png` at `:108`
- `backend/app/services/channels/chart_png.py:399-428` — the CPU-bound render, ending in `image.save(buffer, "PNG", optimize=True)` at `:427`

`rg 'to_thread|run_in_executor' app/services/channels/` → nothing.

## How it fails

Channel turns run inline in the polling task / the spawned webhook task, so a chart-bearing reply rasterizes and PNG-encodes (`optimize=True`) on the shared loop. Bounded canvas, so not seconds — but it is the "misconfigured bot starving the event loop" class (e76ac05c) on the happy path, and `mentions.py` is in the 100% gate, which does not see loop-blocking.

## Fix

`image_png=await asyncio.to_thread(drawn_chart, called)` at both sites.

## Acceptance criteria

- [ ] A chart reply does not block the loop (assert the render runs via `to_thread`, mirroring the #1100 worker test)
- [ ] Test: `test_a_channel_chart_is_rendered_off_the_event_loop`

---
Cluster **Channels and bots** in #168. Nearest: #1100 / #1294 (the same class, fixed elsewhere). Pre-existing, found in the 2026-09-03 channels audit.

OchnikBartek:
## What changed

The PNG a channel reply carries was rasterised and PNG-encoded by Pillow **synchronously on the event loop**: `AnsweredTurn(..., image_png=drawn_chart(called))` inside the `async def` turn methods in `app/services/channels/mentions.py`, right after `await self.runner.execute`. A channel turn runs on the loop every poller and webhook task on the worker shares, so each chart stalled every other channel turn for its duration - the same class as the upload parse (#25) and the worker's file hash (#1100), both already moved off the loop.

Both call sites (`answer` and `answer_default`) now read `image_png=await asyncio.to_thread(drawn_chart, called)`. `drawn_chart`'s docstring says it blocks and that callers reach it through a thread.

**Default executor, not the bounded file pool** (`app/core/blocking.run_blocking`): a chart is a fixed-size canvas, tens of milliseconds, and nothing about it queues large buffers the way an upload storm does - which is what that pool's admission gate exists for.

## How it was verified

`tests/test_channel_charts.py::TestTheReplyAChannelTurnBuilds::test_a_channel_chart_is_rendered_off_the_event_loop`, parametrised over both paths: the render is wrapped to record `threading.get_ident()` and the test asserts it is not the loop's thread - the shape `tests/test_blocking_io_offload.py` uses for the upload offload. **Fails with the `to_thread` removed.**

- `uv run pytest tests/test_channel_charts.py tests/test_channel_mentions.py tests/test_surface_transcripts.py` - 105 passed
- `ruff` / `ty` clean on the changed files; `mentions.py` stays in the 100% gate (no new branch)

## Notes

- No page describes where the render runs, so `docs/channels.md` owes nothing here (the drift hook names it because `mentions.py` moved).
- The test stubs both `member_repo.get` and `member_repo.get_active` for the sender, so it passes before and after #1436 (#1427) lands.
- #1392 rewrites `mentions.py`; this is two one-line hunks, trivial to carry across in either order.

Closes #1429

### DIFF
--- a/backend/app/services/channels/mentions.py
+++ b/backend/app/services/channels/mentions.py
@@ -38,6 +38,7 @@ handle typed into a public channel is a name, not a key.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import re
 from dataclasses import dataclass, field
@@ -97,6 +98,10 @@ def drawn_chart(called: list[RecordedToolCall]) -> bytes | None:
     a reply carries one image. A result that no longer parses is skipped rather
     than raised on: the payload is whatever the tool returned, and a chart that
     cannot be drawn must not cost somebody the answer it came with.
+
+    Blocking: Pillow rasterises and PNG-encodes here, and a channel turn runs on
+    the loop the pollers and the other webhook tasks share - so the callers reach
+    it through `asyncio.to_thread`, never inline.
     """
     for call in reversed(called):
         if call.tool_name != _CHART_TOOL or call.result is None:
@@ -370,7 +375,7 @@ class ChannelAgentRouter:
             ),
             attachments=produced,
             refused=refused,
-            image_png=drawn_chart(called),
+            image_png=await asyncio.to_thread(drawn_chart, called),
             awaiting_approval_run_id=(
                 run.id if run.status == RunStatus.AWAITING_APPROVAL else None
             ),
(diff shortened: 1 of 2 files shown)


OchnikBartek:
@codex review

PR #1450, opened 2026-09-05
fix(channels): one supervised reconnect loop for the three adapters

### LINKED ISSUE #1430: One supervised reconnect loop for the three adapters; Telegram retries a bad token forever
**Severity: Medium** · AUD-CHAN-05

## What's wrong

The "restart the inbound stream on failure" loop is hand-rolled separately in each of the three adapters, and the three disagree — and Telegram's has no config-error branch, so a bot with a bad token retries **forever**, every 5s, logging a traceback each time.

## Where

| | Slack `slack.py:408-440` | Telegram `telegram.py:266-280` | Mattermost `mattermost.py:603-641` |
|---|---|---|---|
| Backoff | fixed `sleep(5)` (`:440`) | fixed `sleep(5)` (`:280`) | exponential 5→60s (`:606,636,641`) |
| `sleep` vs `except` | outside (`:411-416`) | **inside** the `except` (`:280`) | outside (`:634-635`) |
| config-error branch | `record_down` + `return` (`:423-432`) | **absent** | `record_down` + `return` (`:613-624`) |

(all under `backend/app/services/channels/`)

## How it fails

Slack and Mattermost both carry a paragraph about a 100%-CPU spin from a branch that returns without awaiting, and both fix it by moving `sleep` outside the `except`. Telegram puts it inside and dodges the spin only because `dp.start_polling` always blocks. Concretely: a Telegram bot with an unfixable config error has no `ChannelNotConfigured` branch, so it loops the generic `except` every 5s indefinitely, re-recording `down` each time, where Slack/Mattermost record down once and stop. #565 lifted the transport boilerplate into `base.py`; the supervisor loop was never lifted, so the next reconnect fix must find and edit three places.

## Fix

One supervised-run helper (on the base, or in `supervisor.py`) taking the per-adapter session coroutine, a backoff policy and a config-error predicate; the three adapters call it.

## Acceptance criteria

- [ ] All three adapters share one reconnect loop with one backoff policy
- [ ] A Telegram bot with a bad token records `down` once and stops, like the other two
- [ ] Test: `test_a_telegram_config_error_records_down_and_stops_rather_than_looping`

---
Cluster **Channels and bots** in #168. Nearest: #565 (the lift this completes), #1351 (the `record_down` contract), #1392 (rewrites all three adapters — land after it). Pre-existing, found in the 2026-09-03 channels audit.

OchnikBartek:
## What changed

Each adapter carried its own "restart the inbound stream on failure" loop, and the three disagreed:

| | Slack | Telegram | Mattermost |
|---|---|---|---|
| Backoff | fixed 5s | fixed 5s | 5s → 60s |
| `sleep` vs `except` | outside | **inside** | outside |
| `ChannelNotConfigured` branch | records `down`, stops | **absent** | records `down`, stops |

So a Telegram bot whose token Telegram rejects was retried **for ever**, a traceback and a fresh `down` record every five seconds, where the other two stop after one. And any reconnect fix had to find and edit three places (#565 lifted the transport boilerplate into `base.py`; the loop was never lifted).

**`supervise_stream` in `app/services/channels/base.py` is the one loop now.** The three supervisor methods (`_socket_supervisor`, `_polling_supervisor`, `_supervise`) keep their names and call it with their session and their failing-row message. One policy:

- wait 5s, doubling to 60s, on a crash; a clean session resets the wait to 5s
- the wait is **outside** the `except` (a session that returns without awaiting must still yield - the 100% CPU spin)
- `ChannelNotConfigured` → `record_down(bot_id, exc.message)` once, log once, stop
- `CancelledError` propagates (`stop_polling` cancels and awaits the task)
- the log line before each wait names the delay it is about to wait

**Telegram's session** (`_run_polling_once`) raises `ChannelNotConfigured` for the `TelegramUnauthorizedError` a rejected token produces - that is what gives it the branch the other two had.

Docs: `docs/channels.md` "A bot that cannot start stops, rather than retrying" now names Telegram and the shared policy; the `channel-bot` skill tells a new adapter to run under `supervise_stream` rather than hand-roll a loop.

## How it was verified

`tests/test_channel_supervisor_yields.py` now holds all three adapters to the one contract:

- each adapter suspends between sessions that return immediately (the existing sleep-count tests, extended to Telegram)
- each adapter records `down` **once** and stops on `ChannelNotConfigured` - including `test_a_telegram_config_error_records_down_and_stops_rather_than_looping`
- `test_a_rejected_token_is_a_configuration_error_not_a_crash` - Telegram's 401 becomes `ChannelNotConfigured`
- `test_all_three_adapters_run_the_same_loop` - a fix in one place lands in all three
- backoff `[5, 10, 20]`, reset to 5 after a clean session, cap at 60; a cancelled session ends the loop cancelled

Against the **previous** Telegram supervisor (checked by loading `origin/main`'s `telegram.py`), the same configuration error gives 3 attempts and 3 `down` records in 3 sleeps - i.e. it never stops.

- 197 adapter/channel tests pass locally
- `make lint-backend` clean (`ty` warnings are pre-existing, none in these files)

## Notes

- **One behaviour change beyond Telegram:** Slack now backs off 5→60s on repeated crashes instead of a flat 5s. That is the Mattermost policy applied to all three, and it is what the issue asks for.
- Mattermost's "not configured" row message is now the exception's own text rather than a second fixed sentence - one message, written where the refusal is decided, as Slack already did. Review caught that the fixed sentence had carried a hint the exception did not ("or switch it to webhook mode"); the refusal names it now, and the test asserts it.
- Rebased on `main` after #1392 merged; no conflicts.

Closes #1430

### DIFF
--- a/.claude/skills/channel-bot/SKILL.md
+++ b/.claude/skills/channel-bot/SKILL.md
@@ -78,7 +78,11 @@ Access modes: `open`, `whitelist`, `jwt_linked`, `group_only`.
 2. Wire it into `router.py` so inbound reaches the agent and replies stream back.
    **Do not fork the agent pipeline** — one runner behind every surface is the whole
    design.
-3. Add a signature-verified webhook route and/or a polling entrypoint.
+3. Add a signature-verified webhook route and/or a polling entrypoint. A polling
+   entrypoint runs its session under `supervise_stream` from `base.py` — the one
+   reconnect loop — and raises `ChannelNotConfigured` for anything a retry cannot
+   fix (a missing token or URL, a token the platform rejects); do not hand-roll a
+   `while True` with its own sleep.
 4. Reuse the existing conversation/session model.
 5. Respect the per-group/per-thread concurrency controls already in the adapters.
 
--- a/backend/app/services/channels/base.py
+++ b/backend/app/services/channels/base.py
@@ -1,4 +1,7 @@
+import asyncio
+import logging
 from abc import ABC, abstractmethod
+from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import Any, Literal
 
@@ -9,6 +12,10 @@ from app.agents.capabilities.channel_tools import (
     ChannelPost,
     ChannelSummary,
 )
+from app.services.channels import connection_state
+from app.services.channels.exceptions import ChannelNotConfigured
+
+logger = logging.getLogger(__name__)
 
 # Handles every chat platform reserves for addressing the room, rather than one
 # member of it. They match the shape of an agent slug, so `@channel deploying at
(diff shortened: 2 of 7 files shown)


OchnikBartek:
@codex review

Issue #1434, opened 2026-09-05
Channels hygiene: seven small items from the 2026-09-03 audit


OchnikBartek:
**Severity: Low** · AUD-CHAN-08 · AUD-CHAN-13 · AUD-CHAN-14 · batch

Seven small items from the 2026-09-03 channels audit, batched so they do not bury the issues that matter. Each is independent and roughly sub-hour. Most touch files #1392 rewrites — take them after it merges.

- [ ] **Slack fetches a payload-supplied URL with redirects, unvalidated.** `download_attachment` fetches `attachment.handle` = `url_private_download`/`url_private` straight from the (signed) event payload (`slack.py:693-701`) at `:715-717`, on a client built `follow_redirects=True` (`:83`), sending `Authorization: Bearer {bot_token}`, with no `app/core/sanitize.py` validation (`rg sanitize app/services/channels/` → nothing). Exploitation needs the signing secret and httpx strips `Authorization` on cross-host redirects, so low — but it is the one inbound-payload URL fetched unvalidated. Pin the host via `resolve_pinned_url`. Test: `test_a_slack_download_url_off_slacks_domain_is_refused`.
- [ ] **`router.py` threads its domain objects as `Any`.** 43 occurrences — `db: Any` (→ `AsyncSession`), `bot: Any` (→ `ChannelBot`), `identity: Any`, `session: Any` — at `router.py:175,207,388,440-443,958,980,1147-1150,1223`. The repositories it calls are concretely typed, so these recover under a `TYPE_CHECKING` import. Passes CI only because `services/channels/**` is outside the strict-`ty` gate.
- [ ] **`list_platforms()` has no consumers.** `channels/__init__.py:19,42-44` — defined and in `__all__`; `rg 'list_platforms' app/ tests/` returns only the definition and the export, which is what shields it from vulture. Delete it.
- [ ] **HTTP-client lifecycle diverges three ways (#952 half-applied).** Mattermost reuses one `httpx.AsyncClient` (`mattermost.py:114,148-149`); Slack reuses `_http` for downloads but builds a fresh `AsyncWebClient` per send (`slack.py:83` vs `:97-108`); Telegram builds a fresh `aiogram.Bot`+session every call (`telegram.py:51-68`, documented — aiogram leaks the session otherwise). Record the decision on the base class rather than three ad-hoc answers.
[...]

Issue #952, opened 2026-08-19
One HTTP client per channel adapter, not one per call


DEENUU1:
Every helper in the channel adapters opens its own `httpx.AsyncClient`, so every
call pays a fresh TCP connection and TLS handshake against a host it just talked
to.

**Severity: Medium** · AUD-006, audit of 2026-08-19

## Where

- `backend/app/services/channels/mattermost.py` — ten sites: `:144`, `:186`,
  `:208`, `:287`, `:342`, `:386`, `:408`, `:449`, `:550`, `:792`
- `backend/app/services/channels/slack.py:521`
- `backend/app/agents/capabilities/web_research/_search.py:175`, `:195`
- `backend/app/services/model_catalog.py:243`

Each is `async with httpx.AsyncClient(timeout=...) as client:`. The timeout is
right and the cleanup is right; what is missing is reuse — the connection pool
httpx maintains is discarded before it is ever used twice.

## How it fails

A streamed channel turn is not one call. `LiveReply` pushes an edit every
`EDIT_INTERVAL = 1.0` seconds, so a thirty-second answer is roughly thirty
`update_reply` calls, plus `typing`, plus `begin_reply`, plus the final
`_deliver` edit, plus an attachment download per file. Each one is a new
handshake against the same host.

At a typical 60-120 ms handshake to a remote Mattermost or Slack, a one-minute
turn spends several seconds of its wall clock re-establishing connections it
already had — and the bot host sees the socket churn of a client that never keeps
one open.

## Fix

One client per adapter, built when the adapter registers and closed at shutdown.
The adapters already have a lifecycle (`start_polling` / `stop_polling`,
registered in `main.py`).

```python
class MattermostAdapter:
    def __init__(self) -> None:
        self._http = httpx.AsyncClient(timeout=_HTTP_TIMEOUT)

    async def aclose(self) -> None:
        await self._http.aclose()
```

`web_research/_search.py` wants the same for a different reason: a search tool
called three times in one run pays three handshakes to the same provider.

## Acceptance criteria

- [ ] No `httpx.AsyncClient(...)` is constructed per call in the channel adapters
- [ ] Each client is closed on shutdown, where the adapters are already stopped
- [ ] Regression test: two consecutive sends reuse one client instance

OchnikBartek:
Plan and sequencing, because this overlaps #565 (PR #1173).

[...]

PR #1458, opened 2026-09-05
refactor(channels): seven hygiene items from the 2026-09-03 audit

OchnikBartek:
## What changed

The seven items of #1434, batched as the issue asks. Each is independent; the commit body has a paragraph per item.

- [x] **Slack sends the bot token only to Slack.** `download_attachment` fetched `url_private_download` straight from the (signed) event payload with `Authorization: Bearer` and no check on the host. The host is now checked against `slack.com` / `slack-files.com` over TLS before anything is sent; anything else is refused with the client untouched. Second lock, not the first - but a token posted to a host somebody else named is a token gone.
- [x] **`router.py` names its domain objects.** `db`/`bot`/`identity`/`session` were `Any` 43 times; now `AsyncSession`/`ChannelBot`/`ChannelIdentity`/`ChannelSession` under `TYPE_CHECKING`, and `_resolve_identity`/`_resolve_session` say what they return. First thing the checker found: `/start` read `bot.welcome_message` behind a `hasattr`, and no model, schema, page or test has ever had that field - dead branch, removed.
- [x] **`list_platforms()` deleted** - defined, exported, called by nothing.
- [x] **The HTTP-client decision is written once**, on `ChannelAdapter`: one `httpx.AsyncClient` per adapter for the adapter's own REST (#952); a vendor SDK object per call, closed on the way out (aiogram's `Bot`, slack-sdk's `AsyncWebClient`); never a client per message.
- [x] **`prepare_connection` replaces the `getattr` duck-typing** in `supervisor.py`. A no-op hook on the base, called on every adapter the same way; Mattermost keeps the server, Slack the app token, Telegram nothing. The webhook route's direct `remember_server` stays (it already narrows to `MattermostAdapter`).
- [x] **Telegram's secret compare uses `encode_untrusted`**, as Slack and Mattermost do. Safe today (Starlette decodes headers as latin-1); the same defence in depth.
- [ ] **Two duplicated router shapes - half done.** `_message` and `_post_placeholder` are now the one copy of the `OutgoingMessage` and the guarded `begin_reply` that `_lazy_reply`, `_open_reply` and `_deliver` each built. **Left as is:** the try/`AppException`/`Exception` shape `_answer_mention` and the default path share. Unifying it changes what a crash on the mention path does (today it propagates to `route` and releases the dedupe claim; the default path apologises) - a behaviour decision, so its own change: **#1459**, split out on review rather than left as a note here.

## How it was verified

- `test_a_slack_download_url_off_slacks_domain_is_refused` (4 shapes: another host, plain http, a look-alike suffix, a look-alike name) - refused, `client.get` never awaited. The two existing download tests moved to a real `files.slack.com` URL.
- `TestOpening` rewritten for the hook (order: prepare → stop → start; the exact kwargs), plus `TestWhatEachAdapterPrepares` - each adapter keeps only its own fact; Telegram keeps nothing; `None`s keep nothing.
- `test_a_lone_surrogate_telegram_secret_is_refused_rather_than_crashing` + the right secret still matches.
- Live-reply suite over the unified builders; 519 channel tests pass.
- `make lint-backend` clean; `ty` on `router.py` is down to the one pre-existing warning (`get_for_inbound` str/UUID). `make check` - see CI.

## Notes

- `docs/channels.md` owes nothing: none of this changes behaviour a page describes (the Slack host check refuses only what Slack never sends). The `channel-bot` skill's "Adding an adapter" names `prepare_connection`.
- The two observations the issue lists as "not filed" (`/new` attribution in a group; a session whose conversation was hard-deleted) are untouched here.

Part-of #1434
Refs #1459

### DIFF
--- a/.claude/skills/channel-bot/SKILL.md
+++ b/.claude/skills/channel-bot/SKILL.md
@@ -75,6 +75,9 @@ Access modes: `open`, `whitelist`, `jwt_linked`, `group_only`.
 
 1. Implement `services/channels/<platform>.py` against `base.py`: parse inbound into
    the normalized message, send outbound.
+   Anything per bot the connection needs before it opens - a server address, an
+   app token - is registered in `prepare_connection`, the base hook the supervisor
+   calls on every adapter; do not add a differently named method for it.
 2. Wire it into `router.py` so inbound reaches the agent and replies stream back.
    **Do not fork the agent pipeline** — one runner behind every surface is the whole
    design.
--- a/backend/app/services/channels/__init__.py
+++ b/backend/app/services/channels/__init__.py
@@ -16,7 +16,6 @@ __all__ = [
     "SECRET_MINTED_BY_US",
     "get_adapter",
     "inbound_webhook_url",
-    "list_platforms",
     "register_adapter",
 ]
 
@@ -37,8 +36,3 @@ def get_adapter(platform: str) -> "ChannelAdapter":
     if platform not in _adapters:
         raise KeyError(f"No channel adapter registered for platform '{platform}'")
     return _adapters[platform]
-
-
-def list_platforms() -> list[str]:
-    """Return a list of all registered platform names."""
-    return list(_adapters.keys())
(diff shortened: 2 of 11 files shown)


OchnikBartek:
@codex review

PR #1463, opened 2026-09-05
chore: cut 0.0.363

DEENUU1:
Release commit for **0.0.363**. Bumps `backend/pyproject.toml`, `backend/uv.lock` and `frontend/package.json`, and adds the CHANGELOG entry.

The release is #1430, merged as #1450. Each adapter carried its own reconnect loop and the three disagreed — a fixed five seconds against a 5s-to-60s backoff, the sleep inside the `except` in one and outside it in the others, and a stop-on-misconfiguration branch in two of the three. A Telegram bot whose token Telegram rejects was therefore retried for ever, logging a traceback and writing a fresh `down` record every five seconds, where the same bot on Slack or Mattermost recorded `down` once and stopped. One supervised loop serves all three now.

**It also corrects the previous release's notes.** The 0.0.362 release commit was cut before #1454 merged and squashed in after it, so `v0.0.362`'s tree already carries the `jwt_linked` fix while its notes credited only #1390 — and this section originally announced that fix as new, which would have told anyone reading the notes to upgrade for something they already had. The entry moves to 0.0.362 and the published `v0.0.362` notes are amended to match. Caught by the reviewer on this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@ Two things are versioned separately from this file and worth knowing about:
 
 ## [Unreleased]
 
+## [0.0.363] - 2026-09-05
+
+### Fixed
+
+- **A bot saved as `jwt_linked` admitted senders with no linked account.** The
+  mode decided nothing on its own: with `require_link` off, which is the default,
+  it admitted an unlinked room sender under the binding's creator exactly as
+  `open` did - the access check enforced only `whitelist` and `group_only`, and
+  the one place that read the mode required both switches. An operator who picks
+  a mode named for a linked account has asked for one, so the mode requires it.
+
 ## [0.0.362] - 2026-09-05
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


DEENUU1:
Right, and thank you — this was a bookkeeping error, not a wording one.

The 0.0.362 release commit was cut from `main` **before** #1454 merged and squash-merged **after** it, so `v0.0.362` (7dfdb09) carries `_admits_unlinked` already rejecting `jwt_linked` while its notes credit only #1390. Announcing it here would have told anyone reading the notes to upgrade for something their tree already had.

Fixed in `e9379fef`: the entry moves under 0.0.362, the published `v0.0.362` notes are amended to match, and 0.0.363 now documents #1450 — the one supervised reconnect loop — which landed after that tag and is genuinely unreleased.

The cause was cutting a release and merging the next feature concurrently. Not doing that again for the rest of this batch: the release merges before the next feature does.

PR #1514, opened 2026-09-07
perf(dashboard): read a window's slices in one scan, not five

### LINKED ISSUE #949: Collapse the dashboard's fifteen sequential aggregates into a handful
`StatsService.usage` awaits fifteen repository calls one after another. Ten of
them are independent aggregates over the same window of the same table with the
same WHERE clause.

**Severity: Medium** · AUD-003, audit of 2026-08-19

## Where

- `backend/app/services/stats.py:171-332` — `StatsService.usage`
- `backend/app/repositories/agent_run.py:893` — `_window_conditions`
- `frontend/src/hooks/use-usage-stats.ts`, `frontend/src/lib/query-freshness.ts:17`

The ten: `count_runs` twice (window and previous window), `runs_by_day`,
`runs_by_dimension` three times (surface, status, model), `runs_by_agent`,
`sum_cost_window` twice, `cost_by_provider_window`, `count_distinct_users`,
`latency_percentiles_ms`. No `asyncio.gather` and no `TaskGroup` anywhere in the
module.

The endpoint's own docstring has the right instinct one layer up — *"One composed
response on purpose: the cards reading it share one query, one loading state and
one failure, instead of eight half-loaded answers drifting apart"* — and then
issues fifteen queries to build it.

## How it fails

Not a failure, a cost, and it compounds three ways.

1. `(organization_id, started_at)` **is** indexed (`ix_agent_runs_org_started`),
   so these are range scans rather than table scans — which is why this is Medium
   and not High. But most read columns the index does not cover (`cost_usd`,
   `status`, `surface`, `model_label`, the duration pair), so each is a scan plus
   a heap fetch per row: ten passes over the same rows for one card.
2. They are **serial**. Fifteen round trips on one connection, so the response
   time is fifteen latencies plus fifteen scans, not the slowest of them.
3. **Four widgets call this endpoint per dashboard load** — `useUsageStats`,
   `useVersionUsage`, `usePeopleUsage`, `useUsageByHour` — on top of the nine
   queries in `use-dashboard-data.ts`. And `DASHBOARD_FRESHNESS` sets
   `staleTime: 0` with `refetchOnWindowFocus: true`, for a good reason it
   documents, so every alt-tab back to the dashboard repeats the whole burst.

## Fix — fewer queries, not `gather`

`AsyncSession` is not concurrency-safe, and this repo has paid for learning that
(#169: two gated tool calls in one model step raced the request's session and
took the parent run row down with it). Running these concurrently means separate
sessions on separate connections, which is a bigger change than it looks.

Most of them are one scan with conditional aggregates:

```sql
-- one pass instead of six: totals, statuses, cost, latency
SELECT
  count(*)                                                 AS runs,
  count(*) FILTER (WHERE status = 'completed')             AS completed,
  sum(cost_usd)                                            AS cost_usd,
  percentile_cont(0.5)  WITHIN GROUP (ORDER BY duration_ms) AS p50,
  percentile_cont(0.95) WITHIN GROUP (ORDER BY duration_ms) AS p95
FROM agent_runs WHERE <the window>;

-- and one GROUPING SETS pass for the slices that need a key
SELECT surface, status, model_label, agent_id, date_trunc('day', started_at) AS day, count(*)
FROM agent_runs WHERE <the window>
GROUP BY GROUPING SETS ((surface), (status), (model_label), (agent_id), (day));
```

Fifteen becomes three or four: the composed pass, the grouping-sets pass, the
previous-window pass, and the ingestion-spend pass (a different table).
Recommended over concurrency because it is also less code, and because the
previous-window comparison is the only genuinely separate question in the set.

Worth an `EXPLAIN ANALYZE` against a populated database before sizing the work —
the ten-scans claim is read from the code, not measured.

## Acceptance criteria

- [ ] `/stats/usage` issues no more than four queries for the composed response
- [ ] The response is byte-identical for a fixture window — this is a refactor,
      and the test is the existing suite passing unchanged
- [ ] Regression test: a query-count assertion, so the next dimension added does
      not silently become the sixteenth round trip

OchnikBartek:
`StatsService.usage` composed the dashboard's window from a run of aggregate
queries. #1187 already collapsed the window's scalars (count, cost, distinct
users, latency) into one `window_aggregates` scan; this collapses the five that
were left — the day buckets, the surface, status and model splits, and the
provider spend — into one `GROUPING SETS` scan over the same rows.

## What changed

- New `agent_run_repo.window_breakdown`: one query whose `GROUP BY GROUPING SETS`
  returns the day, surface, status, model and provider slices together. A
  `GROUPING` flag per set tags each row, so a genuine `NULL` surface or model is
  told apart from a row that is not that slice at all. Each slice is returned in
  the order its own query gave it: days ascending, dimension counts largest
  first, the provider bill biggest first.
- `runs_by_day`, `runs_by_dimension` and `cost_by_provider_window` are removed —
  each was called only by `usage`, and `window_breakdown` replaces all five
  calls.
- `StatsService.usage` now reads its keyed slices from that one call. The
  composed response issues four `agent_runs` queries instead of eight:
  `window_aggregates`, `window_totals`, `window_breakdown` and `runs_by_agent`
  (which stays separate — it counts delegations and joins the agent's name, a
  different WHERE the others do not share).

## Verified

- The existing `test_usage_stats_sql.py` suite passes unchanged against a real
  database — this is a refactor, so the byte-identical output is the test. Its
  direct calls to the removed functions now read the matching `window_breakdown`
  slice.
- New `test_a_real_null_lands_in_its_own_slice_not_another`: a run with a `NULL`
  model label lands in the model slice, not misread as a surface row missing its
  surface — the property `GROUPING` buys.
- New `test_the_composed_response_reads_no_more_than_four_run_queries`: a listener
  counts the `agent_runs` statements one `usage` call issues and asserts four, so
  the next dimension added cannot silently become a ninth round trip.
- `make check` green.

Closes #949

### DIFF
--- a/backend/app/repositories/agent_run.py
+++ b/backend/app/repositories/agent_run.py
@@ -4,7 +4,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import UTC, date, datetime
 from decimal import Decimal
-from typing import Any, Literal, cast
+from typing import Any, cast
 from uuid import UUID
 
 from sqlalchemy import ColumnElement, and_, case, func, or_, select, tuple_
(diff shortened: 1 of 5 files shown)


OchnikBartek:
@codex review

PR #1515, opened 2026-09-07
fix(onboarding): keep the Agents "?" walk offering create-agent

### LINKED ISSUE #910: Onboarding: the Agents "?" walk stops offering create-agent once the org has any agent
## What happens

After the first-run onboarding, the "Shall we build an agent together?" offer is
suppressed when the organization already has at least one agent. That part is
intended (#624, commit `754bfca29`) — offering "your first agent" to an org that
has six makes no sense.

But the suppression went too far: clicking **"?"** in the **Agents** section —
the explicit help walk, where asking to create one is the whole point — no longer
shows the offer either, in any org that has an agent. In **Skills** and **KB** the
offer still appears (only `create-agent` is gated by count).

## Repro

1. Any org that has at least one agent (e.g. the seeded org).
2. Go to Agents, click **"?"**.
3. Walk through the steps → the "create an agent?" offer never appears.
   (The same walk in Skills / KB → offer appears.)

## Root cause

`frontend/src/components/onboarding/creation-offer.tsx` gates the offer purely on
agent count, blind to where the offer came from:

```ts
offer === "create-agent" && agents.total > 0  →  return null
```

Both the first-run tour ending and the Agents "?" walk end by setting
`offer = "create-agent"` (`flowForPage(ROUTES.AGENTS)` in
`frontend/src/lib/onboarding/flows.ts`), so the gate suppresses both. This
contradicts the code's own docstring in `flows.ts` ("an Agents '?' that walked
into the builder still offers create-agent") and the #624 commit message
("anyone else reaches the same flow from the Agents '?' walk, where asking for it
is the point").

## Direction (to think about)

- **Distinguish the source:** the explicit "?" walk always offers; only the
  automatic first-run-tour ending is gated.
- **Better count condition:** we always create a **getting-started** agent, so
  gate on "at least one agent *other than* getting-started", not `total > 0`.
  After the initial onboarding and clicking through the panels, offering to create
  a real agent is still fine when the only agent is getting-started.

## Done when

Clicking "?" in Agents in a seeded org (getting-started + others) shows the
create-agent offer again, while the first-run tour ending stays suppressed for an
org that already has a real agent.

---

Placeholder for later consideration (per discussion with @DEENUU1). Found while
testing onboarding; not caused by any current branch. Refs #624.

OchnikBartek:
After the first-run tour, the "Shall we build an agent together?" offer is
suppressed for an organization that already has an agent (#624) — right for the
tour, which has just walked the reader through an existing agent's builder.

But the suppression was blind to where the offer came from, so it also swallowed
the offer at the end of the **Agents "?" walk**, the explicit help walk where
asking to build one is the whole point. In an org with any agent, clicking "?"
in Agents and walking to the end showed no create-agent offer; the same walk in
Skills and KB still did.

## Fix

`creation-offer.tsx` gated `create-agent` purely on the agent count. The store
already records how the walk was opened — `mode === "tour"` for the first-run
walkthrough, `mode === "page"` for a section's "?" replay — and `close`/`openOffer`
preserve it, so at offer time it still says which walk produced the offer. The
count gate now applies only when `mode === "tour"`; the "?" walk (`mode === "page"`)
offers regardless, matching the flow's own docstring ("an Agents '?' that walked
into the builder still offers create-agent").

## Verified

- New test: the Agents "?" walk (`mode: "page"`) offers create-agent even in an
  org with six agents.
- Existing tests, now explicit about mode: the first-run tour ending
  (`mode: "tour"`) stays suppressed for an org with agents and still offers for an
  org with none.
- `make lint-frontend`, `make test-frontend-cov`, `make build-frontend` green.

## Noted, not changed

The offer's copy is "Create your first agent?", the tour's framing. Shown from the
"?" walk in an org that already has agents, "first" reads slightly off. Left as-is
— the issue scopes the fix to distinguishing the source, not the copy — and worth a
follow-up if the wording matters.

Closes #910

### DIFF
--- a/frontend/src/components/onboarding/creation-offer.integration.test.tsx
+++ b/frontend/src/components/onboarding/creation-offer.integration.test.tsx
@@ -34,7 +34,14 @@ function renderOffer(
 
 beforeEach(() => {
   rig.can = () => true;
-  useOnboardingStore.setState({ isOpen: false, index: 0, mode: "tour", flowId: null, offer: null });
+  useOnboardingStore.setState({
+    isOpen: false,
+    index: 0,
+    mode: "tour",
+    flowId: null,
+    offer: null,
+    offerFromTour: false,
+  });
 });
 
 describe("CreationOffer", () => {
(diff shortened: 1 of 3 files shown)


OchnikBartek:
@codex review

Issue #1439, opened 2026-09-05
Changing a password from Settings does not revoke the account's other sessions


OchnikBartek:
Pre-existing; found reviewing #1044 (PR #1435). Same cluster as #943 and #1044 in #168 (governance / admin); the sessions machinery is #1337's neighbourhood.

## What breaks

Changing a password from Settings → Account goes through `PATCH /users/me` (`UserService.update`, `backend/app/services/user.py`, the `"password" in update_data` branch). It hashes and stores the new password and stops. Only the *email* reset flow (`confirm_password_reset`, same file) calls `session_repo.deactivate_all_user_sessions`.

So a person who changes their password because they believe somebody else has access leaves every existing session valid - every other browser's refresh token, and since #1044 every impersonation row too. `docs/deployment.md` and the impersonation docstrings had to say "resets their password *by email*" to stay true, which is the tell.

## Sequence

1. Sign in as B in two browsers.
2. In one, change the password from Settings.
3. The other browser keeps working - its next refresh is honoured.

## Fixed when

A password change from Settings deactivates every other session of the account (keeping the one that made the change), with a test; the docs can then say "changes their password" without the qualifier.

PR #1518, opened 2026-09-08
feat(auth): self-service password change that revokes sessions and versions credentials

### LINKED ISSUE #1517: Self-service password change has no working endpoint, so #1439's session revocation never runs
**Severity: High** · found reviewing #1439 (PR #1498)

#1439 revokes the account's other sessions when a password changes, wired into
`UserService.update` and reached through `PATCH /users/me`. Two gaps mean it does
not yet protect the action it is meant to:

## 1. The real password-change flow has no endpoint, so the revocation never runs

`frontend/src/app/[locale]/(dashboard)/settings/account/page.tsx:49` posts to
`/auth/password/change`:

```ts
await apiClient.post("/auth/password/change", { current_password, new_password });
```

There is **no handler** for it — `frontend/src/app/api/auth/` has only
`password-reset/*`, and the backend has no `password/change` route. The page even
carries a `passwordChangeRequiresBackend` fallback toast, so the change is a known
stub: a user changing their password from Settings gets a 404, and the
`deactivate_all_user_sessions` call in `app/services/user.py:355-360` never
executes for the one action a person takes when they believe they are compromised.

`PATCH /users/me` does carry `password` (`UserUpdate.password`,
`app/schemas/user.py:49`), so the revocation is reachable — but the UI never calls
it for a password change, and `UserUpdate` has **no `current_password`**, so that
path changes a password without proving the caller knows the old one. Pointing the
Settings form at it would ship a password change a stolen access token can make
freely, which is why the dedicated endpoint the form already targets is the right
home.

## 2. The revocation races refresh-token rotation

Even reached, `deactivate_all_user_sessions` and a concurrent `/auth/refresh` are
not serialized. A stolen refresh token being rotated can validate its old session
row before the bulk update and insert its replacement after the update's statement
snapshot, so the replacement is never deactivated and the attacker keeps a valid
token past the completed password change (`app/services/user.py:355-360` vs the
refresh path in `app/api/routes/v1/auth.py`). Closing it wants a credential/token
version bumped on a password change and checked at refresh, or an equivalent
serialization — not just the bulk deactivate.

## Fix (a subsystem, to design)

- A dedicated `POST /auth/password/change` that verifies `current_password`,
  changes the hash, and revokes the account's other sessions — the flow the
  Settings form already posts to.
- A credential version (or equivalent) so a rotation in flight during that change
  cannot mint a surviving token.
- Wire #1439's revocation into it, and keep the `PATCH /users/me` revocation for
  the admin-reset path.

## Done when

Changing a password from Settings succeeds, revokes every other session including
one whose refresh is racing the change, and refuses a wrong `current_password`.

---

Cluster **Auth and sessions** in #168. Found reviewing #1439 (PR #1498); the
missing endpoint predates it, the revocation it would run is #1439's.

OchnikBartek:
> **Stacked on #1498 (#1439).** The base is `fix/1439-revoke-sessions-on-password-change`,
> so this diff is only #1517's commits; it retargets to `main` once #1498 merges.
> #1517's revocation lives in `UserService.update`, which #1498 adds, so this
> cannot stand alone.

#1439 revokes the account's other sessions on a password change, but two gaps
meant it protected nothing in practice. This closes both, and the review's
findings on top.

## 1. A working password-change endpoint

The Settings account form has always posted to `/auth/password/change`, and no
handler existed — it 404'd behind a "requires backend" toast, so the one action a
person takes when they think they are compromised did nothing, and #1439's
revocation was never reached.

`POST /auth/password/change` proves the current password — which `PATCH /users/me`
cannot ask for — under a lock on the user row, then changes it. It revokes **every**
session (the caller's own included, since the version gate below would refuse a
spared session's stale token anyway) and opens a fresh session at the new version,
returning its tokens; the BFF swaps the cookies to them. The caller stays signed
in while every other device is logged out, and `PATCH /users/me` now **refuses** a
password so it cannot be used as a bypass.

## 2. Closing the revocation-vs-rotation race

A refresh rotating a stolen token could read its session as active before the
deactivate and insert its replacement after. A `credential_version` on the user,
bumped on every password change under the row lock and stamped into each refresh
token as `cv`, closes it: the refresh path refuses a token whose `cv` is behind
the account's. A token minted before the claim reads as 0, so nothing is logged
out on deploy until the first change.

## Verified

- Service: refuses a wrong current password and an OAuth-only account with none,
  proves against the *locked* row, revokes every session, and bumps the version;
  a profile edit does neither and `/users/me` refuses a password.
- Route: answers 200 with a fresh refresh token carrying the new version, mints
  no session on a wrong password, and rate-limits by account as well as address.
- Refresh: refuses a token below the version; one at it rotates and carries the
  version forward.
- Frontend: the BFF swaps both cookies; a password-change 401 does not refresh and
  retry; the proxy-coverage list has no stale entry.
- Migration verified against a real Postgres — applies over the chain, `alembic
  check` sees no model drift, and it reverses. `make check` green.

Closes #1517

### DIFF
--- a/backend/alembic/versions/0075_user_credential_version.py
+++ b/backend/alembic/versions/0075_user_credential_version.py
@@ -1,0 +1,41 @@
+"""A credential version on the user, bumped when a password changes.
+
+A password change revokes the account's other sessions (#1439), but a refresh
+racing that revocation could read its old session as still valid and mint a
+replacement the bulk deactivate never saw - so the stolen refresh token outlived
+the change (#1517). This column is the version a refresh token is minted with,
+and the refresh path refuses a token whose version is behind the user's: a change
+that bumps it invalidates every token issued before it, whatever the race did to
+the session rows.
+
+`0` for every existing row, and a token minted before this carries no version and
+reads as `0` - so nothing is logged out on deploy, until the first password change
+moves that user to `1` and its earlier tokens stop refreshing.
+
+Revision ID: 0075_user_credential_version
+Revises: 0074_message_search_vector
+Create Date: 2026-09-08
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "0075_user_credential_version"
+down_revision: str | None = "0074_message_search_vector"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "users",
+        sa.Column("credential_version", sa.Integer(), nullable=False, server_default="0"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("users", "credential_version")
--- a/backend/app/api/routes/v1/auth.py
+++ b/backend/app/api/routes/v1/auth.py
@@ -20,6 +20,7 @@ from app.core.exceptions import AuthenticationError
 from app.core.security import (
     create_access_token,
     create_refresh_token,
+    verify_token,
 )
 from app.schemas.password_reset import (
     MagicLinkRequest,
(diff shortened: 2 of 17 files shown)


OchnikBartek:
@codex review

PR #1491, opened 2026-09-07
fix(impersonation): refuse identity binding while impersonating

OchnikBartek:
## What this fixes

Two routes bind an external identity to the *current* account, and neither knew whether that account was being impersonated (#1438):

- **`POST /me/channel-link/{token}`** confirms a chat-link code and attaches the confirmer's Telegram / Slack / Mattermost identity to the signed-in user.
- **Personal MCP OAuth** (`POST /me/mcp-connections/oauth/start`) begins an authorization-code flow that stores the consenting person's grant as their connection.

While an app admin acts as B, either one fastened the **administrator's own** chat account or OAuth grant onto B's account. That binding outlived the impersonation's hour, was recorded against nobody who consented, and broke the flow's own promise that it ends when the session row ends (#1044).

## The fix

- New `refuse_binding_while_impersonating(action)` in `app/services/impersonation.py`: raises `AuthorizationError` (403) when `current_impersonation()` is set, with a message naming the action.
- Called at the two binding chokepoints — `ChannelLinkService.confirm` and `McpConnectionService.oauth_start` — **before** the token is read or the OAuth client is registered, so nothing is spent and no pending row is written.

The MCP OAuth **callback** carries no session (it is authenticated by the `state` token so the provider redirect can reach it), so the personal *start* is the only seam where the impersonation context is visible. The connection's owner is fixed at start time, so blocking the start closes the flow — the callback can never find a valid state.

## Not in scope (filed as #1490)

The **organization**-scoped OAuth starts (`oauth_start_for_org`, and the GitHub / portal variants on a different chokepoint, `_stage_pending_oauth`) take the same class of action but bind to the *organization*, not to the impersonated person — recording `created_by = ctx.subject_id`. Whether that should refuse or instead record the impersonator is a separate governance decision, so it is filed rather than folded in.

**Manual token entry is deliberately out of scope.** `POST /me/mcp-connections` and the `PATCH` both let an admin type a bearer token that would store under the impersonated member — but a manually typed token is opaque configuration an admin may legitimately enter *as* the member, not the admin's own captured identity the way an OAuth grant or a chat account is. The residual (a member's connection backed by the admin's token) is filed as #1492 for a refuse-or-audit decision.

## Verification

- `refuse_binding_while_impersonating` refuses under an impersonation and is a no-op otherwise (`tests/test_impersonation.py`).
- `ChannelLinkService.confirm` under an impersonation raises and **neither reads nor spends** the token (`tests/test_channel_link_requests.py`).
[...]

Issue #1438, opened 2026-09-05
Identity-binding routes are allowed while impersonating, so an admin can attach their own chat identity to somebody's account


OchnikBartek:
Pre-existing; found reviewing #1044 (PR #1435). Same cluster as #943 and #1044 in #168 (governance / admin).

## What breaks

Two routes bind an external identity to the *current* account, and neither knows whether the current account is being impersonated:

- `POST /me/channel-link/{token}` (`backend/app/api/routes/v1/me_channel_link.py`) confirms a channel-link code and attaches the confirmer's Telegram / Slack / Mattermost identity to the signed-in user. No audit entry.
- MCP OAuth completion (`backend/app/api/routes/v1/me_mcp_connections.py`, `oauth.py`) stores a personal connection on the signed-in user.

While an app admin is acting as B, either one attaches the *administrator's* chat identity or OAuth grant to B's account. That binding outlives the impersonation's hour, is not recorded anywhere, and is not what "the impersonation ends when the row ends" promises. #1044 made impersonation a first-class flow, so this is now a deliberate path rather than an accident.

## Sequence

1. As an app admin, impersonate B from `/admin/users`.
2. In Telegram, ask the bot for a link code; as B, confirm it at `/link`.
3. End the impersonation. The administrator's Telegram account now speaks as B, indefinitely.

## Fixed when

Identity-binding routes refuse under an impersonation (`current_impersonation()` is set in `app/services/impersonation.py`) with a 403 that says why — or, if an administrator fixing somebody's connection is a wanted flow, the binding is audited with the impersonator named. Decide which; either way a test covers the refusal or the entry.

PR #1520, opened 2026-09-08
fix(mcp): refuse an org OAuth start under an impersonation

### LINKED ISSUE #1490: Organization MCP OAuth starts are allowed while impersonating, binding an admin grant as a member
Pre-existing; found fixing #1438 (identity-binding during impersonation). Same cluster as #943, #1044 and #1438 in #168 (governance / admin).

## What breaks

#1438 refuses the two *personal* identity-binding routes under an impersonation (`ChannelLinkService.confirm`, `McpConnectionService.oauth_start`). The **organization-scoped** OAuth starts take the same class of action and are not guarded:

- `POST /api/v1/organizations/{id}/mcp-connections/oauth/start` → `McpConnectionService.oauth_start_for_org`
- `.../oauth/start/github` → `oauth_start_for_org_github`
- `.../oauth/start/portal` → `oauth_start_for_org_portal`

`oauth_start_for_org` funnels through `_oauth_start`; the GitHub/portal variants funnel through `_stage_pending_oauth` instead, so #1438's guard on `oauth_start` (the personal method) reaches none of them.

While an app admin acts as B, they can start an org OAuth flow, consent with the **administrator's own** provider account, and the org connection is created with `created_by_user_id = ctx.subject_id` — i.e. attributed to B. The grant backing a shared org integration is then the administrator's, recorded as the member's, and it stops working when the administrator's provider access is revoked, with nothing pointing at why.

## Sequence

1. As an app admin, impersonate B from `/admin/users`.
2. As B, open an org integration and run its OAuth (`/oauth/start`), consenting with your own provider account.
3. End the impersonation. The org connection is attributed to B but backed by the administrator's grant.

## Fixed when

The organization OAuth starts refuse under an impersonation the way the personal one now does — `refuse_binding_while_impersonating(...)` at `_oauth_start` and `_stage_pending_oauth` (or at each org entry point), with a 403 that says why — **or** the flow is explicitly allowed and the org connection records the impersonator rather than the subject as `created_by`. Decide which; either way a test covers the refusal or the attribution. This is the org half of the decision #1438 made for personal connections.

Note the different chokepoint: `_stage_pending_oauth` (github/portal) is not on `_oauth_start`, so a single guard on `_oauth_start` would miss two of the three routes.

OchnikBartek:
> **Stacked on #1491 (#1438).** The base is
> `fix/1438-refuse-identity-binding-while-impersonating`, whose
> `refuse_binding_while_impersonating` this uses; it retargets to `main` once
> #1491 merges.

#1438 refuses the two *personal* identity-binding routes under an impersonation.
The **organization-scoped** OAuth starts take the same class of action and were
not guarded:

- `POST /organizations/{id}/mcp-connections/oauth/start` → `oauth_start_for_org`
- `.../oauth/start/github` → `oauth_start_for_org_github`
- `.../oauth/start/portal` → `oauth_start_for_polled_portal`

While an app admin acts as B, they could start an org OAuth flow, consent with
the **administrator's own** provider account, and the org connection was created
with `created_by_user_id = ctx.subject_id` — the administrator's grant recorded
as B's, and it stops working when the administrator's provider access is revoked,
with nothing saying why.

## Fix

The consistent half of #1438's decision: **refuse**. Each org start now calls
`refuse_binding_while_impersonating` at its entry — the personal `oauth_start`
already guards there, and the github/portal starts skip `_oauth_start`, so a
single chokepoint guard would have missed two of the three (the issue's own
note). Guarding each entry refuses before any discovery, credential read or row,
and keeps the tests trivial.

## Verified

A test that each of the three org starts raises `AuthorizationError` under an
impersonation and creates no org-scoped row. `make check` green.

Closes #1490

### DIFF
--- a/backend/app/services/mcp_connection.py
+++ b/backend/app/services/mcp_connection.py
@@ -667,7 +667,14 @@ class McpConnectionService:
         connection stops working, and the fix is for somebody to authorize it
         again. An organization that wants this should consent with an account it
         controls, not with a member's personal one.
+
+        Raises:
+            AuthorizationError: When the request runs under an impersonation. The
+                grant that comes back is the administrator's own, and the org
+                connection would record it as the member's - the org half of the
+                refusal #1438 made for personal connections (#1490).
         """
+        refuse_binding_while_impersonating("Connecting an integration")
         return await self._oauth_start(
             name=name,
             url=url,
@@ -866,7 +873,10 @@ class McpConnectionService:
                 connect through GitHub (no `mcp_catalog_key`).
             NotFoundError: If the organization has stored no `github_oauth_app`
                 secret - a 4xx the connect UI shows, never a 500.
+            AuthorizationError: When the request runs under an impersonation - the
+                administrator's own grant would be bound as the member's (#1490).
         """
+        refuse_binding_while_impersonating("Connecting an integration")
         portal = portal_catalog.get_portal(portal_key)
         if portal is None or portal.mcp_catalog_key is None:
             raise BadRequestError(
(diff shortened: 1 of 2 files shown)


OchnikBartek:
@codex review

Issue #1456, opened 2026-09-05
Refuse a deactivated linked sender before attachments are stored and transcribed


OchnikBartek:
**Severity: Low** · `security` · found reviewing #1454

## What's wrong

A deactivated or departed member whose chat account is still linked (`channel_identities.user_id` set) is refused by `_membership_context` **at the run** — after the router has already fetched the message's attachments, stored them, and sent any voice note to the paid `TranscriptionService` with the organization's credential. The refusal #1436 added is correct and late.

## Where

- `backend/app/services/channels/router.py:268-271` — the admission gate asks only whether the identity is *linked*, not whether the link still points at an active member
- `backend/app/services/channels/router.py:295-310` — `_receive_files`, `_transcribe` and `_thread_files` run before `_answer_mention` / `answer_default`
- `backend/app/services/channels/mentions.py` — `_membership_context` reads `member_repo.get_active` and refuses; nothing before it does

## How it fails

An offboarded account with a still-linked Telegram identity sends a voice note. The file is downloaded and stored, the transcription is billed, and *then* the turn is refused with the link-first message. Bounded by the per-sender rate limit and the transcription byte cap, so a slow leak rather than a hole — but it is spend by an account the platform says is refused everywhere.

## Fix

Read the membership once in the router, before the attachment work, with the same joined read (`member_repo.get_active`), and treat a linked identity whose member is gone as unlinked from that point on — refused in a DM, admitted under the binding in a room — so the run-level refusal becomes the second lock rather than the first.

## Acceptance criteria

- [ ] A deactivated linked sender's voice note is refused before `TranscriptionService` is called and before any file row is written
- [ ] Test: `test_a_deactivated_linked_sender_costs_no_transcription`

---
Cluster **Channels and bots** in #168. Nearest: #1427 / #1436 (the run-level refusal this moves earlier), #1433 / #1454. Pre-existing, found reviewing #1454.

PR #1523, opened 2026-09-08
fix(channels): rate-limit the invite-on-refusal path

### LINKED ISSUE #1516: Channel invite-on-refusal is not rate-limited: a refused sender can churn link requests
## What's wrong

`ChannelMessageRouter._route_inner` refuses a sender who may not be answered
(unlinked in a link-required room, or — since #1456 — a linked account whose
member was deactivated) by calling `_invite_to_link`, and this refusal returns
**before** `_check_rate_limit`. So every refused message mints a link request
(a delete + insert on `channel_link_requests`) and, in a DM or a message that
names the bot, sends a platform reply — with no per-sender allowance consumed.
A refused sender can therefore churn the invite path as fast as they can type.

## Where

- `backend/app/services/channels/router.py:335-345` — the admission refusal
  (`_refuse_if_named(... _invite_to_link(...))`) returns before the rate limit
- `backend/app/services/channels/router.py:350` — `_check_rate_limit`, keyed on
  `identity.id`, runs only on the path that continues to a turn

## How it fails

1. An offboarded account (or anyone, in the link-required room's unlinked case)
   sends repeated DMs.
2. Each one hits the admission refusal, mints a fresh link-request row and sends
   an invitation, and returns before the rate limit is consulted.
3. The `rate_limit_rpm` control the bot is configured with never sees them.

## Pre-existing, and its blast radius

The unlinked-sender half is **pre-existing**: an unlinked sender in a
link-required room was always refused here, before the rate limit. #1456 only
extends the same early refusal to a linked-but-deactivated sender, which is what
surfaced it. The impact is modest — the platform throttles bot replies itself,
and the DB write is one indexed row — which is why this is Low rather than a
DoS.

## Fix (to decide)

Consume the per-sender allowance on the refusal/invite path too, without undoing
#1456's "refuse before the attachment work" ordering. Options: move
`_check_rate_limit` above the admission refusal (it only needs `identity.id`,
already resolved), or throttle `_invite_to_link` on its own counter. Both change
the first-contact invite UX slightly (a legitimate new user's repeated messages
would be throttled too), so it wants a deliberate choice rather than folding into
a security bugfix.

## Done when

A refused sender's repeated messages stop minting invitations once the bot's
`rate_limit_rpm` is reached, and a first-contact user still gets one invitation.

---

Cluster **Channels and bots** in #168. Found reviewing #1456 (PR #1500); the
deactivated-sender half arrived with it, the unlinked-sender half predates it.

OchnikBartek:
## What this fixes

A sender refused at the channel router's admission gate — unlinked in a link-required room, or (since #1456) a linked account whose member was deactivated — is answered by `_invite_to_link`, which mints a `channel_link_requests` row and, in a DM or a message that names the bot, sends an invitation. That branch in `_route_inner` returned **before** `_check_rate_limit`, so a refused sender could churn the invite path — a delete + insert and a platform reply — as fast as they could type, and the bot's `rate_limit_rpm` never saw them.

## The fix

`_route_inner` now consumes the per-sender allowance **inside** the admission-refusal branch, before `_invite_to_link`, routed through the invite's own channel (`_refuse_if_named`):

- the throttle is shown where the invite would show and stays **silent where the invite would stay silent**, so a room message naming a colleague is not talked over;
- the continuing path's own `_check_rate_limit` is unchanged, and the two branches are mutually exclusive, so a message is counted exactly once;
- the limiter **consumes then checks**, so a new sender's first message still gets its one invitation and only repeated messages past the allowance mint nothing.

This is the behaviour the issue's "Done when" asks for. The choice the issue flagged — that a legitimate new user's rapid repeat messages are throttled too — is accepted by that criterion (one invitation, then throttle).

## Tests

Two tests driving `_route_inner` for a refused sender (`tests/test_channel_link_requests.py`):

- a first message still sends the invitation;
- one past the allowance never calls `_invite_to_link` and answers "slow down" instead.

## Verification

`make check` green in substance — 7331 passed, 0 failures; lint (ruff/ty/deptry/eslint) clean. The reported 99.00% total is the local integration-skip artifact (no database on this machine); `channels/router.py` is template-inherited and not on the 100% gate. The channel suites (`link_requests`, `router_limits`, `addressing`, `membership`) all pass.

## Scope

The unlinked-sender half is pre-existing; #1456 extended the same early refusal to a deactivated linked sender, which surfaced it. A single hoist covers both halves.

## Stack

Branches off `fix/1456-reject-deactivated-sender-before-write` so the fix covers the deactivated-sender refusal #1456 adds as well as the pre-existing unlinked one; cannot stand alone until #1456 merges, at which point GitHub retargets this PR to `main`.

Closes #1516
Refs #1456

### DIFF
--- a/backend/app/services/channels/router.py
+++ b/backend/app/services/channels/router.py
@@ -1079,6 +1079,17 @@ class ChannelMessageRouter:
             # carrying a code somebody copied. Kept because "how do I connect
             # this?" is a question people ask in words, and because a URL that
             # expired needs a way to ask for another.
+            if incoming.chat_type == "private":
+                # The other door onto `_invite_to_link`, and it runs before the
+                # admission gate's allowance is consulted, so a refused sender
+                # could churn link requests through `/link` where the plain
+                # message is now throttled (#1516). Only in a direct message,
+                # where the invite actually mints a request.
+                identity = await self._resolve_identity(incoming, bot, db)
+                try:
+                    await self._check_rate_limit(bot, str(identity.id))
+                except BadRequestError as exc:
+                    return exc.message
             try:
                 return await self._invite_to_link(incoming, db)
             except Exception:
--- a/backend/app/services/channels/router.py
+++ b/backend/app/services/channels/router.py
@@ -342,6 +342,16 @@ class ChannelMessageRouter:
             # attachment is fetched, stored or transcribed for a turn
             # `_membership_context` refuses anyway - which is now the second lock
             # rather than the first (#1456).
+            try:
+                await self._check_rate_limit(bot, str(identity.id))
+            except BadRequestError as exc:
+                # Consumed on the refusal path too, not only before a turn:
+                # otherwise a refused sender mints a link request and an
+                # invitation on every message with no allowance ever spent
+                # (#1516). Rides the invite's own channel - shown where the
+                # invite would show, silent where it would stay silent.
+                await self._refuse_if_named(bot, incoming, exc.message)
+                return
             await self._refuse_if_named(bot, incoming, await self._invite_to_link(incoming, db))
             return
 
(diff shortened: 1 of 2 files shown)


OchnikBartek:
@codex review

PR #1524, opened 2026-09-08
fix(mcp): stop an unlinked guest borrowing the owner's personal MCP

### LINKED ISSUE #1469: mcp: confirm publisher-fallback + acts_for_sender lets personal-MCP borrow the owner's credentials for an unrecognized guest
**Found while reviewing #1383 (agent memory); pre-existing behaviour from #1392, already on `main` — not introduced by #1383.** Flagged by the adversarial reviewer as "worth confirming separately"; filing so it is not lost. Needs confirmation of intent before it is treated as a defect.

## What was observed

On the **publisher-fallback** admission surface, when `AdmittedAs.acts_for_sender` is `True` and the sender is an **unrecognized guest** (no linked user), a **personal-MCP** connection resolves its user via:

```
personal_mcp_user_id = (owner_user_id or ctx.user_id) if acts_for_sender else None
```
(`backend/app/services/agent_runner.py` ~l2031; resume restore ~l3541-3546)

So for an unrecognized guest on a publisher-fallback channel, personal-MCP would act with the **owner/publisher's** MCP credentials rather than refusing or running credential-less.

## Why it may matter

An anonymous/unlinked channel participant could cause the agent to reach a third-party service (the MCP server) authenticated as the **owner/publisher**, i.e. use someone else's connected-account credentials without being that account. Whether that is intended depends on the publisher-fallback contract (the memory side is safe here: `derive_end_user_scope_key` maps publisher-fallback → owner ignored → `chan`/`None`, so no cross-user memory leak — verified in the #1383 review).

## What "done" looks like

- Decide whether personal-MCP under `publisher_fallback` + `acts_for_sender` for an unrecognized guest is **intended** (publisher lends their tool access to anonymous channel users) or should **refuse / run without personal-MCP**.
- If unintended: gate personal-MCP resolution so a publisher-fallback / unrecognized-guest turn does not borrow the owner's MCP identity; add a refusal test.
- If intended: document it in `docs/mcp.md` / `docs/channels.md` so the credential-lending is explicit.

Refs #1392 (source of the behaviour), #1383 (where it was found), #168 (channels/governance cluster).

OchnikBartek:
## Summary

On the publisher-fallback admission surface an unidentified channel visitor runs under the binding's publisher, and `acts_for_sender` is set there too — so personal-MCP resolution fell back to the owner (`personal_mcp_user_id = (owner_user_id or ctx.user_id) if acts_for_sender`). An anonymous, unlinked channel guest could therefore make the agent reach a third-party MCP server authenticated as the **owner/publisher**, using their connected-account credentials.

`docs/mcp.md` already documented the opposite — an unlinked channel sender "has no account to speak through" and "the run proceeds without that server" — so this makes the code match the documented contract rather than introducing a new rule. Decision on the issue was to **gate**, not to document credential-lending as intended.

## Related issue

Closes #1469. Refs #1392 (source of the behaviour), #1383 (where it was found).

## Changed

- `app/services/agent_runner.py`: gate personal-MCP resolution on the subject being a real person, never a publisher standing in.
  - **Fresh run:** read from the request (`ctx.subject_is_publisher_fallback`).
  - **Resume:** the approver's context cannot tell (its `user_id` is the publisher either way), so the fact is carried in the parked terms — a new `AdmittedAs.subject_is_publisher_fallback` field, threaded back through `_assemble`. This is the same #788 pattern already used for `owner_user_id` and the audience.
  - Defaults `false`, so a run parked before this was recorded resumes exactly as before, rather than losing its tools.

## Testing

- Two regression tests — a fresh publisher-fallback guest and a resumed one — assert the personal identity is **not** resolved; both fail without the gate (verified by reverting it).
- `test_agent_runner.py` full suite green; adjacent suites (channels, chat, audience, mcp, triggers, subagents) green; `test_publisher_standing.py` integration green against a real DB.
- Backend 100% coverage gate green (8110 passed) against a real database.
- No migration: the flag is a field on the run's JSONB paused-state payload, not a column.

## Notes for reviewers

- The gate keeps the resolved value `(owner_user_id or ctx.user_id)` unchanged for every real-subject case (linked channel sender, API key holder, dashboard user); it only forces `None` for a publisher stand-in.
- The resume signal is the stored flag rather than `restored_audience.user_id is None`, because an old `private_to_user` payload legitimately has no stored audience yet was an identified person — keying on the audience there would wrongly strip a real person's tools.


OchnikBartek:
@codex review

Issue #1459, opened 2026-09-05
Unify the mention and default turn paths' refusal-and-crash handling in the channel router


OchnikBartek:
**Effort: small** · the second half of the last item on #1434

## What's wrong

`_answer_mention` (`backend/app/services/channels/router.py:439-516`) and the default path at the end of `_route_inner` (`:333-384`) run the same turn in two hand-written copies of the same shape: try the agent, on `AppException` discard the turn's files and answer the refusal, otherwise stitch the notes and `_deliver`. They differ in three places that read as accidents rather than decisions:

- the default path catches `Exception`, logs it and apologises ("Sorry, something went wrong"); the mention path catches only `AppException`, so a crash there propagates out of `route`, releases the dedupe claim and answers nothing
- the mention path posts a refusal only where the bot was named (`_names_the_bot`); the default path posts it unconditionally
- `UnaddressedMessage` is the mention path's "not for me" signal and has no equivalent on the other

"Fix one, miss the other" is the shape that produced #718; #1434 unified the placeholder and edit builders the two share (`_message`, `_post_placeholder`) and left this half alone because it changes behaviour.

## Where

- `backend/app/services/channels/router.py:439-516` — `_answer_mention`
- `backend/app/services/channels/router.py:333-384` — the default path in `_route_inner`

## How it fails

A crash inside a mention turn (a provider error the runner did not wrap) answers the sender with silence and a released claim - so the platform's redelivery runs it again - where the same crash on the default path answers an apology once. Two people asking the same bot get two policies depending on whether they typed a handle.

## Fix

One `_run_turn(...)` that takes the answering coroutine and the refusal-posting policy, with the crash behaviour decided once - most likely the default path's apology, posted through `_refuse_if_named`. Decide and write down whether a mention crash should keep releasing the claim; today the two paths disagree and neither says why.

## Acceptance criteria

- [ ] Both paths share one try/except; a crash on either answers the same way
- [ ] Test: `test_a_crash_on_the_mention_path_answers_like_one_on_the_default_path`

---
Cluster **Channels and bots** in #168. Nearest: #1434 (the batch this is split from), #718 (the two-copies class), #1458 (unified the other half). Found in the 2026-09-03 channels audit; split out of #1434 on review.

Issue #690, opened 2026-08-13
A successful channel turn never links its ChatFile rows to a message


DEENUU1:
## What breaks

A channel turn never links its `ChatFile` rows to a message, even when the run
succeeds. `link_files_to_message` is called from exactly one place —
`ChatService` in `app/services/agent.py`, the web chat path — so a file sent to a
Slack, Telegram or Mattermost bot is stored, fed to the agent, and then left with
`message_id` NULL for ever. `TranscriptService.record` writes the user turn for a
channel run with `content=prompt` and nothing about the files that came with it.

`chat_files` has no `organization_id`, so a row with no message is scoped by
`user_id` alone: it is reachable through `GET /files/{id}` by its owner and by
nothing else. The conversation it belongs to does not know the file exists, so a
transcript of a channel thread shows the question and not the spreadsheet it was
asked about.

## The sequence

1. Drop a `.csv` on a channel bot with any question.
2. The agent answers using the file — `ChannelAttachmentService.receive` stored it
   and `AttachmentRouter` folded it into the prompt.
3. `SELECT message_id FROM chat_files WHERE id = …` → NULL, for a turn that ran
   perfectly.

## Where

`backend/app/services/channels/router.py` (`_receive_files`, and the two calls to
`ChannelAgentRouter`), `backend/app/services/transcript.py` (`record`, the user
turn it writes), `backend/app/services/conversation.py`
(`link_files_to_message`, the only linker), `backend/app/services/agent.py` (its
only caller).

## How I would know it is fixed

A channel turn that ran leaves its `ChatFile` rows carrying the `message_id` of
the user message the transcript wrote for that run, the same way a web chat upload
does — asserted by a test that runs a channel turn with an attachment and reads
the row back.

## Related

Pre-existing, found fixing #661 (a turn *refused* before the run orphaning the
same rows; PR #689 deletes them on a refusal and deliberately leaves this half
alone). #660 touches the same `_receive_files` calls. Checked #168; this belongs
with the #660 / #661 channel-attachments cluster.

PR #1525, opened 2026-09-08
fix(channels): discard a crashed turn's files, not only a refused turn's

### LINKED ISSUE #1503: A crashed channel turn leaves its attachments orphaned in chat_files
Pre-existing; found unifying the channel router's turn paths (#1459). Cluster **Channels and bots** in #168; same #690 class (a turn with no run leaves orphaned `chat_files` rows).

## What's wrong

When an agent run **crashes** (any non-`AppException`), the channel router apologises and returns, but does **not** discard the attachments the turn already fetched and stored - unlike an `AppException` refusal, which discards them (#690).

- `backend/app/services/channels/router.py` — `_run_turn`'s `except Exception` branch apologises but calls no `_discard_files`; the `except AppException` branch does.

Both the mention path and the default path behave this way (after #1459 unified them - before it, only the default path apologised on a crash; the mention path propagated and, via redelivery, could store a fresh set on each retry, which was worse).

## How it fails

A voice note or file is fetched and stored, the run crashes on a provider error, the sender gets an apology, and the `chat_files` row is left pointing at a turn that never completed. `chat_files` carries no organization, so an unlinked row is scoped by `user_id` alone and nothing collects it (#690). A slow leak, bounded by the per-sender rate limit and the transcription/file caps.

## Fix

Discard the turn's files in `_run_turn`'s crash branch too, the way the refusal branch already does - a crashed turn produced no run, so its files are as orphaned as a refused turn's. A test that a crashed turn leaves no stored file (the crash sibling of `test_a_refused_turn_leaves_no_stored_file`).

Deliberately left out of #1459 to keep that change a behaviour-preserving unification of the *answer* handling rather than also changing what a crash does with files.

OchnikBartek:
## What this fixes

When a channel agent run crashes on any non-`AppException`, `_run_turn`'s `except Exception` branch apologises and consumes the delivery claim — but, unlike the `except AppException` refusal branch, it does **not** discard the attachments the turn already fetched and stored. A voice note or file fetched for a turn that then crashes on a provider error is left as a `chat_files` row pointing at a turn that never completed. `chat_files` carries no organization, so an unlinked row is scoped by `user_id` alone and nothing collects it (#690) — a slow leak, on both the mention and default paths after #1459 unified them.

## The fix

The crash branch now discards the turn's files too, the way the refusal branch already does — a crashed turn produced no run, so its files are as orphaned as a refused turn's. One line, plus the docstring corrected (it had explicitly flagged the crash orphan as tracked by this issue).

## Tests

`test_a_crashed_turn_leaves_no_stored_file` (`tests/test_channel_attachments.py`) — the crash sibling of `test_a_refused_turn_leaves_no_stored_file`: a `RuntimeError` on either the mention or the default path leaves `rows == []`. It fails without the fix (the crash branch would keep the stored file).

## Verification

`make check` green in substance — the full backend suite passes; the reported sub-100% total is the local integration-skip artifact (no database on this machine). `channels/router.py` is template-inherited and not on the 100% gate. Channel suites (`attachments`, `mentions`, `supervisor`, `voice_messages`, `live_reply`) all pass.

## Scope

Deliberately left out of #1459, which was a behaviour-preserving unification of the *answer* handling rather than a change to what a crash does with files. This is that follow-up.

## Stack

Branches off `fix/1459-unify-refusal-and-crash-paths`, since the unified `_run_turn` crash branch it fixes is a #1459 construct; cannot stand alone until #1459 merges, at which point GitHub retargets this PR to `main`.

Closes #1503
Refs #1459, #690

### DIFF
--- a/backend/app/repositories/chat_file.py
+++ b/backend/app/repositories/chat_file.py
@@ -87,6 +87,26 @@ async def link_to_message(
     return result.rowcount  # ty: ignore[unresolved-attribute]
 
 
+async def unlinked_ids(db: AsyncSession, file_ids: Iterable[UUID]) -> set[UUID]:
+    """Of these files, the ids whose row is not yet attached to a message.
+
+    A turn's attachments are linked when the runner records its transcript - a
+    failed turn included, which persists and commits before it re-raises - so a
+    file already carrying a `message_id` belongs to a stored turn and must not be
+    swept up as an orphan (#1503). Read as a column select rather than through the
+    mapped rows: the link is a bulk UPDATE the session never saw and
+    `expire_on_commit` is off, so a loaded instance still reads `message_id` as
+    None.
+    """
+    ids = list(file_ids)
+    if not ids:
+        return set()
+    result = await db.execute(
+        select(ChatFile.id).where(ChatFile.id.in_(ids), ChatFile.message_id.is_(None))
+    )
+    return set(result.scalars().all())
+
+
 async def delete(db: AsyncSession, *, db_file: ChatFile) -> None:
     """Delete a chat file row."""
     await db.delete(db_file)
--- a/backend/app/services/channels/router.py
+++ b/backend/app/services/channels/router.py
@@ -21,6 +21,7 @@ from app.repositories import (
     channel_bot_repo,
     channel_identity_repo,
     channel_session_repo,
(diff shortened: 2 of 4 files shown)


OchnikBartek:
@codex review

PR #1526, opened 2026-09-08
fix(users): treat an explicit null password as a no-op, not a 500

### LINKED ISSUE #1497: PATCH /users/me with {"password": null} returns 500 instead of a 4xx
Pre-existing; found reviewing #1439 (session revocation on password change). Adjacent to the `UserService.update` password branch that #1439 touches.

## What breaks

`PATCH /users/me` (and the admin `PATCH /users/{id}`) with a body of `{"password": null}` returns a **500**, not a 4xx.

`UserUpdate.password` is `str | None` (`backend/app/schemas/user.py`), so `{"password": null}` passes schema validation. `UserService.update` (`backend/app/services/user.py`) then branches on `"password" in update_data` — which is True for an explicit `null` — and calls `get_password_hash(None)`, which reaches `bcrypt` with `None` and raises `AttributeError`. The global handler turns that into a 500.

`writable()` does not drop the `None`: the null-drop rule keys on the model column, and `User` has `hashed_password`, not `password`, so `password=None` survives into `update_data`.

Not a security hole and not a fail-open: it raises *before* any write and the request transaction rolls back (nothing is persisted, no sessions are revoked). It is a malformed-input 500 that should be a 422/400 — or a no-op.

## Sequence

1. `PATCH /api/v1/users/me` with `{"password": null}` and a valid session.
2. Response is 500 with an unhandled `AttributeError` logged.

## Fixed when

The branch guards on a present, non-null password — `update_data.get("password") is not None` rather than `"password" in update_data` — so an explicit `null` is a no-op (or the schema rejects `null` for `password` with a field error). A test covers `{"password": null}` returning a 4xx / no-op instead of 500. This is the same block #1439 edits, so whoever takes it should rebase on #1439's shape (the `password_changed` flag).

Cluster: same file/branch as #1439; see #168 governance/admin only loosely — this is an input-validation defect, not an authz one.

OchnikBartek:
## What this fixes

`PATCH /users/me` and the admin `PATCH /users/{id}` with a body of `{"password": null}` returned a **500**, not a 4xx or a no-op. `writable()` keeps the explicit `None` — the row stores `hashed_password`, so there is no `password` column for its null-drop rule to key on — and `UserService.update` branched on `"password" in update_data`, which is true for an explicit null, so `get_password_hash(None)` reached bcrypt and raised. It raised before any write (the transaction rolled back — nothing persisted, no sessions revoked), so it was a malformed-input 500 rather than a data hazard.

## The fix

`update()` pops `password` unconditionally and only hashes it, bumps the credential version and revokes sessions when a real value was sent. An explicit null falls through untouched — a no-op.

## Tests

`test_an_explicit_null_password_is_a_no_op_not_a_crash` (`tests/test_services.py`): `update` with `{"password": null}` hashes nothing, bumps no version, revokes no session, and keeps the other fields. It raised without the fix.

## Verification

`make check` green in substance — 7339 passed, 0 failures; lint clean. The reported 98.99% total is the local integration-skip artifact (no database on this machine); `services/user.py` is not on the 100% platform gate.

## Stack

Branches off `fix/1517-password-change-endpoint`, which evolved this same `UserService.update` password branch (the `credential_version` bump on top of #1439's session revocation); basing here keeps the null-guard composing with that shape rather than conflicting. Cannot stand alone until #1517 merges, at which point GitHub retargets this PR to `main`.

Closes #1497
Refs #1517, #1439

### DIFF
--- a/backend/app/services/user.py
+++ b/backend/app/services/user.py
@@ -342,10 +342,14 @@ class UserService:
         user = await self.get_by_id(user_id)
 
         update_data = writable(user_in, over=User)
-        password_changed = "password" in update_data
+        # `password` has no column (the row stores `hashed_password`), so writable
+        # keeps an explicit null rather than dropping it. Popped unconditionally so
+        # a null is a no-op, not a hash of None that reaches bcrypt as a 500 (#1497).
+        new_password = update_data.pop("password", None)
+        password_changed = new_password is not None
         if password_changed:
             update_data["hashed_password"] = await asyncio.to_thread(
-                get_password_hash, update_data.pop("password")
+                get_password_hash, new_password
             )
             # Bump the credential version alongside the hash, so a refresh token
             # minted before this change is refused even if it raced the session
--- a/backend/tests/test_services.py
+++ b/backend/tests/test_services.py
@@ -385,6 +385,32 @@ class TestUserServicePostgresql:
             )
 
     @pytest.mark.anyio
+    async def test_an_explicit_null_password_is_a_no_op_not_a_crash(
+        self, user_service: UserService, mock_user: MockUser
+    ):
+        """`PATCH /users/{id}` (and /users/me) with `{"password": null}` reaches
+        here as an explicit None that `writable` keeps - there is no `password`
+        column. It must be a no-op: nothing hashed, no version bump, no session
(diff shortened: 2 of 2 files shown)


OchnikBartek:
@codex review

Issue #1492, opened 2026-09-07
Manual personal MCP token entry is allowed while impersonating, storing an admin token as a member's


OchnikBartek:
Pre-existing; found reviewing #1438 (identity binding during impersonation). Same cluster as #943, #1044, #1438 and #1490 in #168 (governance / admin).

## What breaks

#1438 refuses the two routes that capture the *administrator's own* identity under an impersonation — a chat-account link (`ChannelLinkService.confirm`) and a personal OAuth grant (`McpConnectionService.oauth_start`). Manual bearer-token entry on the same personal surface is **not** refused:

- `POST /api/v1/me/mcp-connections` → `McpConnectionService.create` (`backend/app/services/mcp_connection.py`)
- `PATCH /api/v1/me/mcp-connections/{id}` → `McpConnectionService.update`

Both persist a caller-supplied `auth_token` under `VaultScope.user(user_id)` where `user_id` is the impersonated target. While an app admin acts as B, they can paste their own MCP server token and store it as **B's** personal connection: B's agents then call that server as the administrator's account, the credential outlives the hour-bounded impersonation, and it stands recorded against B.

## Why it was left out of #1438 (the decision to make)

The two flows #1438 refused capture the administrator's *ambient authenticated* identity — whoever consented at the provider or messaged the bot is necessarily the admin. Manual token entry is explicit value-entry configuration, and doing configuration *as B* is much of what impersonation is for; a manually typed token is opaque data, not provably the admin's identity. So excluding it is defensible — but the residual harm above is real, so the choice should be recorded, not silent.

## Sequence

1. As an app admin, impersonate B from `/admin/users`.
2. As B, `POST /me/mcp-connections` with your own MCP server's bearer token.
3. End the impersonation. B's agents keep calling that server as your account.

## Fixed when

A decision is taken and a test covers it:
- **Refuse:** call `refuse_binding_while_impersonating("Adding an integration token")` in `create`/`update` when `data.auth_token` is set, with a 403; or
- **Allow, audited:** record the connection with the impersonator named, so the credential is attributable to the administrator who entered it rather than to B.

This is the manual-token half of the line #1438 drew for captured-identity bindings; see also #1490 for the organization-scoped OAuth half.

PR #1527, opened 2026-09-08
fix(mcp): refuse a manual org token while impersonating

### LINKED ISSUE #1521: Manual org MCP token entry is neither refused nor fully audited while impersonating
Pre-existing; found fixing #1492 (the personal-surface half). Same cluster as #943, #1044, #1438, #1490 and #1492 in #168 (governance / admin).

## What breaks

#1492 refuses a manually-entered bearer token on the *personal* surface (`POST/PATCH /api/v1/me/mcp-connections`) while impersonating. The *organization* surface has the same shape and no such guard:

- `POST /api/v1/orgs/{org}/mcp-connections` → `McpConnectionService.create_for_org`
- `PATCH /api/v1/orgs/{org}/mcp-connections/{id}` → `McpConnectionService.update_for_org`

Both persist a caller-supplied `auth_token` under `VaultScope.organization(ctx.organization_id)` (`backend/app/services/mcp_connection.py:1284`, `:1343`). While an app admin acts as B (an org owner/admin), they can paste their own MCP server token and store it as the **organization's** shared credential. The org's agents then call that server as the administrator's account, and the credential outlives the hour-bounded impersonation.

## Why this is a separate decision from #1492

#1492's harm is a credential stored as *B's personal* connection. Here the credential is the *organization's* shared one — which is meant to be admin-entered — so the "stored as B" framing is weaker, and the two halves of the audit trail already differ:

- `create_for_org` calls `record_audit`, which stamps `impersonator_user_id` from the audit ContextVar (`app/core/audit.py:66`). So a **created** org connection is already attributable to the administrator, not to B — the "allow, audited" outcome #1492 describes is, for creation, already in place.
- `update_for_org` writes **no audit entry at all** (`mcp_connection.py:1382`). A token pasted over an existing org connection under an impersonation is therefore neither refused nor attributed.

What neither audit addresses is the residual harm that *is* shared with #1492: the administrator's token keeps speaking as the administrator's own account for every org agent bound to the connection, after the hour-bounded impersonation has ended.

So the choice here should be taken deliberately, as it was for #1438 / #1490 / #1492 — not left silent.

## Sequence

1. As an app admin, impersonate B (an org owner) from `/admin/users`.
2. As B, `POST /orgs/{org}/mcp-connections` (or `PATCH .../{id}`) with your own server's bearer token.
3. End the impersonation. The org's agents keep calling that server as your account; a token set via `PATCH` left no audit trail at all.

## Fixed when

A decision is taken and a test covers it, mirroring #1492:

- **Refuse:** call `refuse_binding_while_impersonating("Adding an integration token")` in `create_for_org` / `update_for_org` when `auth_token` is set (403); or
- **Allow, audited:** ensure both paths record the impersonator — add a `record_audit` to `update_for_org` (it has none today) so a manual token set under an impersonation is attributable to the administrator, and accept the shared-credential case as within what org administration allows.

The residual "the credential still speaks as the administrator after the hour ends" harm is the same on both surfaces; whether the shared, deliberately-admin-entered nature of an org credential makes it acceptable where a personal one is refused is the actual question to settle.

This is the organization-scoped manual-token corner of the matrix that #1438 (personal OAuth), #1490 (org OAuth) and #1492 (personal manual token) fill.

OchnikBartek:
## Summary

The organization MCP surface had the same shape as the personal one #1492 guarded, and no guard. `create_for_org` / `update_for_org` seal a caller-supplied `auth_token` under the org's vault scope, so while acting as an org owner an app admin could paste their own MCP server's bearer token and store it as the **organization's shared credential** — every org agent then calls that server as the administrator's account, long after the hour-bounded impersonation ends (#1521).

This is the org-scoped manual-token corner of the same matrix as #1438 (personal OAuth / identity), #1490 (org OAuth) and #1492 (personal manual token).

## Stacking

Based on **`fix/1492-refuse-manual-token-while-impersonating`** (PR #1522), not `main`: it needs the `refuse_binding_while_impersonating` helper (added by #1438, carried through #1492) and extends #1492's governance paragraph to cover the org connection. Retarget to `main` once #1492 (and #1438 beneath it) merge.

## Related issue

Closes #1521. Refs #1492.

## Changed

- `app/services/mcp_connection.py`
  - **Refuse** a typed org token under an impersonation in `create_for_org` and `update_for_org`, via `refuse_binding_while_impersonating("Adding an integration token")` — mirroring #1492's personal-surface guard. A tokenless connection and clearing a token stay allowed (they store no credential).
  - **Audit** `update_for_org`, which wrote no entry at all: an `mcp_connection.updated` record naming the changed fields (never their values, never the staleness resets), so a token rotated onto an existing org connection is attributable — matching `create_for_org`, which already was. `record_audit` stamps the impersonator from the audit context.
- `docs/governance.md`: extend the impersonation binding-refusal paragraph to state the org connection is refused the same way, and that updates now leave a trail.

## Testing

- Regression tests (org surface): a manual token is refused on **create** and **update** under an impersonation — both fail without the guard (verified by reverting it); a tokenless create and a token clear stay allowed; a no-op update writes nothing; and an update records who changed the shared credential.
- `test_mcp_connections.py` + impersonation/audit suites green (254 passed).
- Backend 100% coverage gate green (8106 passed) against a real database.
- `make lint-backend` and `make docs-build` (--strict) green.
- No migration — the flag lives in the audit context; nothing changed on a table.

## Notes for reviewers

- The refusal fires only on a **non-empty** token (`if token:`), matching #1492: clearing a token under an impersonation is allowed, since it stores no credential.
- The update audit lists the caller's fields snapshotted before the internal staleness/OAuth resets add their own keys, so the trail names what was actually edited.

### DIFF
--- a/backend/app/services/mcp_connection.py
+++ b/backend/app/services/mcp_connection.py
@@ -1282,6 +1282,8 @@ class McpConnectionService:
                 details={"name": data.name},
             )
         token = data.auth_token.strip() if data.auth_token else None
+        if token:
+            refuse_binding_while_impersonating("Adding an integration token")
         sealed = seal(token, scope=VaultScope.organization(ctx.organization_id)) if token else None
         try:
             connection = await mcp_connection_repo.create_org_scoped(
@@ -1323,6 +1325,12 @@ class McpConnectionService:
         update_data: dict[str, Any] = writable(
             data, over=McpConnection, exclude={"clear_allowed_tools"}
         )
+        # What the caller changed, snapshotted before the staleness resets below
+        # add their own keys - those are bookkeeping, not an edit the audit should
+        # report (#1521).
+        edited_fields = set(update_data)
+        if data.clear_allowed_tools:
+            edited_fields.add("allowed_tools")
 
         if "url" in update_data:
             update_data["url"] = await _checked_url(update_data["url"])
@@ -1342,6 +1350,8 @@ class McpConnectionService:
 
         if "auth_token" in update_data:
             token = (update_data["auth_token"] or "").strip()
+            if token:
+                refuse_binding_while_impersonating("Adding an integration token")
             # "" clears the stored token; a non-empty value replaces it, sealed at
             # the row's version - one version column covers every envelope in the
             # row, so bumping it would orphan the OAuth siblings (#552).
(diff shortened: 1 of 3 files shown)


OchnikBartek:
@codex review

PR #1539, opened 2026-09-09
refactor(api): collapse the orgs/me/admin forwarders into platformProxy

### LINKED ISSUE #564: Collapse the backendFetch forwarder family (~40 routes) into platformProxy

OchnikBartek:
## Summary

Collapses the remaining hand-rolled `backendFetch` forwarder routes under
`orgs/**`, `me/**` and `admin/**` into `platformProxy()` mounts — the same shape
`kb`, `rag` and (already merged) `sessions` use. Each converted file dropped the
block #564 is about: read the cookie → 401, set the bearer, stringify, map
`BackendApiError`, `NextResponse.json`. `platformProxy()` does all of that plus
the active-organization header, byte-accurate bodies, error-body passthrough and
`no-store` — the three things the hand-rolled copies silently dropped
(#546, #553, #106).

Net **−1200-ish lines** across ~29 files, all behaviour-preserving except two
deliberate, safe changes called out below.

The `sessions/**` cluster this issue also names was done in an earlier batch and
is already on `main`; this branch is the orgs + me + admin follow-up that batch
left open.

## Changed

- **orgs**: `list/one/members/member/invitations/invitation` → one-line
  `platformProxy` mounts (per file, so the `[id]` segment keeps working — an
  `orgs/[[...path]]` catch-all would collide with `orgs/[id]/avatar`).
- **me**: `me/mcp-connections` (`list`, `[id]`, `[id]/test`, `oauth/start`) →
  the same mounts; `me/channel-link` and `me/slash-commands` → one `[[...path]]`
  catch-all each, replacing six files.
- **admin**: `conversations, organizations, ratings/summary, stats, system,
  users, users/[userId], users/[userId]/detail` → one-line mounts.

Two deliberate behaviour changes, both safe:

- Admin routes drop the frontend `requireAdmin` pre-check. Every admin endpoint
  is gated by `CurrentAppAdmin` on the backend, and `admin/settings` already
  proxies this way — the pre-check was redundant, not a second gate.
- Dropping the admin query-param allowlist forwards the whole query string.
  FastAPI ignores a param it does not declare, so nothing changes for a declared
  one — and it fixes the latent bug where the users table's `sort_by`/`sort_dir`
  were dropped while the screen kept sending them.

## Kept hand-rolled (genuinely special)

- `orgs/[id]/avatar` — binary, pins an image content-type under this origin's CSP.
- `orgs/[id]/integrations/*` (four routes) — address `/org/integrations` with the
  tenant in a header, not the path; not a plain forward.
- `me/mcp-connections/oauth/callback` — sets a cookie and redirects.
- `admin/users/[userId]/impersonate` — mints an access token into an HttpOnly
  cookie rather than a body; keeps `requireAdmin`.

## Testing

- `bun run test:coverage` — statements/functions/lines 100%, branches 97.69% (gate ≥97.5).
- `make lint-frontend` (eslint, tsc, prettier, i18n, knip) and `build-frontend` green.
- Each mount joins `proxy-routes.test.ts` (five verbs + auth refusal); its cases
  leave `cookie-routes.test.ts`, whose remaining routes `platform-proxy.test.ts`
  covers generically. `admin-routes.test.ts` is now the impersonate suite alone.

## Notes for reviewers

- Rebased on current `main`; the two files main touched under this change-set
  (`impersonate`, `oauth/callback`) are both in the kept-hand-rolled set and are
  unmodified here.
- Two commits, each independently green (orgs+me, then admin).

Closes #564.


OchnikBartek:
@codex review

PR #1541, opened 2026-09-09
test(lifespan): stub the web-research client the drain test still closed

### LINKED ISSUE #1338: Flaky: the lifespan drain test fails under xdist with a closed event loop
`tests/test_lifespan_drain.py::test_the_lifespan_waits_for_in_flight_background_work`
failed once on CI with `RuntimeError: Event loop is closed`, then passed on a
plain re-run of the same commit, and passes locally in isolation
(`uv run pytest tests/test_lifespan_drain.py -q -p no:randomly`).

Seen on the `test` job of #1333 (`fix/1274-durable-org-teardown`), run
33169284651. The suite runs under `-n auto` with `pytest-randomly`, so this is
order- or worker-dependent rather than a property of the test alone — the two
candidates are what shares its worker and what the run's seed put before it.

Worth chasing before it costs somebody a red branch they trust: a flake in the
lifespan drain test is exactly the one people re-run without reading, and the
drain is what #1137's and #1274's deferred cleanups depend on.

The failing seed and worker split identified, and either the test made
independent of it or the leak in the code it exercises fixed. `--randomly-seed`
from the failing run's header is the way in; the run above still has its log.

Found while merging the W4 branch backlog. Related to the deferred-cleanup
cluster (#1137, #1274), which is what this test guards.

OchnikBartek:
## The flake

`tests/test_lifespan_drain.py::test_the_lifespan_waits_for_in_flight_background_work` failed once on CI with `RuntimeError: Event loop is closed`, then passed on a plain re-run of the same commit — order/worker-dependent under `-n auto` + `pytest-randomly`.

## Root cause

The test runs the real application shutdown in-process to exercise `background.drain()`. Shutdown closes **three** module-global httpx clients: `close_db`, `close_listing_client` and `close_http_client` (web research). The first two were stubbed for exactly this reason — a client another test opened on its own event loop, closed here on this test's loop, raises `Event loop is closed` under xdist — but `close_http_client` (`app/main.py:210`) was left out, so the drain test still flaked on it whenever a web-search test that left `_search._client` open shared its worker.

## The fix

Stub `close_http_client` the same way as its two siblings, so the drain test no longer depends on the shared `_search._client` global across xdist workers.

## Verification

- `uv run pytest tests/test_lifespan_drain.py -q -p no:randomly` — 1 passed.
- Confirmed the mechanism deterministically: `close_http_client()` propagates `RuntimeError: Event loop is closed` when `_search._client` is a client bound to a closed loop — the exact failure the drain test hit.

Closes #1338

### DIFF
--- a/backend/tests/test_lifespan_drain.py
+++ b/backend/tests/test_lifespan_drain.py
@@ -74,11 +74,15 @@ def _stub_startup(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(main, "RedisClient", lambda *a, **k: _FakeRedis())
     monkeypatch.setattr(main, "_start_channel_polling", AsyncMock())
     monkeypatch.setattr(main, "close_db", AsyncMock())
-    # A shutdown collaborator, like `close_db`: the real one closes a module-global
-    # httpx client that another test may have opened on its own event loop, and
-    # closing it here - on this test's loop - raises "Event loop is closed" under
-    # xdist. The lifespan runs in-process, so stub it out the same way.
+    # Shutdown collaborators like `close_db`: each closes a module-global httpx
+    # client that another test may have opened on its own event loop, so closing
+    # it here - on this test's loop - raises "Event loop is closed" under xdist.
+    # The lifespan runs in-process, so stub every one of them out the same way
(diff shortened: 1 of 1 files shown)


OchnikBartek:
@codex review

PR #1542, opened 2026-09-09
docs(configuration): document SMTP settings and what breaks without them

### LINKED ISSUE #1482: docs: SMTP settings are undocumented, and nothing says what breaks without them
`SMTP_HOST`, `SMTP_PORT`, `SMTP_USER`, `SMTP_PASSWORD`, `SMTP_TLS`, `EMAIL_FROM`
and `EMAIL_FROM_NAME` are in `backend/.env.example` and read by
`app/services/email/`, and appear nowhere in `docs/configuration.md` — which is
the settings reference, and the page a deployment reads to find out what it has
to set.

## Why it matters more than a missing table row

A deployment with no mail service does not fail. It runs, and three things
silently stop working:

- **invitations** — the address is never mailed (#1479 is the half of this that
  was visible: the dialog claimed to have sent one);
- **password resets** — the only self-service way back into an account;
- **notifications** — a budget breach, an approval request, a usage report.

None of those announces itself. The first is now honest in the console, but
somebody setting up a deployment still has no page saying "configure this, or
these three things do not happen".

## What it needs

A row per setting in the configuration reference, and a short section saying what
depends on email — probably beside the production checklist, since a deployment
that skips SMTP deliberately should be making that choice knowingly rather than
discovering it when somebody forgets their password.

Found writing #1481. Out of scope there: that PR makes the invite path honest,
and documenting the whole email surface is a different change.

OchnikBartek:
## What this adds

`SMTP_HOST`, `SMTP_PORT`, `SMTP_USER`, `SMTP_PASSWORD`, `SMTP_TLS`, `EMAIL_FROM` and `EMAIL_FROM_NAME` are read by `app/services/email/` and live in `backend/.env.example`, but appeared nowhere in `docs/configuration.md` — the settings reference a deployment reads to find out what it must set.

A deployment with no mail service does not fail. It runs, and **invitations**, **password resets** and **notifications** all go silently unsent, none of them announcing itself.

## Changes

- A new **Email (SMTP)** section: a row per setting (defaults matched to `app/core/config.py`), and a short list of what depends on email.
- A pointer beside the production checklist saying email is deliberately not on it, so a deployment that skips SMTP does so knowingly rather than discovering it when somebody forgets their password.

## Verification

`make docs-build` (strict) succeeds and the `#email-smtp` anchor resolves; the paragraph-length and docs-drift guards pass. Edited by hand — no prettier on `docs/*.md`.

Closes #1482

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -167,24 +167,33 @@ Computed properties:
 
 ## Email (SMTP)
 
-The deployment sends mail through an SMTP server, and a deployment with none
-configured does not fail — it runs, and three things silently stop working, none
-of which announces itself:
+The deployment sends mail through an SMTP server, and one with none configured
+does not fail — it runs, and every mail-dependent flow silently stops, none of
+them announcing itself:
 
+- **passwordless sign-in and password resets** — the magic-link and reset emails
+  are the self-service ways into an account;
 - **invitations** — an invited address is never mailed (the console now says so
-  rather than claiming it sent one, #1479);
-- **password resets** — the only self-service way back into a locked-out account;
-- **notifications** — a budget breach, an approval request, a usage report.
+  rather than claiming it sent one, #1484);
+- **notifications** — a budget breach, an approval request, a usage report, the
+  notice sent when an administrator acts as another account.
 
 | Variable | Default | Description |
 |----------|---------|-------------|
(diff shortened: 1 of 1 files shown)


OchnikBartek:
@codex review

Issue #1512, opened 2026-09-07
The sandbox image was private, so nobody outside the org could run `make dev`


DEENUU1:
`ghcr.io/vstorm-co/sandboxd` was a **private** package, and every compose file
in this repository references it. `COMPOSE_DEV_PROFILES ?= --profile sandbox` is
on by default, so `make dev` — the first command the README gives anybody — pulls
it:

    Image ghcr.io/vstorm-co/sandboxd:0.2 Error error from registry: unauthorized

Nobody outside the organization could start the stack. Not a degraded start: the
pull fails, compose aborts, and no container comes up. `docs/sandbox.md` says the
daemon is "shipped as `ghcr.io/vstorm-co/sandboxd`", which was true of the name
and not of the access.

Found when the production deploy failed on the same pull — the host had no GHCR
credentials either, which is the same defect wearing a deployment's clothes.

**Already fixed**: the package is public and pulls anonymously (`docker pull
ghcr.io/vstorm-co/sandboxd:0.2` with no login; manifest answers 200). The
organization also had public package creation disabled, which greyed out the
per-package toggle — that is off now too.

Filed rather than closed silently because the state was invisible from inside the
organization: everyone who could test it was authenticated. Worth two things:

- a check that catches it — the quickstart or `agenticos cmd doctor` proving the
  sandbox image is reachable, rather than a compose abort naming a registry;
- the same question asked of anything else this repository points at by URL.

Found operating the deployment from #1476, in the same sitting as #1504 and
#1506.

PR #1546, opened 2026-09-10
feat(compose): publish the images on every release and pull them

### LINKED ISSUE #1544: Frontend: NEXT_PUBLIC_* are baked at build time, so a published frontend image cannot exist
`frontend/Dockerfile` takes `NEXT_PUBLIC_API_URL`, `NEXT_PUBLIC_WS_URL`,
`NEXT_PUBLIC_SITE_URL`, `NEXT_PUBLIC_CHAT_MAX_UPLOAD_SIZE_MB`,
`NEXT_PUBLIC_OAUTH_PROVIDERS` and `NEXT_PUBLIC_RAG_ENABLED` as **build
arguments**, and Next inlines them into the browser bundle. So one built image
carries one deployment's hostnames, and a published `agenticos-frontend` image
would send every browser to `localhost:8000`. The compose files and
`docs/install.md` warn about exactly this ("changing one means rebuilding the
image, not restarting it"), which is the right warning about the wrong design.

Six variables, and only the browser-side ones are the problem:

| Variable | Read by | Actually browser-side? |
|---|---|---|
| `NEXT_PUBLIC_API_URL` | `constants.ts` → `hosted-chat.tsx` (embed uploads), `oauth-buttons.tsx` (provider redirect), `command-palette.tsx` (`/docs` link) | yes |
| `NEXT_PUBLIC_WS_URL` | `constants.ts` → `use-chat.ts`, `hosted-chat.tsx` | yes - a route handler cannot upgrade a WebSocket, so the browser reaches the API directly |
| `NEXT_PUBLIC_SITE_URL` | `seo.ts` → root layout metadata, `robots.ts`, `sitemap.ts` | no - server only, but `robots`/`sitemap` are prerendered at build unless told otherwise |
| `NEXT_PUBLIC_CHAT_MAX_UPLOAD_SIZE_MB` | `utils.ts` → `chat-input.tsx` | yes |
| `NEXT_PUBLIC_OAUTH_PROVIDERS` | `oauth-buttons.tsx` | yes |
| `NEXT_PUBLIC_RAG_ENABLED` | **nothing** - passed by three compose files and read by no source file | dead |

Everything HTTP already goes through the BFF proxy, whose `BACKEND_URL` is
read at runtime; only these five values are baked.

## Done when

- The root layout reads `PUBLIC_API_URL`, `PUBLIC_WS_URL`, `PUBLIC_SITE_URL`,
  `CHAT_MAX_UPLOAD_SIZE_MB` and `OAUTH_PROVIDERS` from the **server's**
  environment per request and hands them to the client through a context, the
  way `BrandingProvider` already hands down the deployment's name. Components
  read `usePublicConfig()`; the module constants `WS_URL`, `BACKEND_URL` and
  `CHAT_MAX_UPLOAD_SIZE_MB` go.
- The same names the deployed compose files already interpolate
  (`PUBLIC_API_URL`, `PUBLIC_WS_URL`, `PUBLIC_SITE_URL` - what
  `scripts/server-init.sh` writes), so `backend/.env` needs no new rows.
- `robots.ts` and `sitemap.ts` render per request, or the published image
  advertises `http://localhost:3000` in its sitemap for ever.
- `frontend/Dockerfile` has no `ARG NEXT_PUBLIC_*`; `bun run build` in it
  produces one image for every deployment. `NEXT_PUBLIC_RAG_ENABLED` is
  deleted everywhere.
- `frontend/.env.example`, `frontend/README.md`, `docs/install.md`,
  `docs/configuration.md` and `docs/file-processing.md` stop describing build
  arguments.

## Where it sits

Blocks #A (no frontend image without it). Cluster **The frontend's own API
surface** in #168 - the same runtime-vs-build boundary `BACKEND_URL` already
sits on the right side of.

### LINKED ISSUE #1545: Publish the backend and frontend images to GHCR on every release, and make docker compose up pull them
Asked by a prospective user, in as many words: *"Is vstorm-co publishing the
docker images? The docker compose setup is too complicated. That can be heavily
simplified by publishing the docker images during each release."*

## What is true today

- **No image is published.** The `docker` job in `ci.yml` builds the backend
  image on every push to `main` with `push: false`, only so Trivy can scan it,
  and the release (`chore: cut 0.0.x` + a `v*` tag) publishes nothing. The only
  image under `ghcr.io/vstorm-co/` is `sandboxd`, from another repository.
- **Every one of the nine compose files carries `build:`.** A first
  `docker compose up` compiles the backend image (LibreOffice, Tesseract, the
  whole dependency set) and, for the console, a Next.js production build;
  `scripts/quickstart.sh` says so itself: *"First run builds an image, so this is
  the slow part — a few minutes."*
- Running it at all needs a clone plus `make`, `uv` and `bun`, because
  `make dev` is the entry point and the frontend is a second compose file.
- The backend image cannot be published as it is built: its context is
  `./backend`, so `emails/compiled/` is bind-mounted from the checkout rather
  than shipped in the image.
- The frontend image cannot be published at all until #B: `NEXT_PUBLIC_*` are
  build arguments, so one image carries one deployment's hostnames.

## Done when

- A push of a `v*` tag publishes `ghcr.io/vstorm-co/agenticos-backend` and
  `ghcr.io/vstorm-co/agenticos-frontend`, tagged with the version and `latest`;
  a push to `main` publishes `edge` and `sha-<short>`. Both images build for
  `linux/amd64` and `linux/arm64` on native runners, and Trivy scans the
  published backend image the way the `docker` job scans an unpublished one now.
- Both packages are **public** and pull anonymously - the state #1512 found the
  sandbox image not in, and the one thing nobody inside the organization can see
  from inside it.
- `docker-compose.yml` pulls the published images and starts the whole product -
  API, worker, console, Postgres, Redis, Prefect - with no `.env` to write, and
  runs the migrations itself. A clone keeps building from source through
  `docker-compose.override.yml`, which Compose merges on its own, so `make dev`
  behaves as it does today.
- The production and dev-server files pull a pinned `AGENTICOS_VERSION` instead
  of building; `scripts/deploy.sh` pulls the image the commit produced rather
  than rebuilding on the host.
- `scripts/quickstart.sh` needs Docker and nothing else: no clone, no `make`,
  no `uv`, no `bun`.
- `docs/install.md`, `docs/deploy.md` and the README describe the image path
  first and the source build second.

## Where it sits

Cluster **CI, the suite and the tools** in #168. Blocked by #B (the frontend
image). Neighbours: #1512 (a GHCR package is private until somebody flips it),
#1415 (an SBOM per release - the same release event, the same artefact), #143
(`test_ci_parity.py` - the `docker` job leaves `ci.yml`, and the parity test's
out-of-scope list has to follow).

DEENUU1:
A user asked whether we publish Docker images, because "the docker compose setup
is too complicated". We did not: the `docker` job in `ci.yml` built the backend
with `push: false` for Trivy to scan, every compose file carried `build:`, and
running the product meant a clone plus `make`, `uv` and `bun`. This branch
publishes both images on every release and makes one downloaded
`docker-compose.yml` the whole product.

## What changed

**The images.** `.github/workflows/images.yml` builds
`ghcr.io/vstorm-co/agenticos-backend` and `agenticos-frontend` for amd64 and
arm64 on native runners (no QEMU over LibreOffice), pushes by digest and joins
one manifest list per tag: `<version>` + `latest` from a `v*` tag, `edge` from
`main`, `sha-<short>` from both. Trivy scans the published backend and frontend
the way the old `docker` job scanned an unpublished one; that job is gone from
`ci.yml`. No `pull_request` trigger, deliberately - the jobs hold
`packages: write`. Three rules, each added after review: **only a commit on
`main` is published** (a `resolve` job refuses anything else before a job with
`packages: write` runs, so a `v*` tag from a branch cannot move `latest`);
**one concurrency group per commit**, never per ref (a group per ref with
`cancel-in-progress: false` cancels the pending run when a third merge lands,
and that commit then has no `sha-` images to deploy - the trap `ci.yml` already
documents); and **`workflow_dispatch` takes a `sha`** to publish a commit that
has none - a rollback target older than the workflow, or a lost run. A release
on a commit `main` already built re-tags its `sha-` manifest rather than
rebuilding, so the release names the digests hosts pinned to. `backend/Dockerfile` is built from the repository root so
`emails/compiled/` is in the image instead of bind-mounted from a checkout;
`.dockerignore` at the root is an allowlist, and it excludes `backend/.env`,
which the old `backend/.dockerignore` did not.

**The frontend reads its public URLs at runtime (#1544).** `NEXT_PUBLIC_*` were
build arguments, so one image carried one deployment's hostnames. The root
layout now reads `PUBLIC_API_URL`, `PUBLIC_WS_URL`, `PUBLIC_SITE_URL`,
`CHAT_MAX_UPLOAD_SIZE_MB` and `OAUTH_PROVIDERS` from the server's environment
per request and hands them down through `PublicConfigProvider`, the way
`BrandingProvider` already hands down the deployment's name; `usePublicConfig()`
replaces the `WS_URL` / `BACKEND_URL` / `CHAT_MAX_UPLOAD_SIZE_MB` constants.
`robots` and `sitemap` are `force-dynamic` so the published image does not
advertise `localhost:3000` for ever. `NEXT_PUBLIC_RAG_ENABLED` was read by
nothing and is deleted.

**Compose.** `docker-compose.yml` pulls both images and is the product on its
own: API, worker, console, Postgres, Redis, Prefect, and a `migrate` service
the API waits on, every variable defaulted, two optional env files (`.env`
beside it, `backend/.env` in a clone). `docker-compose.override.yml` is what
Compose merges in a clone - build from the tree, bind mounts, the reload
supervisor, the console behind a `console` profile, the database ports - so
`make dev` behaves as before and `docker-compose.frontend.yml` is gone. The
dev-server and production files pull `edge` / `latest` at `AGENTICOS_VERSION`
and grow the same `migrate` service; `make prod` no longer sleeps five seconds
and hopes.

**Scripts and docs.** `scripts/quickstart.sh` needs Docker and nothing else -
the Compose plugin at 2.24 or later, which it now checks. Outside a clone it
downloads the one file **at the latest release** (so the file and the `latest`
images are one release), into `./agenticos` - never adopting another project's
`docker-compose.yml` that happens to be in the current directory - writes a
`.env` beside it with mode 0600 and a generated `SECRET_KEY`, `VAULT_MASTER_KEY`
and sandbox token, and pulls; inside a clone it builds and touches nothing.
`scripts/deploy.sh` pins `AGENTICOS_VERSION=sha-<short>`, waits for the
manifest CI publishes for that commit, and pulls - nothing is built on the host
any more. `docs/install.md`, `docs/deploy.md`, the README, `docs/commands.md`
and `docs/branching.md` describe the image path first and the source build
second.

## Verified

- `backend/tests/test_images_workflow.py`: no PR trigger, `packages: write`
  only on the pushing jobs, only a commit on `main` is published, one
  concurrency group per commit, a release re-tags rather than rebuilds, every
  job bounded, and every compose file that pulls an application image pulls one
  of the two names the workflow publishes at the `AGENTICOS_VERSION`
  interpolation.
- `quickstart.sh` dry-runs: an empty directory, a directory holding another
  project's compose file (installs beside it), one whose `./agenticos` holds a
  stranger's (refuses), one already set up (keeps it), and the clone (touches
  nothing); the 0600 mode and the key shape by hand. `test_ci_parity.py`,
  `test_ci_workflow.py`, `test_sandbox_runtime_catalog.py` green.
- `docker compose config` on every file and on base + override with every
  profile; the backend image builds from the root context locally.
- Frontend: scoped specs, `type-check`, `lint`, `format:check`, `check:i18n`,
  the coverage gate and `bun run build` (robots and sitemap render `ƒ`).
- `zizmor`, `yamlfmt`, `codespell`, the docs guards and `docs_drift.py` pass.
- Quickstart dry-runs from a clone and from an empty directory.

## Not verified here, and what to do after the merge

- **The first publish.** The workflow has never run; the first push to `main`
  is its first run, and `gh run watch` on it is the review. Then cut a release:
  `latest` does not exist until a `v*` tag, and the standalone compose file and
  `quickstart.sh` both default to it. GHCR creates a new
  package **private** (#1512) - after that run, `docker manifest inspect
  ghcr.io/vstorm-co/agenticos-backend:edge` from a logged-out shell, and flip
  both packages to public if it refuses. Until then the standalone
  `docker compose up` in the docs cannot work.
- **The production deploy** goes through `scripts/deploy.sh`, which now waits
  for the `sha-` tag and pulls. The first deploy after this merge should be
  watched. A rollback to a commit *older* than this one has no images: publish
  it first with `gh workflow run images.yml --ref main -f sha=<commit>` - the
  deploy prints that command when it gives up waiting.
- `ubuntu-24.04-arm` runners are GitHub-hosted and free for public repositories;
  if the arm64 build is refused, drop that matrix leg and the images ship amd64
  only until it is sorted.

Closes #1545
Closes #1544


.github/workflows/images.yml
```diff
@@ -0,0 +1,233 @@
...
+permissions:
+  contents: read
+
+# One publish per ref at a time, queued rather than cancelled: a manifest list
+# is assembled from four digests pushed by four jobs, and cancelling the run
+# between them leaves digests nothing references and a tag still pointing at the
+# previous build - which is harmless, but a queued run is simpler to reason about
+# than a half-finished one.
+concurrency:
+  group: images-${{ github.ref }}
```
DEENUU1: **Blocking.** One group per ref with `cancel-in-progress: false` is one running and *one* pending run: merge A running, B pending, C landing cancels B - the exact rule `ci.yml` quotes at lines 42-46 and keys `main` on `github.run_id` to avoid. B's commit then has no `sha-` images, and `scripts/deploy.sh` for it waits twenty minutes and exits 1, which closes both its auto-deploy and the rollback path to it.

```suggestion
  group: images-${{ github.ref }}-${{ github.sha }}
```

Each run pushes by digest and its own `sha-` tag, so runs do not collide; `edge`/`latest` become last-finisher-wins, which is fine for `edge` (or skip it in `publish` when `github.sha` is no longer `origin/main`'s head). Codex's thread below is the same finding.

scripts/quickstart.sh
```diff
@@ -284,74 +282,167 @@ wizard() {
...
+    ok "in an AgenticOS clone — the images are built from this tree"
     return 0
   fi
-  step "Getting the code"
-  if [ -d "$REPO_DIR" ]; then
-    note "$REPO_DIR exists — using it"
+  step "Getting docker-compose.yml"
+  # A directory this script already set up is used in place - re-running it from
+  # there must not nest another install under it.
+  if [ -f docker-compose.yml ]; then
```
DEENUU1: **Blocking.** Outside a clone this adopts *any* `docker-compose.yml` in the current directory. The README runs `curl … | bash` from wherever the reader is; from a project with its own compose file the script runs `pull`/`up -d` on that stack, appends `SANDBOXD_TOKEN` to that project's `.env`, then `exec`s `agenticos cmd bootstrap` into a service that is not ours. The old script refused here (`in_repo || die "… does not look like an AgenticOS clone"`).

Adopt only a file that is ours, otherwise fall through to `$INSTALL_DIR`:

```bash
ours() { grep -q 'ghcr.io/vstorm-co/agenticos-' docker-compose.yml 2>/dev/null; }
if ours; then
  note "docker-compose.yml already here — keeping it"
else
  run mkdir -p "$INSTALL_DIR" …
```

Same check on the `$INSTALL_DIR` branch a few lines down.

scripts/quickstart.sh
```diff
@@ -284,74 +282,167 @@ wizard() {
...
+ensure_sandbox_token() {
+  if grep -q '^SANDBOXD_TOKEN=.' "$ENV_FILE" 2>/dev/null; then
+    note "SANDBOXD_TOKEN already in $ENV_FILE"
+    return 0
+  fi
+  # Every stage reads its whole input: a `head` at the end of a pipe kills `tr`
+  # with SIGPIPE, which `set -o pipefail` turns into this script exiting.
+  local token
+  token="$(head -c 64 /dev/urandom | base64 | LC_ALL=C tr -dc 'A-Za-z0-9' | cut -c1-43)"
+  run_sh "printf '\n# Authorises opening a sandbox session, and a session runs commands\n# on this host. Treat it like the Docker socket it sits in front of.\nSANDBOXD_TOKEN=%s\n' '$token' >> '$ENV_FILE'"
```
DEENUU1: Suggestion. The installer creates `.env` here under the caller's umask, usually 0644, and the value is the one the compose file says to treat like the Docker socket. `make sandbox-token` has the same shape so this is not new, but here the file is born: `( umask 077; : >> "$ENV_FILE" )` before the append is one line. (Codex's thread above is this.)

scripts/quickstart.sh
```diff
@@ -172,10 +173,8 @@ check_prerequisites() {
...
   esac
 
-  need git
-  need make
-  need python3
   need docker
+  in_clone || need curl
 
   if command -v docker >/dev/null 2>&1; then
     if docker compose version >/dev/null 2>&1; then
```
DEENUU1, line 212: Suggestion. This proves `docker compose` exists, not that it is ≥ 2.24, which the `env_file: { path, required: false }` form needs and which `docs/install.md` now states. Ubuntu 22.04's own `docker-compose-v2` is older and fails on the compose schema rather than with a message naming the version - `docker compose version --short` and a compare would turn that into a hint like the others here.

.github/workflows/images.yml
```diff
@@ -0,0 +1,233 @@
...
+# from a shell that is not logged in, and the same for the frontend.
+name: Images
+
+on:
+  push:
+    branches: [main]
+    tags: ["v*"]
+  # For publishing `edge` before `main` has moved, and for re-running a publish
+  # whose runner died. It reads the ref it is dispatched on and nothing else.
+  workflow_dispatch:
```
DEENUU1, line 44: Suggestion. Give this a `sha` input the build checks out. It is the backfill for a commit whose pending run was cancelled (the concurrency thread) and for a rollback to anything older than this PR - nothing before it has an image, and `deploy.sh` no longer builds. `deploy.yml`'s `resolve` job already has the ancestor-of-`main` check to copy, and the same check on a `v*` tag would answer Codex's security thread on line 30 (a tag from any branch currently publishes `latest`).

docker-compose.yml
```diff
@@ -1,89 +1,115 @@
...
+# them and where the rest of this repository says to look.
 
 services:
+  # Runs the migrations and exits. `app` and `prefect-runner` wait for it to have
+  # *completed*, not for the database to be up, so a first `up` on an empty
+  # volume is one command and a second `up` on a migrated one is a no-op that
+  # takes a second. `docker compose logs migrate` is where a failed migration
+  # says why the API never started.
+  migrate:
+    image: ghcr.io/vstorm-co/agenticos-backend:${AGENTICOS_VERSION:-latest}
```
DEENUU1, line 42: Note for the merge sequence rather than a change: `latest` does not exist until a `v*` tag is pushed, and the first `Images` run creates both packages private. So merge → watch the run → flip both packages public → cut a release, and only then does the README's standalone path work. Between releases a `main`-hosted compose file and a `latest` image can drift; having `quickstart.sh` fetch the file at the latest release tag would pin the two together. I would resolve Codex's thread here with that sequence rather than default to `edge`.

docs/deploy.md
```diff
@@ -388,7 +417,7 @@ otherwise get an empty file and an error nobody reads on the way past.
 
 | | How |
 |---|---|
-| **Code** | Deploy the previous commit: `workflow_dispatch` with its sha, or `scripts/deploy.sh` |
+| **Code** | Deploy the previous commit: `workflow_dispatch` with its sha, or `scripts/deploy.sh`. The images are still in the registry, so this is a pull, not a build |
```
DEENUU1: Nit. True from this PR onward, but a rollback to any commit older than it has no image, and `deploy.sh` will wait twenty minutes and fail rather than build. One clause here - "for commits published after images.yml landed; older ones need a dispatch of `Images` on that sha" - once that input exists.

backend/tests/test_images_workflow.py
```diff
@@ -0,0 +1,134 @@
...
+        """`docker compose up` on the one file has to be the whole product, console included (#1545)."""
+        text = (REPO_ROOT / "docker-compose.yml").read_text()
+        assert {image for image, _ in _IMAGE_LINE.findall(text)} == PUBLISHED
+
+    def test_the_override_builds_rather_than_pulls(self) -> None:
+        override = yaml.safe_load((REPO_ROOT / "docker-compose.override.yml").read_text())
+        services = override["services"]
+        assert "build" in services["app"] and "build" in services["frontend"]
+        for name, service in services.items():
+            assert not str(service.get("image", "")).startswith("ghcr.io/"), (
```
DEENUU1: Nit. This is the CodeQL alert on the PR (incomplete URL substring sanitization). It is a compose image name, not a URL, so a false positive - dismiss it as such, or `image.partition("/")[0] == "ghcr.io"` if you would rather not carry the alert.

DEENUU1 (commented):
Reviewed at `bec00828`. Local tree matches the head; CI on that sha is green on every required check (lint, test, test-frontend, e2e, docs, Security Scan). CodeQL is red on one alert in a test file (below) and is not a required check.

**Outcome: two things I would fix before merging, both small.** The design is right and the frontend half of it is verified end to end: `bun run build` on this head renders every page route `ƒ` (only `/icon` is static), so `readPublicConfig(process.env)` in the root layout really is read per request and a published image cannot bake a hostname. The compose split (base pulls, override builds) validates standalone and merged with every profile. What follows is what the diff itself does not yet cover.

## Blocking

### 1. `images.yml` reintroduces the one-pending-run trap `ci.yml` documents

`concurrency: group: images-${{ github.ref }}` with `cancel-in-progress: false` means one running and **one** pending run per group. Merge A running, B pending, C lands: GitHub cancels B - the rule `ci.yml` quotes at lines 42-46, which is why `ci.yml` keys `main` on `github.run_id`. B's commit never gets its `sha-` images, and `scripts/deploy.sh` for B now waits twenty minutes and exits 1 - both the auto-deploy and the documented rollback path to B are closed. Batch merges of three or four PRs are how this repository lands work, so this is not a corner case.

Fix: key the group on the commit, the way `ci.yml` does - `group: images-${{ github.ref }}-${{ github.sha }}` - or drop the block. Each run pushes by digest and its own `sha-` tag, so runs do not interfere; the only shared state is `edge`/`latest`, where the last run to finish wins. That is acceptable for `edge`; if you want it exact, have `publish` skip `edge` when `github.sha` is no longer `origin/main`'s head. Codex's thread on line 48 is the same finding - resolve it with this.

### 2. Outside a clone, `quickstart.sh` adopts *any* `docker-compose.yml` in the current directory

`obtain_compose` checks `in_clone` (override + `backend/`), then `[ -f docker-compose.yml ]` and keeps whatever it finds (`scripts/quickstart.sh:311`). The README says to run `curl … | bash` from wherever you are; a developer running it from a project with its own compose file gets `docker compose pull && up -d` on *that* stack, a `SANDBOXD_TOKEN` appended to *that* project's `.env`, and then `docker compose exec app agenticos cmd bootstrap` against a service that is not ours. The old script refused here (`in_repo || die "… does not look like an AgenticOS clone"`); this one does not.

Fix: adopt an existing file only when it is ours, otherwise fall through to `$INSTALL_DIR`:

```bash
ours() { grep -q 'ghcr.io/vstorm-co/agenticos-' docker-compose.yml 2>/dev/null; }
if ours; then
  note "docker-compose.yml already here — keeping it"
else
  run mkdir -p "$INSTALL_DIR" && cd "$INSTALL_DIR" …
```

Same check on the `$INSTALL_DIR` branch (line 318), and a test line in `--dry-run` is enough to prove it.

## Suggestions (non-blocking)

- **Merge sequence, because `latest` does not exist until a `v*` tag.** The downloaded `main` compose file defaults to `latest`; after this merge nothing has published it, and the first run also creates both packages private. So: merge → watch the `Images` run → flip both packages public → cut a release → then the README's standalone path works. Between releases a `main` compose file with a `latest` image can drift from each other; a cheap guard is for `quickstart.sh` to fetch `docker-compose.yml` at the latest release tag rather than at `main`. Codex's `docker-compose.yml:42` thread is this; I would resolve it with the sequence rather than change the default to `edge`.
- **Give `images.yml`'s `workflow_dispatch` a `sha` input** that the build checks out. It is the backfill path for both a commit whose pending run was cancelled and a rollback to a commit older than this PR (nothing before it has an image, and `deploy.sh` will not build). `deploy.yml`'s `resolve` job already has the "is it an ancestor of `main`" check to copy. Codex's `deploy.sh:138` thread is this gap; a one-row note in `docs/deploy.md`'s rollback table would close the documentation half.
- **A `v*` tag from any branch publishes `latest`.** The tag ruleset only forbids update/delete. Either restrict `v*` creation in the ruleset, or have `publish` refuse a tag whose commit is not an ancestor of `main` (again the `resolve` pattern). Cheap, and it answers Codex's security thread on `images.yml:30`.
- **`.env` created by the installer is mode 0644** (`quickstart.sh:343`), and it holds the sandbox token, which the compose file describes as worth the Docker socket. `make sandbox-token` has the same shape, so this is not new, but the installer creates the file from nothing - `( umask 077; : >> "$ENV_FILE" )` before the append costs one line.
- **The installer checks that `docker compose` exists, not that it is ≥ 2.24** (`quickstart.sh:180`). The `env_file: { path, required: false }` form needs 2.24, and the docs now say so; Ubuntu 22.04's own `docker-compose-v2` is older and fails with a schema error rather than a message naming the version.
- **Standalone install runs with the default `SECRET_KEY` and no `VAULT_MASTER_KEY`.** Pre-existing for `make dev`, but this PR makes it the advertised product path, and the README says keys are "stored encrypted" - encrypted under a constant that is in the repository. `server-init.sh` already generates both; the installer could write them into `.env` beside the sandbox token.

## Nits

- `ci.yml:434` still sets `NEXT_PUBLIC_WS_URL` for the e2e job; nothing reads it now. Harmless only because the default equals it - rename to `PUBLIC_WS_URL` so the comment above it stays true.
- CodeQL's "incomplete URL substring sanitization" on `test_images_workflow.py:132` is a false positive (a compose image name, not a URL). Dismiss it as such, or `image.partition("/")[0] == "ghcr.io"` if you would rather not carry the alert.
- A `v*` tag rebuilds the commit `main` already built minutes earlier. `imagetools create -t <version> -t latest <image>:sha-<short>` on a tag event would publish the release in seconds and keep the digests identical.

## Verified

- `uv run pytest tests/test_images_workflow.py tests/test_ci_parity.py tests/test_ci_workflow.py tests/test_reload_supervisor.py tests/test_sandbox_runtime_catalog.py` - 114 passed.
- `bunx vitest run` on the five specs the frontend change touches - 52 passed.
- `docker compose config` on the base file alone in an empty directory, and on base + override with `sandbox` and `console` - both valid; the merged stack runs `agenticos_backend:dev` under the reload supervisor, the standalone one pulls `:latest`.
- `bun run build` from `frontend/` - every page route `ƒ`, `robots` and `sitemap` included.

Not verified here: an actual `make dev` or image build from the root context, `make lint`, the coverage gates (CI is green for those), and the workflow itself, which has never run.

backend/tests/test_images_workflow.py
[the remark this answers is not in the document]
DEENUU1: False positive - a compose image name, not a URL - but gone anyway: b8ad31bb compares the registry (`image.partition("/")[0] == "ghcr.io"`) instead of a prefix.

docker-compose.yml
```diff
@@ -1,89 +1,106 @@
...
```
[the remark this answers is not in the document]
DEENUU1: Agreed on the gap, resolved differently: the default stays `latest` (the release is the product), and `quickstart.sh` now fetches `docker-compose.yml` at the latest release tag rather than at `main` (f61fad46), so the file and the images it pulls are one release. The window before the first post-merge release is closed by cutting that release right after the merge - same window as the packages being private, same fix.

scripts/deploy.sh
```diff
@@ -110,24 +118,53 @@ git fetch --prune --quiet origin
...
+
+# `images.yml` starts when the commit lands on `main`, the same moment this
+# workflow does, and finishes some minutes later - so an approval given at once
+# would pull a tag that does not exist yet. Wait for the manifest rather than
+# fail on it; twenty minutes is a stalled publish, not a slow one.
+say "Waiting for the images tagged $AGENTICOS_VERSION"
+for image in agenticos-backend agenticos-frontend; do
+  ref="ghcr.io/vstorm-co/$image:$AGENTICOS_VERSION"
+  for attempt in $(seq 1 80); do
+    if docker manifest inspect "$ref" >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
DEENUU1: Fixed in b8ad31bb: `images.yml`'s `workflow_dispatch` takes a `sha`, checks it is on `main`, and publishes that commit's `sha-<short>` tag and nothing that moves - the backfill for a commit older than the workflow or one whose run was lost. `deploy.sh` prints the exact `gh workflow run` when it gives up waiting, and `docs/deploy.md`'s rollback row says the same.

.github/workflows/images.yml
```diff
@@ -0,0 +1,233 @@
...
+  contents: read
+
+# One publish per ref at a time, queued rather than cancelled: a manifest list
+# is assembled from four digests pushed by four jobs, and cancelling the run
+# between them leaves digests nothing references and a tag still pointing at the
+# previous build - which is harmless, but a queued run is simpler to reason about
+# than a half-finished one.
+concurrency:
+  group: images-${{ github.ref }}
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
DEENUU1: Fixed in b8ad31bb: the concurrency group is the commit (`images-${{ inputs.sha || github.sha }}`), the rule `ci.yml` already keys `main` on `run_id` for. Runs neither queue nor cancel; `publish` checks where `main` is before moving `edge`, so a run for an older commit that finishes last leaves `edge` alone. Test: `TestOnePublishPerCommit`.

scripts/quickstart.sh
```diff
@@ -284,36 +282,124 @@ wizard() {
...
+  if grep -q '^SANDBOXD_TOKEN=.' "$ENV_FILE" 2>/dev/null; then
+    note "SANDBOXD_TOKEN already in $ENV_FILE"
+    return 0
   fi
-  ok "code in $(pwd)"
+  # Every stage reads its whole input: a `head` at the end of a pipe kills `tr`
+  # with SIGPIPE, which `set -o pipefail` turns into this script exiting.
+  local token
+  token="$(head -c 64 /dev/urandom | base64 | LC_ALL=C tr -dc 'A-Za-z0-9' | cut -c1-43)"
+  run_sh "printf '\n# Authorises opening a sandbox session, and a session runs commands\n# on this host. Treat it like the Docker socket it sits in front of.\nSANDBOXD_TOKEN=%s\n' '$token' >> '$ENV_FILE'"
```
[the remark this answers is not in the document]
DEENUU1: Fixed in f61fad46: the env file is created with `umask 077` before anything is appended, values go to `printf` as argv rather than through `sh -c`, and `--dry-run` masks them.

.github/workflows/images.yml
```diff
@@ -0,0 +1,233 @@
...
+# the organization can tell from inside it - every pull they try is
+# authenticated. #1512 found the sandbox image in that state after a month.
+# After the first run: `docker manifest inspect ghcr.io/vstorm-co/agenticos-backend:edge`
+# from a shell that is not logged in, and the same for the frontend.
+name: Images
+
+on:
+  push:
+    branches: [main]
+    tags: ["v*"]
```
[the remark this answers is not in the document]
DEENUU1: Fixed in b8ad31bb: a `resolve` job runs before anything holding `packages: write`, refuses a commit that is not an ancestor of `origin/main`, and `build` checks out the commit it resolved - so a `v*` tag from a branch publishes nothing. Test: `test_only_a_commit_already_on_main_is_published`. Restricting `v*` creation in the tag ruleset would be a second lock; the workflow no longer depends on it.

.github/workflows/images.yml
DEENUU1: Fixed in b8ad31bb - group per commit, `edge` decided against `main`'s head at publish time.

scripts/quickstart.sh
DEENUU1: Fixed in f61fad46 - a file is adopted only when it names our images; otherwise the install goes to `$INSTALL_DIR`, and a stranger's file there is refused with a pointer to `--dir`. Dry-runs for all four cases in the commit body.

scripts/quickstart.sh
DEENUU1: Fixed in f61fad46 - created 0600, appended as argv, masked under `--dry-run`.

scripts/quickstart.sh
DEENUU1: Fixed in f61fad46 - `check_compose_version` reads `docker compose version --short` and refuses below 2.24, naming the version and the fix.

.github/workflows/images.yml
DEENUU1: Fixed in b8ad31bb - `sha` input, on-`main` check in `resolve`, and a release on a built commit re-tags the existing `sha-` manifest instead of rebuilding.

docker-compose.yml
DEENUU1: Kept `latest`; `quickstart.sh` fetches the compose file at the latest release (f61fad46). The merge sequence stands: merge → watch the `Images` run → flip both packages public → cut a release.

docs/deploy.md
DEENUU1: Fixed in b8ad31bb - the row names the backfill command, and the deploy prints it when it gives up.

backend/tests/test_images_workflow.py
DEENUU1: Fixed in b8ad31bb.

DEENUU1:
The two review-fix commits were rewritten to drop a commit trailer; the trees are identical. Thread replies name the old shas: `b8ad31bb` is now `30fdcaf5` (the workflow) and `f61fad46` is now `5be20cc0` (the quickstart). CI was green on the old shas and is re-running on the new ones.

PR #1569, opened 2026-09-11
fix(agents): hold the stale-reference notice until its lists load

### LINKED ISSUE #1472: agents: the stale-reference warning flashes on every agent page load, before the lists it checks have answered
## What you see

Open (or refresh) any agent whose spec references a collection, an MCP
connection, a context file or a skill. For about a second the page shows the
orange warning above the tabs:

> This draft still references … this organization no longer has. Publishing is
> refused until they are cleared.

Then it disappears on its own. Nothing is actually stale.

## Why

`staleReferences` (`src/components/agents/stale-references.tsx:44`) decides a
reference is dead by asking whether its id is in a list. **It cannot tell "the
organization has none" from "the query has not answered yet"**, and every one of
those lists defaults to `[]` while loading:

```
page.tsx:186  const { kbs: collections } = useKnowledgeBases();   // isLoading not read
page.tsx:197  const { connections: mcpConnections } = useOrgMcpConnections();
page.tsx:206  const { servers: mcpCatalog } = useMcpCatalog();
```

`useKnowledgeBases` returns `data: kbs = []` and an `isLoading` the page does not
destructure. So on first paint every `spec.collection_ids` entry is missing from
an empty set, the notice renders, the query resolves and it unmounts.

The component's own docstring names the assumption that fails: *"Collections and
connections arrive whole"*. True once loaded, false for the first render.

The guard the other two kinds have does not close it either. `contextComplete` is
`contextTotal <= contextFiles.length`, which is `0 <= 0` while loading - so
context files and skills flash for the same reason, by a different route.

## Why it matters more than a flicker

The sentence it flashes is *"Publishing is refused until they are cleared."* It is
an accusation about the agent's configuration, in the alarm colour, on every load
of a page people open all day. It also carries a button that edits the draft, so a
fast click during the flash would strip live references out of a spec.

Refresh an agent that references a collection ten times: no notice at any point.
And with a genuinely deleted collection, the notice still appears once the lists
have loaded.

## Fix shape

Pass the loading state in and render nothing - and compute nothing - until every
list the check consults has answered. A `known.loaded: boolean` on
`StaleReferenceSet`'s input is enough; the four ids should not be computed from a
list that is merely absent.

Found while working on #1470 (memory), which does not touch this file: the branch
changes `agents/[id]/page.tsx` only where the Memory tab was wired. Pre-existing.
Nearest existing issue: #909, the same class - a list read before it has answered
showing the wrong thing.

Refs #168

OchnikBartek:
## What this fixes

Opening or refreshing any agent whose spec references a collection, context file, skill or MCP connection flashed the orange stale-reference banner above the tabs for about a second:

> This draft still references … this organization no longer has. Publishing is refused until they are cleared.

Nothing was actually stale. `staleReferences` reads a reference as dead when its id is absent from a list — but each list it consults (`useKnowledgeBases`, `useContextFiles`, `useSkills`, `useOrgMcpConnections`, `useMcpCatalog`) defaults to `[]` while its query loads, so on first paint every reference resolved to nothing, the banner rendered, and it unmounted once the queries answered. The `contextComplete = contextTotal <= contextFiles.length` guard the file already had did not close it either: `0 <= 0` holds while loading.

It mattered more than a flicker: the banner is in the alarm colour, accuses the agent's configuration on every load, and carries the button that strips those references from the draft — so a fast click during the flash would have removed **live** references.

## The change

- `staleReferences` takes a `loaded` flag and computes nothing (returns an empty set) until every list it consults has answered. Absent-because-still-loading is no longer read as absent-because-deleted.
- `agents/[id]/page.tsx` derives `loaded` from the five hooks' `isLoading` and passes it to `StaleReferences`.

A genuinely deleted reference still surfaces, once the lists are whole.

## Verification

- `stale-references.test.tsx`: added a function-level case (genuinely stale ids report nothing while `loaded: false`) and a component-level case (renders nothing while loading); the existing "still names a deleted collection/file/skill/connection once loaded" cases stay green. 8/8 pass, the changed file at 100% lines/statements/functions.
- `tsc`, `eslint`, `prettier` and the i18n guard pass (pre-commit).

Closes #1472

### DIFF
--- a/frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx
+++ b/frontend/src/app/[locale]/(dashboard)/agents/[id]/page.tsx
@@ -179,13 +179,23 @@ export default function AgentBuilderPage({ params }: PageProps) {
   // The Builder holds the set rather than paging it: the gallery has to know
   // which selected skills still exist, and it can only tell that from what it
   // has. 100 is the endpoint's ceiling; `total` says when that is not all.
-  const { skills, total: skillCount, isLoading: skillsLoading } = useSkills({ limit: 100 });
+  const {
+    skills,
+    total: skillCount,
+    isLoading: skillsLoading,
+    error: skillsError,
+  } = useSkills({ limit: 100 });
   const {
     files: contextFiles,
     total: contextCount,
     isLoading: contextLoading,
+    error: contextError,
   } = useContextFiles({ limit: 100 });
-  const { kbs: collections, isLoading: collectionsLoading } = useKnowledgeBases();
+  const {
+    kbs: collections,
+    isLoading: collectionsLoading,
+    listError: collectionsError,
+  } = useKnowledgeBases();
   // Only the newest, and only for the number below: the history card pages the
   // rest itself, so a page that fetched fifty to read one would be fetching a
   // list nothing on it renders.
@@ -200,6 +210,7 @@ export default function AgentBuilderPage({ params }: PageProps) {
     connections: mcpConnections,
     test: probeMcpConnection,
     isLoading: connectionsLoading,
+    error: connectionsError,
   } = useOrgMcpConnections();
   const [connectingServer, setConnectingServer] = useState<McpCatalogEntry | null>(null);
   const [toolPicker, setToolPicker] = useState<ToolPickerState | null>(null);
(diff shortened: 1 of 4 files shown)


OchnikBartek:
@codex review

Issue #700, opened 2026-08-13
change_role lets an Admin demote a peer Admin that remove refuses to remove


DEENUU1:
`MemberService.remove` refuses an Admin removing another Admin. `change_role`
does not refuse an Admin *demoting* another Admin to `viewer`, which strips the
same authority and leaves the row behind.

## What breaks

The two methods disagree about the same peer:

- `backend/app/services/member.py:138` — `if requester.role == OrgRole.ADMIN.value
  and target.role == OrgRole.ADMIN.value: raise AuthorizationError(...)`.
- `backend/app/services/member.py:76-84` — `change_role` inspects the target only
  for `owner`. Below that line the target's role is never consulted, so the
  ceiling bounds what may be *offered* and says nothing about who it is offered
  *to*.

An Admin who may not remove a peer Admin can therefore demote them to Viewer and
then remove them, or simply leave them demoted. The audit entry reads
`member.role_changed`, which is true but not what happened.

## The sequence

1. Two Admins, A and B, in one organization.
2. A calls `PATCH /orgs/{org_id}/members/{B}` with `{"role": "viewer"}`.
3. `viewer` is in `assignable_roles("admin")`, so the ceiling passes. B is now a
   Viewer, having been un-removable a moment earlier.

## Why it is worth fixing

The refusal at `:138` exists because peer Admins are not each other's to
administer. Demotion is the same decision with a different verb, so the pair as
it stands is a rule enforced on the path somebody would take by accident and not
on the path somebody would take deliberately.

`assignable_roles` (added by #672, `app/core/permissions.py`) answers "what may
this requester offer". The missing half is "whose role may this requester touch",
and the natural rule is the same one: a requester may only change the role of a
member their own role strictly outranks - which also lets the `owner` special case
at `:76` and the Admin-vs-Admin rule at `:138` be stated once rather than three
times.

## How I would know it was fixed

A test in `backend/tests/test_services_members.py` asserting that an Admin
requester with an Admin target is refused by `change_role`, sitting beside
`test_remove_admin_cannot_remove_admin`, which already asserts the other half.

## Context

Pre-existing; found reviewing my own diff on #672 (PR #697), which rewrote the
ceiling in this method but deliberately did not widen its scope. Same cluster as
#672, #551 and #696 in the permission model - "who may hand out which role" - and
this is the fourth: #551/#671 the invitation schema, #672 the membership service,
#696 the invitation service, this one the membership *target*. Nothing blocks it,
[...]

PR #1573, opened 2026-09-11
fix(permissions): decide member removal by the catalog, not an admin literal

### LINKED ISSUE #1066: MemberService.remove decides the Admin-vs-Admin rule with a literal, not the catalog
## What breaks

`MemberService.remove` decides the Admin-vs-Admin rule with a literal:

```python
if requester.role == OrgRole.ADMIN.value and target.role == OrgRole.ADMIN.value:
    raise AuthorizationError(message="Admin cannot remove another Admin")
```

`change_role` decides the same question with the catalog instead, since #700:

```python
if target.role not in assignable_roles(requester.role):
    raise AuthorizationError(...)
```

So the two methods answer "may this requester act on this member" by two
different mechanisms. For the four built-in roles they agree, which is why
nothing is failing: `assignable_roles("admin")` excludes `admin` and `owner`, and
the Owner case is caught by the branch above. There is no live defect today.

## Why it matters anyway

#700's whole argument was that the Admin-vs-Admin rule should *be* the assignment
ceiling rather than a second rule beside it — because `assignable_roles` is
derived from the permission catalog, so a custom role is bounded automatically.
Half of that landed. A custom role (Phase 2) with `members:manage` and a
permission set that does not outrank an Admin's would be refused by `change_role`
and allowed by `remove`, which is the same hole #700 closed, in the other method.

## The sequence that would trigger it

1. A deployment defines a custom role holding `members:manage` whose permission
   set does not strictly outrank `admin`.
2. A member with that role calls `DELETE /orgs/{id}/members/{user_id}` against an
   Admin.
3. `requester.role == "admin"` is false, so the literal does not fire and the
   Admin is removed.

`remove` asks `target.role not in assignable_roles(requester.role)`, the Owner
branch above it keeps its own message and its last-Owner check, and
`test_remove_admin_cannot_remove_admin` still passes unchanged — the built-in
roles must behave exactly as they do now.

`backend/app/services/member.py:170`.

Pre-existing; found sweeping the siblings of #1017. Not folded in, because it
changes a refusal for roles that do not exist yet and #1017 is a fix with a
behaviour reversal in it already.

Related: #700 (the half that landed), and the members cluster in #168.

OchnikBartek:
## The drift

`MemberService.remove` decided the Admin-vs-Admin rule with a literal:

```python
if requester.role == OrgRole.ADMIN.value and target.role == OrgRole.ADMIN.value:
    raise AuthorizationError(message="Admin cannot remove another Admin")
```

`change_role` decides the same question — "may this requester act on this member" — with the catalog instead, since #700: `target.role not in assignable_roles(requester.role)`. For the four built-in roles the two agree (there is no live defect), but #700's whole point was that the Admin ceiling should *be* the assignment ceiling, so a custom role is bounded automatically. Half of it landed. A custom role (Phase 2) holding `members:manage` whose permission set does not strictly outrank an Admin's would be refused by `change_role` and **allowed by `remove`** — the same hole #700 closed, still open on this side.

## The change

`remove` now bounds every non-Owner target by `assignable_roles(requester.role)`, the same ceiling `change_role` uses. The Owner branch above it is unchanged — its last-Owner check and its "Only an Owner can remove another Owner" message stay — and it is decided *before* the ceiling, because an Owner may remove a peer Owner and `assignable_roles` never offers Owner (an `elif`).

## Verification

- `test_remove_admin_cannot_remove_admin` passes **unchanged** — the built-in roles behave exactly as before.
- A new test, `test_a_custom_role_that_does_not_outrank_an_admin_cannot_remove_one`, gives a custom role `members:manage` (so it clears the requester gate) whose set does not outrank Admin and asserts removing an Admin is refused. It **fails against the literal** (I reverted the fix locally to confirm) and passes with the catalog.
- `make test` — 100.00% coverage, gate reached.

No behaviour change for the built-in roles, and no catalog change; `docs/permissions.md` is unaffected (`scripts/docs_drift.py` reports nothing owed).

Closes #1066

### DIFF
--- a/backend/app/services/member.py
+++ b/backend/app/services/member.py
@@ -172,9 +172,18 @@ class MemberService:
                 )
             if requester.role != OrgRole.OWNER.value:
                 raise AuthorizationError(message="Only an Owner can remove another Owner")
-
-        if requester.role == OrgRole.ADMIN.value and target.role == OrgRole.ADMIN.value:
-            raise AuthorizationError(message="Admin cannot remove another Admin")
+        # Every other target is administrable only by a role that strictly
+        # outranks their current one - the catalog-derived ceiling `change_role`
+        # applies, so a custom role that does not outrank an Admin cannot remove
+        # one either, rather than the literal `admin`-vs-`admin` rule that left
+        # that hole open on this side (#700, #1066). The Owner case is decided
+        # above, because an Owner may remove a peer Owner and `assignable_roles`
+        # never offers Owner.
+        elif target.role not in assignable_roles(requester.role):
+            raise AuthorizationError(
+                message="You cannot remove a member your own role does not outrank",
+                details={"target_role": target.role},
+            )
 
         removed_role = target.role
         await member_repo.delete(self.db, target)
--- a/backend/tests/test_services_members.py
+++ b/backend/tests/test_services_members.py
@@ -299,6 +299,34 @@ class TestMemberService:
             await service.remove(uuid.uuid4(), uuid.uuid4(), requester_id=uuid.uuid4())
 
     @pytest.mark.anyio
+    async def test_a_custom_role_that_does_not_outrank_an_admin_cannot_remove_one(
+        self, service, monkeypatch
+    ):
+        """The hole the literal `admin`-vs-`admin` rule left, that `change_role`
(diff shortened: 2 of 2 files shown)


OchnikBartek:
@codex review

PR #1574, opened 2026-09-11
fix(channels): give Telegram the same chat_type vocabulary as the others

### LINKED ISSUE #556: chat_type vocabulary differs per adapter, so the column holds mixed vocabularies
`telegram.py:326,425` emit the raw platform `chat.type` (`"supergroup"`, `"channel"`) while `slack.py:422` and `mattermost.py:662,742` collapse to `"private"`/`"group"`. Contract at `base.py:95`; persisted to `channel_sessions.chat_type`; one live consumer at `router.py:452` (`!= "private"`, accidentally safe).

**How it fails.** The stored `chat_type` holds a different vocabulary per platform. The first consumer that writes `chat_type == "group"` — the natural thing given the field — silently mishandles every Telegram supergroup and channel.

**Fix.** Normalise Telegram to `"private"`/`"group"` in the parser (private→private, else→group), matching the other two; or make all three emit the finer distinction and update the consumer. Pick one and pin it in a test.

**Done when.** All three adapters emit the same vocabulary; a test pins it per platform.

---

OchnikBartek:
## The mixed column

`telegram.py` emitted the raw platform `chat.type` (`supergroup`, `channel`) while `slack.py` and `mattermost.py` collapse everything but a DM to `group`. So `channel_sessions.chat_type` held a **different vocabulary per platform**. The one live consumer (`router.py`, `!= "private"`) is accidentally safe, but the first consumer to write the natural `chat_type == "group"` would silently mishandle every Telegram supergroup and channel.

## The change

I picked normalisation over the finer distinction — the two-word vocabulary the other two adapters already emit:

- **Parser.** The Telegram parser now folds every room type to `group` and keeps `private` (`"private" if chat.get("type") == "private" else "group"`), matching Slack and Mattermost. The contract comment on `IncomingMessage.chat_type` narrows to `"private" | "group"`.
- **Backfill.** Migration `0076` folds the rows already written — `UPDATE channel_sessions SET chat_type = 'group' WHERE chat_type NOT IN ('private', 'group')` — so the column stops holding `supergroup`/`channel` beside `group`. Every value that is not `private` was a room, exactly the reading the live consumer takes, so **no row changes meaning** and the consumer is not disturbed. Data-only and irreversible (the folded-away platform distinction is not recorded), so downgrade is a no-op.

## Verification

- New `test_channel_chat_type.py` pins Telegram's mapping per input type (`private`→private; `group`/`supergroup`/`channel`→group). Mattermost's own mapping is already pinned in `test_mattermost_channel.py`, and Slack's `im`→private/else→group is unchanged.
- The channel router/addressing/mentions suites (205 tests) stay green — the `!= "private"` consumer and the `memory_room_key` "anything not private is a room" logic are unaffected.
- `make test` — 100.00% coverage, gate reached. The migration is data-only with a single linear head (`0076` → `0075`); it runs no local DB here, so **CI `test-migrations` is the gate** for the forward/back cycle.

Closes #556

### DIFF
--- a/backend/alembic/versions/0076_normalize_channel_chat_type.py
+++ b/backend/alembic/versions/0076_normalize_channel_chat_type.py
@@ -1,0 +1,37 @@
+"""Normalise the mixed `channel_sessions.chat_type` vocabulary to two words.
+
+Telegram used to persist its own `supergroup` and `channel` while Slack and
+Mattermost collapsed everything but a DM to `group`, so the column held a
+different vocabulary per platform. The adapters now all emit `private` or
+`group` (#556); this folds the rows written before that so a consumer reading
+`chat_type == "group"` sees every room, not only the ones Slack and Mattermost
+wrote. Every value that is not `private` was a room, which is exactly the reading
+the one live consumer already took (`!= "private"`), so no row changes meaning.
+
+Revision ID: 0076_normalize_channel_chat_type
+Revises: 0075_user_credential_version
+Create Date: 2026-09-11
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+revision: str = "0076_normalize_channel_chat_type"
+down_revision: str | None = "0075_user_credential_version"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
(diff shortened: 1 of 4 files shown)


OchnikBartek:
@codex review

Issue #1444, opened 2026-09-05
The personal-service briefing builds a frontend URL in the runner


DEENUU1:
`personal_service_gap` (`backend/app/services/agent_runner.py:1282`) builds `"{FRONTEND_URL}/mcp-servers?connect={key}"` and hands that string both to the model's briefing and to the chat card, which opens it with `window.open` (`frontend/src/components/chat/connect-services-card.tsx:95`).

**What breaks.** The console's URL layout now has a second home in a Python service. The frontend already owns the route and the parameter name (`CONNECT_PARAM`, `frontend/src/components/mcp/mcp-server-list.tsx:60`). A locale-prefixed deployment serves the page at `/pl/mcp-servers`, so the link the agent quotes 404s — and nothing tests it. The card also leaves the app to reach its own page rather than navigating.

**Why it is not simply wrong.** A channel reply genuinely needs an absolute URL: the reader is in Slack. The chat does not.

**Fixed when** the frame carries `catalog_key` and `gap` and the chat derives its own href (it already resolves the catalog entry), the model's briefing names the page in words for the dashboard, and any absolute URL a channel needs is built next to `channel_link.py` rather than inline in the runner.

Found reviewing #1392 (`Part-of #1391`).

PR #1575, opened 2026-09-11
fix(mcp): decide a personal connection's usability once, on the server

### LINKED ISSUE #1443: ownAccountStatus re-implements the server's account rule, and they disagree
`ownAccountStatus` (`frontend/src/lib/mcp-servers.ts:339`) decides whether a person can reach a service through their own account. So does the backend, in `_nominated` + `_resolve_auth_headers` (`backend/app/services/mcp_connection.py`). The two already disagree.

**What breaks.** The server reports `unauthorized` whenever `_resolve_auth_headers` yields `None` — an OAuth grant that no longer refreshes *or* a bearer token that no longer unseals. The client says `unauthorized` only for `auth_type === "oauth" && !oauth_authorized`. A token connection whose ciphertext no longer opens therefore reads **Connected** in the chat's controls, and the next turn's `personal_services_unavailable` frame says it is not.

**Sequence.** Rotate `SECRET_KEY` (or hand-edit the sealed token). Open the chat controls: "Connected". Ask the agent for that service: it says the connection needs authorizing again.

**Fixed when** one side computes it. The backend already does, per turn; exposing it once — a `status` per `catalog_key` on `GET /me/mcp-connections`, or a small per-agent endpoint — and deleting the client copy makes the settings panel and the run agree by construction.

Found reviewing #1392 (`Part-of #1391`).

OchnikBartek:
## The drift

`ownAccountStatus` (`frontend/src/lib/mcp-servers.ts`) decided whether a person can reach a service through their own account, and so does the backend in `_resolve_auth_headers`. They disagreed: the server reports **unauthorized** whenever `_resolve_auth_headers` yields no headers — an OAuth grant that no longer refreshes *or* a bearer token that no longer unseals — while the client said unauthorized only for `auth_type === "oauth" && !oauth_authorized`. So a token connection whose sealing key was rotated away read **Connected** in the chat controls and unavailable in the very next turn's `personal_services_unavailable` frame.

## The change — one place decides usability

- **Backend.** `McpConnection.account_authorized` is now the single, **side-effect-free** answer to "can this credential be used": OAuth is authorized once its payload is written; a bearer token while the master key that sealed it is still configured (`vault.is_key_version_available`); a connection with no token is usable. `_resolve_auth_headers` uses it (so a run and this answer cannot drift), and `McpConnectionRead` carries it to the client as `authorized`.
- **Frontend.** `ownAccountStatus` reads `chosen.authorized` instead of re-deriving the credential check. Which connection answers — one, the default among several, or none chosen — stays a reading of the connection list here, because that is a pure function of what the client already holds and cannot drift from the server.

## A design note, for the reviewer

The issue suggested "exposing it once … and deleting the client copy." Two constraints shaped how:

- **The list read deliberately does not decrypt.** `McpConnectionRead.from_model` already avoids decrypting the OAuth payload "just to render the list". So `account_authorized` answers the bearer case by asking whether the sealing key version is still configured (the `SECRET_KEY`-rotation failure the issue describes) rather than by unsealing — no decryption on a list render. The one residual gap from perfect run-parity is a bearer token whose ciphertext was *tampered with* under a still-configured key; that reads authorized until the run actually unseals it. Hand-editing a sealed token is not an operational path, and decrypting every row on every list read is the cost the existing design refuses.
- **OAuth resolution has side effects.** `_resolve_auth_headers`'s OAuth path can spend a refresh token and write the DB, so it cannot run on a `GET`. `account_authorized` uses the same `oauth_payload is not None` signal the list already exposed — no refresh.

If you would rather have exact run-parity via a dedicated status endpoint that performs the full resolution, say so and I will build that instead.

## Verification

- Backend: new tests pin `account_authorized` (bearer no-token / configured key / dropped key; OAuth payload present / absent), `_resolve_auth_headers` refusing a token whose key is gone, `is_key_version_available`, and `McpConnectionRead.authorized`. `make test` — 100.00% coverage.
- Frontend: `ownAccountStatus` tests now assert the token-unseal drift reads unauthorized; the health-check test still reads connected (usable credential, failed probe). `bun run test:coverage` — 6132 passed, 97.71% branches ≥ the gate.

## Sibling note

The connect-services-card **test** file is also touched by PR #1572 (#1444); a merge-tree check confirms the two auto-merge cleanly (different regions). Both are on HOLD.

Closes #1443

### DIFF
--- a/backend/app/core/vault.py
+++ b/backend/app/core/vault.py
@@ -124,6 +124,18 @@ def current_key_version() -> int:
     return max(_configured_master_keys())
 
 
+def is_key_version_available(key_version: int) -> bool:
+    """Whether a secret sealed at this version can still be unwrapped.
+
+    A `SECRET_KEY` rotation, or a version dropped from `VAULT_MASTER_KEYS` before
+    every row moved off it, leaves sealed rows whose version no longer has a
+    master key - `unseal` then fails at `_master_key`. This answers that without
+    decrypting anything, so a caller can tell a usable credential from one the
+    deployment can no longer open.
+    """
+    return key_version in _configured_master_keys()
+
+
 def _master_key(key_version: int) -> str:
     """The master key that sealed envelopes at this version.
 
--- a/backend/app/db/models/mcp_connection.py
+++ b/backend/app/db/models/mcp_connection.py
@@ -49,6 +49,7 @@ from sqlalchemy import (
 from sqlalchemy.dialects.postgresql import JSONB, UUID
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
+from app.core.vault import is_key_version_available
 from app.db.base import Base, TimestampMixin
 
 if TYPE_CHECKING:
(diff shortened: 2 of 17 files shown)


OchnikBartek:
@codex review

PR #1613, opened 2026-09-14
fix(channels): refuse an unverifiable Slack event with 403, not 500

### LINKED ISSUE #555: Three channel webhook routes drift on a missing secret: 500 (Slack) vs 403 (others)
`api/routes/v1/{slack,telegram,mattermost}_webhook.py` all run load → unseal → verify → `parse_incoming` → `spawn(process_channel_event)` → 200, and disagree on the secret-missing branch: **Slack 500s** (`HTTPException`), Telegram and Mattermost return **403**.

**How it fails.** A Slack bot saved without a signing secret answers unverified events with a bodiless 500 to Slack's retrier; the other two answer 403. Three hand-rolled copies of one flow guarantee the copies diverge (the duplication angle on the AUD-004 security finding), and any change to the dispatch contract must be made three times.

**Fix.** Extract `receive_channel_webhook(platform, bot_id, request, bot_service)` with the two genuine per-platform bits as parameters (Slack's `url_verification` challenge; a "does this platform sign the body" flag). Pick 403 for secret-missing.

**Done when.** One receiver; the same misconfiguration returns the same status on all three; Slack's challenge hook still works.

---

OchnikBartek:
## What

The three channel webhook receivers drifted on a missing or unverifiable secret:

- `telegram_webhook.py` and `mattermost_webhook.py` refuse an unverified event with **403** — a bot saved without a secret is an unauthenticated endpoint that would otherwise run an agent on an organization's budget.
- `slack_webhook.py` alone raised **500** when the bot had no signing secret — a bodiless error to Slack's retrier rather than a clean refusal.

Slack now logs which bot to configure and refuses with **403** like its siblings, so all three routes answer an unverified event the same way. The path is unchanged otherwise (load → verify → parse → spawn → 200), and a wrong signature was already a 403 on all three.

`unseal_slack_signing_secret` already documented this as the caller's contract — *"a webhook that cannot be verified must refuse, and the refusal should say which bot to configure"* — so the route now honours its own helper.

## How verified

- `backend/tests/test_slack_webhook.py` — the no-signing-secret unit test now asserts 403 (was 500).
- `backend/tests/test_channel_webhook_secret.py::TestTheReceiversRefuse` — extended to cover all three receivers: a missing secret and a wrong signature each answer 403 for Telegram, Mattermost **and** Slack.
- `uv run pytest tests/test_slack_webhook.py tests/test_channel_webhook_secret.py tests/test_slack_socket_lifecycle.py tests/test_channel_webhook_urls.py -q` → all green.

## Notes

- No docs page documents these status codes, so none is owed; `architecture.md` documents the routes→services layering, which is unchanged by a status-code fix.

Closes #555


OchnikBartek:
@codex review

Issue #1616, opened 2026-09-14
Add the 'redacted' trace-content mode: PII-filter span content before export


OchnikBartek:
Follow-up to #1413, which added `ObservabilitySpec.content` with `full` and `none`. This adds the third mode from the original spec: `redacted`.

## Why

`none` protects a run by sending no message text at all, and `full` sends everything. `redacted` is the middle ground the security programme asked for: keep the message and tool content in the trace for debugging, but run the same PII filter the log pipeline uses (`app/core/logging.py`, `RedactingFilter`) over it first, so an email, a token or a key never lands in the Logfire project.

## What to build

- Widen `TraceContent` in `app/agents/spec.py` to `Literal["full", "redacted", "none"]` (widening a Literal is additive; no migration, no `SPEC_VERSION` bump — `none`/`full` specs are unaffected).
- Enforce it where the agent is instrumented (`app/agents/observability.py` / `factory.py`), not in the Builder. `include_content` is a bool on `instrument_pydantic_ai`, so `redacted` cannot be expressed by that flag alone: it needs an OpenTelemetry `SpanProcessor` that scrubs the content attributes pydantic-ai writes (the message/output/tool-argument span attributes) through `RedactingFilter._redact` before export. This is the fragile part — it depends on the exact attribute names pydantic-ai's instrumentation version emits, so pin the version behaviour with a test that feeds a known span.
- Builder: add the third option to the observability card (`observability-card.tsx`), copy in en and pl.
- A test that a `redacted` run's spans carry the message *shape* but not a planted email/token; `none` still carries neither; `full` carries both.
- `docs/reference/spec.md` (via the docstring) and `docs/environments.md` updated to describe all three.

## Done when

- [ ] `redacted` records spans whose content attributes have been through the PII filter
- [ ] The span-attribute names are pinned by a test, so a pydantic-ai upgrade that renames them fails loudly rather than silently leaking
- [ ] The Builder offers all three modes, en + pl

Refs #1413. Part of the security programme (#1424). Cluster in #168: *Security programme*.

DEENUU1:
Closing as not planned.

Logfire is opt-in — a deployment that sets no token sends nothing, and a run still records its trace id locally. When it is on, its job is to record prompts and outputs; that is the product. Cutting the content down to a filtered middle ground makes the traces worth less for the thing they exist for, and hands a deployment a guarantee nobody can audit: a PII filter that misses one identifier has exported it, and the operator believes it did not.

[...]

Issue #1413, opened 2026-09-04
Traces sent to Logfire carry full prompts and outputs; add a content mode to AgentSpec.observability


DEENUU1:
Part of the security programme (#1424). Pre-existing; found planning W6.

## What happens

`AgentSpec.observability` lets an agent send its traces to a Logfire project, including a client's own (`app/agents/spec.py`, `ObservabilitySpec`). Nothing in `app/agents/observability.py` or `app/agents/factory.py` narrows what the instrumentation records - there is no `include_content` anywhere in the tree - so Pydantic AI's default applies and every span carries the user's message, the model's output and every tool argument and result.

For a deployment handling health, legal or HR data that is a copy of the protected content leaving the machine to a third party, per run, with no switch. A security review will find it in the first hour.

## What to build

- `ObservabilitySpec.content: Literal["full", "redacted", "none"]`, default `full` so nothing stored changes behaviour. `none` records spans with timing, tokens, cost and tool *names* only; `redacted` runs the same PII filter the log pipeline uses over message text before it is attached.
- Enforced where the agent is instrumented, not in the Builder - a spec that says `none` must produce spans with no message text however it is run.
- Publish validation unchanged; this is a per-agent decision the Builder shows next to the token field.
- A deployment-wide floor in `deployment_settings` (an app admin may forbid `full`) is a follow-up, not this issue.

## Done when

- [ ] The field exists, `SPEC_VERSION` bumped, `docs/reference/spec.md` via the docstring
- [ ] A test asserts that with `none` no span attribute contains the prompt text, and with `full` one does
- [ ] `docs/models.md` or `docs/governance.md` (whichever holds the observability section) says what each mode sends
- [ ] The Builder's observability card offers the three, with copy in en and pl

Related: #1424, #561 (the observability secret validated at run time - same card). Cluster in #168: *Security programme*.

DEENUU1:
Done — closing. Delivered in #1617:

- `ObservabilitySpec.content`, typed `TraceContent = Literal["full", "none"]` (`backend/app/agents/spec.py:77`), default `full` so nothing stored changes behaviour. `SPEC_VERSION` is at 11, and `docs/reference/spec.md` picks the field up from the docstring.
- Enforced where the agent is instrumented, not in the Builder. Covered by `backend/tests/test_agent_observability.py` and `test_agent_runner.py`.
[...]

PR #1617, opened 2026-09-14
feat(agents): add a trace-content mode so a run traces without its prompts

OchnikBartek:
## Problem

`AgentSpec.observability` redirects an agent's traces to a Logfire project — often a client's own. But nothing narrowed what the instrumentation recorded: Pydantic AI's default applies, so every span carried the user's message, the model's output and every tool argument and result, with no switch. For a deployment whose runs touch health, legal or HR data, that is a copy of the protected content leaving the machine to a third party, per run.

## What this does

Adds `ObservabilitySpec.content`, a `Literal["full", "none"]`:

- **`full`** (default) — records everything, exactly as before.
- **`none`** — records timing, tokens, cost and tool *names* only; no message text or tool arguments.

Enforced **where the agent is instrumented**, not in the Builder: `factory._instrument` passes `include_content=(content != "none")` into `instrument_agent`, which hands it to `instrument_pydantic_ai`. So a spec that says `none` produces content-free spans however the run is started, and the choice is preserved through the environment-tracing merge (an environment redirects *where* traces go, not *how much* they carry).

The Builder's observability card offers the two modes beside the token, service name and environment (locked until a token is chosen, since the mode only bites on a per-agent project). Copy in en and pl.

## Scope

The original issue lists a third mode, **`redacted`** (run the log pipeline's PII filter over the content before export). That needs a custom OpenTelemetry span-processor scrubbing pydantic-ai's content attributes and is version-fragile, so it is tracked as a follow-up: **#1616** (`Refs #1413`). Widening the `Literal` later is additive.

## Why no `SPEC_VERSION` bump

`content` is an optional field with a default. Stored specs omit it and load as `full`, unchanged, so no document needs migrating — `spec_version` stays 11 (per the `agent-spec` skill: an additive field with a default is free).

## Verification

- `make test` — backend, **100% coverage** on the platform layer.
- `make lint`, `make test-frontend-cov` (100% stmts/funcs/lines, 97.83% branch), `make build-frontend`, `make docs-build` — all green.
- New tests: `none`/`full` at the factory and at `instrument_agent` (the flag reaches Logfire); the content choice survives the environment merge (both branches); the field defaults to `full` and an old YAML without it loads as `full`; the Builder card defaults to full, locks without a token, and records `none`.
- Docs: `docs/reference/spec.md` (via the field docstring) and `docs/environments.md`.

## Notes for reviewers

- `none` is the enforceable contract here: `include_content=False` is Pydantic AI's own switch for content-free spans. The span-level assertion that no attribute holds the prompt is part of the `redacted` follow-up's span-processor test (#1616), where the attribute names are pinned.

Refs #1413


OchnikBartek:
@codex review

Issue #1622, opened 2026-09-14
Audit trail tamper evidence: a per-entry hash chain and audit-verify


OchnikBartek:
Follow-up to #1422, which added the audit-trail export (`GET /audit/export`, CSV + JSONL). This is the other half: making the trail **tamper-evident**.

## Why

Audit entries are plain rows. An operator with database access can edit or delete one and nothing notices. HIPAA §164.312(b) audit controls and (c) integrity; SOC 2 CC7.2. The export is what an auditor takes away; the hash chain is what makes it worth taking.

## What to build

- A hash chain on `app_admin_audit_logs`: each entry stores `prev_hash` and `hash = H(prev_hash || canonical row)`, written in the same transaction as the entry, in the one place that writes audit rows (`app/core/audit.py::record_audit`).
- **The concurrency is the hard part.** Two audit writes racing for the same `prev_hash` fork the chain. Serialize per organization — a `SELECT ... FOR UPDATE` on the org's chain-head row, or an advisory lock — so `prev_hash` is read and the new head written under one lock. Decide whether the chain is per-organization or deployment-wide (per-org is the natural tenant boundary and avoids one global lock on every audited action).
- A **backfill migration** computing the chain over existing rows in `created_at` order, per organization.
- `agenticos cmd audit-verify [--org]` walks the chain and reports the first break. A detection control, not a prevention one — the docs say exactly that.
- Optional: a daily flow that verifies and writes its own audit entry.

## Done when

- [ ] A test edits a row in the middle and `audit-verify` names it
- [ ] Two concurrent audited actions produce a valid, unforked chain (an integration test against real Postgres)
- [ ] `docs/governance.md#audit` and `docs/commands.md` updated

Refs #1422. Part of the security programme (#1424). Related: #20 (make an audit write failure loud). Cluster in #168: *Security programme*.

Issue #1422, opened 2026-09-04
Audit trail export and tamper evidence


DEENUU1:
Part of the security programme (#1424). Pre-existing.

## What is there

Runs, approvals and spend each have an export (`GET /runs/export`, `/approvals/export`, `/spend/export` - `docs/governance.md:758`). The audit trail has one `GET` in `app/api/routes/v1/audit.py` and no export. Entries are plain rows: an operator with database access can edit or delete one and nothing notices.

HIPAA §164.312(b) audit controls and (c) integrity; SOC 2 CC7.2. The export is what an auditor takes away; the chain is what makes it worth taking.

## What to build

- `GET /audit/export` with the same filters as the list and the same gate (`audit:read`), CSV and JSONL.
- A hash chain: each entry stores `prev_hash` and `hash = H(prev_hash || canonical row)`. Written in the same transaction as the entry by the one place that writes audit rows (`record_audit`).
- `agenticos cmd audit-verify [--org]` walks the chain and reports the first break. It is a detection control - it does not prevent tampering, it makes it visible, and the docs say exactly that.
- Optional: a daily flow that verifies and writes its own audit entry.

## Done when

- [ ] Export works for an organization with more entries than one page
- [ ] A test edits a row in the middle and `audit-verify` names it
- [ ] `docs/governance.md#audit` and `docs/commands.md` updated

Related: #1424, #20 (make an audit write failure loud), #983 (sync sources leave no audit entry). Cluster in #168: *Security programme*.

DEENUU1:
Done — closing. All three "Done when" boxes hold:

- **Export**: `GET /audit/export`, CSV and JSONL, behind the same `audit:read` gate as the list (`backend/app/api/routes/v1/audit.py:36`). Past one page by a wide margin: a 100-row page cap against a 10 000-row export cap, and a match over it is refused rather than silently truncated. Delivered in #1623.
- **Hash chain**: `prev_hash` / `entry_hash` per entry (`backend/app/db/models/audit_log.py:64`), written by `record_audit` in the entry's own transaction. Delivered in #1647, via the split-out #1622.
- **`agenticos cmd audit-verify [--org]`**: `backend/app/commands/audit_verify.py:45`. A row edited in the middle is detected and named — `tests/integration/test_audit_hash_chain.py::test_editing_a_stored_row_is_detected`, plus `test_a_broken_chain_is_named_and_exits_non_zero` for the exit code.
[...]

PR #1623, opened 2026-09-14
feat(audit): export the audit trail as CSV or JSONL

OchnikBartek:
## Problem

Runs, approvals and spend each have a CSV export (`GET /runs/export`, `/approvals/export`, `/spend/export`). The audit trail had a list route (`GET /audit`) and **no export** — an auditor could read it a page at a time in the console but not take the whole thing away.

## What this does

`GET /audit/export` — the organization's audit trail over a window, as **CSV or JSONL** (`?fmt=`):

- Same `audit:read` gate and organization scope as `GET /audit`.
- A **mandatory date range** (a refusal names the missing bound) and a **10,000-row cap** (a match over it is refused, not paged) — an export has no ceiling by nature, so it is given one by design.
- CSV for a spreadsheet, JSONL for a log pipeline; the two describe the same entries, with `details` flattened to a JSON string in the cell and kept as a nested object in the lines.
- Ships **exactly the fields the read model exposes** — the stored `ip_address` is not on the tab, so it is not in the export either (the same "what the tab shows" rule the run export follows).
- **Records its own read** in the trail (`audit.export`, naming the window, the format and the row count) — reading a whole trail is a privileged act.

The change is in two commits:

1. **`refactor(export)`** — the CSV cell/injection guard, the row cap, the date-range check and the download-name stamp were private to `run_export`. A second export domain needs all of them, so they move to `app/services/exporting.py` (one source of truth) and `run_export` imports them. Adds a JSONL writer and `jsonl_response`. Pure refactor: `run_export`'s behaviour and tests are unchanged.
2. **`feat(audit)`** — the repo window query, `AuditService.export`, the route, tests and docs.

## Scope

This is the **export** half of #1422. The **hash chain** that makes the trail tamper-evident (a `prev_hash`/`hash` per entry, a backfill migration, `audit-verify`) is a schema change with a concurrency-sensitive `prev_hash` and does not belong in the same change — filed as **#1622** (`Refs #1422`). So this PR is `Refs #1422`, not `Closes`.

## Verification

- `make test` — backend, **100% coverage** (8364 passed).
- `make lint`, `make docs-build` — green.
- New tests: `test_exporting.py` (the shared primitives, including every `_json_default` branch and the empty-JSONL case); `AuditService.export` service tests (CSV/JSONL content, date-range and cap refusals, tenant scope, the self-audit); an **integration test against Postgres** (window filter, a neighbour organization's entries excluded, `details` round-tripping through JSONB, the export recording its own read); and `/audit/export` in the platform-route gate table.
- Docs: `docs/governance.md#exporting-to-csv`.

Refs #1422


OchnikBartek:
@codex review

PR #1629, opened 2026-09-14
fix(agents): stop the exposure form showing the catalog's short blurb

### LINKED ISSUE #1473: capabilities: the Toolbox shows a different description than the one the model reads, and editing it truncates the prompt
## What is wrong

The Toolbox shows a tool's description under the heading **"Description the model
reads"**, and it is not the description the model reads.

Two strings, two sources:

- **What the UI shows and edits** — `CapabilityToolInfo.description`, typed by
  hand in `@register(...)`.
- **What the model is actually sent** — the tool method's docstring, which
  pydantic-ai turns into the tool schema.

Checked on `memory_mem0`, but the shape is every capability's: the registry
declares a description *and* the toolset carries a docstring, and nothing compares
them.

```
$ ... await toolset.get_tools(ctx) ... print(t.tool_def.description)

<summary>Remember a fact you will want to recall later by its meaning.

Use this for something durable you learned - a preference, a decision, a
fact about this person or subject - that a future conversation should be
able to find without knowing the exact words. Do not store secrets, or
anything the person would not expect you to keep. Facts are found with
`recall`, not by name.</summary>
<returns>
<description>A confirmation. It is kept in this conversation's own memory - one
person's when you are alone with them, this group chat's when you are
not.</description>
</returns>
```

The panel, for the same tool, shows one sentence: *"Remember a fact you will want
to recall later by its meaning."*

## Why it is worse than a display bug

The field is **editable, and an edit does reach the model** through
`tool_overrides`. So the editor is pre-filled with a value that is not the current
value, and saving after changing one word replaces the whole docstring with the
one-liner - silently cutting most of the prompt that was steering the tool.

"Reset" restores the declared blurb, not the docstring, so there is no way back
from the UI once it has been overridden.

The heading makes the promise explicit, and the panel repeats it: *"A tool's name
and description are the prompt the model reads before it decides to call it, so
both steer it."* That is exactly right, and exactly why showing a different string
there is a defect rather than a cosmetic mismatch.

## Fix shape

One source. The catalog should serve each tool's **real** description - read off
the built toolset's tool defs, the same value the model gets - instead of a
hand-typed duplicate in `register(...)`. `CapabilityToolInfo` would keep `id`,
`side_effecting` and the approval metadata, which nothing else can derive.

That is a platform-wide change, so the cheap interim is a drift test:
`test_capability_registry.py` already compares declared tool *ids* against offered
ones in both directions (`TestToolDeclarations`), and the same pass can assert the
declared description is the docstring's summary line. It would fail today on
whichever capabilities have drifted, which is itself worth knowing.

Open any tool in the Toolbox and see the text the model is sent, including the
paragraphs after the first line. Change one word, save, reopen: the rest is still
there.

Found while working on #1470 (memory), which does not cause it - the mechanism
predates the branch and applies to every capability in the registry.

Refs #168

MichalZelasko:
### Summary

Closes #1473 by adding the regression test the issue itself proposed as the interim fix, and fixes a same-class bug found while confirming it: the exposure form's channel-lookups checklist was still showing the catalog's short, hand-typed blurb instead of a tool's real description.

### Business Context

#1473 reported that the Toolbox panel showed a different (and editable, truncating) description than the one the model actually reads. Investigating found that specific defect already fixed by #1383, before this branch existed — `capability_contracts.py` now serves every tool's real, docstring-derived description, and the Toolbox panel already prefers it. What was missing was #1473's own suggested "cheap interim": a test guarding that the catalog's declared one-liner stays consistent with the real thing. While confirming that, a second, undiscovered instance of the same defect turned up in the exposure/channel-binding form, which this PR also fixes.

### Changes

- **fix(agents):** `agent_exposure.py`'s channel-lookups checklist now sources each tool's description from the same `tool_contracts()` the Toolbox panel reads, instead of the catalog's hand-typed blurb. Two defects surfaced fixing this: the real description arrives wrapped in `<summary>`/`<returns>` markup for the model's benefit, which was about to render as literal XML tags in a checkbox label (stripped by a new `_prose()`); and the fallback to the catalog blurb only checked whether a tool's id was present in the contracts dict, not whether the value was truthy, so a tool with no docstring would have rendered blank.
- **test(capabilities):** adds the drift-check #1473 asked for — every capability's declared `CapabilityToolInfo.description` is asserted to be a prefix of the real, built-toolset description. Passes today with no other changes (there was no live drift), so this is purely a guard against it recurring.
- **test(capabilities):** `capability_contracts.py` has existed since #1383 but was never added to either of `pyproject.toml`'s platform-layer gate lists (100% coverage, strict typing) — added it to both, plus the two tests needed to close the resulting coverage gap.
- **fix(ci):** cherry-picked from a sibling branch — routes the `check-routes` pre-commit hook through the pinned backend interpreter rather than the host's system `python3`, which was crashing the hook outright on a machine whose system Python predates 3.10.

### Verification

```bash
cd backend && uv run pytest tests/test_agent_exposures.py tests/test_capability_registry.py tests/test_capability_contracts.py -q
make lint-backend   # ruff, ty, vulture, deptry, guard scripts all clean
make test           # 8337 passed, 100% coverage, capability_contracts.py now gated
python3 scripts/docs_drift.py   # clean
```

### Linked Issues

Closes #1473

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MichalZelasko:
@codex review

PR #1625, opened 2026-09-14
docs(security): add the threat model and a controls matrix

OchnikBartek:
## Problem

A client's security review — HIPAA §164.312 or SOC 2 shaped — asks for a threat model, a data-flow statement, an at-rest statement and a controls matrix. `docs/rollout.md` answered the six headline questions in a table, but the long form lived scattered across `SECURITY.md`, `docs/models.md`, `docs/mcp.md`, `docs/governance.md` and `docs/secrets.md`, or nowhere.

## What this does

Adds `docs/security.md`, one page synthesizing the **existing** security model — it describes what is, not what would be nice:

- **Threat model** — the trust boundaries (browser → BFF → API → store; API/worker → model providers, MCP, search, Logfire, connectors), what crosses each, and what is trusted on the far side.
- **What leaves the deployment** — prompts/outputs to the model provider, traces to Logfire (reducible per agent), tool calls to MCP, queries to search, documents to an embedding provider — each with its "when".
- **What is encrypted where** — the vault's per-organization envelope, and honestly what it does **not** cover: files, message bodies, `rag_documents`, sandbox workspaces are the deployment's disk to encrypt.
- **A controls matrix** — one row per control across HIPAA §164.312 and SOC 2 CC6–CC8, each naming the mechanism (a `file`) and the test that holds it true. Every mechanism and test citation was verified against the code.

Three gaps are marked honestly, each linking its issue: audit-log export (#1422), tamper evidence (#1622), app-level file encryption (#1423). HSTS is named as the reverse proxy's job (the bundled `nginx.conf` sets it) rather than claimed in the app — correcting a stale line `SECURITY.md` carried ("set in middleware", which no code emits).

`SECURITY.md` now keeps only the disclosure policy and the hardening checklist and points here for the model — one copy. `docs/rollout.md`, `llms.txt`, the nav (`Learn → Keep it under control`) and `CLAUDE.md`'s topic map all link the page.

This **unblocks #1417** (the `security` pytest marker), which maps the same refusal tests.

## Verification

- `make docs-build` (`--strict`) — green (no dead links, nav valid).
- `scripts/check_docs_paragraphs.py` and `scripts/check_backticks.py` — clean; sentence-case headings, a recap.
- Every `path`/test in the matrix was located in the code before it was cited.

Closes #1412

Issue #1412, opened 2026-09-04
docs/security.md: threat model and a controls matrix mapped to mechanisms (HIPAA §164.312, SOC 2 CC6–CC8)


DEENUU1:
Part of the security programme (#1424). Found planning W6; pre-existing gap.

## What is missing

A client's security review asks the same six questions and `docs/rollout.md#what-your-security-review-will-ask` answers them in a table. What it does not have, and what a HIPAA- or SOC 2-shaped review actually hands over, is:

- **A threat model.** Trust boundaries (browser → BFF → API → Postgres/Redis; API → model providers, MCP servers, search vendors, Logfire; worker → connectors), what crosses each one, and what is trusted on the far side.
- **A data-flow statement.** What leaves the deployment and to whom: prompts and outputs to the configured model provider, traces to Logfire when a token is set, tool calls to MCP servers, queries to a web-search vendor, documents to an embedding provider. Today this is spread over `docs/models.md`, `docs/mcp.md`, `docs/governance.md` and `SECURITY.md`.
- **What is encrypted where.** Vault envelopes per organization (`docs/secrets.md`) are covered; message bodies, `rag_documents`, uploaded files (`app/services/file_storage.py` - local disk) and sandbox workspaces are not stated anywhere. The honest answer for most of them is "disk encryption is the deployment's job", and it has to be written down.
- **A controls matrix.** One row per control, one column saying which mechanism in this codebase satisfies it and which test holds it true. HIPAA §164.312 technical safeguards (access control, audit controls, integrity, person or entity authentication, transmission security) and SOC 2 CC6 (logical access), CC7 (system operations), CC8 (change management). A row with no mechanism says so - that is the list the rest of the programme works from.

## Done when

- [ ] `docs/security.md` exists, is in `nav` under Learn → Keep it under control, and `docs/rollout.md` links to it from the security-review table
- [ ] The matrix names a file or a test for every row that has one, and says "not yet" for every row that does not, each linked to its issue
- [ ] `SECURITY.md` at the root keeps the disclosure policy and the hardening checklist, and points to the page for everything else - one copy
- [ ] `llms.txt` gets one line about it
- [ ] Sentence-case headings, no paragraph over ~115 words, a recap

Related: #1424, #1415, #1417. Cluster in #168: *Security programme*.

PR #1630, opened 2026-09-14
A security pytest marker naming every refusal test, and a report a client can read

### LINKED ISSUE #1417: A `security` pytest marker naming every refusal test, and a report a client can read
Part of the security programme (#1424).

## Why

"Does the code meet a standard" is answered by showing the tests, not by asserting it. The platform's refusals are already tested - tenant isolation including when the caller owns the row, permission scopes and grants, a budget checked before the model request, an approval decided once, no plaintext secret in any response, log or audit entry, a spec refused at publish - but they are scattered across the suite and nothing names them as a set.

## What to build

- `@pytest.mark.security` on each refusal test, registered in `pyproject.toml` so an unknown marker fails.
- `make test-security` runs only those, and `make check` keeps running everything - the marker is for the report, not a second gate.
- A step in CI that writes the count and the list (test ids, one line each) as a job summary and as `security-tests.txt` in the release assets.
- `docs/security.md` links the list from the controls matrix: each matrix row that has a test names it.

## Done when

- [ ] Every test that covers an item in `CLAUDE.md` → "Cover the refusal" carries the marker
- [ ] `tests/test_security_marker.py` fails if a test whose name or module says `tenant`, `permission`, `budget`, `approval`, `secret` or `plaintext` lacks it - a nudge, not a gate on the wording
- [ ] The report is attached to the next release

Related: #1424, #1412, #1415. Cluster in #168: *Security programme*.

OchnikBartek:
## Problem

The platform's refusals — tenant isolation, permission and grant refusals, a budget checked before the model call, an approval decided once, secret confidentiality, spec publish validation — are already tested, but scattered across the suite with nothing naming them as a set. "Does the code meet a standard" is answered by showing the tests, not asserting it.

## What this does

- Registers a `security` pytest marker and turns on `--strict-markers`, so a misspelled marker name fails the run instead of silently dropping a test from the set.
- `make test-security` runs only the marked tests; `make check` still runs everything — the marker is for the report, not a second gate.
- `make security-report` writes the count and the list to `backend/security-tests.txt` and, in CI, to the job summary; a CI step runs it and uploads the file as an artifact.
- `tests/test_security_marker.py` is a static guard: it fails when a test whose function or module name mentions a tenant, a permission, a budget, an approval, a secret or plaintext neither carries the marker nor appears in its `EXEMPT` set. A nudge, not a gate on wording.
- `docs/security.md` links the set from the controls matrix.

## How the marking was done

- **Module level** for security-themed files (tenant isolation, secrets, permissions, approvals, auth, audit, vault, budget/spend, TLS, rate limits, security headers) — "if it's in a secret/tenant/permission module, mark it," which is the guard's own intent.
- **Per test** for refusals living in broader modules (e.g. `test_platform_flows.py`, `test_agent_triggers.py`), classified one by one from each test's body.
- **Exempted**, with a recorded reason, for coincidental keyword matches — a budget rendered on a dashboard, an approval notification, plain secret CRUD, TLS `plaintext` config, catalog/meta invariants.

`make test-security` currently names **699** tests.

## Verification

- `make lint-backend` — clean (ruff, ruff format, ty, vulture, deptry, guard scripts, codespell).
- `tests/test_security_marker.py` — passes (completeness, no stale exemptions, marker registered, non-empty set).
- `tests/test_ci_parity.py`, `tests/test_coverage_gate.py` — pass (the CI report step is a make target, exempted in `CI_ONLY_TARGETS`).
- Full `--strict-markers` collection exits 0 (no unregistered marker anywhere); a sample of the marked modules runs green.
- `docs-build --strict`, paragraph and backtick checks — clean.

## Limitations

- The repository cuts releases by hand (no release workflow), so attaching `security-tests.txt` to a release is a release-time step; CI publishes it as a per-build artifact and job summary today.
- A couple of borderline TLS-config citations (e.g. `test_doctor_sandbox.py`'s Redis check) are not marked; the marker set covers the refusals, and the docs note says exactly that rather than claiming every cited test carries it.

## Stacking

Based on `docs/1412-security-model` (#1625) because the docs link lands in that branch's `docs/security.md`, which is not yet on main. **Retarget to `main` once #1412 merges.**

Closes #1417
Refs #1412


OchnikBartek:
@codex review

PR #1633, opened 2026-09-14
Keep the MCP OAuth token out of the logs on a bad token response

### LINKED ISSUE #1626: A malformed MCP OAuth token response logs the token in plaintext
## Problem

`app/agents/mcp_oauth.py::_token_request` parses the provider's token payload with a Pydantic model. When the server names a field wrongly the `ValidationError` carries the rejected input — which *is* the token — and the handler logs it with `logger.exception`. The exception is kept out of the HTTP response, but the raw token reaches the logs.

This is proven, not hypothetical: `tests/test_mcp_connections.py::test_an_unreadable_token_response_does_not_echo_its_input` asserts both that the token is absent from the raised error's string **and** that `"at-secret-9f2c" in caplog.text`.

## Impact

The platform's confidentiality guarantee — no plaintext secret in any response, log or audit entry — does not hold for this path. An operator reading logs (or shipping them to a log aggregator) can see a third-party MCP access token in the clear.

## Fix

Sanitize the `ValidationError` before logging (log the field names / error types, not the input), or catch and re-raise without `logger.exception` on the raw payload. Update the test to assert the token is **absent** from `caplog.text` once fixed.

## Notes

Surfaced while documenting the controls matrix in `docs/security.md` (#1412): the 'no secret in logs' row had to be narrowed to responses and audit entries because of this path. This issue is the tracked gap the doc links.

Refs #1412

OchnikBartek:
## Problem

`_token_request` parses the provider's token response with `OAuthToken.model_validate_json`. A malformed response raises a pydantic `ValidationError`, whose `str()` and traceback echo the input they rejected — and for a token response that input is the token. `logger.exception` beside the raise therefore wrote a live credential to the logs. `test_mcp_connections.py::TestOAuthRefusalsDoNotQuoteTheServer::test_an_unreadable_token_response_does_not_echo_its_input` asserted exactly that (`"at-secret-9f2c" in caplog.text`).

## Fix

- Log the field locations and error types, never the values — `exc.errors(include_url=False, include_input=False)` via a new `_validation_detail` helper.
- Drop from `logger.exception` to `logger.error`, so no traceback carrying the raw payload is appended.

The refusal shown to the caller was already the class name only (`_flow_failed`), and the routes convert `OAuthError` with `str(exc)` / the refresh path logs `%s`, so nothing downstream re-logs the chained `ValidationError`.

## Verification

- The named test now asserts the token is **absent** from `caplog` and the failure is still described.
- Two unit tests cover `_validation_detail`: a `ValidationError` names the field and type but not the value; a plain `ValueError` reports its class.
- `app/agents/mcp_oauth.py` stays at 100% line+branch coverage; ruff, ruff format and ty clean.

## Notes

Surfaced while writing the security controls matrix (#1412); it is the gap the "logs are not part of the no-secret guarantee" row links.

Closes #1626


OchnikBartek:
@codex review

PR #1634, opened 2026-09-14
Lock the tool-contract cache build and stop caching a transient failure

### LINKED ISSUE #1621: capability_contracts: tool_contracts()'s process cache has no lock and no self-healing after a transient build failure
## What is wrong

`tool_contracts()` (`backend/app/services/capability_contracts.py:156-183`) caches its whole-catalog build in a bare module global, `_CACHED`, with no lock around the build-and-write step:

```python
global _CACHED
if _CACHED is not None:
    return _CACHED
contracts: dict[str, dict[str, ToolContract]] = {}
for definition in all_capabilities():
    try:
        contracts[definition.id] = await _contracts_for(definition)
    except Exception:
        logger.exception(...)
        contracts[definition.id] = {}
_CACHED = contracts
return contracts
```

Two consequences, both real and both currently untested:

1. **Duplicated cold-start work.** Two concurrent requests hitting a fresh process before the cache warms (right after a worker restart or redeploy) both see `_CACHED is None`, both iterate every registered capability and build its documentation-stub toolset, and both write `_CACHED` - redundant, not corrupting (the two dicts are content-equivalent), but real duplicated CPU/latency on whichever request loses the race.
2. **A transient failure is cached forever.** If one capability's build raises on the very first call in a process's life (a flaky import-time resource, a dependency not yet warm), the `except Exception` branch caches `contracts[id] = {}` for that capability - and because `_CACHED` is then set, every later call returns the same empty result for the rest of the process's life. Nothing re-attempts it until the next redeploy.

## Why it matters more now

Until #1383 (which introduced `capability_contracts.py`), the only caller was `GET /capabilities` - a Builder deliberately opening the Toolbox or the capability catalog. #1473's fix (sourcing the exposure form's channel-lookups checklist from the same real descriptions) added a second, more routine caller: `GET /agents/{id}/exposures`, hit on every load of an agent's "where this agent is available" page. That moves both the cold-build cost and the failure-caching risk onto a much more frequently hit, lower-friction path.

- Two concurrent calls to `tool_contracts()` against a cold cache do the documentation-stub build exactly once (an `asyncio.Lock` around the check-and-build is the standard fix for this shape).
- A capability whose build fails once does not stay empty forever - either the failure is not cached (retried on the next call), or there is an explicit, deliberate reason a transient failure should stick, stated in the code.

## Found reviewing

Found reviewing #1473 - pre-existing since #1383/#1470 (the memory-capability reshape that introduced this module), not introduced by that branch, but #1473 is what gives it a second, hotter caller. Flagged independently by three separate review agents during #1473's own multi-angle review.

Refs #168

OchnikBartek:
## Problem

`tool_contracts()` (`app/services/capability_contracts.py`) cached its whole-catalog build in a bare module global with no lock:

- **Duplicated cold-start work.** Two requests hitting a fresh process before the cache warms both see `_CACHED is None` and both build every capability's documentation-stub toolset.
- **A transient failure cached forever.** If a capability's build raised on the first call in a process (a flaky import-time resource), the `except` branch cached `{}` for it — and since `_CACHED` was then set, every later call returned that empty result until the next redeploy.

Both matter more since #1473 added `GET /agents/{id}/exposures` as a second, routine caller (hit on every load of an agent's availability page).

## Fix

- Build under an `asyncio.Lock` with a re-check inside it, so concurrent cold-start callers build the catalog once.
- Cache only a build that completed every capability. A capability that fails still degrades to an empty contract set for that one answer (unchanged), but the result is not cached, so a transient failure is retried on the next call rather than frozen for the process.

## Verification

- Three tests (`tests/test_capability_contracts_cache.py`): concurrent cold-cache callers build each capability once; a transient failure is not cached and is retried, then the complete build is cached; a complete build is returned as-is without rebuilding.
- Existing `tests/test_tool_text_shape.py` (23) still pass; ruff, ruff format, ty clean.

Closes #1621
Refs #168


OchnikBartek:
@codex review

PR #1636, opened 2026-09-14
Prefill the MCP tool prefix with a valid slug, not the namespaced registry key

### LINKED ISSUE #1628: MCP connect dialog: Tool prefix is prefilled with the registry key and fails validation
## What happens
Settings → MCP servers → pick a registry entry (e.g. Snitcher, `com.snitcher/snitcher`) → "Connect for the organization". The **Tool prefix** field is prefilled with the registry key `com.snitcher/snitcher`. Submitting answers:

> Name must be lowercase letters, digits and hyphens (max 32 characters).

The pattern is `^[a-z0-9][a-z0-9-]{0,31}$` (`schemas/mcp_connection.py::NAME_PATTERN`), so the prefill can never pass. The same dialog for HubSpot prefilled `hubspot` correctly, so it seems to depend on the registry key having a dotted namespace.

## Expected
Prefill the prefix with a slug derived from the key or the display name (`snitcher`), or leave it empty.

## Workaround
`POST /api/v1/mcp-connections/oauth/start` with `{"name": "snitcher", "url": "...", "catalog_key": "com.snitcher/snitcher"}` works and the connection appears with the right name.

Seen 2026-09-14 on agenticos.testapp.ovh.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JJdUFmtmsUuenRem2naBuq

OchnikBartek:
## Problem

Settings → MCP servers → pick a registry entry with a dotted namespace (e.g. Snitcher, `com.snitcher/snitcher`) → Connect. The **Tool prefix** field is prefilled with the raw key, and submitting is refused:

> Name must be lowercase letters, digits and hyphens (max 32 characters).

The pattern is `^[a-z0-9][a-z0-9-]{0,31}$` (`schemas/mcp_connection.py::NAME_PATTERN`), so the prefill can never pass. HubSpot worked only because its key (`hubspot`) is already its own name.

## Fix

Add `slugForPrefix` (`src/lib/mcp-servers.ts`): the segment after the last `/`, lower-cased, every other run of characters reduced to a single hyphen, bounded to 32 — so `com.snitcher/snitcher` → `snitcher` and `hubspot` stays `hubspot`. Seed from it at the three sites that used the raw key: `freeName` (the suggested name), the connect dialog's fallback, and the OAuth-start fallback. A key with nothing usable seeds the empty string, leaving the field blank rather than prefilling a refusal.

## Verification

- `src/lib/mcp-servers.test.ts`: 5 unit tests for `slugForPrefix` (namespaced key, already-a-name, punctuation, the 32-char bound, an empty result), each valid slug asserted against `NAME_PATTERN`.
- `mcp-server-list.integration.test.tsx`: a namespaced registry entry now prefills `thing`, not `com.example/thing`; the existing `github-2` conflict-avoidance test still passes.
- `make lint-frontend` clean (eslint, prettier, tsc, i18n, knip).

Closes #1628


OchnikBartek:
@codex review

PR #1641, opened 2026-09-14
Let a deployment admin read one organization's metadata (app-admin gate)

### LINKED ISSUE #1245: An admin can see another tenant's organizations and conversations, and open neither
An app admin can *see* another tenant's organizations and conversations and
cannot **open** either of them. There is no admin-readable destination for one.

Found wiring the links #942 asked for, and it is why that PR ships the rows as
text.

## The two dead ends

**An organization.** `/orgs/{id}` resolves through
`OrganizationService.get_for_user`, which looks up a membership row and raises
`NotFoundError` when there is none — with no `is_app_admin` bypass. So a
deployment admin following a link to an organization they do not personally
belong to gets a 404, and **the target's own personal organization is the
common case**: every account has one, and the admin is a member of none of
them. `/admin/organizations` lists every tenant and has no per-organization
page behind it.

**A conversation.** `/admin/conversations?user_id=` lists one account's threads
and `docs/architecture.md` records, deliberately, that *"there is no way to read
a conversation across tenants any more"* — the sentinel that allowed it went
with the deployment-wide browser. Activity is the surface that replaced it, and
Activity is scoped to the admin's **active** organization through
`X-Organization-Id`, while that listing is deployment-wide. So a link from one
of those rows usually lands on an empty page, or worse, on the same person's
runs in a different tenant.

## Why this is a decision rather than a bug

Not reading another tenant's transcripts is the *point* of that removal, and
the drawer should not undo it. What is missing is narrower and defensible:

- **An organization an admin can inspect without joining it.** Its name, its
  members and their roles, its agent count, its budget — the same row
  `/admin/organizations` already lists, with a page behind it. No agents, no
  conversations, no transcripts.
- **A run-level hand-off that carries its tenant.** Activity already adopts an
  organization named in a URL (#1032 established the shape); a link from an
  admin surface would have to carry the conversation's own
  `organization_id` and let `ActiveOrgGuard` switch to it, which is a
  deliberate act by the admin rather than a silent cross-tenant read.

Either one, or neither — but not a link that 404s.

## Done when

- Following an organization from the admin user drawer shows something, or the
  row is honestly not a link.
- The same for a conversation row.
- Whichever is chosen, the tenant boundary in `docs/architecture.md` still says
  what it says, and a test holds it.

## Related

Found by the reviewer on **PR #1239** (#942), which is why that drawer's rows
are text. **#1032** is the precedent for a page adopting the organization its
URL names. **#63** (Admin redesign) is the umbrella. **#1204** is the same
tenant-in-the-URL question one surface over, for alert emails.

OchnikBartek:
## Problem

An app admin can *see* every tenant in `/admin/organizations` but can open none of them. `/orgs/{id}` resolves through `OrganizationService.get_for_user`, which 404s for anybody who is not a member — and the target's own **personal** organization, which the admin belongs to none of, is the common case (#1245). There is no admin-readable destination for an organization.

## What this does (backend slice)

`GET /admin/organizations/{id}` (`CurrentAppAdmin`): the organization's name, members and their roles, size, owner and budget — the same row the listing carries, with a page's worth behind it. Read-only and metadata only: it reaches no agent, conversation or secret.

## Why a dedicated endpoint, not an `is_app_admin` bypass

The app-admin `AuthContext` already carries **every permission at `Scope.ALL`**, and `resolve_access` blocks cross-tenant only by an org-id equality check. So widening `get_for_user` / `get_active_organization` for `is_app_admin` would **cascade into full read *and write*** of the foreign tenant — the opposite of a read-only view. A dedicated `CurrentAppAdmin` read path avoids that entirely, and the tenant boundary in `docs/architecture.md` (no cross-tenant conversation reads) is untouched. The cross-tenant read is itself audited (`admin.organization.read`).

## Verification

- API (mocked service): the route hands the service the org from the path and the admin as the actor; app-admin-only (403 otherwise).
- Integration (real Postgres): the detail carries members+roles, counts, owner and budget for a foreign org; the read writes an audit entry naming the org and the admin; a missing org is `NotFoundError`.
- `make lint-backend` clean; `test_platform_routes` recognises the new route as guarded.

## Follow-up (not in this PR)

The frontend link from the admin user-drawer's organization row to this endpoint is the second half and closes #1245. The **conversation** row stays an honest non-link: reading a conversation cross-tenant would relax the deliberate tenant boundary, which the issue's own "Done when" requires to hold.

Refs #1245


OchnikBartek:
@codex review

PR #1644, opened 2026-09-14
Link the admin drawer's organization row to a per-tenant page

### LINKED ISSUE #1245: An admin can see another tenant's organizations and conversations, and open neither

OchnikBartek:
## Problem

The admin user-drawer showed a person's organizations and conversations as **text**, because there was no destination an app admin could open — `/orgs/{id}` 404s for anybody who is not a member, and the target's personal organization (which the admin belongs to none of) is the common case (#1245).

## What this does (frontend — the other half)

Each organization row now links to `/admin/organizations/{id}`, a new page that reads the metadata endpoint added in the backend PR: the organization's name, members and their roles, size, owner and budget. A dedicated app-admin read — **not** an `is_app_admin` bypass on the tenant-scoped path (which would cascade into full tenant access).

The **conversation** rows stay text, by decision: there is no admin-readable destination for a single conversation, and reading one across tenants would relax the boundary `docs/architecture.md` keeps ("no way to read a conversation across tenants"). Both drawer comments now record that decision.

## Pieces

- `AdminOrganizationDetail` type; `qk.admin.organizationDetail` key; `useAdminOrganizationDetail` hook (100% covered)
- `/admin/organizations/[id]` page (loading/error/not-found states; members list); no onboarding stop, by the same rule the rest of `/admin` follows
- The drawer org row → `Link`; its integration test flipped from "not a link" to asserting the href
- The BFF proxy route `api/admin/organizations/[id]/route.ts`, registered in `proxy-routes.test.ts` (the guard that caught its absence)
- 9 i18n keys under `pages.admin`

## Verification

`make lint-frontend` and `make test-frontend-cov` both green (418 files, 6147 tests, coverage threshold met).

## Stacking

Based on `feat/1245-admin-org-read` (the backend PR #1641). **Retarget to `main` once that merges.**

Closes #1245


OchnikBartek:
@codex review

PR #1647, opened 2026-09-14
feat(audit): tamper-evidence hash chain and audit-verify

### LINKED ISSUE #1622: Audit trail tamper evidence: a per-entry hash chain and audit-verify

OchnikBartek:
## Problem

`app_admin_audit_logs` is the record of every privileged action, and the app-admin
bypass story in `docs/governance.md` leans on it — but a row was just a row. An
operator with the database could rewrite or delete an entry and leave nothing that
said so.

## What this does

Each entry now joins a **per-organization hash chain**: it carries its own hash
over its canonical fields with the previous entry's hash folded in, plus a
deployment-wide `seq` that gives the chain a deterministic order even when one
transaction writes two entries on the same (transaction-stable) `created_at`.
Editing, reordering, inserting or deleting any entry diverges every hash after it.

- **Write path** — `record_audit` reads the chain head and appends under a
  per-organization advisory lock (`LockScope.AUDIT_CHAIN_PER_ORG`), so two audited
  writes for one organization cannot read the same head and fork the chain.
- **`agenticos cmd audit-verify`** — walks each chain (or one `--org`), recomputes
  the hashes, and names the first entry whose stored hash no longer matches its
  contents or its link to the entry before it. Exits non-zero on any break, so a
  cron can gate on it.
- **Migration `0079`** — adds `seq` / `prev_hash` / `entry_hash`, fills `seq` with
  `ROW_NUMBER() OVER (ORDER BY created_at, id)` before making it an identity column,
  and backfills every existing entry per organization in `seq` order. It imports
  `chain_hash` from `app.core.audit` (the same one `record_audit` and the verifier
  use) so the three cannot disagree.

## Boundary

**Detection, not prevention.** An operator with the database can rewrite a row and
recompute every hash after it, so a clean run is evidence of no tampering by anyone
who did not also re-forge the chain — not proof the rows are immutable.
`docs/governance.md` states this explicitly.

## Verification

- `make test` (full backend suite + 100% coverage gate) — green; the three gated
  files (`core/audit.py`, `services/audit.py`, `commands/audit_verify.py`) at 100%.
- `make lint-backend`, `make docs-build --strict` — green.
- `alembic check` against a freshly-migrated database — no model/migration drift;
  the migration chain round-trips forwards and back.
- Integration (real Postgres): a written chain reads back and verifies (details
  round-trip through JSONB), an edited row and a deleted row are each detected and
  named, and **six concurrent audited writes for one organization serialize into a
  single unbroken chain** rather than forking.
- Migration: the backfill is exercised against a database of pre-existing entries
  across three tenants and the deployment-wide chain, all verifying afterwards.

Closes #1622


OchnikBartek:
@codex review

OchnikBartek:
Thanks — addressed the review:

- **P1, audit lock deadlock (audit.py):** fixed in f49beb32. `record_audit`'s per-org chain lock is transaction-scoped, and the expiry sweep held it while going on to lock more rows. The sweep now does all its row work (decide, settle) first and audits **last**, so the chain lock is the final lock it acquires — matching how a single request audits at its end. The invariant is documented in `record_audit`.
- **P2, deployment-chain contents (docs/comments):** fixed in 9b876512. The null-organization chain holds tenant-less actions — deployment settings, impersonation, app-admin user management — not the expiry sweep / RAG shell, which carry their row's organization. Corrected in `governance.md`, the command, the model and the migration.
- **P2, id/seq in the hash:** kept out by design, docstring clarified. Both are metadata, not the recorded action; reordering **is** caught (the walk follows `seq`, so a changed order makes `prev_hash` disagree), and renumbering `seq` without changing order changes nothing the trail attests to. Neither value exists before the row is flushed, where the hash is built.
- **P3, truncation / whole-chain deletion:** real and inherent to a bare chain — documented as a known limitation in `governance.md`, the command and `verify_chain`, and tracked for a protected terminal checkpoint in #1648.
- **P2, N+1 in `verify_all_chains`:** acceptable for this occasional operator command (one chain in memory at a time), streaming optimization noted on #1648.

The four github-code-quality "unused global variable" findings on the migration (`revision`, `down_revision`, `branch_labels`, `depends_on`) are false positives — Alembic reads those module-level globals by contract; every migration in the tree declares them.

Issue #1593, opened 2026-09-12
Extend RAG tool and retrieval metadata filters for FA-039


DEENUU1:
Extend RAG retrieval and its agent tool to support the metadata filters required by FA-039: source, document type, organizational unit, tenant, date and permissions. Existing document/organization restrictions and metadata provide the starting point.

Scope and acceptance:
- Define typed, validated optional filter arguments and expose the permitted business filters through the RAG API/tool.
- Populate the required metadata during ingestion and preserve it on documents/chunks; define date semantics and missing-field behavior.
- Derive tenant and authorization restrictions from trusted server context. The model/user must not be able to widen access by supplying filter parameters.
- Apply the same restrictions before returning results from vector and keyword/hybrid retrieval. The current _bm25_search path does not forward parent_doc_id; include this consistency fix and regression coverage.
- Cover combined filters, missing metadata, cross-tenant/access denial and both hybrid/non-hybrid retrieval with focused tests; document the API/tool contract.
- Keep the filter contract usable by the planned OpenSearch adapter. This issue does not itself deliver the whole OpenSearch integration or replace the document ACL design in FA-037.

Requested by Kacper for W6+W7. FA-039 is a committed delivery item in the tender matrix.

DEENUU1:
Delivered in **0.0.471** — #1656, merged as `45bdaf37`.

Against this issue's own acceptance list:

- **Typed, validated optional filter arguments, exposed through the API and the
  tool.** `RetrievalFilters` — `source`, `document_type`, `organizational_unit`,
  `date_from`/`date_to`, `parent_doc_id` — on `POST /rag/search` and the
  knowledge tool. `extra="forbid"`, closed vocabularies for `source` and
  `document_type`, `date_from <= date_to`, and a supplied-but-empty list
  refused rather than read as "match everything".
- **Metadata populated at ingestion and preserved on chunks; date semantics and
  missing-field behaviour defined.** `source` and `doc_date` are stamped by the
  upload and both sync paths, `document_type` is derived from the parsed
  filetype. A chunk lacking the dimension under filter fails closed, and
  `rag_safe_to_date` never raises, so a malformed stored date reads as NULL
  rather than breaking the query or the index build. **`organizational_unit` is
  the exception and is tracked separately — see below.**
- **Tenant and authorization derived from trusted server context; a caller
  cannot widen.** `RetrievalScope` is a different type, built in a different
  place, and carries no field a caller or a model can name. The store ANDs it
[...]

PR #1656, opened 2026-09-14
feat(rag): metadata filters and server-derived tenant scope (FA-039)

MichalZelasko:
## What this changes

Adds structured, narrowing-only **metadata filters** to RAG retrieval and its agent tool (FA-039, #1593): `source`, `document_type` (the stored filetype/mime), `organizational_unit`, and an inclusive `date_from`/`date_to` range, as a typed `filters` object on `POST /rag/search` and the knowledge tool (OR within a field, AND across fields; a missing dimension fails closed under a filter for it). The security-bearing **tenant conjunct is derived only from the caller's trusted context** — never the request body or a model parameter — so a filter can only *narrow*, never *widen*, access, closing the latent cross-tenant read on a collection name two orgs happen to share. Ingestion stamps `organization_id` and the business metadata onto every chunk, and `GET /rag/collections/{name}/filter-values` exposes the in-scope `organizational_unit` values. Migration `0091_rag_metadata_prereqs` adds the `rag_safe_to_date` helper plus JSONB and partial-date indexes.

## Why this way

- **Two separate types** that meet only in the store: caller/model-supplied `RetrievalFilters` (narrowing-only, `extra="forbid"`, `[]→None` at the tool but 400 on the API body) vs. server-trusted `RetrievalScope` (tenant now; an FA-037 authorization slot later). Tenant/auth are *structurally inexpressible* by a caller.
- **Enforced in SQL `WHERE` on both the vector and BM25/hybrid paths** — fixes `_bm25_search` not forwarding scope and removes an unscoped probe. Filtered-recall under pgvector HNSW is handled via **iterative index scan** (requires pgvector ≥ 0.8; raised `ef_search` fallback).
- **Fail-closed** missing-field/date semantics via a non-raising `rag_safe_to_date` + partial index. The legacy scalar `filter` string is retired to a **full-match `parent_doc_id` shim** — a deliberate, safe 400 for other strings (noted in the CHANGELOG).
- **Scope boundaries kept:** the ingestion write-path cross-tenant clobber is FA-037/#1684's; the tenant-ownership backfill (P3) and a semantic `document_category` are deferred.

## Checks

- [x] Tests cover the new behaviour, and would fail without the change — combined filters, cross-tenant denial, access-widening rejection, empty-set and impossible-date, hybrid + non-hybrid parity, the `_bm25_search` regression, the HNSW recall guarantee, and the facet.
- [x] `make check` passes (lint, format, types) — **CI is green**: lint, the 100% coverage `test` gate, `test-frontend`, `e2e`, `docs` and the migration chain all pass.
- [x] Platform-layer coverage is still 100% — `filters.py` and the tool/capability code are gated at 100%.
- [x] If a capability was added or changed, its `README.md` says why — the knowledge-capability route/tool/schema docstrings carry the filter grammar and date semantics.
- [x] If a migration was added, `alembic downgrade base && alembic upgrade head` works — `0091_rag_metadata_prereqs`; the forward/back chain passes in `test-migrations` (CI green).

## Notes for the reviewer

- Reviewed with **three codex (gpt-5.6-sol) rounds**; the access-widening invariants were confirmed clean throughout. Fixed along the way: `source`/`document_type` closed-vocabulary validation, lower-casing `document_type` so uppercase extensions filter, and stamping `source` on CLI-synced connector documents.
- **Open items, as they stand at merge:**
    - **#1684 lands first** — done: it merged as #1688 (0.0.465), and the collision it created with this branch (an app-scoped base stamps no tenant, so `TenantScope`'s equality matched none of its rows) is resolved here by `AppScope`, which ANDs `tenant IS NULL`. `scope_for_tenant` builds the right shape from the base's own resolved tenant, never from the caller's organization id.
    - **P1 `document_type` = filetype/mime** — confirmed; semantic `document_category` stays deferred.
    - **P2 pin pgvector ≥ 0.8** — `pgvector/pgvector:pg16` ships 0.8.6, so the iterative scan is live on the image this repository runs. An older image is handled rather than assumed: the knobs are set inside a savepoint, so an unknown GUC rolls back and the search proceeds on `ef_search` alone.
    - **P3 the tenant backfill** — still its own `effort:l` task.
    - **`organizational_unit` has no production population source** — filed as #1777. The filter, the facet, the index and the fail-closed semantics all ship; only the writing end is missing, so the dimension is inert until one is wired.
- Migration renumbered to `0091_rag_metadata_prereqs`, on main's current head (`0090_notification_created_idx`), one head.
- The four `github-code-quality` bot findings (Alembic module globals) are false positives (triaged in a PR comment).

## Review (DEENUU1)

Reviewed the whole diff against `main` at 0.0.470. One change made on the branch, in `f559ecb8`:

- **`retrieve_multi` took a `scopes` map and a single `scope` beside it**, and read a name absent from the map under that shared one. Both callers already build the map from the same bases they build the name list from, so the fallback could not fire — but what it would do if it ever did is read one collection under a *different* collection's tenant, which is the one mistake the map exists to prevent. `scopes` is now the only parameter, a name without one is refused before the first collection is read, and both call sites lost the `scope=scopes[names[0]]` boilerplate.

Two things filled in rather than asked for:

- **The feature was undocumented.** `docs/file-processing.md` gained a *Narrowing a search* section — the four dimensions, OR-within/AND-across, the fail-closed missing dimension, why a filter cannot reach another tenant, the facet endpoint, and the deprecated `filter` string's 400 — with the Polish, German and Spanish pages beside it.
- **The changelog entry had landed inside the released `0.0.469` section**, where merging main put it when the cut promoted `[Unreleased]` underneath it. Moved back up to `[Unreleased]`.

Verified on the merged tree: `make lint` clean, `make test` 9867 passed at 100.00% on the platform layer, `make test-frontend-cov` 438 files / 6394 tests at 100% statements and lines.

Refs #1593


.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+`organization_id` in the `ingestion` form is ignored; the replace path scoped so it
+cannot delete another tenant's document; a local sync writes no `organization_id`;
+`doc_date` normalized to ISO from a source timestamp.
+
+### 9.6 Phase 6 — migration + tenant backfill (distinct workstream, P3/M4)
+
+Own `effort:l` sub-issue. Ships behind the ordered rollout so no window leaves
+untagged chunks searchable.
+
+**Migration numbering (codex-3 M11).** `0077_drop_allow_byo` is the current single head
```
DEENUU1: **The numbers are already taken.** `0077_drop_allow_byo` is no longer the head — `main` carries `0078_local_services.py` and `0079_audit_hash_chain.py`.

M11 says to regenerate against the then-current head, and that is the right rule, but §9.6 and the §9.9 dependency graph still name `0078_rag_metadata_prereqs` and `0079_rag_tenant_backfill` in prose. An implementer reading the plan literally produces duplicate revision ids — which is exactly how duplicate revisions have reached `main` before from parallel branches, and it is invisible until `upgrade head` runs on a real database.

Suggest dropping the numbers from the text entirely: "the prereq migration" and "the backfill migration", with the revision generated at implementation time against whatever the head is then. (Worth also noting in passing that revision ids are capped at 32 characters in this repo's setup, so the generated slug needs a glance before merge.)

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+recall guarantee this section claims must be *verified*, not inferred from the
+predicate living in `WHERE`. The exact-scan alternative (drop the ANN index) is
+noted and rejected on latency grounds. The check pgvector version supports iterative
+scan is a deployment prerequisite (§2.5 rollout).
+
+`BaseVectorStore.search(...)` grows a structured parameter (the composed
+scope+filters object) replacing the bare `parent_doc_id`. `PgVectorStore.search`
+translates it to parameterized conjuncts on the metadata JSONB, e.g.:
+
+- tenant: `metadata->>'organization_id' = :org`
```
DEENUU1, line 279: **The one security-bearing field ends up as untyped JSONB text.** Everything else in this design is deliberately typed and structurally constrained; the tenant conjunct becomes `metadata->>'organization_id' = :org` on a hash index.

§2.4 reaches for a real typed column as "the sturdier alternative" for `doc_date` — but never applies the same reasoning to the field that actually carries the security guarantee. A real `organization_id uuid` column on each `rag_*` table would be type-safe, `NOT NULL`-enforceable once backfilled (so the backfill can prove it finished rather than being audited by a count), and cheaper to filter.

The part that matters most is H1: §9.2 rejects a per-tenant partial index because tenants are created dynamically, and rejects partitioning implicitly, leaving iterative scan as the only lever. A good deal of that is downstream of the tenant key living in JSONB on a single global HNSW graph. With a real column, partitioning and partial indexes are at least *expressible* choices rather than ruled out before they are weighed.

Migration cost looks comparable — both are an iteration over the runtime `rag_*` tables, and `0058_backfill_rag_lookup_indexes` already established that pattern. Worth adding as a **P5** precondition, since it changes Phase 2 and Phase 6.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+These gate the work; none is a code change, and two would change scope.
+
+- **P1 — `document_type` meaning (self-review M2).** Confirm with the issue owner
+  whether #1593's "document type" means the stored **filetype/mime** string (the
+  portable technical canonical this design adopts) or a **semantic** class
+  (contract/invoice/policy). *Plan proceeds on the mime/filetype decision.* If
+  semantic typing is required, it becomes a separate `document_category` dimension
+  (new business metadata + a populate story) — a scope increase to be re-planned, not
+  silently satisfied by mime.
+- **P2 — pgvector version.** Iterative index scan (the H1 mitigation, 9.2) requires
```
DEENUU1: **This cannot be a one-off check, because the image is not pinned to a version.** `docker-compose-dev.yml` and `docker-compose-prod.yml` both use `pgvector/pgvector:pg16` — a floating tag. "Does the image ship ≥ 0.8" therefore resolves to whatever that tag pointed at on the day a given host last pulled, and it can regress on the next pull without anything changing in the repo.

If the H1 recall guarantee depends on `hnsw.iterative_scan`, then pinning the image to a concrete pgvector version in both compose files and in CI **is part of FA-039's deliverable**, not a manual precondition someone ticks once.

Related, and worth making a decision rather than a footnote: §9.2 says the `ef_search`-only fallback "does not claim the same recall". So if the pin does not happen, FA-039 ships with a known recall shortfall on precisely the conjunct that is now mandatory on every single query — the tenant filter, which on a shared table is also the most selective one. That is the trade to put in front of the issue owner explicitly.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+- **`app/services/rag/ingestion.py::ingest_file`:** accept the trusted
+  `organization_id` and business metadata (`source`, `document_type`,
+  `organizational_unit`, `doc_date`) and stamp them onto `document.metadata` **before**
+  `insert_document` (so `_build_chunk_metadata` carries them). `organization_id` is
+  **never** taken from the uploader's `ingestion` form or any model input.
+- **Per-source provenance mapping (codex-3 H6) — where each field is filled, so fields
+  do not silently stay `None`:**
+  - `document_type` = the existing stored `filetype`/mime (a pure derivation, computed
+    for every path; P1).
+  - `source` = a canonical origin string set at the call site: `upload`, the
```
DEENUU1, line 1355: **This line answers "how does the agent know which `source` values exist" — and it contradicts §2.1.**

Here `source` is a canonical string the *code* sets per origin. And the code has exactly four origins: `upload`, `local` (the directory sync), and the two connectors registered in `app/services/rag/sources/__init__.py` — `GoogleDriveSource` and `S3Source`. There is no path by which an author or a model writes an arbitrary `source`.

So `source` is a **closed vocabulary known at build time**, not the free-form dimension §2.1 describes ("otherwise free-form", requiring a runtime facet lookup to be usable). If that holds, it belongs in the tool schema as a `Literal`/enum: the model then reads the legal values straight out of the function schema, cannot guess wrong, and pays nothing at runtime. A new connector adds a value in the same commit that adds the connector.

That in turn leaves `organizational_unit` as the only dimension that genuinely needs discovery at runtime, which changes the shape of M3 substantially (see the comments on §2.1 and §9.4). Worth reconciling the two sections before Phase 4, because it decides whether `list_filter_values` needs to exist at all.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+    connector's name (e.g. `gdrive`, `s3`), or `local` for the directory sync.
+  - `doc_date` (§2.4 precedence): upload → author/uploader-supplied date if the form
+    carries one, else file mtime; connector sync → the source file's modified time
+    (`app/services/rag/sources/base.py` `SourceFile` modified field, S3
```
DEENUU1: **`SourceFile` has no modified-time field, and neither connector fetches one.**

`SourceFile` is `id, name, mime_type, size, path` (`app/services/rag/sources/base.py`). Upstream of it: the Drive connector requests `fields="files(id, name, mimeType, size)"` (`google_drive.py:89`) so `modifiedTime` is never even asked for, and the S3 lister constructs `SourceFile(...)` without `LastModified` (`s3.py:69`).

§9.11 H6 corrected codex on the class *name* — `SourceFile`, not `RemoteFile` — but the field it points at does not exist in either. So step 2 of the `doc_date` precedence is not "read the source's modified time", it is three changes: add the field to `SourceFile`, widen the Drive `fields=` projection, and wire `LastModified` through in S3.

That matters beyond bookkeeping: it is on the critical path for every document whose date should come from the source rather than from ingestion time, which is most of a synced corpus. As written the plan bills it as already available, so the estimate for Phase 5 is light by two connectors.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+contains **no tenant field and no authorization field** — those dimensions are
+structurally *inexpressible* here, so a caller cannot even name them (see 2.2).
+
+Whitelisting for the tool: the agent tool exposes only the whitelisted business
+dimensions (`source`, `document_type`, `organizational_unit`, `date_from`,
+`date_to`). `parent_doc_id` is not offered to the model (the model chooses *what*,
+not *which stored id*); the API keeps it for programmatic callers. Every exposed
+argument is optional and typed on the tool signature so PydanticAI validates it.
+
+**Filter-value discoverability (self-review M3).** `document_type` is validated
```
DEENUU1: Half the premise of this section does not survive §9.5 (see the comment there): `source` is set by the code from a closed set of origins, so it is statically discoverable and belongs in the schema as an enum, not behind a runtime lookup.

That leaves `organizational_unit` as the only genuinely free-form, author-supplied dimension — and `document_type` is already validated against a closed vocabulary by this same section. Scoping M3 to one dimension removes most of the justification for a second tool operation, and makes the remaining affordance much cheaper to justify.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+**translated to its `vector_document_id`** before it reaches the store, or the
+predicate matches nothing. This is stated so an implementer cannot compare the two
+namespaces by accident, and so permission intersection (2.7) and the backfill (2.5)
+join on the same key. A typed `VectorDocumentId` (a thin `str` newtype) is
+preferred over a bare `str` so the distinction is visible in signatures.
+
+Semantics: **within a field, multiple values are OR; across fields, AND.**
+**Tri-state, explicit (codex-2 R1):** for a business filter dimension, `None` (the
+field is absent) imposes no restriction; a **non-empty** list is an allow-list; an
+**empty list is rejected at validation** (a supplied-but-empty filter is a caller
```
DEENUU1, line 113: **Right for the API body, wrong at the tool boundary — and the plan treats them as one boundary.**

The reasoning holds where it was aimed: a filter body arriving over HTTP with `[]` is genuinely ambiguous, and reading it as "match everything" is a fail-open. But models routinely emit `[]` instead of omitting an optional list, and there it unambiguously means "I am not filtering this dimension". As written, that costs a retry on an intention that was already correct — and the retry is spent on a model that has to guess what it did wrong.

Suggest: normalize `[] → None` in the tool before constructing `RetrievalFilters`, and keep the hard rejection on the API schema where the ambiguity is real. Same type, two boundaries, different handling — which is worth saying explicitly in §2.1, because §9.4 currently just says "assemble a `RetrievalFilters` inside the tool" and inherits this rule by silence.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+  (`distinct_metadata_values(collection, keys, scope)` — a `RetrievalScope`, not a bare
+  nullable UUID, for consistency with §9.2's "every scoped op takes a scope" rule →
+  `SELECT DISTINCT metadata->>'X' … WHERE metadata->>'organization_id' = :org`). A
+  `RAGFilterValues` response schema in `app/schemas/rag.py`.
+- **Agent tool `app/agents/capabilities/knowledge/_toolset.py` + `_search.py`
+  (+ `_capability.py`):**
+  - Expose the **whitelisted** business filters as optional, typed tool parameters on
+    `search_documents` (`source`, `document_type`, `organizational_unit`, `date_from`,
+    `date_to`) so PydanticAI validates them; **do not** expose `parent_doc_id`
+    (§2.1). Assemble a `RetrievalFilters` inside the tool.
```
DEENUU1, line 1283: **Two things this needs to say, because both are invisible in the current phrasing.**

**(a) A validation failure must not reach the model as an outage.** `_toolset.py` wraps the whole call in a single `except Exception` returning `steer(ctx, "Knowledge base temporarily unavailable, please try again.")` — deliberately, so an error never reads as "nothing found". But if `RetrievalFilters(...)` is constructed inside the tool body, an unknown `document_type` or a reversed date range now lands in that same handler. The model is told the knowledge base is down, which it cannot act on, and which is not true. It needs its own `except ValidationError` that steers with the actual field error, so the next attempt is a corrected call rather than a repeat.

**(b) `extra="forbid"` is not what protects this path.** On the tool the arguments are named parameters on a typed signature, so the signature is the whitelist; `extra="forbid"` is the guard for the API body, where a caller controls the JSON. Worth stating, so the §9.4 test for smuggled keys is written against the API route (where it is meaningful) and nobody reads the tool as covered by the same mechanism.

**(c)** Minor, same area: if `document_type` is validated against a closed vocabulary, give the model the vocabulary in the schema (`Literal`) rather than a bare `str` it can fail. It will otherwise write `"PDF"` or `"pdf document"` and learn the legal set one rejection at a time.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+  - Build `RetrievalScope` from `ctx.deps.organization_id` (never from parameters);
+    thread it and the filters through `search_knowledge_base` →
+    `retrieve`/`retrieve_multi`.
+  - **Discoverability for the model (M3) — one concrete choice (codex-3 M10):** add a
+    **second whitelisted tool operation** `list_filter_values` on the knowledge
```
DEENUU1: **A permanent second tool is the expensive option, not the cheap one.**

It sits in the schema of every agent with knowledge enabled — including every agent that will never filter — so it is tokens on every turn of every run, forever. The alternative rejected here (resolving values at run setup) costs one lookup per run. The stated reason for rejecting it, "extra resolution per run for data the model rarely needs", actually argues the other way: rarely needed is the case for *not* making it a permanent schema entry.

And per the §9.5 comment, if `source` becomes an enum in the schema, this tool exists for `organizational_unit` alone — one free-form dimension, on the corpora that actually use it.

If it stays, consider registering it conditionally: only when the agent's bound collections actually carry non-empty `organizational_unit` values. Agents with untagged corpora then pay nothing, and the ones that do get the affordance where it means something.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+---
+
+## 9. Implementation plan
+
+Sections 1–8 are settled. This section turns them into an ordered, file-level plan.
+It is grounded in the code as it stands on `feat/1593-rag-metadata-filters`
+(migration head `0077_drop_allow_byo`, verified). It adds **no production code** —
+it is the plan the implementation PR(s) will follow.
+
+### 9.0 Preconditions / open decisions (resolve before coding)
```
DEENUU1, line 976: **The split should go four ways, not two.** P3 correctly lifts out the backfill, but there are four separable pieces in this document with quite different urgency:

1. **The cross-tenant write clobber (R2)** — `IngestionService` looks a document up unscoped by `source_path`/`filename`/`content_hash` and then deletes the `parent_doc_id` it gets back.
2. **The filter contract and its enforcement** — the actual FA-039 feature.
3. **The tenant backfill, rollout ordering and the enforcement flag** (already P3).
4. **Ingestion provenance (H6)** — which, per the §9.5 comment, means changes in both connectors and the upload route.

(1) is a live security defect with no dependency whatsoever on metadata filters, and right now it is the most urgent item in the document while being sequenced behind a tender feature's backfill and flag flip. It wants its own fix PR, now, ahead of FA-039 — scoping the dedup lookup and the delete is a small, self-contained change with a regression test.

Doing that also shrinks what the §9.6 flag has to protect, which is the subject of a separate comment.

.planning/1593-design.md
```diff
@@ -0,0 +1,1595 @@
...
+     date predicate or builds the partial date index is deployed** (codex round-2 R1):
+     `_ensure_collection` runs `CREATE INDEX … rag_safe_to_date(…)` on every ingest and
+     a date business filter is not gated by the enforcement flag, so "before flag-on" is
+     not late enough. Then ship the tenant **dual-write** (Phase 5: new ingests stamp
+     `organization_id`) and deploy everywhere;
+  2. **drain/upgrade old workers** so nothing still writes untagged chunks;
+  3. run `0079` **backfill and verify** (report reviewed, quarantine understood);
+  4. **enable tenant enforcement across every scoped operation — search, listing,
+     counts, chunk fetch, dedup lookup, deletion and the insert conflict policy — not
+     reads alone.** A single `RAGSettings` flag gates all of them **within a process**,
```
DEENUU1, line 1487: **Worth asking whether this flag should exist at all.**

Its "off" state is defined as today's behaviour: no tenant conjunct. So the design deliberately keeps a switch whose off position is the vulnerability, plus — per §9.7 — a doubled test matrix ("enforcement off vs on", across backfilled and untagged rows), plus step 5's promise to remove it later in a follow-up that has no owner named.

What the flag buys is a safe rollout across an installed base being migrated. AgenticOS does not have one yet. Without deployments carrying legacy untagged chunks, there is nothing to roll: the backfill runs, it is verified, enforcement is on. No flag, no off-state, no off/on test axis, and R2's monotonic-mix argument becomes unnecessary because there is no mix.

If some deployment genuinely does need the staged path, the flag is worth its cost — but that should be a named case, not the default assumption. And if it stays, step 5 needs an owner and a target release now, for the same reason §9.4's shim removal was given one: a temporary switch in a security path with no removal date is a permanent switch.

DEENUU1 (commented):
Reviewed the design and the plan, and checked every claim it makes about the code against `main` rather than taking them on trust. The verified ones all hold: `_parent_doc_id_from_filter` honouring one clause and discarding the rest, `search()` taking `organization_id` only to resolve the embedding key and never as a `WHERE` conjunct, `ON CONFLICT (id) DO UPDATE` with no owner predicate, `BaseSchema` not forbidding extras, `AgentDeps.organization_id` being nullable, the local sync ingesting with no org, and `PLATFORM_PACKAGES = ("app/agents",)` with per-file exceptions. That is a high hit rate for a document this size, and the two-type separation (`RetrievalFilters` vs `RetrievalScope`) is the right shape — widening is structurally inexpressible rather than defended by a check someone can forget.

The most valuable thing here is not the filter contract: it is the cross-tenant write clobber in R2. `rag_<collection_name>` is keyed by name only, and `IngestionService` looks a document up by `source_path`/`filename`/`content_hash` with no `organization_id`, then deletes the `parent_doc_id` it gets back. One tenant's ingest can replace and delete another's document. That is a live defect and it is currently sequenced behind a tender feature's backfill and rollout flag.

Comments inline, roughly in order of how much they change. The short version:

- The two migration numbers the plan fixes are already taken on `main`.
- The tenant conjunct lands in JSONB, which is the one place the design settles for less than it does everywhere else — and it is what makes the H1 recall mitigation hard.
- `source` is a closed set the code controls, which contradicts §2.1 and changes what the discoverability affordance needs to be.
- `SourceFile` has no modified-time field and the Drive connector does not fetch one, so `doc_date` provenance step 2 is unimplemented work billed as available.
- The scope should split four ways, not two.

Two process notes that do not fit a line anchor:

**The review log is now most of the document.** §5, §6, §8, §9.11 and §9.12 are roughly 600 of 1595 lines, and every one of their findings is already applied to §1–§9 — that is the point of them. As merged repo content they will rot on first contact with implementation, and nobody updates a critique log. For an implementer the useful document is §1–§4 plus §9. I would move the verdict logs into the PR conversation or a collapsed appendix and keep the design sections as the artifact.

**`.planning/` has no convention yet.** This PR and its two siblings (#1654, #1655) introduce a new tracked top-level directory. Worth deciding once, for all three: is it long-lived repo content, or a working artifact that belongs in `.gitignore`? Design documents describing a future state, living next to `docs/` which documents the current one, will drift the moment implementation deviates — and `scripts/docs_drift.py` does not watch this path, so nothing will say so.

.planning/1593-design.md
MichalZelasko: Addressed. §2.3 (lines 258-273) records a real `organization_id uuid` column as the sturdier alternative for the one security-bearing dimension — `NOT NULL`-provable after backfill, cheaper to filter, and it makes the H1 partitioning / partial-index mitigations at least expressible rather than ruled out because the key lives in JSONB. It is filed as precondition P5 in §9.0 (lines 1010-1018), noted as changing Phase 2's WHERE/index shape and Phase 6's backfill.

.planning/1593-design.md
MichalZelasko: Addressed. §2.1 (lines 151-177) now classifies `source` as a build-time-closed vocabulary (the four code-set origins `upload`/`local`/`gdrive`/`s3`) alongside `document_type`, and both ship as a `Literal`/enum in the tool schema per §9.4 (lines 1301-1308). Runtime facet discovery (M3) is scoped to `organizational_unit` alone.

.planning/1593-design.md
MichalZelasko: Addressed. §2.1 "One rule, two boundaries" (lines 117-126) keeps the hard `[] → 422` on the API body and normalizes `[] → None` at the agent-tool boundary; §9.4 (lines 1309-1314) applies that normalization before constructing `RetrievalFilters`, so a model's not-filtering `[]` costs no retry.

.planning/1593-design.md
MichalZelasko: Addressed on all three points: (a) §9.4 (lines 1315-1324) adds a dedicated `except ValidationError` that steers the model with the actual field error instead of routing it through the broad "temporarily unavailable" handler; (b) §2.1 (lines 136-141) states `extra="forbid"` guards the API body, and the smuggled-key rejection test belongs on the API route — the tool's typed signature is its whitelist; (c) `document_type` ships as a `Literal` in the tool schema per §9.4 (lines 1301-1308).

.planning/1593-design.md
MichalZelasko: The four-way split and the "fix R2 first" decision are recorded as precondition P6 in §9.0 (lines 1019-1033): the cross-tenant write clobber is a live defect with no dependency on metadata filters and should ship as its own fix PR ahead of FA-039. That split has now been filed as its own issue, #1684 (cited at line 1027), so it is tracked outside this design doc rather than fixed inside it.

.planning/1593-design.md
MichalZelasko: Fixed in this branch (15a03e17). §9.0 adds precondition P7 (lines 1035-1050): the staged enforcement flag's off-state is today's no-tenant-conjunct behaviour (the vulnerability), and its only benefit is a rolling migration of an installed base holding pre-FA-039 untagged chunks — which AgenticOS does not have yet. So the flag is now a named case for deployments carrying legacy untagged chunks, confirmed with the issue owner, not the default; where none exists, enforcement ships on with no off-state and no off/on test axis. §9.6 step 5 (lines 1518-1524) now requires the flag-removal follow-up to name an owner and a concrete target release, mirroring §9.4's shim removal.

MichalZelasko:
@DEENUU1 fixes to plan applied. For flagged preexisting issues, two fixes prepared: #1687 and #1688. Let me know, if I am good to go with implemantation.

MichalZelasko:
Triaged the four `github-code-quality` "Unused global variable" findings on `backend/alembic/versions/0081_rag_metadata_prereqs.py` (`revision`, `down_revision`, `branch_labels`, `depends_on`) — all **false positives**.

These are Alembic's **required module-level attributes**: Alembic's migration-script loader imports each revision module and reads them *by name* to establish the revision's identity and its position in the chain. They're consumed via runtime attribute access the analyzer's static "unused" check can't follow. Every migration under `backend/alembic/versions/` declares the identical four, and removing any (e.g. `revision`/`down_revision`) breaks `alembic upgrade`. No change.

MichalZelasko:
@codex review

backend/app/services/rag/vectorstore.py
```diff
@@ -282,6 +307,74 @@ def _group_documents(self, results: list[dict[str, Any]]) -> list[DocumentInfo]:
...
+
+    A `TenantScope` contributes the mandatory tenant conjunct and, when the
+    authorization slot is populated, the per-document restriction. The unscoped
+    maintenance marker contributes nothing. Every value is bound.
+    """
+    conjuncts: list[str] = []
+    params: dict[str, Any] = {}
+    if isinstance(scope, TenantScope):
+        conjuncts.append("metadata->>'organization_id' = :scope_org")
+        params["scope_org"] = str(scope.organization_id)
```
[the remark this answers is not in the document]
MichalZelasko: By design, not premature enablement. FA-039 chose fail-closed: a chunk missing `metadata.organization_id` is never returned to a tenant-scoped caller — silence, never leakage (plan \S2.4, lines 399-403). The ownership backfill onto pre-existing chunks is the deferred **P3** item; migration `0083` intentionally ships indexes only. Disabling the tenant conjunct until backfill would invert that invariant — legacy chunks would become tenant-searchable with no verified owner. Legacy collections stay invisible until (re-)ingested or backfilled under a real org, which is the documented recovery path.

backend/app/api/routes/v1/rag.py
```diff
@@ -258,27 +260,61 @@ async def search_documents(
     """
     names = request.collection_names or [request.collection_name]
     collections = [kb.collection_name for kb in await access.readable_all(ctx, names)]
+    # The tenant conjunct, built only from trusted context after collection
+    # access is resolved. The request body carries business filters only, so a
+    # supplied filter can only narrow, never widen, access (FA-039 §2.2).
+    scope = TenantScope(organization_id=ctx.organization_id)
```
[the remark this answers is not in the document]
MichalZelasko: By design under FA-039. The scope is derived from trusted context per plan \S2.2; documents ingested with `organization_id=None` (app-scoped / local-directory) are **deployment-scoped, not tenant-scoped**, and reachable only through the explicit unscoped maintenance path (CLI / app-admin marker), never an ordinary tenant search or agent tool (plan \S2.5, lines 423-435). A deployment that wants such content tenant-searchable must (re-)ingest it under a real organization — the plan states this explicitly. Deployment-wide tenant-route search of app-scoped bases is out of scope for this PR; the related cross-tenant `get_documents`/`delete_document` concerns are tracked under #1684.

backend/app/services/rag/ingestion.py
```diff
@@ -122,6 +135,23 @@ async def ingest_file(
                 document.metadata.source_path = source_path
                 document.metadata.filename = Path(source_path).name
 
+            # Trusted, security-bearing tenant plus the business dimensions. Set
+            # before insert_document so _build_chunk_metadata carries them.
+            document.metadata.organization_id = organization_id
+            document.metadata.source = source
+            document.metadata.organizational_unit = organizational_unit
+            document.metadata.doc_date = doc_date
```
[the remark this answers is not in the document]
MichalZelasko: Recorded gap, by design for this PR. `organizational_unit` is author-supplied business metadata (plan \S2.5, lines 436-446); the filter dimension, its validation, the store conjunct and the distinct-value discovery endpoint (`GET .../filter-values`) are wired end-to-end. The production population source — an author/connector input threaded through the worker — is deliberately future work, and the field defaults to `None` until then. This is a not-yet-populated dimension, not a defect in the retrieval path this PR adds.

backend/app/services/rag/config.py
```diff
@@ -227,6 +227,23 @@ class RAGSettings(BaseModel):
...
+    # bounded by `max_scan_tuples`; the raised `ef_search` is the fallback.
+    #
+    # `iterative_scan` and `max_scan_tuples` require pgvector >= 0.8.0. On an
+    # older image set `hnsw_iterative_scan = false`: the store then tunes only
+    # `ef_search` (available on every pgvector), a stopgap that does not claim
+    # the same recall. The store applies these per search transaction, so
+    # unfiltered paths and other statements are unaffected.
+    hnsw_iterative_scan: bool = True
+    hnsw_max_scan_tuples: int = 20000
+    hnsw_ef_search: int = 100
```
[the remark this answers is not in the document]
MichalZelasko: Out of scope under precondition **P2** (pgvector >= 0.8), which this PR keeps intact. On a supported image all three knobs apply cleanly in the per-search `SET LOCAL` savepoint, nothing rolls back and nothing is logged. `hnsw_iterative_scan=false` is a documented stopgap for an unsupported older image — not a production configuration path, since supported deployments run pgvector >= 0.8 where iterative scan is always available. Auto-detecting the extension version or wiring an env-backed override is a possible future enhancement, not required given P2.

backend/app/agents/capabilities/knowledge/_toolset.py
```diff
@@ -24,30 +39,70 @@ def build_knowledge_toolset(*, default_top_k: int) -> FunctionToolset[AgentDeps]
     """
 
     async def search_documents(
-        ctx: RunContext[AgentDeps], query: str, top_k: int | None = None
+        ctx: RunContext[AgentDeps],
+        query: str,
+        top_k: int | None = None,
+        source: list[Source] | None = None,
+        document_type: list[str] | None = None,
+        organizational_unit: list[str] | None = None,
```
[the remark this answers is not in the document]
MichalZelasko: Fixed in `bce10c91`. `document_type` now ships as a closed `DocumentType` StrEnum (derived from `DOCUMENT_TYPE_VOCABULARY`) on the tool signature, mirroring `source`, so the generated tool schema advertises the same closed filetype set the API validator enforces. This matches the FA-039 plan decision (\S2.1, sharpened for PR #1656) that both build-time-closed dimensions ship as a `Literal`/enum in the tool schema. Thanks.

backend/alembic/versions/0083_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+
+from collections.abc import Sequence
+
+from sqlalchemy import text
+from sqlalchemy.engine import Connection
+
+from alembic import op
+
+revision: str = "0083_rag_metadata_prereqs"
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `revision` is a module-level global Alembic **requires** on every migration script — the framework reads it by name (via reflection over the module) to build the revision graph, so it is consumed by Alembic, not by code in this file. Every script in `backend/alembic/versions/` declares it the same way; it cannot be removed.

backend/alembic/versions/0083_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+down_revision: str | Sequence[str] | None = "0082_portal_account_id"
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `down_revision` is a module-level global Alembic **requires** on every migration script — the framework reads it by name to build the revision graph (it is what points this migration at its parent). It is consumed by Alembic, not by code in this file, and every script in `backend/alembic/versions/` declares it the same way.

backend/alembic/versions/0091_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+from collections.abc import Sequence
+
+from sqlalchemy import text
+from sqlalchemy.engine import Connection
+
+from alembic import op
+
+revision: str = "0083_rag_metadata_prereqs"
+down_revision: str | Sequence[str] | None = "0082_portal_account_id"
+branch_labels: str | Sequence[str] | None = None
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `branch_labels` is a module-level global Alembic reads by name on every migration script (part of the revision-graph contract). It is consumed by the framework, not by code in this file, so `unused` does not apply; every script in `backend/alembic/versions/` declares it.

backend/alembic/versions/0091_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+depends_on: str | Sequence[str] | None = None
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `depends_on` is a module-level global Alembic reads by name on every migration script (part of the revision-graph contract). It is consumed by the framework, not by code in this file, so `unused` does not apply; every script in `backend/alembic/versions/` declares it.

MichalZelasko:
@codex review

.planning/1593-design.md
MichalZelasko: Addressed in implementation. The migration was regenerated against the then-current head per M11 rather than the numbers this prose carries: it ships as `0085_rag_metadata_prereqs` with `down_revision = "0084_retention_policies"` (rebased from the original 0081/0083 as parallel branches reached `main` first — see the cross-branch note in the migration file). The revision id is 25 characters, within the 32-character cap you flagged, and there is no `0078_rag_metadata_prereqs` / `0079_rag_tenant_backfill` collision on the branch. The §9.6 / §9.9 prose numbers are historical design-doc text and out of scope to edit on this PR; the code path is correct.

.planning/1593-design.md
MichalZelasko: Out of scope under precondition P2, which this PR keeps intact: supported deployments run pgvector >= 0.8, where `hnsw.iterative_scan` is always available, so the mandatory tenant conjunct's recall is not at risk on a supported image. Pinning the compose images to a concrete pgvector version (and a matching CI pin) is an infra/deploy change outside FA-039's retrieval scope; the `ef_search`-only fallback is the documented stopgap for an unsupported older image, not a production path. No code change on this PR.

.planning/1593-design.md
MichalZelasko: Recorded gap, deferred by design for this PR — the same boundary as the `organizational_unit` population gap. `SourceFile` still carries no modified-time field and neither connector fetches one, so the source-modified-time step of the `doc_date` precedence is intentionally not wired here. The retrieval/filter path this PR ships is complete: the `date_from`/`date_to` predicate over the stored `doc_date`, the IMMUTABLE `rag_safe_to_date` helper, and the partial date index. Adding the `SourceFile` field, widening the Drive `fields=` projection and threading S3 `LastModified` is future connector work, tracked with the ingestion-provenance items rather than part of this PR's retrieval contract.

.planning/1593-design.md
MichalZelasko: Addressed. `source` now ships as a build-time-closed `Source` StrEnum (`upload`/`local`/`gdrive`/`s3`) on the agent tool and is validated by `RetrievalFilters` at the API boundary, so it is statically discoverable and needs no runtime lookup — exactly as this comment argued. Runtime facet discovery is therefore scoped to `organizational_unit` alone, and it is an API endpoint (`GET /collections/{name}/filter-values`), not a second permanent agent tool. Same reconciliation recorded on the §2.1 / §9.4 threads.

.planning/1593-design.md
MichalZelasko: Addressed — no second permanent agent tool was added. The knowledge capability ships exactly one tool (`search_documents`); `source` and `document_type` are closed enums in its schema, so they cost nothing extra at runtime and the model reads the legal set straight from the function schema. Facet discovery for the one genuinely free-form dimension (`organizational_unit`) is an API endpoint (`GET /collections/{name}/filter-values`), not a schema entry present on every turn of every run — which is the cheaper option this comment favoured.

backend/alembic/versions/0085_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+
+from collections.abc import Sequence
+
+from sqlalchemy import text
+from sqlalchemy.engine import Connection
+
+from alembic import op
+
+revision: str = "0085_rag_metadata_prereqs"
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `revision` is a module-level global Alembic **requires** on every migration script — the framework reads it by name (via reflection over the module) to build the revision graph, so it is consumed by Alembic, not by code in this file. Every script in `backend/alembic/versions/` declares it the same way; it cannot be removed. (Same disposition as the already-resolved threads on this file's earlier revision numbers.)

backend/alembic/versions/0085_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+down_revision: str | Sequence[str] | None = "0084_retention_policies"
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `down_revision` is a module-level global Alembic **requires** on every migration script — the framework reads it by name to build the revision graph (it is what points this migration at its parent, `0084_retention_policies`). It is consumed by Alembic, not by code in this file, and every script in `backend/alembic/versions/` declares it the same way. (Same disposition as the already-resolved threads on this file's earlier revision numbers.)

backend/alembic/versions/0091_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+from collections.abc import Sequence
+
+from sqlalchemy import text
+from sqlalchemy.engine import Connection
+
+from alembic import op
+
+revision: str = "0085_rag_metadata_prereqs"
+down_revision: str | Sequence[str] | None = "0084_retention_policies"
+branch_labels: str | Sequence[str] | None = None
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `branch_labels` is a module-level global Alembic reads by name on every migration script (part of the revision-graph contract). It is consumed by the framework, not by code in this file, so `unused` does not apply; every script in `backend/alembic/versions/` declares it. (Same disposition as the already-resolved threads on this file's earlier revision numbers.)

backend/alembic/versions/0091_rag_metadata_prereqs.py
```diff
@@ -0,0 +1,124 @@
...
+depends_on: str | Sequence[str] | None = None
```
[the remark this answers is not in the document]
MichalZelasko: False positive. `depends_on` is a module-level global Alembic reads by name on every migration script (part of the revision-graph contract). It is consumed by the framework, not by code in this file, so `unused` does not apply; every script in `backend/alembic/versions/` declares it. (Same disposition as the already-resolved threads on this file's earlier revision numbers.)

DEENUU1:
**Blocked on a design decision, not on the merge.** #1688 landed (0.0.465) and I tried to rebase this onto it. Most of the 23 conflict hunks resolve cleanly in this branch's favour — the typed `RetrievalQuery` subsumes main's `organization_id` + `tenant` arguments, and this branch's rerank-over-filtered-candidates even removes the unscoped `get_documents()` probe that #1688 had to work around. But one case has no expressible answer in the merged code, so I stopped rather than guess.

**An app-scoped knowledge base reads nothing.** #1688 stamps chunks with `KnowledgeBase.vector_tenant`, which is `NULL` for an app-scoped base — the deployment-wide collection every organization reads — and its `resolve_tenant` returns `None` for one, so the store matches exactly those untagged rows. This branch gives an ordinary caller exactly one shape, `TenantScope(organization_id: UUID)` (non-null by construction), which ANDs `tenant = <org>`; that matches none of the untagged rows. `UnscopedScope` is not the answer either — it ANDs no tenant predicate at all, which on a shared table is a cross-tenant read.

So a naive merge makes every shared, app-scoped collection silently return an empty result set for ordinary callers. Neither PR's tests catch it, because each is internally consistent: #1688 has no typed scope, and this branch has no app-scoped base in its fixtures.

Three ways out, and the choice is yours since you wrote both:

1. `TenantScope` gains an app-scoped variant that ANDs `tenant IS NULL` (keeps #1688's stamping, keeps "narrowing-only", adds one scope shape).
2. App-scoped bases stamp the reading organization at ingest — contradicts #1688's `vector_tenant`, and means the same shared document is stored per tenant.
3. The store treats an app-scoped collection as unscoped for reads — simplest, and the one that needs the hardest look, because it is a cross-tenant predicate on a shared table.

Option 1 looks right to me and is the smallest change, but it is a security-bearing decision on a tender-committed item (FA-039), so I have left this branch exactly as you pushed it. Two mechanical things also wait for whoever rebases: the migration is `0085`, which main now uses (`0085_refresh_reuse`), so it needs `0086`+; and `Refs #1593` wants to become `Closes #1593` if this is the whole of FA-039 — with `organizational_unit` still having no production population source, that may deserve a follow-up issue instead.

MichalZelasko:
> **Blocked on a design decision, not on the merge.** #1688 landed (0.0.465) and I tried to rebase this onto it. Most of the 23 conflict hunks resolve cleanly in this branch's favour — the typed `RetrievalQuery` subsumes main's `organization_id` + `tenant` arguments, and this branch's rerank-over-filtered-candidates even removes the unscoped `get_documents()` probe that #1688 had to work around. But one case has no expressible answer in the merged code, so I stopped rather than guess.
> 
> **An app-scoped knowledge base reads nothing.** #1688 stamps chunks with `KnowledgeBase.vector_tenant`, which is `NULL` for an app-scoped base — the deployment-wide collection every organization reads — and its `resolve_tenant` returns `None` for one, so the store matches exactly those untagged rows. This branch gives an ordinary caller exactly one shape, `TenantScope(organization_id: UUID)` (non-null by construction), which ANDs `tenant = <org>`; that matches none of the untagged rows. `UnscopedScope` is not the answer either — it ANDs no tenant predicate at all, which on a shared table is a cross-tenant read.
> 
> So a naive merge makes every shared, app-scoped collection silently return an empty result set for ordinary callers. Neither PR's tests catch it, because each is internally consistent: #1688 has no typed scope, and this branch has no app-scoped base in its fixtures.
> 
> Three ways out, and the choice is yours since you wrote both:
> 
>     1. `TenantScope` gains an app-scoped variant that ANDs `tenant IS NULL` (keeps [fix(rag): scope RAG ingestion and reads by tenant (#1684) #1688](https://github.com/vstorm-co/agenticos/pull/1688)'s stamping, keeps "narrowing-only", adds one scope shape).
> 
>     2. App-scoped bases stamp the reading organization at ingest — contradicts [fix(rag): scope RAG ingestion and reads by tenant (#1684) #1688](https://github.com/vstorm-co/agenticos/pull/1688)'s `vector_tenant`, and means the same shared document is stored per tenant.
> 
>     3. The store treats an app-scoped collection as unscoped for reads — simplest, and the one that needs the hardest look, because it is a cross-tenant predicate on a shared table.
> 
> 
> Option 1 looks right to me and is the smallest change, but it is a security-bearing decision on a tender-committed item (FA-039), so I have left this branch exactly as you pushed it. Two mechanical things also wait for whoever rebases: the migration is `0085`, which main now uses (`0085_refresh_reuse`), so it needs `0086`+; and `Refs #1593` wants to become `Closes #1593` if this is the whole of FA-039 — with `organizational_unit` still having no production population source, that may deserve a follow-up issue instead.

Thanks, good catch. I will go with option 1.

DEENUU1 (dismissed):
Reviewed the whole diff and merged `main` (0.0.470) into the branch — it merges
cleanly, and on the merged tree `make lint` is clean, `make test` is 9867 passed
at 100.00% on the platform layer, `make test-frontend-cov` is 438 files / 6394
tests at 100% statements and lines, and `make test-migrations` cycles the chain
forwards and back with `0091_rag_metadata_prereqs` as the single head.

**The design decision is resolved, and resolved well.** `AppScope` is option 1 as
discussed: it ANDs `tenant IS NULL` rather than an equality, so it narrows
exactly as `TenantScope` does and reads no other organization's rows, and
`scope_for_tenant` builds the shape from the *base's* own resolved tenant rather
than from a caller's organization id. `e3407abf` closes the half I had not seen
coming — `ingest_file` was stamping `document.metadata.organization_id` from the
caller-supplied paying organization, which for an app-scoped base disagrees with
the ingester's own `None` and left those chunks reachable by neither scope.
Mirroring it from `self._tenant` alone means the two cannot disagree at all.

The rest reads clean where it matters most. Every value in `_retrieval_conjuncts`
is bound and the only literals are fixed key names; `_FACET_KEYS` whitelists the
one place a key name *is* interpolated; the empty authorization set becomes
`FALSE` rather than being conflated with an unset slot; the search tuning is
`SET LOCAL` inside a savepoint, so an older pgvector degrades to `ef_search`
instead of failing the search; and `resolve_legacy_filter` refuses rather than
ignores, which is the whole point of retiring that string.

---

**One change I made on the branch** (`f559ecb8`), and it is the only finding:

**`retrieve_multi` took a `scopes` map and a single `scope` beside it**, reading
a name absent from the map under that shared one. Both callers already build the
map from the same bases they build the name list from, so the fallback could not
fire — but what it would do if it ever did is read one collection under a
*different* collection's tenant, which is the single mistake the whole parameter
exists to prevent. There was even a test pinning that behaviour, which is what
made it worth changing rather than leaving: a security-bearing default that reads
as deliberate. `scopes` is now the only parameter, a name without one is refused
before the first collection is read, and both call sites lost the
`scope=scopes[names[0]]` boilerplate.

Two smaller ones in `f8a27405`: `0091`'s `downgrade` dropped `_org_idx` along
with the three indexes it creates, but `0086_scope_rag_rows_by_org` owns that one
and drops it in its own `downgrade` — so a chain rolled back only past `0091`
lost an index an earlier revision is still responsible for. And
`filters.__all__` now names `AppScope` and `scope_for_tenant`, which the search
route and the knowledge capability both import.

**Two things I filled in rather than asked for:**

- **The feature was undocumented.** `docs/file-processing.md` gained a *Narrowing
  a search* section — the four dimensions, OR-within/AND-across, the fail-closed
  missing dimension, why a filter cannot reach another tenant, the facet
  endpoint, and the deprecated `filter` string's 400 — with the Polish, German
  and Spanish pages beside it. A BREAKING change to a public route wants a page
  saying so, not only a changelog line.
- **The changelog entry had landed inside the released `0.0.469` section.**
  Merging `main` put it there: the cut promoted `[Unreleased]` underneath an
  entry written before it. Moved back up to `[Unreleased]`.

**On the open items from your own notes:** #1684 landed as #1688; `document_type`
= filetype/mime is confirmed; `pgvector/pgvector:pg16` ships 0.8.6, so the
iterative scan is live on the image this repository runs and the savepoint covers
anything older; the tenant backfill stays its own task. `organizational_unit`
having no production population source is filed as #1777 — the filter, the facet,
the index and the fail-closed semantics all ship, only the writing end is
missing, so `Refs #1593` is right and the issue stays open.

Issue #1416, opened 2026-09-04
Frontend security headers: a real Content-Security-Policy, not only frame-ancestors


DEENUU1:
Part of the security programme (#1424). Pre-existing.

## What is there

`SECURITY.md` says the frontend sets `frame-ancestors 'none'` and that HSTS is set in production. `frontend/next.config.ts` holds the headers block. That is the whole policy.

## What a review expects

- A `Content-Security-Policy` with `default-src 'self'`, an explicit `connect-src` (the API origin and the WebSocket), `img-src` allowing `data:` for the brand glyphs and avatars, `script-src` without `unsafe-inline` where Next allows it (nonce-based), `object-src 'none'`, `base-uri 'self'`.
- **The embedded widget and the hosted page are the exception, on purpose:** they are meant to be framed by a customer's site, so their routes need a different `frame-ancestors` (the embed's allowed origins, which the embed row already knows) rather than `'none'`. Getting this wrong breaks every widget install; test it.
- `Referrer-Policy: strict-origin-when-cross-origin`, `Permissions-Policy` denying camera/microphone/geolocation except where the speech-to-text feature needs the microphone, `X-Content-Type-Options: nosniff`.

## Done when

- [ ] The headers are set, the widget still loads inside a third-party page in the e2e suite, and the console has no CSP violations in the browser log
- [ ] A vitest spec reads the config and asserts each header
- [ ] `docs/deployment.md` documents the policy and what a reverse proxy must not override

Related: #1424, #702 (proxy routes echoing content type - same family). Cluster in #168: *Security programme*.

OchnikBartek:
Delivered by #1580 (merged): the console now serves a real Content-Security-Policy from `src/middleware.ts` — `default-src 'self'`, an explicit `connect-src` built per-deployment from `PUBLIC_API_URL`/`PUBLIC_WS_URL`, `img-src` with `data:`, `object-src 'none'`, `base-uri 'self'` and `frame-ancestors 'none'`, all asserted by `csp.test.ts`.

The one remaining bullet — `script-src` without `'unsafe-inline'` — is the riskiest and can only be verified in a running browser (a missed nonce is a blank page, and `make check` cannot catch it). It is tracked as its own issue: #1624. Closing this as delivered-bar-the-nonce, with the nonce work carried forward there.

PR #1659, opened 2026-09-14
feat(security): drop 'unsafe-inline' from script-src via a per-request nonce

### LINKED ISSUE #1624: CSP: drop 'unsafe-inline' from script-src via a per-request nonce
Follow-up to #1416. PR #1580 shipped the real Content-Security-Policy — `default-src 'self'`, an explicit `connect-src` built from `PUBLIC_API_URL`/`PUBLIC_WS_URL`, `img-src` with `data:`, `object-src 'none'`, `base-uri 'self'`, `frame-ancestors 'none'`, served per request from `src/middleware.ts`. The one bullet of #1416 it did not do is the last, riskiest one: **`script-src` without `'unsafe-inline'`.**

## Why it's separate

`script-src` still carries `'unsafe-inline'` because Next's app router inlines flight data. Removing it needs a **per-request nonce**: the middleware (which already stamps the CSP) generates a nonce, puts `'nonce-<value>' 'strict-dynamic'` in `script-src`, and Next stamps the nonce on its own inline scripts. `'unsafe-eval'` stays where Next's dev runtime needs it.

The reason this is a follow-up and not part of #1580: **it can only be verified in a running browser.** A single inline script that misses the nonce is a blank page, and nothing in `make check` catches it — the gate is opening the console (chat, the Builder, the dashboard, the hosted embed page) and confirming zero CSP violations in the browser console. Shipping the `unsafe-inline` removal without that check risks a broken app, so it wants a session that can stand the stack up and drive a browser.

## Done when

- [ ] `script-src` no longer contains `'unsafe-inline'`; a per-request nonce is generated in the middleware and reaches Next's inline scripts
- [ ] No CSP violation in the browser console across chat, the Builder, the dashboard and the hosted embed page
- [ ] `frontend/src/lib/csp.test.ts` asserts `script-src` carries a nonce placeholder and not `'unsafe-inline'`
- [ ] `SECURITY.md` / `docs/deployment.md` updated

Refs #1416. Part of the security programme (#1424). Cluster in #168: *Security programme*.

OchnikBartek:
## Problem

Follow-up to #1416 / #1580. The console's CSP shipped with everything locked down
except the last, riskiest bullet: `script-src` still carried `'unsafe-inline'`,
because Next's app router inlines its flight data.

## What this does

`script-src` no longer allows `'unsafe-inline'`. The middleware mints a 128-bit
per-request nonce, writes `'nonce-…' 'strict-dynamic'` into the directive, and
forwards it on the **request** headers so Next stamps that nonce onto its own
inline scripts; the response carries the same policy. `'unsafe-eval'` stays for
Next's dev runtime, and `connect-src` (which governs the chat WebSocket) is
untouched.

- `frontend/src/lib/csp.ts` — `cspDirectives`/`contentSecurityPolicy` take the nonce.
- `frontend/src/middleware.ts` — generate the nonce, forward it through next-intl.
- Tests: `csp.test.ts` asserts `script-src` carries the nonce placeholder and no
  `'unsafe-inline'`; `middleware.test.ts` asserts the header is fresh per request.

## Verification (this is the "needs a live browser" issue)

Standing the frontend up and driving it:

- **Response header**: `script-src 'self' 'unsafe-eval' 'nonce-<b64>' 'strict-dynamic'` — no `'unsafe-inline'`.
- **Rendered page**: every one of the ~50 `<script>` tags carries the request's nonce; **0 inline scripts without a nonce**, 0 external scripts without one.
- **Live browser** (sign-in): page hydrates, **zero CSP violations** in the console (only the expected 401s from the backend not running).
- **Static analysis**: no `dangerouslySetInnerHTML` with scripts, no inline `<script>`, no `next/script`, no third-party scripts anywhere in `src/` — so every surface shares this Next-only, auto-nonced profile, and `'strict-dynamic'` extends the nonce's trust to the chunks those scripts load.
- `make lint-frontend`, `make test-frontend-cov` (421 files, 6183 tests, coverage gate held) and `make docs-build --strict` all green.

## Limitation

Browser verification covered the sign-in surface directly; the authenticated
surfaces (chat, Builder, dashboard) and the hosted page use the **same** middleware
policy and the same Next auto-nonce mechanism, with no page-specific inline or
third-party scripts, so they share the verified profile. A full authenticated
walk-through would need a seeded backend.

## Docs

`SECURITY.md` and `docs/deployment.md` (+ de/es/pl) updated.

Closes #1624


OchnikBartek:
@codex review

PR #1674, opened 2026-09-14
chore: cut 0.0.429

DEENUU1:
Release 0.0.429: version, lock and changelog only.

Ships #1647 - a per-organization hash chain over the audit trail, so an entry
that was edited, reordered, inserted or deleted stops verifying.

Carries migration 0079_audit_hash_chain.


DEENUU1:
Closing: #1647 has conflicts with main, so this cut names a merge that did not happen. Reopening the cut after the merge lands.

PR #1675, opened 2026-09-14
chore: cut 0.0.429

DEENUU1:
Release 0.0.429: version, lock and changelog only.

Ships #1620 - MCP OAuth accepts a pre-registered client id and secret, which is
the only way to connect a server that publishes no registration endpoint, and
refuses the three shapes of that client which used to fail only after consent.


DEENUU1:
Closing: the PR it names still has conflicts with main.

PR #1686, opened 2026-09-15
fix(tests): exempt the deployment-wide audit chain test from the security marker

MichalZelasko:
### Summary

Exempt `test_a_tenant_less_write_chains_and_verifies` from the `security` marker check so CI's `test` job is green on `main` and on every branch built from it.

### Business Context

`test_security_marker.py` draws a keyword net over the suite and fails any test whose name matches `tenant`/`permission`/`budget`/`approval`/`secret`/`plaintext` and carries neither the `security` marker nor an `EXEMPT` entry. The tamper-evidence hash chain work (#1647) added `TestTheDeploymentChain::test_a_tenant_less_write_chains_and_verifies`, which matches on "tenant" but exercises the deployment-wide (`organization_id=None`) audit chain, not a refusal — the same accept-complement shape as other existing `EXEMPT` entries. This left the `test` CI job failing on `main`, and every open PR inherited the same failure since they all run against `main`'s test suite.

### Changes

- Add the test's nodeid to `EXEMPT` in `backend/tests/test_security_marker.py` with a one-line reason, following the file's existing convention.

### Verification

- `uv run pytest tests/test_security_marker.py -q` — 4 passed (previously 1 failed).
- `uv run ruff check` / `ruff format --check` on the changed file.


MichalZelasko:
@codex review

MichalZelasko:
Fixed #1690 - closing

PR #1690, opened 2026-09-15
test(audit): mark the hash-chain tests as security

DEENUU1:
The `security` marker (#1630, 0.0.435) and the audit hash chain (#1647, 0.0.433) landed within an hour of each other, each green against a `main` that did not yet have the other. `main` went red where they meet:

```
tests/integration/test_audit_hash_chain.py::TestTheDeploymentChain::test_a_tenant_less_write_chains_and_verifies  (matched 'tenant')
```

The match is on `tenant-less`, so the word is incidental — but the test is not. Tamper evidence over the audit trail is a control `docs/security.md` names in the matrix, and the two tests beside it, `test_editing_a_stored_row_is_detected` and `test_deleting_a_row_is_detected`, are the least ambiguous security tests in the file and trip no keyword at all.

So the marker goes on the module, not on the one test the guard happened to catch. Exempting it would have been the smaller diff and the wrong answer: a set that exists to be counted would have been missing its tamper-detection half.

## Verification

`uv run pytest tests/ -q` — 8657 passed, 16 skipped. `tests/test_security_marker.py` green, ruff clean.


MichalZelasko:
Duplicate of #1686 but comment is better here. When this PR is merged, I will close #1686.

PR #1712, opened 2026-09-15
ci: run CodeQL on pull requests, audit bun.lock, ship an SBOM

### LINKED ISSUE #1415: CI: CodeQL on pull requests, a frontend dependency audit, and an SBOM per release
Part of the security programme (#1424). Pre-existing gaps, found planning W6.

## What CI does today, and what it does not

| Control | State |
|---|---|
| Static analysis | CodeQL **default setup, weekly** (`gh api .../code-scanning/default-setup`: `schedule=weekly`, five languages). A finding lands after merge, on `main`, up to a week late |
| Backend dependency audit | `make audit` - `uv export` of the locked set through `scripts/audit_dependencies.py` (`Makefile:460`) |
| Frontend dependency audit | **None.** `bun audit` is on the `SECURITY.md` checklist and in no job |
| Image scan | Trivy on the built image, push to `main` only |
| Secrets | GitHub push protection and secret scanning enabled at the repository; `detect-private-key` in pre-commit |
| Supply-chain lint | `zizmor` in pre-commit; actions pinned by SHA |
| SBOM | **None** |

## What to add

- A `codeql.yml` workflow running on `pull_request` for `python` and `javascript-typescript`, so a finding blocks the PR that introduced it. Keep default setup for the weekly full run, or replace it - one, not both reporting twice.
- `make audit` extends to `frontend/bun.lock` (`bun audit --audit-level=high`, or `osv-scanner` over both lockfiles if bun's output is not stable enough to gate on). `tests/test_ci_parity.py` updated with it.
- A CycloneDX SBOM (`syft`) for both images, attached to each GitHub release by the release workflow, so a client can answer "what is in it" without building it.

## Done when

- [ ] A PR introducing a known-bad pattern goes red on CodeQL before merge
- [ ] `make check` includes the frontend audit and the parity test knows it
- [ ] The next release carries `sbom-api.cdx.json` and `sbom-frontend.cdx.json`
- [ ] `docs/branching.md` (what protects `main`) and `docs/security.md` updated

Related: #1424, #1412, #143 (CI parity). Cluster in #168: *Security programme*.

## SD-010 — documented component inventory (additional W6+W7 scope)

Kacper requests documenting the complete component inventory. Reuse this issue's release SBOM work rather than generating a separate competing inventory.

- [ ] Publish a human-readable documentation page covering application components, backend/frontend dependencies, runtime images, bundled assets and configured model/service components; state the inventory's release/version and deployment-specific additions.
- [ ] Distinguish self-developed, third-party open-source and proprietary components, with source/provider, version, purpose and links to license/terms evidence.
- [ ] Link machine-readable release SBOM artifacts to the documentation and explain their scope, including components not captured by image scanning.
- [ ] Define how the inventory is regenerated/updated for a release and extended for the client's deployment.
- [ ] Cross-link the SD-012 license review and include this documentation in the planned Polish/German/Spanish translations.

Documentation and inventory must identify gaps explicitly; an SBOM for two images alone does not inventory all external models/services in the client's future stack.

DEENUU1:
## What this changes

Three CI controls a security review asks about, and the readable inventory that goes with them.

- **CodeQL on the pull request.** `.github/workflows/codeql.yml` analyses `actions`, `javascript-typescript`, `python` and `rust` with the `security-extended` pack on every pull request, on pushes to `main`, and weekly. Before this, CodeQL ran only from GitHub's **default setup** on a weekly schedule, so a finding arrived on `main` up to a week after the merge that introduced it.
- **`make audit-frontend`.** `bun audit --audit-level=high` over `frontend/bun.lock`, in the `Security Scan` job and in `make check`. Nothing in CI read that lockfile before — `bun audit` was a line on the `SECURITY.md` checklist and in no job.
- **An SBOM per release, per image, per architecture.** A new `sbom` job in `images.yml` runs syft against the manifest `publish` just pushed, once per platform, and attaches four CycloneDX documents to the GitHub release on a `v*` tag (`sbom-api-amd64`, `sbom-api-arm64`, `sbom-frontend-amd64`, `sbom-frontend-arm64`). Per architecture because syft resolves a manifest list to one variant, and the two do not contain the same packages. `make sbom` writes something else and is named so: `sbom-source-*.cdx.json`, an inventory of what the tree declares.
- **`docs/reference/components.md`** (SD-010), in all four languages: what this project writes, what it depends on, what each image contains, how the inventory is regenerated, and the models, providers, MCP servers and stores a deployment adds that no image SBOM can see.

## Why this way

**The frontend audit was red before it was a gate.** Adding it without fixing what it found would have landed a permanently red required check, so this raises `next` (past two unauthenticated-RCE advisories) and `postcss` (past two source-map path-traversal advisories), and pins `nanoid` and `js-yaml` forward with `overrides` because their parents — `postcss` and `@eslint/eslintrc` — have not moved yet. That is the whole of the lockfile churn: a targeted `bun update next postcss` rather than a full refresh, which belongs to #1485.

**The SBOM reads the published image, not the source tree.** An inventory generated from a checkout describes a different thing under the same file name, and the base-image layers — the ones a Trivy finding is usually about — would be missing. `make sbom` exists for the local case and its own docs say what it omits.

**One CodeQL configuration, not two.** The issue asks for one reporter. GitHub refuses an advanced upload while default setup is configured, so this workflow **replaces** it rather than joining it, and the weekly full run moves into the workflow so nothing is lost.

**The workflow makes the analysis reach the pull request; it does not by itself refuse the merge.** `codeql-action/analyze` uploads its results and exits 0 whatever is in them, so requiring the job proves only that the analysis ran. Blocking a pull request on a new alert is code-scanning merge protection on `main`'s ruleset — a repository setting, like the six required status checks beside it. `docs/branching.md` names both settings this workflow needs and cannot make, and nothing in the diff claims the job is a gate.

## Checks

- [x] Tests cover the new behaviour, and would fail without the change — `backend/tests/test_codeql_workflow.py` (new) and `TestTheReleaseCarriesAnInventory` in `test_images_workflow.py`; `test_ci_parity.py` already fails on a gating step `make check` does not run, which is what holds `audit-frontend`
- [x] `make check` passes — ran `make lint`, `make docs-build`, `make test-frontend-cov`, `make build-frontend`, `make audit-frontend` and the four workflow test files individually
- [x] Platform-layer coverage is still 100% — no `app/` module changed
- [ ] If a capability was added or changed, its `README.md` says why — n/a
- [ ] If a migration was added — n/a

## Notes for the reviewer

**This needs two repository settings changed, and the workflow can make neither.**

The first, without which the new jobs cannot upload at all:

```bash
gh api -X DELETE repos/vstorm-co/agenticos/code-scanning/default-setup
```

The second is what turns the analysis into a gate: **code-scanning merge protection** on `main`'s ruleset, with `CodeQL` as the tool and a severity threshold. Without it an alert is visible on the pull request and in the Security tab, and the merge button stays green. Both are written down in `docs/branching.md` so they are not only in a PR description.

Until default setup is off, the `CodeQL` jobs will run and their **upload will be refused** — "CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled". So expect the new workflow to be red on this pull request itself. The command is in the workflow header and in `docs/branching.md` so it is not only in a PR description.

Two things deliberately left out:

- **Gating the image publish on Trivy's CVE count.** The `scan` job's comment points at this issue for that decision; the issue's own scope is the SBOM, and turning a base-image CVE into a failed publish is a policy change worth its own discussion.
- **`SECURITY.md`'s production checklist** still says "Run `pip-audit` / `bun audit` in CI". It is a checklist for a deployment operator rather than a claim about this repository, and editing it makes its three translations stale for one line.

Closes #1415


DEENUU1:
All four Codex findings were correct and are fixed in `ebe1a7f`.

**P1 · An SBOM per published architecture.** Right, and the frontend is the sharp example: `@img/sharp-libvips-linux-x64` and `-arm64` are different packages, so an arm64 deployment was being assessed against an inventory that named neither the packages it has nor omitted the ones it does not. The `sbom` job's matrix is now `image × arch` and the release carries four documents. `anchore/sbom-action` does not expose `--platform`, so syft is installed through `download-syft` and invoked directly against `registry:<ref> --platform linux/<arch>`. Held by `test_images_workflow.py::…::test_each_published_architecture_gets_its_own`.

**P2 · The release has to exist before the upload.** Right in the general case. In this repository the cut runs `gh release create`, which makes the tag and the release in one call, so ordinarily there is a release by the time the tag push triggers this — but a tag pushed by hand has none, and that is a missing inventory rather than a failed publish. The step now checks `gh release view` first and, finding nothing, writes a notice to the job summary and exits 0; the run artifact is still the copy. Held by `test_a_tag_with_no_release_behind_it_does_not_fail_the_publish`.

**P2 · `make sbom` was claiming to be the release document.** Right, and the claim was in the docs as well as in the Makefile comment. It writes `sbom-source-api.cdx.json` and `sbom-source-frontend.cdx.json` now, and both the target's comment and `docs/reference/components.md` say plainly that these inventory what the source tree *declares* — dev and docs groups included, base image and built artifacts excluded — and are not evidence of what a release ships.

**P1 · A green `analyze` is not a merge gate.** Right, and the wording was the defect: `codeql-action/analyze` uploads its SARIF and exits 0 whatever is in it. What refuses the merge is code-scanning merge protection on `main`'s ruleset, which is a repository setting this workflow cannot make — the same class as deleting default setup. `docs/branching.md` gains a `### CodeQL runs on the pull request` section naming both settings and stating explicitly that the job's status proves only that the analysis ran; the enforcement table gains a row for it; the controls-matrix row in `docs/security.md` and the changelog entry no longer say "blocks".

Still expected to be red on this pull request: the four `analyze` jobs, whose upload GitHub refuses while default setup is configured. That is the first of the two repository settings above, and it is in the PR description and in `docs/branching.md`.

Issue #1598, opened 2026-09-12
Add role-targeted in-app notifications, delivery channels and app-admin announcements


DEENUU1:
Build a notification center with a bell in the application and optional email delivery. Different actions notify different users/roles. The deployment/application administrator can also send system announcements. Reuse the existing NotificationService, audience resolution and email preferences.

Scope:
- Persist notifications with type, timestamp, safe summary, context/action link and read/unread state. Provide a bell with unread count, paginated inbox, mark-read/mark-all-read and a sensible empty state.
- Define an extensible event catalog and recipient rules: specific users, roles and relevant organization/agent/resource audiences. Cover agreed events such as approvals, budget/security events, run completion/failure, ingestion and configuration changes.
- Resolve permissions at delivery/read time so links and content cannot expose another tenant's resources. Role membership changes and removed users must be handled.
- Provide per-user/per-event channel preferences for in-app and email, with a delivery-channel abstraction for future channels. Respect existing email opt-outs; explicitly define any mandatory system notice policy instead of silently bypassing preferences.
- Give app admins an audited composer for system announcements with explicit recipient scope and channels. An organization admin must not gain deployment-wide broadcast rights.
- Queue delivery with bounded retries, deduplication/idempotency and delivery/failure visibility. Email outages must not fail the originating business action.
- Define notification retention and integrate with #1420. Use safe content, validated links and authorized deep links.

Acceptance:
- Two users in different roles receive the configured different notifications; users outside the intended tenant/audience receive none.
- Bell/inbox/read state works across reloads, and email follows preferences without duplicate delivery.
- App-admin announcements are delivered only to the selected audience; unauthorized attempts are rejected and privileged sends are audited.
- Add tests for routing, permission changes, channel preferences, retry/dedup and UI states; document the event catalog and extension points.

Related: #1424, #1420 and current approvals/budget notification mechanisms. This covers notifications to people; arbitrary outgoing webhooks to external systems under IA-014–017 are a separate integration unless explicitly implemented here.

Requested by Kacper for W6+W7.

PR #1713, opened 2026-09-15
feat(notifications): GET/PATCH the per-user channel preferences

MichalZelasko:
### Summary

Adds the backend half of Decision 4's preference surface — `GET`/`PATCH
/notifications/preferences`. The frontend page extending
`settings/notifications/page.tsx` with these toggles is the follow-up piece,
not yet built.

### Business Context

`docs/design/notification-center-plan.md`, Decision 4: both channels are
preferences, except a named mandatory set (`security_event`,
`configuration_changed`, which bypass preference entirely) and the four
legacy-column pairs (the email channel of `budget_exceeded`,
`approval_requested`, `usage_report`, `agent_usage_report`, still governed by
the three existing boolean columns on `User` and `PATCH /users/me`).

### Changes

- `NotificationCenterService.list_preferences`/`.update_preference`, backed
  by two new repository functions (`list_channel_preferences`,
  `upsert_channel_preference` — the latter an `INSERT ... ON CONFLICT
  (user_id, event_type, channel) DO UPDATE`, so two first-time writes for
  the same pair leave one row rather than racing an insert).
- The togglable set (`_TOGGLABLE_PAIRS`) is computed once at import time:
  every non-mandatory event type's in-app channel, plus every non-mandatory,
  non-legacy event type's email channel. `PATCH` on anything outside that
  set — a mandatory event type or a legacy pair — is refused with a 400
  rather than silently writing a row nothing ever reads.
- The two new routes are registered ahead of the existing `PATCH
  /notifications/{notification_id}`: Starlette matches path routes in
  registration order, and the dynamic one would otherwise swallow
  `preferences` as its own path parameter (caught by the route tests, which
  is how the ordering bug was found before this shipped).
- `NotificationPreferenceUpdate` is declared in `test_update_nulls.py`'s
  `UPDATE_TARGETS` registry as `None` — its three fields are all required,
  an upsert of one exact triple rather than a `writable()`-style partial
  patch, so there's no null-to-`NOT NULL` question for that gate to check.

### Verification

- `uv run pytest tests/api/test_notification_routes.py
  tests/integration/test_notification_center.py tests/test_update_nulls.py -q`
- `make lint` and `make test` (the 100% platform-layer coverage gate, which
  covers `app/services/notification_center.py`,
  `app/api/routes/v1/notifications.py`, `app/repositories/notification.py`
  and `app/schemas/notification.py`): 8810 passed, 100% coverage.


MichalZelasko:
@codex review

DEENUU1:
Merged as part of #1760, which carries all fourteen steps of #1598 in one
squash - see 2369aaef on `main`.

Not a rejection of this step, and not a decision about how the work was split:
GitHub's own "stacked pull requests" feature refused to merge the chain one
pull request at a time, answering
`This pull request is part of a stack and must be merged using the
asynchronous merge REST API` - and that API answers 403 for this repository,
while retargeting anything below the top three was refused outright. The tip
branch already contained every commit of every branch in the stack, so the
whole feature landed from there.

This branch's commits are in `main` with their own history intact; only the
pull request closes unmerged. The review threads opened here were triaged
rather than bulk-resolved: each one has a reply saying what was fixed, or the
evidence for rejecting it.

PR #1715, opened 2026-09-15
feat(chat): name the delegate a persisted question came from

### LINKED ISSUE #1042: Name which delegate asked a persisted ask_user question
Split out of #502 (which persists an answered `ask_user` question as a turn
part — PR #1041). The persist-and-render half is done; **naming which
delegate asked** is not, and cannot be closed inside this repo without an
upstream change first.

## Why it is blocked here

A sync delegate's `ask_parent` reaches the parent through the shared
channel with no asker identity:

- `subagents_pydantic_ai/toolset.py` `_create_ask_parent_toolset` calls
  `ctx.deps.ask_user(question, [])` — the question and options only;
- `AskUserCallback` is fixed as `(question, options) -> answer`;
- `SubAgentState` (`_state.py`) carries `ask_callback` / `task_manager` /
  `task_id` / `questions` but **no name**.

So `AgentSession._ask_one`, where the `ask_user` part is built
(`app/services/agent_session.py`), has no way to know a question came from
a delegate rather than the main agent, or which one.

## What it needs

1. Upstream `subagents-pydantic-ai`: thread the subagent name to the ask
   channel — extend the callback, or add `SubAgentState.name` and read it
   via the context var in `ask_parent`.
2. Here: read that name in `_ask_one`, store it on the `ask_user`
   `MessagePart` (an `asked_by` field), and render "asked by <delegate>"
   in `AskUserBlock`.

## Done when

- A delegate's persisted question records which delegate asked.
- The transcript names the delegate on a replayed `ask_user` part.
- A test asserts a delegated question carries the delegate's name into the
  stored part.

Related: #502 (parent), #168 (issue map — same chat-persistence cluster as
#509's timeline work).

DEENUU1:
## What this changes

A stored `ask_user` part now records **which delegate asked**, and both transcripts say so: "Asked by researcher" in the chat and in the run timeline, where before every replayed question read "Asked you" whatever put it.

- `MessagePart.asked_by` on the stored timeline entry.
- `asking_delegate()` in `app/agents/ask_user.py`, read inside `AgentSession._ask_one`.
- `AskUserBlock` and the run timeline render the name when there is one.
- `subagents-pydantic-ai` floor raised to **0.2.22**.

## Why this way

**The upstream half had to come first, and it is a field rather than a signature change.** `ask_parent` reaches the surface through `ask_callback`, which is handed the question and nothing else — so this issue was blocked here, exactly as it says. [subagents-pydantic-ai#89](https://github.com/vstorm-co/subagents-pydantic-ai/pull/89) puts `name` on the delegation's bound state rather than widening `AskUserCallback`: the state is already bound for the whole delegation, so it is bound inside the callback too, and every other caller of that library is untouched. Released as [0.2.22](https://github.com/vstorm-co/subagents-pydantic-ai/releases/tag/v0.2.22).

**The name is read where the question is put, not where the answer arrives.** That is the one thing easy to get wrong here and the reason `asking_delegate` has a docstring about it: the answer comes back on the socket's receive loop, a different task with no delegation bound, so reading it there would answer `None` for every question and the feature would quietly do nothing. `test_a_delegated_question_records_which_delegate_asked` binds a state around the *ask* and not around the answer, which is what makes it a real test of that.

**`None` means the main agent asked**, and the transcript then says nothing extra — no "asked by the agent" placeholder. Every question stored before this field existed reads the same way, which is honest: nothing recorded who asked.

## Checks

- [x] Tests cover the new behaviour, and would fail without the change — `test_agent_session.py` (a delegated question carries the name; one the main agent asked carries none), `test_ask_user.py::TestWhichDelegateIsAsking`, `conversation-to-chat.test.ts`, `message-item.test.tsx`
- [x] `make check` passes — `make lint`, `make test` (8679 passed, gate at 100%), `make test-frontend-cov` (gate met), `make licenses`
- [ ] If a capability was added or changed — n/a
- [ ] If a migration was added — n/a. `parts` is JSONB; an older row simply has no `asked_by` key, and the schema's default is `None`

## Notes for the reviewer

**No migration and no backfill.** The stored timeline is JSONB, so a row written before this reads back with `asked_by` absent and defaults to `None` — which is also the honest answer, since nothing recorded who asked.

**Two tests import `bind_subagent_state` from `subagents_pydantic_ai._state`.** The library exports the *reader* (`current_subagent_state`) because that is what a host needs; binding is the library's own job around a real delegation. A test that wants to be inside one either reaches for the binder or runs a delegation, and the comment at each import says so.

**`pl.json` gains no key.** The `chat` namespace there is partial and already falls back to English for `askUserAsked` and `askUserAnswered`; adding only the new key would have been the odd one out.

Closes #1042


DEENUU1:
Both Codex findings were correct and are fixed in `887014c`.

**P2 · The attribution was not serialized with the round.** Right, and the race predates this branch for `_pending_question` — `_ask_lock` held the wire round but not the mark saying which question was open, so a second delegate reaching `_ask_one` while the first was parked overwrote it. What made it worth fixing now rather than noting is that this branch gives the mark a *name*: the first delegate's answer was persisted as the second's question, asked by the second.

`_send_and_wait` is `_ask_user`'s body with the lock lifted out, so `_ask_one` takes the lock itself and sets both fields inside it; `_ask_user` keeps its signature and still wraps the same body for the batch path. `test_two_delegates_asking_at_once_each_keep_their_own_name` fails against the previous commit with `('Which cluster?', 'eu', 'deployer')` where the first row should be — the defect exactly.

**P2 · The notices were generated against the wrong tree.** Right, and the cause is worth recording: `license_inventory.py` reads `frontend/node_modules` for the frontend half rather than `frontend/bun.lock`, so `make licenses` run after a branch switch writes whatever the previous branch's `bun install` left on disk. Regenerated here from this branch's own frozen install; `make licenses-check` is clean.

Issue #132, opened 2026-08-03
E2E: two seed/setup steps assert a just-created row is on screen and flake when it is not


DEENUU1:
Two intermittent failures in the same shape, both seen on `feat/sandbox` and both passing on a re-run of the same code:

1. `e2e/seed.setup.ts:48` — "a skill exists". Fills the create dialog, presses **Create**, then `expect(skillCard(page, …)).toBeVisible()`. Failed with `element(s) not found` after 15s in run [30842209736](https://github.com/vstorm-co/agenticos/actions/runs/30842209736); the same commit's code passed in the next run.
2. `e2e/vault.spec.ts:125` — "rotating a secret…" — filed separately as #130. Same shape: create, then wait for the row.

**Why one issue for both.** A failure here fails the *whole* e2e job, and neither spec can say whether the create was refused or the list simply had not refetched — they assert on the row that should have appeared and never on the refusal that might have. When it flakes we learn nothing except that it flaked, which is why #130 has no diagnosis after two runs.

**What would make them diagnosable.** After the submit, and before waiting for the row:
- assert the dialog closed (a dialog still open means the submit was refused, and the refusal is on screen);
- assert nothing is on the error toast.

Then a flake names the cause instead of timing out on a missing element.

**And what might actually be wrong.** Both screens list from a TanStack query and both mutations invalidate it. If the assertion runs between the mutation resolving and the refetch landing, the row is legitimately absent for a moment — in which case the fix is in the spec (wait for the network idle or for the mutation's own confirmation), not in the app. If instead the create is being refused under load, the assertions above are what will show it.

**How you would know it was fixed.** Twenty consecutive e2e runs green, and a deliberately broken create (e.g. a duplicate name) fails these specs with the refusal text rather than with "element(s) not found".

Pre-existing; found watching CI on #36, not caused by it.

DEENUU1:
Hit again on `main` at `ca7410f` (the 0.0.5 release commit), same step: `seed.setup.ts:48 › a skill exists`, `element(s) not found` on `skillCard(page, SEEDED_SKILL_NAME)` after 16.3s. Passed on a re-run of the identical commit, and had passed on that commit's own PR run minutes earlier — so three data points on one sha.

Worth recording because of *which* commit it was: 0.0.5 changed three version literals and a markdown file. Nothing it touched can reach the skills page, which makes this about as clean an isolation of the flake from any real cause as you will get.

[...]

PR #1717, opened 2026-09-15
fix(e2e): say what the cost journey was waiting for when it flakes

### LINKED ISSUE #162: e2e: the cost journey spec flakes on main after passing on the PR

DEENUU1:
## What this changes

`journey.spec.ts`'s last step — the one that failed on `main` at `5780012` and passed on a re-run of the same commit — now proves the run was **priced through the API** before asking Activity to draw its row. Every other wait in that spec carries a message. `nowThere` leaves `seed.setup.ts` as `nowListed` / `nowMatching` in `e2e/helpers.ts`, so specs and fixtures share one "the write has landed" step. And a failing `e2e` job uploads `test-results/` beside the HTML report.

## Why this way

**The failure carried no information, and that was the actual defect.** The step crosses five things: the model provider answered, the run row was written, the usage was metered into a spend record, the list refetched, the row rendered. `expect(row).toBeVisible()` failing says one of those did not happen — and `element(s) not found` names the only thing that cannot be the cause. Three passes and one failure across the same tree bought no diagnosis because there was nothing to read.

Polling the collection first splits that in two. A failure in `nowMatching` is a run that was never priced, and it prints the rows that *were* there. A failure in the `toBeVisible()` after it is a page that did not draw a row the API was serving — which is #230's shape, and a different defect.

**The shared primitive is the issue's own suggestion, and it is right.** Four specs flake this way; individually each reads as bad luck, together as one shape. `seed.setup.ts` already had the answer and kept it private. Moving it rather than copying it means the next spec that needs it does not write a fifth version.

**The poll is scoped to this journey's own agent** (`/api/runs?agent_id=…`) rather than reading the first page of everything: CI's organization is shared with every other spec, and a run pushed past `limit` by a busy list would read as a run that never happened — which is the class of bug this whole change is about.

## Checks

- [x] Tests cover the new behaviour — this *is* a test change; the backend workflow parity tests cover the CI half
- [x] `make check`'s relevant parts pass — `make lint` (eslint, prettier, tsc over `e2e/`, the guards, codespell, the translation gate), `make docs-build`, `tests/test_ci_parity.py` and `tests/test_ci_workflow.py`
- [x] Platform-layer coverage is still 100% — no `app/` module changed
- [ ] n/a — no capability, no migration

## Notes for the reviewer

**This does not claim the flake is gone.** The issue's bar is twenty consecutive green runs on `main`, and one pull request cannot show that. What it claims is narrower and checkable: the next failure says which of the five steps did not happen, and the artifacts to diagnose it are attached to the run. If the poll turns out to be the thing that fails, that is a product defect — a run whose cost is never recorded — and a much better issue than this one.

**The e2e suite itself runs in CI on this pull request**, not on my machine: it needs a migrated database, a seeded organization and a running backend, and the journey it changes is the one job that exercises all three.

**The other three flaking specs are untouched** (#130, #132, #154). They have their own issues; this one adds the primitive they would use.

Closes #162


DEENUU1:
Correct, and it undercut the change's own point — fixed in the commit above. `nowMatching` polled a boolean, so Playwright reported `Received: false` and threw away the rows `rowsAt` had just fetched; "no run at all" and "a run that is there and not priced yet" then read identically, which is exactly the distinction this step exists to make. It polls the rows now — empty means matched, anything else is what was there instead — so the failure prints them. That is the shape `nowThere` already used and the reason its docstring gives for it.

Issue #1108, opened 2026-08-25
Harden the request-path file-IO offload: cancellation-safety and a bounded executor


OchnikBartek:
## What's wrong

The request-path file work now offloads to `asyncio.to_thread` (#25 / #1099), which
frees the event loop but leaves two robustness gaps on the shared default executor,
raised in review.

**1. A cancelled save can orphan its file.** `LocalFileStorage.save` does
`await asyncio.to_thread(file_path.write_bytes, data)`. If the request or agent task
is cancelled while that await is pending, `asyncio.to_thread` cancels the *await* but
cannot stop the running `write_bytes` in the worker thread. The thread finishes
creating the file after `save()` has raised, so the caller never receives
`storage_path` and cannot delete or record it - cancelled uploads, avatar changes and
generated-image saves accumulate unreachable files. (Before #25 the write ran
synchronously on the loop, so it was equally uninterruptible but always *returned* the
path; the new cancellation point is what opens the window.)

**2. CPU-heavy parses share the default executor.** `parse_content` submits pymupdf /
openpyxl / docx parses to asyncio's shared default `ThreadPoolExecutor`, with no
admission control. A burst of concurrent uploads can occupy all its workers and leave
an unbounded queue holding upload buffers, while unrelated `to_thread` users on the
same pool - bcrypt password hashing, pinned-host DNS resolution - wait behind them.

## Where

- `backend/app/services/file_storage.py` (`save`/`load`)
- `backend/app/services/file_upload.py` (`parse_content`)

## Fix

1. Make the write cancellation-safe: `await` the submitted write to a terminal state
   before letting cancellation propagate (or shield it) so its path can be recorded or
   cleaned up.
2. Route parsing (and ideally the storage byte ops) through a dedicated bounded
   executor / capacity limiter, so uploads cannot starve unrelated request work.

## Acceptance criteria

- [ ] A cancelled `save` does not leave an unreachable file (path recorded or cleaned up)
- [ ] Parser work runs on a bounded executor that cannot exhaust the pool bcrypt/DNS use

## Notes

Found reviewing #1099 (codex P2 ×2). Both are refinements to the #25 offload, not
regressions - #25 strictly improves on the pre-existing loop-blocking behaviour - so
they were split out rather than folded into that branch. Same file-processing area.

Issue #254, opened 2026-08-05
A nested delegate's create_agent specialist does not survive an approval park


DEENUU1:
## What breaks

#175 (PR #251) makes a `create_agent` specialist kept by the **run's own agent**
survive an approval park: the registrations are snapshotted into
`PausedRunState.dynamic_specialists` and re-registered on resume. A specialist a
**nested delegate** keeps is still lost across the same park.

Each level of the delegation tree that binds `allow_dynamic` gets its own
`DynamicAgentRegistry` (one per `build_agent`). `PausedRunState.dynamic_specialists`
is a *flat* list of the top-level agent's registrations only, and the stash reads
and writes are guarded to `depth == 0` on purpose — a nested level's specialists
would have to hang off that delegate's `DelegationFrame` the way its conversation
already does, which a flat run-level list cannot represent.

## The sequence

1. An agent A delegates to a published/inline delegate B, and B's spec has
   `allow_dynamic: true` (`max_depth >= 2`).
2. Inside the delegation, B calls `create_agent('helper')` and `task(..., 'helper')`.
3. Anything parks the run on an approval.
4. Approve, resume. B's `task(..., 'helper')` answers "Unknown subagent 'helper'",
   because B is rebuilt with a fresh, empty registry — the exact failure #175 fixed
   one level up.

## Where

- `backend/app/agents/capabilities/subagents/_capability.py` — `build_delegation`
  seeds the registry from `runtime.stash.to_register` only when `depth == 0`, and
  `Delegation.wrap_run` snapshots only at `depth == 0`.
- `backend/app/agents/capabilities/subagents/_journal.py` —
  `DelegationJournal.record_created_specialists` returns early when `self.depth != 0`.
- `backend/app/services/agent_runner.py` — `PausedRunState.dynamic_specialists` is a
  flat list; `_parked_tree` fills it from `prepared.stash.registered` (top level only).

## Fix (sketch)

Carry a nested level's kept specialists on its `DelegationFrame`, the same shape the
frame already uses for the delegate's conversation and spend, and re-register per
level on resume rather than only at depth 0. This is a per-level tree walk, not a
flat list — hence its own issue rather than a widening of #251.

## Done when

A test parks a run inside a delegate that kept a specialist, resumes, and asserts the
delegate's `task` still reaches it — the depth-1 analogue of
`TestAKeptSpecialistSurvivesTheParkThatCreatedIt::test_task_reaches_a_resumed_specialist_and_still_meters_it`.

[...]

PR #1721, opened 2026-09-15
feat(capability): keep a compacted history's pictures out of the database

### LINKED ISSUE #55: Add media

DEENUU1:
## What this changes

A `media` capability that writes a compacted conversation's large parts out to the organization's media store and leaves a `media+sha256://…` reference in the stored history. No tools: it rewrites what is *stored*, not what the model can do.

- `app/agents/capabilities/media/` — the capability, an `OrganizationMediaStore` over `BaseFileStorage`, a README.
- `BaseFileStorage.save_at` and `.exists` — the one caller whose key is the digest of its own bytes.
- Offload at the write point in `agent_chat`; restore, unconditionally, in `ConversationService.model_history`.

## Why this way

**Most of #55 already works, and [the analysis on the issue](https://github.com/vstorm-co/agenticos/issues/55#issuecomment-3194727891) is why I did not rebuild it.** An attachment is inlined or written to the workspace and referenced; the ordinary history is rebuilt from the transcript's *text*, so a picture reaches the model once. Part A of that analysis — parsing and storage off the event loop — has landed since, through #1108. What was left is B1.

**A compacted conversation is the one place media genuinely piles up.** The library's own dump of `all_messages()` is stored whole and replayed exactly as the model last saw it, base64 and all, until the next summary replaces it. That is rows in Postgres and bytes on the wire, every turn in between, and it is what this empties.

**Offloading is optional, restoring is not.** Binding the capability is the decision to offload. Re-inlining happens for every conversation whether or not it is still bound, because markers outlive the binding — an agent can be unbound, or the next turn can run a different agent, and a marker nobody re-inlines is a picture the model is handed in a language it does not read. The walk over a history carrying none is a tree traversal that changes nothing.

**Keyed per organization, and that is the whole of the isolation.** A media URI is a content hash, so two tenants holding the same picture compute the same URI. The organization is part of the path and comes from the run through `resources` — never from configuration, which a builder could choose, and never from the URI, which a caller could present. `test_one_organizations_digest_does_not_resolve_in_another` is that boundary.

**The store issues no public URL, deliberately.** The harness's forthcoming externalizer uses one to hand a *model* a URL it fetches itself; these bytes are a tenant's, behind this deployment's authentication, and a URL a model provider can fetch is a URL anybody can.

**Both directions fail soft.** Losing a summary a model was paid to produce, because a store hiccuped, is worse than a history larger than it needed to be.

**`save_at` is concrete, not abstract.** A backend that cannot honour a caller-chosen key then says so at the one call site rather than failing to import — and adding a backend does not mean implementing a method it may have no use for. It also keeps this branch from breaking [#1714](https://github.com/vstorm-co/agenticos/pull/1714)'s `S3FileStorage` on merge; see the note below.

## Checks

- [x] Tests cover the new behaviour — `tests/test_media_offload.py` (16): what leaves the history and comes back identical, dedup, the tenant boundary, restore without the binding, both fail-soft paths, and what an unconfigured capability contributes. Plus the wiring test in `test_agent_chat.py`
- [x] `make check` passes — `make lint`, `make test` (8694 passed, gate at 100%), `make docs-build`
- [x] Platform-layer coverage is still 100% — `app/agents/**` is in the gate and the new package is at 100%
- [x] The capability's `README.md` says why
- [ ] n/a — no migration

## Notes for the reviewer

**This does not reduce what the model is sent**, and the README says so. The parts are re-inlined before the request goes out, because that is what keeps the run correct. Rewriting them to a URL the model fetches itself is the harness's own #254 and would need a URL this store deliberately does not issue.

**B2 from the analysis is not here.** A `view_image` / `describe_media` tool would satisfy the Done criteria's "declared tools" more literally, and it is a different feature: a metered vision call, with its own cost story and its own approval question. Kacper's lean was A + B1 and B2 "if we want a model-facing tool"; I left that decision where he left it rather than making it inside this PR.

**It touches `BaseFileStorage`, which [#1714](https://github.com/vstorm-co/agenticos/pull/1714) also extends.** Whichever merges second needs `S3FileStorage.save_at` — one method, the S3 branch already has the `create_cancel_safe` pattern it wants. I made `save_at` concrete rather than abstract so the intermediate state fails loudly at the call site instead of at import, which is the failure mode a batch merge would otherwise hide.

Closes #55


DEENUU1:
All five were correct and are fixed in `37e8e0b`. The first two were design holes rather than slips, and I am glad they were caught before this merged.

**P1 · The bytes had no lifetime.** Exactly right, and it is the worst thing a feature about not accumulating bytes could do. A content hash records nothing about who still references it, so replacing a summary, deleting a thread or deleting a tenant left the objects for ever.

Rather than reference tracking, the answer is structural: the path is `media/<organization>/<conversation>/<digest>` now. The organization is the isolation (unchanged); the conversation is the *lifetime* — `ConversationService.delete_conversation` removes the thread's prefix and `OrganizationService.purge` removes the tenant's, both after the commit for #1293's reason. `BaseFileStorage.delete_prefix` is what that needed. A run with **no** conversation now offloads nothing, because there would be nothing to delete it with.

The cost is storing the same picture twice if it appears in two threads, and one honest residue: a summary replaced by a later one leaves its objects until the thread goes. That is bounded rather than immediate, and the README says so instead of implying otherwise.

**P1 · Only the WebSocket chat offloaded.** Right — `AgentRunnerService._run` dumps the same history and calls the same `keep_summary`, so the API, channel mentions, embeds and triggers kept storing base64 while the capability reported itself bound. Both paths go through one `offloaded_history` helper, and `TestOffloadingReachesEveryRunSurface` is about the helper rather than about one caller.

**P1 · `save_at` undid itself on cancellation.** Right, and the reasoning generalises: for a content address the cleanup is *wrong*, not merely racy. Two callers writing one digest write the same bytes to the same path, so unlinking on cancellation removes a file another writer is already depending on — and what it would have saved is a file identical to the one that belongs there. It is a plain pooled write now and the docstring says exactly how it differs from `save`.

**P1 · Store construction outside the guard.** Right. `LocalFileStorage` does `mkdir`, so an unavailable volume failed a turn that had already paid a model for the summary this was going to shrink. Inside the `try` now.

**P2 · `exists` on the event loop.** Right, and it is the one call in that module that was not on the bounded pool.

One test in `test_org_purge_reservation.py` changed shape: it asserted that *nothing* was spawned, which the unconditional media teardown now breaks without saying anything about the collection reservations that file is for. It asserts on the `org_purge_cleanup` dispatch specifically, with a comment saying why.

PR #1725, opened 2026-09-15
feat(surfaces): close the accidental gaps between chat, socket and API

### LINKED ISSUE #936: Surface parity: what /chat offers that the raw WebSocket and the Public API do not
`/chat`, the raw WebSocket and the Public API all go through the same runner, and
they do not offer the same run. Some of the differences are deliberate and
written down; several are accidental and nothing says which is which. Establish
the table, close the accidental ones, record the rest.

**Not "make them identical".** A public widget must not offer a model override,
and a stranger holding a link must not be handed a URL into somebody's console -
both are decided, with reasons, in the code. The goal is that every difference is
a decision somebody can point at, and that a surface does not silently do less
than the one beside it.

## Where the three diverge today

| | `/chat` (WEB) | Raw WebSocket (EMBED) | Public API |
|---|---|---|---|
| Entry | `ChatAgentRunner.run` | `EmbedSession` → `AgentRunnerService.execute` | `POST /agents/{id}/run` → same `execute` |
| Streaming | yes | yes, filtered by `visible_frames(embed)` | **no** |
| Attachments | yes | yes (`file_ids`) | **no** |
| Conversation continuity | yes | yes (`continuity_key`) | yes (`conversation_id`) |
| `environment_id` | yes | **no** | yes |
| Model override | yes | no, deliberately | no |
| `ask_user` | yes | **no** | no (correct: not live) |
| Compaction notice | yes | **no** | n/a |
| Delegation frames | yes | no, deliberately | no |
| Approvals: parks the run | yes | yes | yes |
| Approvals: a path to the decision | yes (`/runs`) | one sentence, deliberately | **partial - see below** |

## The accidental ones

- [ ] **The API cannot attach a file.** `AgentRunRequest` (`app/schemas/agent.py:373`)
  is `prompt`, `conversation_id`, `environment_id`. `execute()` already takes
  `attachments`, `AttachmentRouter` already decides where a file goes by whether
  the agent has a workspace, and the widget socket already sends `file_ids`. One
  field, and the surface whose whole purpose is "run it from your own backend" is
  the one that cannot send a document.
- [ ] **The API's run result never fills `parked`.** `AgentRunResult` has the field
  (`:468`), `AgentRunnerService.parked_calls` (`agent_runner.py:2902`) computes it,
  and `POST /runs/{id}/resume` returns it - but `run_agent`
  (`api/routes/v1/agents.py:450`) constructs the result with seven fields and none
  of `steps`, `settled` or `parked`. So an API caller whose run parks gets an empty
  string, `status: awaiting_approval`, and nothing saying what is waiting; they
  have to already know to call `GET /runs/{id}/approvals`.
- [ ] **The raw socket has no `environment_id`.** The API has it, for the reason its
  docstring gives - exercising a dev environment before promoting it. An
  integration building its own client has the same need and no way to say it.
- [ ] **The raw socket cannot ask a question.** `ask_user`'s module docstring says
  the pause and resume "live in the WebSocket session" - the *chat's* session. The
  embed socket is equally live, with a person equally sitting there, and leaves
  `AgentDeps.ask_user` at `None`, so a delegate that would ask is told nobody could
  be reached. That is the right answer for a scheduled run and the wrong one here.
- [ ] **No compaction notice on the socket.** `CompactionSink`'s docstring names the
  failure it was written for: "the chat simply stopped for the length of it with
  nothing said". A widget conversation long enough to compact does exactly that
  today, on a surface that streams and could say so.

## The deliberate ones, to be written down rather than changed

- No model override on a public surface.
- No console URL to a stranger (`_no_answer` in `embed_session.py:263`).
- `subagent_events` left `None` by default: attaching a sink makes the library open
  a *streamed* request per child, so a delegate whose provider cannot stream breaks
  the moment somebody watches it. A first-party raw-socket client may want them -
  that is a decision to record, not a default to flip.
- The API POST is the non-streaming path. Whether it should grow an SSE variant is
  a real question and belongs on its own issue, not folded in here.

## One open question worth deciding, not assuming

**Approvals on the raw socket.** Today a parked run answers a visitor with "That
needs somebody to approve it before it can run", which is right for a stranger on
a widget. But the raw socket is sold as "an interface you build yourself: a mobile
app, a kiosk" - a first-party client with a signed-in user. Whether such a client
may be handed the parked call and a way to decide it is a product decision this
audit should force, not leave to whichever surface was written first.

## Done when

- `docs/channels.md` carries the parity table - every capability, per surface,
  with a reason on each "no". It has a surface table already (`:8`) that answers
  "what it needs" and "who the visitor is", and this is the third column it is
  missing.
- The five accidental gaps are closed or explicitly declined in the table.
- A test asserts the table rather than trusting it - the shape
  `test_capability_registry.py::TestFrontendToolCatalog` uses to keep two lists in
  agreement, applied to what each surface passes into the runner.
- While in there: `docs/channels.md:212` says the `history` frame is "on a hosted
  page only", which `continuity_key` and `_ALWAYS` contradict - any socket with a
  visitor key gets it.

## Related

- **#516** and **#517** (both closed) published the socket and the hosted page as
  surfaces in their own right; this is the follow-up question those raised - what
  a published surface owes a caller.
- **#39** (other channels setup) is the wider cluster and this is not it: that is
  about wiring surfaces up, this is about what they do once wired.
- **#925** (the per-conversation approval mode) touches the same seam from the
  chat's side - if it lands first, the table gains a row.

DEENUU1:
## The problem

Three surfaces run the same agent through the same runner — the dashboard's
chat, the raw WebSocket and the public API — and they did not offer the same
run. Some of those differences were deliberate; some were accidents nobody had
noticed. Nothing in the repository said which, so every one of them read as an
accident to whoever found it next.

## What changed

**Two accidents, closed.**

- The public API could not send the agent a document, although its whole
  purpose is running the agent from your own backend. `AgentRunRequest` now
  takes `file_ids`, resolved against the caller's own files exactly as `/chat`
  and the widget resolve them.
- `AgentRunResult.parked` existed and the route never filled it. A caller whose
  run stopped for an approval got a status and nothing to act on; it now carries
  the waiting call and the approval to post the decision to.

**One gap nobody had named.** A compaction takes tens of seconds, and only
`/chat` passed a sink into the runner — so every other streaming surface simply
went quiet for its whole length with nothing said. `AgentRunnerService.prepare`
and `.execute` take `on_compaction`, the embed session forwards
`compaction_started` / `compaction_finished` / `compaction_impossible`, and
those three are in `_ALWAYS`: that the agent is tidying its own notes is a fact
about the product, not about its reasoning, so a page showing neither reasoning
nor steps still gets them.

**The rest are decisions, now written down.** `docs/channels.md` gains a parity
table with a reason beside every "no", and three sections for the ones that are
declines rather than gaps:

- `environment_id` on the socket would let a stranger holding an embed key pick
  which *version* of the agent answers them, including one the organization has
  not promoted.
- `ask_user` on the socket is missing its other half, not its callback: there is
  no `ask_user` frame, no `ask_user_response` and no widget form, so attaching
  the sink would park runs on a question nothing can answer.
- Deciding an approval is a member's act gated on `approvals:decide`, and an
  embed key identifies a published surface rather than a member.

The `history` frame row was also plainly wrong — it said hosted pages only,
where any socket whose visitor carries a continuity key gets it.

## Verification

- `make lint` — clean
- `make test` — 8688 passed, 100% gate on the platform layer reached
- `make docs-build` — clean, all four languages
- `backend/tests/test_surface_parity.py` asserts the table against the code in
  the shape `test_capability_registry.py::TestFrontendToolCatalog` uses, so
  changing one without the other fails rather than leaving the prose describing
  what the code used to do.

## Limitations

- Streaming over the public API is still the separate question it was; this does
  not add an SSE variant.
- Delegation frames stay off the socket deliberately: attaching a sink makes the
  library open a streamed request per child, so a delegate whose provider cannot
  stream would break the moment somebody watched it.
- `ask_user` on the socket needs frontend work (the two frames and the widget's
  form) plus a decision about hosted pages, where the person being asked is a
  stranger. Declined here rather than half-built.

Closes #936


DEENUU1:
Codex reported nothing on `520328e3` and the code-quality reviewer found one thing, which is fixed in `3b48b7ad`:

- **`backend/tests/test_agent_runner.py:224` — statement has no effect.** Right: the sink stub's body was a bare `...`. It is a sink `prepare` must *attach* and must not *call*, so the body now says that and raises if it is ever awaited — the fix and an extra assertion in one line.

PR #1726, opened 2026-09-15
feat(auth): sign in through the deployment's own identity provider

### LINKED ISSUE #1419: Generic OIDC sign-in (Entra ID, Okta, Keycloak) beside Google
Part of the security programme (#1424). Pre-existing gap; README's comparison table marks SAML/SCIM as "ours to fix".

## What is there

Password sign-in and Google OAuth (`docs/configuration.md#oauth2-google`). A company deploying this on its own infrastructure (the whole pitch) runs an identity provider and will not create local passwords for its staff. Without SSO, MFA and offboarding are also the deployment's problem twice.

## What to build

- A generic OIDC provider configured by discovery: `OIDC_ISSUER`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET` (the secret through the vault mechanism the Google flow already uses, not a second one), an optional display name and icon from the auth glyph table.
- Authorization-code flow with PKCE, `email_verified` required, JIT user creation that **honours the sign-up policy** (`docs/deployment.md#who-may-register`: invite-only and allowed domains apply to an OIDC user exactly as to a password one).
- Group-to-role mapping is a follow-up; this issue signs people in.
- SAML and SCIM are filed separately in Later (SAML/SCIM (not filed - in the roadmap note)); most identity providers a mid-size company runs speak OIDC.

## Done when

- [ ] Sign in with Keycloak in the e2e suite (a container in the compose file for tests), with Entra and Okta documented
- [ ] A user refused by the sign-up policy is refused on the OIDC path too, with a test
- [ ] `docs/deployment.md`, `docs/configuration.md`, `docs/security.md` authentication row
- [ ] The sign-in page shows the provider under its own name, in en and pl

Related: #1424, #914 (sign-up policy), #955 (auth glyphs). Cluster in #168: *Security programme*.

DEENUU1:
## The problem

Password sign-in and Google. A company deploying this on its own infrastructure —
the whole pitch — runs an identity provider already and will not create local
passwords for its staff, so without SSO its MFA and its offboarding are that
company's problem twice.

## What changed

**A generic OIDC provider, configured by discovery.** `OIDC_ISSUER` plus a client
pair, and the authorization, token, userinfo and JWKS endpoints come from
`<issuer>/.well-known/openid-configuration` rather than from four more settings a
deployment can get subtly wrong. Entra ID, Okta, Keycloak, Auth0. Authorization
code with PKCE: a code intercepted between the provider and the callback is
useless without the browser that started the flow.

**One pair of routes, not a second copy of Google's.** `/{provider}/login` and
`/{provider}/callback`; the flow is one flow and the providers differ only in
their discovery document. The OIDC client is built on first use, which is what
makes an unconfigured deployment answer 404 there rather than redirecting a
browser at an empty issuer — and it is the *same* 404 an invented provider name
gets, because which identity provider a company runs is not a fact a stranger
enumerates by guessing paths.

**Two claims are required, of both providers now.** `email_verified` must be
true, with absent counting as false — an unverified address means anybody at that
provider can claim anybody's work address, and the sign-up policy's domain
allow-list is built on an address meaning something. And a stable `sub`, which
the account is keyed on rather than the address, so somebody who marries or whose
domain is bought keeps their history and the next holder of a freed address does
not inherit it.

**A refused sign-in now says why.** `check_may_register` already gated
`get_or_create_oauth_user`, so `invite_only` and the domain list did refuse an
OAuth sign-in — but the refusal was caught by the callback's catch-all and shown
as "Sign-in failed. Please try again." beside every timeout. The policy's own
sentence, the one the registration form shows, is carried to the sign-in page.

**On the page**, the provider appears under the name the deployment gives it
(`OIDC_DISPLAY_NAME`) — "Continue with OIDC" names a protocol at somebody looking
for their employer — with a mark from the table the auth pages already ship
(`OIDC_ICON`) or a plain key. English and Polish.

## Verification

- `make lint`, `make test` (8693 passed, 100% gate), `make test-frontend-cov`
  (100% / 97.72% branches), `make build-frontend`, `make docs-build`
- `backend/tests/test_oidc_sign_in.py` — 18 tests: each claim that refuses, the
  unconfigured and the invented provider answering alike, PKCE on the client, the
  round trip, and the sign-up policy refusing an OIDC sign-in with its own
  sentence
- **Against a real Keycloak.** The `sso` compose profile was started and the realm
  imported; discovery loads through the application's own client
  (`sign_in_client("oidc").load_server_metadata()` returns the realm's issuer and
  endpoints), and the authorization request built with these settings is accepted
  by the provider with `code_challenge_method=S256` — a login form, not a refusal
  about the client or the redirect URI.

## Limitations

- **`frontend/e2e/sso.spec.ts` was not executed.** It drives the browser leg
  against the Keycloak profile and skips without `E2E_OIDC=1`. e2e needs a seeded
  backend and is not a CI job, so this spec is written and reviewed but unrun;
  the container, the realm and the two accounts it signs in as are checked in and
  the discovery half is verified above.
- **Group-to-role mapping is not here.** People sign in; an administrator places
  them. Said so in `docs/security.md` rather than left to be discovered.
- **SAML and SCIM are not implemented.** Most identity providers a mid-size
  company runs speak OIDC.
- The `email_verified` requirement is a behaviour change for Google sign-in too.
  Google always sends the claim, so nothing that worked stops working; applying
  it to one provider and not the other would have been the odd thing.

Closes #1419


DEENUU1:
All four Codex findings accepted and fixed in `c196a551`.

- **P1 — normalize Entra's verification signal.** Correct, and it would have refused every Entra account on a deployment configured exactly as the page described. `email_verified` and `xms_edov` both count now, a string `"true"` counts too because Entra has shipped it that way, and `OIDC_VERIFIED_CLAIM` names a third for a provider that calls it something else. `docs/configuration.md` also says `xms_edov` is an *optional* claim the app registration has to enable — which is the next thing that would have gone wrong.
- **P1 — fetch claims from the discovered UserInfo endpoint.** Correct. `_oauth_claims.claims_for` uses the ID token where it is sufficient and asks UserInfo where it is not, merging UserInfo over the token. Asked second rather than always: it is a round trip per sign-in, and a *present* `email_verified: false` is an answer a fetch would not change, so that case does not pay for one either.
- **P2 — namespace OIDC subjects by issuer.** Correct, and the collision is worse than it first looks because `get_or_create_oauth_user` matches the provider pair *before* it compares the address. `identity_key` stores `<issuer>#<sub>` for `oidc`. Google keeps its bare subject deliberately: its issuer is a constant, and rewriting the key would orphan every account that has already signed in with it.
- **P2 — give the backend a routable Keycloak issuer.** Correct. The fixture is `keycloak:8081` now, published on the same port it listens on so one issuer string resolves from the browser and from the compose network, with the single `/etc/hosts` line in the compose file's comment and in the spec's header.

Tests added for each: the two verification claims and the string form, the namespaced key and Google's unchanged one, and six cases for the UserInfo fallback in `backend/tests/test_oauth_claims.py`.

Issue #1423, opened 2026-09-04
Uploaded files and workspaces at rest: an S3-compatible backend with server-side encryption beside the local disk


DEENUU1:
Part of the security programme (#1424). Pre-existing.

## What is there

`app/services/file_storage.py` - "Supports local filesystem storage" - and that is the one backend. Chat attachments, workspace files and RAG uploads sit on the API container's disk unencrypted; a sandbox workspace is a directory the runtime mounts. The honest answer for a single-host deployment is "encrypt the volume", and `docs/security.md` (#1412) will say so. It stops being enough the moment there are two API replicas, or a client asks for object storage with their own KMS key.

## What to build

- `S3FileStorage` behind the same `BaseFileStorage`, selected by `FILE_STORAGE_BACKEND=s3`, with `SSE-S3` or `SSE-KMS` (key id in settings), path-style for MinIO. The S3 sync connector already carries a client and the vault plumbing for credentials - reuse both.
- Local stays the default. A `doctor` line says which backend and whether encryption is on.
- Not in scope: migrating existing files between backends; document that it is a deployment-time choice.

## Done when

- [ ] Upload, read, delete and the workspace file browser work against MinIO in the integration suite
- [ ] `docs/file-processing.md` and `docs/configuration.md` document the backend; `docs/security.md` at-rest row points here

Related: #1424, #1412. Cluster in #168: *Security programme*.

Issue #1420, opened 2026-09-04
Per-organization data retention: conversations, runs, files and audit on a schedule


DEENUU1:
Part of the security programme (#1424). Pre-existing.

## What happens today

Nothing is ever deleted on a schedule. `grep -ri retention backend/app` finds no code. Conversations, messages, chat files, agent workspaces, run records and audit entries live until somebody deletes the organization. Two standards pull in opposite directions and both need a setting: HIPAA wants audit records kept six years; GDPR wants everything else minimised.

## What to build

- Retention in days, per organization, per class: `conversations` (messages and their files), `runs` (run rows and manifests; spend totals survive as aggregates), `workspaces`, `sandbox sessions`. `audit` has a **floor**, not a setting - an org may lengthen it, never shorten it below the deployment's default.
- A deployment-wide default and ceiling in `deployment_settings`, so an app admin can say "nothing lives longer than N days here".
- A Prefect flow sweeping in batches, one audit entry per sweep naming the class and the count, never the content.
- Purge is a hard delete of content, not a soft flag - a retention policy that keeps the rows is not one.

## Done when

- [ ] Settings page under the organization, gated on `org:settings`, with a tour stop
- [ ] The sweep is a `@flow` on the schedule the other periodic flows use, tested with frozen time
- [ ] Spend for a purged run still counts in the month's figure (test)
- [ ] `docs/governance.md` gains a retention section; `docs/security.md` rows for integrity and audit controls point here

Related: #1424, #1421. Cluster in #168: *Security programme*.

## Expanded W6+W7 scope — FA-109 / NFA-022

Kacper requests extending this security milestone to cover configurable retention/deletion across chat history, memory, documents and logs, not only the existing conversation/run/workspace classes.

- [ ] Include native personal/agent memory and configured external memory stores, with expiry rules and removal/exclusion from retrieval. Coordinate per-entry controls with #1594 and person erasure with #1421.
- [ ] Include RAG source documents, uploaded files, extracted content, chunks, embeddings/indices and derived copies. Define the retention timestamp and synchronization behavior so an expired document is not immediately re-ingested unless policy permits it. Preserve shared resources still valid for other users.
- [ ] Distinguish technical operational logs, business-content logs/traces, run manifests and audit metadata; apply configurable periods and document any justified retention floor separately from content retention.
[...]

PR #1729, opened 2026-09-15
feat(security): a HIPAA deployment profile and a doctor that proves it

### LINKED ISSUE #1448: A HIPAA deployment profile, and a doctor check that proves a deployment matches it
Part of the security programme (#1424). Downstream of #1412, which is its skeleton.

## Why

A client's security review does not ask "is this software compliant" — there is no such thing, HHS certifies no software and OCR does not recognise private certifications. It asks **"can we run this inside our compliant environment, and can you prove it."** Today the answer is a conversation. This issue makes it a configuration plus a command.

The commercial case is sharper than the engineering one: this is what a managed deployment hands a client's security officer in week one instead of scheduling three calls, and it is cheap next to what it opens.

## What a HIPAA profile is

An opinionated, documented deployment configuration shipped with the product, plus a check that proves a running deployment matches it.

**The configuration** — `deploy/profiles/hipaa/`, a compose file and a Helm values file:

- TLS to Postgres and Redis (#1418), encrypted volumes, no egress beyond an allowlist
- A local model profile as the default (Ollama or a LiteLLM proxy), so no PHI reaches a third party and no BAA with a model vendor is needed
- `LOGFIRE_TOKEN` unset, and `content: none` as the deployment floor for `AgentSpec.observability` (#1413)
- Retention preset per class, with the audit floor at **six years** — what HIPAA asks of audit records (#1420)
- OIDC required and local password sign-in disabled (#1419); MFA is the identity provider's job and the profile says so
- Audit export and the hash chain on (#1422)
- File storage on an S3-compatible backend with server-side encryption, or a documented encrypted volume for the local backend (#1423)

**The check** — `agenticos cmd doctor --profile hipaa`:

Reads the running deployment and prints a pass/fail sheet, one row per control, naming the setting that satisfies it or the one that does not. This is the part that matters: it turns "we followed good practices" into something a person can put on the table in front of an auditor, and it fails loudly when a later config change quietly breaks a control.

Exit non-zero on any fail, so it can run in the client's own CI.

## Scope, and the line the docs must not blur

The profile answers the **technical** safeguards, §164.312, and only those. Administrative (§164.308 — risk analysis, workforce training, sanction policy, contingency plan, BAAs) and physical (§164.310) safeguards belong to the operator and always will. `docs/security.md` must say this in the same breath as the profile, or the profile becomes a claim we cannot support.

Two things also belong in the docs rather than in code:

- **Who is the business associate.** A client running this themselves gets software from us and we never touch PHI, so no BAA is needed. A deployment Vstorm operates makes Vstorm a business associate, and that is a contract, not a config flag.
- **The model vendor's BAA has a narrower scope than people expect.** Anthropic's HIPAA-enabled organization blocks code execution and web fetch, which is exactly the shape of the `sandbox`, `code_execution` and `web_fetch` capabilities. A client on Claude under a BAA loses them. The profile's local-model default is what makes that a non-issue, and the docs should say why rather than leaving somebody to find out during an incident.

## Worth building now rather than later

The January 2025 NPRM would remove the required/addressable distinction and make encryption at rest and in transit, and MFA for every system reaching ePHI, mandatory. Final action is currently targeted at July 2027, so it is not law — but the controls it names are the ones already in this milestone, and a profile that presets them is the difference between being ready and retrofitting.

## Done when

- [ ] `deploy/profiles/hipaa/` holds a compose file and Helm values that come up clean on a fresh machine
- [ ] `agenticos cmd doctor --profile hipaa` prints a per-control sheet and exits non-zero on a fail
- [ ] An integration test starts a deployment that violates two controls and asserts the doctor names both
- [ ] `docs/security.md` gains the profile, the §164.312 matrix rows it satisfies, and the scope line above
- [ ] `docs/deployment.md` and `docs/commands.md` updated
- [ ] Nothing anywhere claims compliance or certification — the wording is "deployable inside a compliant environment, and here is the evidence"

Related: #1424 (epic), #1412 (the matrix), #1413, #1418, #1419, #1420, #1422, #1423. Cluster in #168: *Security programme*.

DEENUU1:
## The problem

A client's security review does not ask "is this software compliant" — HHS
certifies no software and OCR recognises no private certification. It asks **can
we run this inside our compliant environment, and can you prove it.** The answer
was a conversation. This makes it a configuration plus a command.

## What changed

```bash
uv run agenticos cmd doctor --profile hipaa
```

One row per control against the running deployment, naming the setting that
satisfies it or the one that does not, and a non-zero exit on any unmet control
so a client's own CI can gate on it. Ten controls:

| Control | Safeguard |
|---|---|
| `postgres-tls` | §164.312(e)(1) — `verify-full`, because `require` verifies no certificate |
| `redis-tls` | §164.312(e)(1) |
| `vault-key` | §164.312(a)(2)(iv) |
| `content-at-rest` | §164.312(a)(2)(iv) — **the operator's**, named not passed |
| `local-model` | §164.312(e)(1) — every model profile, not a default; one with no `base_url` is the vendor's public API by definition |
| `traces-local` | §164.312(e)(1) |
| `sso` | §164.312(d) |
| `signup` | §164.312(a)(1) |
| `audit-retention` | §164.312(b) — §164.316(b)(2)'s six years |
| `audit-chain` | §164.312(c)(1) |

**`deploy/profiles/hipaa/`** — a compose overlay that refuses to start without
the settings it cannot default (an unset `OIDC_ISSUER` stops the stack rather
than producing a deployment that looks configured and accepts passwords), an
annotated env file, and a README carrying the control table and the scope line.

**Three outcomes, and the middle one is the design.** A control that is genuinely
the operator's prints `--`, is named, and does not fail the command. A sheet that
skipped what it cannot see would read as complete and would not be; a command
that failed on volume encryption would be one nobody could ever pass.

**The line the docs do not blur.** `docs/security.md` states, in the same breath
as the profile, that it answers §164.312 and only that — the administrative
(§164.308) and physical (§164.310) safeguards are the operator's and always will
be. It also says who the business associate is (a client running this themselves
gets software; nobody here touches their PHI) and why a hosted model's agreement
is narrower than people expect: a HIPAA-enabled organization at a major vendor
typically excludes code execution and web fetch, which is the exact shape of the
`sandbox`, `code_execution` and `web_fetch` capabilities. That is why local
inference is the profile's default rather than a suggestion.

Nothing anywhere claims compliance or certification.

## Verification

- `make lint`, `make test` (8703 passed, 100% gate), `make docs-build`
- The compose overlay validates against the repository's own `docker-compose.yml`
  and refuses with the intended message when `OIDC_ISSUER` is unset
- The sheet was run against the development deployment and read correctly:
  5 unmet, 2 met, 3 attested, exit 1
- `tests/integration/test_hipaa_profile.py` — the one the issue asks for: a
  deployment breaking two controls, and both named. Plus one that matches the
  profile and fails nothing, and one that an attested control never fails the
  command.
- `tests/test_deployment_profile.py` — 25 cases, each control in both directions

## Limitations

- **No Helm values file.** This repository ships no chart, and values for a chart
  that does not exist would be a file nobody can apply. The README says so, and
  the settings transfer unchanged to whatever chart a deployment uses.
- **Two controls read settings that arrive with sibling PRs.** `sso` reads
  `OIDC_ISSUER` (#1419) and `audit-retention` reads the deployment's audit floor
  (#1420); both are read defensively, so this branch stands alone and the
  controls report honestly — `sso` is unmet on a deployment without OIDC, which
  is the truth — and become satisfiable as those land.
- **`traces-local` checks the token, not a per-agent floor.** A deployment-wide
  floor for `AgentSpec.observability.content` is #1413. With no token, no span
  leaves, which is the control that matters; with one, the check is unmet and
  says why.
- **The profile does not check egress allowlisting or S3 server-side
  encryption**, because neither has a deployment-wide setting to read yet
  (#1423). `content-at-rest` covers the at-rest question as an attested control.

Closes #1448


DEENUU1:
All nine accepted and fixed in `763c4ee0`. Four of them were controls reporting **met** on a deployment that did not satisfy them, which is the one failure mode a sheet put in front of an auditor must not have.

**P1 — parse the model hostname before declaring it local.** The worst of the nine, and entirely my fault: a substring test called `https://ollama.vendor.example/v1` local and passed the sheet while every prompt left the network. The hostname is parsed now and matched as a name or a suffix, with private, loopback and link-local addresses recognised separately because an IP literal has no suffix to match. Tests for the vendor host, a public address, a private one and a cluster name.

**P1 — require hostname verification for PostgreSQL.** Right, and the documentation already said `verify-full` while the code accepted `verify-ca` — so the sheet disagreed with its own page. Only `verify-full` passes; `verify-ca` is refused by name with the reason.

**P1 — reject weak vault master keys.** Right. HKDF derives a correctly sized wrapping key from anything and cannot put entropy into a one-character secret, so the ciphertext was recoverable under a row reporting the credentials protected. Every configured key, rotation set included, must be at least the 64 characters `openssl rand -hex 32` produces.

**P1 — include per-agent exporters in the tracing control.** Right, and I had read only the deployment's own token. An agent's `observability.token_secret_id` and an environment's `logfire_token_secret_id` each attach an exporter with `content` defaulting to `full`. Both are read now, from the **published** spec, and an agent tracing with `content: none` is correctly not counted — timing and tool names leaving is not what this control is about.

**P1 — add an inbound HTTPS control.** Right: store and model transport were checked and the browser's was not, so all ten rows could pass while a sign-in crossed the client boundary in plaintext. `browser-tls` refuses an http `FRONTEND_URL` or `PUBLIC_BASE_URL` and attests the certificate to the operator's proxy.

**P1 — provision TLS on the bundled database and Redis** and **P1 — force production secret validation.** Both right, and both mean the documented command produced a deployment that could not work or should not: `verify-full` against a `db` with no TLS listener cannot connect, and `ENVIRONMENT=local` with the repository's published `SECRET_KEY` signs every session token. The overlay sets production, and refuses to start without a `SECRET_KEY`, a `POSTGRES_HOST` and a `REDIS_HOST` of your own. The README says the bundled stores are for development.

**P1 — implement OIDC before requiring it.** Right — the overlay demanded variables the application does not declare, so the documented command could never come up and the control could never pass. They are passed through rather than required now, and the README, the env example and the sheet all say the `sso` control waits on #1419. Leaving it unmet is the truthful reading of a deployment where people still sign in with passwords.

**P2 — skip the profile sheet when the database probe fails.** Right, and it defeated the command in exactly the case it exists for. The sheet runs inside `_run` after the probe; a failed probe prints a line saying the sheet needs a database too, rather than a traceback.

PR #1734, opened 2026-09-16
feat(desktop): hand sign-in to the system browser, and take the workaround out

### LINKED ISSUE #1532: Desktop shell: hand OAuth sign-in to the system browser with a one-time exchange code
## What breaks

The desktop shell (#1531) signs in inside its WKWebView. Google's authorization endpoint refuses embedded user-agents (`disallowed_useragent`), so #1531 gives the console window Safari's version tokens on the same WebKit engine - a workaround that works today and that Google's OAuth policy says not to rely on: the flow it wants is the system browser, with the result handed back to the app.

## What the proper handoff needs

1. **Shell:** open the provider's authorization URL in the system browser (`tauri-plugin-opener`), and register a deep link (`agenticos://auth/…`, `tauri-plugin-deep-link`) for the return.
2. **Backend:** after the OAuth callback completes in the browser, mint a **one-time, short-lived exchange code** bound to the session, and redirect to the deep link with it. A new endpoint exchanges the code for the session tokens exactly once.
3. **Frontend BFF:** a route the shell's webview can call with that code to set the httpOnly cookies, so the console in the window is signed in without the browser's cookie jar.

Found reviewing #1531 (Codex, P1). Pre-existing in the sense that no desktop client existed before; the workaround ships with #1531 and this issue removes it.

## Done when

`CONSOLE_USER_AGENT` in `desktop/src-tauri/src/lib.rs` is gone, Google sign-in from the shell completes through the system browser, and the exchange code is refused on its second use (a test in `backend/tests/`).

Related: #1419 (generic OIDC) lands on the same handoff; #1414 (oauth-return clear-text storage) is the frontend half of the return leg.

DEENUU1:
## The problem

The desktop shell signed in inside its own WKWebView. Google's authorization
endpoint refuses an embedded user-agent (`disallowed_useragent`), so #1531 gave
the console window Safari's version tokens on the same WebKit engine. That works,
and Google's OAuth policy says not to rely on it.

## What changed

**`CONSOLE_USER_AGENT` is gone.** The window now refuses exactly one navigation:
a start at `/api/oauth/<provider>/login`, which it opens in the system browser
with `client=desktop` appended. Watching for the *start* rather than asking the
console to behave differently is what keeps the frontend out of this entirely —
nothing in it knows it is running in a shell, and nothing has to. Every other
navigation, including the provider's return to the console, loads in the window
as before.

**The backend records `client=desktop` in the session at that start**, and reads
it at the callback — never off the return, because the callback builds a redirect
out of it and a value a caller could hand back would be an open redirect into
whatever URL handler that machine has registered. The redirect goes to
`agenticos://auth/callback?code=<one-time code>`: the same single-use, one-minute
code the browser flow already mints, through the same `POST /oauth/exchange` that
already refuses a second use.

**The return leg is deliberately narrow.** A deep link is something any process
on the machine can fire, so all one may do is send the console window to one
path, on the server the user themselves configured, carrying a code that redeems
exactly once. It cannot name an address; a link with no code, the wrong host, the
wrong path or the wrong scheme does nothing.

The scheme is a deployment setting (`DESKTOP_DEEP_LINK_SCHEME`) matched by
`tauri.conf.json`, so a deployment shipping the shell under a name of its own
says so once rather than patching the callback.

## Verification

- `make lint`, `make test` (8678 passed, 100% gate), `make docs-build`
- **`make desktop-check`** — `cargo fmt --check`, clippy with `-D warnings`, and
  27 Rust tests. The new ones cover what the window refuses and what it does not,
  that a staged invitation's `flow` survives the handoff, that `client` is not
  doubled, and the six refusals the deep link makes.
- `backend/tests/test_oauth_signin_exchange.py` — a desktop sign-in returns
  through the deep link, an ordinary one still returns to the console, and a
  `scheme` a caller puts on the *return* changes nothing. The single-use
  refusal already had a test and still passes.

## Limitations

- **The frontend BFF half was already there.** `POST /api/auth/oauth-callback`
  takes a code, exchanges it and sets the httpOnly cookies, and
  `/auth/callback` already POSTs to it — so the shell navigates to the page that
  exists rather than to a route built for it. That is the smaller change and it
  keeps one exchange path rather than two.
- **Not verified end to end against Google.** That needs a signed build, a
  registered scheme and a real OAuth client; what is verified is every piece
  either side of the operating system's hand-off — the navigation the window
  refuses, the URL it opens, the redirect the callback builds, and what the
  shell does with a link it is given.
- **`tauri-plugin-deep-link` needs the bundled app to be installed** for the
  scheme to be registered on macOS; `cargo run` from a checkout will not receive
  one. `docs/desktop.md` says how the flow works rather than promising it in a
  development run.
- **Merge order:** #1419 generalises these routes to `/{provider}/login`. Both
  touch `oauth.py`; whichever lands second wants a look at the other's shape.

Closes #1532


DEENUU1:
All six accepted; five fixed in `4a213271` and the sixth fixed as a side effect of the first. Thank you — the headline finding is that **the feature did not work at all**, and that is exactly right.

### It never reached the backend

**`client` discarded by the proxy** (`lib.rs:400`). Confirmed: the shell intercepted the console's `/api/oauth/<provider>/login`, but that path builds a *fresh* backend URL carrying only `invitation_handle`, so the marker was dropped and every desktop sign-in finished in the browser at the ordinary frontend URL while the webview stayed signed out.

Rather than teach the proxy to forward the marker, the shell now intercepts one step later — the redirect the proxy produces, which is the deployment's own `/api/v1/oauth/<provider>/login`. That fixes **staged invitations across the handoff** (`lib.rs:935`) in the same move: the proxy hop runs *in the webview*, where the httpOnly cookie is, so the URL the browser is handed already carries the handle. An invitee on an `invite_only` deployment is no longer refused.

### The security pair

**Deep-link codes not bound to the initiating desktop** (`lib.rs:430`). The shell mints a nonce per attempt, passes it on the way out, the deployment returns it beside the code, and a link carrying any other one is ignored. It is taken out of its slot when a link arrives, so a replay of the same link finds nothing either. Compared in constant time — not because this is timing-attackable in practice, but because writing the other comparison invites the next reader to copy it somewhere it matters.

**Interception scoped only by path** (`lib.rs:934`). Right, and the consequence you name is the sharp one: this webview permits navigation to external sites during MCP consent, so a loaded page could drive the privileged opener in a loop. The predicate compares the origin with the configured server, and refuses when no server is configured at all.

### The other two

**The marker was session-global** (`oauth.py:95`). Fixed by generating the OAuth `state` in the route — so the marker can be filed under it — and reading the entry belonging to the state authlib has just validated. A second sign-in started in the same browser no longer erases or consumes the first one's destination; there is a test for each order. Pending markers are capped at four and evicted oldest-first, so an abandoned flow cannot grow the session cookie.

**No single-instance handling** (`lib.rs:1085`). Also right, and a desktop sign-in is precisely the case where an instance is always already running. `tauri-plugin-single-instance` is in, declared under the non-mobile target, and the second process focuses the existing console and hands over its argv.

### Verification

`make lint`, `make test` (8680 tests, platform gate at 100%), `make desktop-check` — rustfmt, clippy with warnings denied, 32 Rust tests.

New Rust tests: the origin check from both sides, that the console's own hop is deliberately *not* intercepted, the nonce appended on the way out, and the way back refused for a wrong nonce, a missing one, and one arriving when nothing is pending. New backend tests: the state binding, two concurrent sign-ins keeping their own destinations, and a callback with no state defaulting to the console rather than to the deep link.

`docs/desktop.md` and its three translations describe the flow again, including why the console's hop is followed rather than intercepted.

PR #1764, opened 2026-09-18
fix(chat): a token refresh no longer cancels the turn it was streaming

henfircreo:
## What this changes

A chat turn no longer dies when the console refreshes its own access token. The
refresh runs on a timer nobody asked for - roughly every twenty minutes per open
tab - and each new token used to close the socket an answer was streaming on;
the server read that close as the reader leaving and cancelled the run, so a
half-written reply was recorded `cancelled` and labelled *stopped* in the
transcript while the composer went on spinning, because the frame that would
have ended the turn on screen went to the socket that had just gone. Turns here
run for minutes, so a large share of them were caught.

Three changes, one per cause:

- **The socket is no longer identified by its credential.** `useWebSocket` takes
  `protocols` as a getter read at `connect()` rather than an array held as part
  of *which socket this is*, and `sigOf` is the address alone. A refreshed token
  now reaches the next handshake without disturbing the live one; an
  organization switch still swaps the socket, because that travels in the URL.
- **A socket that goes away no longer cancels the turn.** `AgentSession.shutdown`
  lets a detached turn reach its own end, bounded by `DETACHED_TURN_GRACE_S`
  (ten minutes), and cancels only if it outruns that. A run parked on `ask_user`
  is answered as nothing rather than left holding the grace open for somebody
  who is not there. `stop` is untouched - a turn somebody asked to stop still
  stops at once.
- **The client says what happened.** A socket that dropped mid-answer and came
  back ends the turn on screen, draws a notice above the composer, and offers a
  re-read of the transcript - which is where a finished turn lands, since it was
  never the socket that carried it there.

## Why this way

The alternative for the socket was to keep swapping it and make the server
tolerate the reconnect, which is more machinery for a change that is not a
change: `get_current_user_ws` verifies the token once, at the handshake, and
never again, so a socket authenticated by the credential it shook hands with is
still that socket after a refresh. Making the token a getter rather than a value
is the smallest statement of that.

For the turn, cancelling on a dropped socket was protecting against exactly one
thing - a run going on for ever with nobody watching - and it was paying for it
with every answer a reader was waiting for. A bound spends the same protection
without the cost. Ten minutes is above the longest turn measured on a small
deployment (416 s) with room to spare, and far below the cost of an afternoon;
the budget capability is the real ceiling, and this is the one that applies when
the reason to stop is that the reader has gone.

The notice is a button rather than a poll because the turn finishes when it
finishes and nothing tells the client when. It holds the outbound queue while it
is up: the lost turn is still being written under a socket this client no longer
has, and a queued message sent now would run a second turn against a history
missing the first one's answer.

## Checks

- [ ] If a capability was added or changed, its `README.md` says why - none was
- [ ] If a migration was added - none was

Verified against `main` after the notification center landed (0.0.469): the
branch merges cleanly, and on the merged tree `make lint` is clean, `make test`
is 9797 passed at 100.00% on the platform layer, and `make test-frontend-cov` is
438 files / 6394 tests at 100% statements and lines.

## Limitations

- The notice is drawn when the socket **comes back**. A reader who loses
  connectivity outright still gets a composer that spins, which is the old
  behaviour rather than a regression.
- `interrupted` is not scoped to the conversation the way `lastUsage` is, so a
  reader who switches threads inside a reconnect sees the notice on the wrong
  one.
- Asking something new during the grace clears the notice by design, and the
  lost turn is still running - so its answer can land in the transcript after
  the later question.


DEENUU1 (approved):
Reviewed the whole diff, and verified it against `main` as it stands after the
notification center landed (0.0.469) - your branch merges cleanly, and on the
merged tree `make lint` is clean, `make test` is 9797 passed at 100.00% on the
platform layer, and `make test-frontend-cov` is 438 files / 6394 tests at 100%
statements and lines.

**The change is right.** The premise it rests on holds where it has to:
`send_event` answers a dead socket with `False` rather than raising, and
`run_stream.FrameSink`'s own docstring already states the contract - *a visitor
who closed the tab must not take the run with them*. So a turn that outlives its
socket keeps iterating and `process_message`'s `finally` still writes the answer
down, which is exactly what the grace is buying. `wait_for(shield(task), ...)`
is the correct shape for keeping the timeout path and the explicit-cancel path
distinguishable, the `CancelledError` branch re-raises after stopping the turn
rather than reporting a clean drain over a turn still going, and moving
`manager.disconnect` ahead of `shutdown` keeps the connection count honest for
the length of the grace.

Dropping the subprotocols from `sigOf` is the right call for the same reason the
comment gives: the token is verified once at the handshake and never again, so a
new one is not a new socket. I checked the one way that could bite - a logout
followed by a sign-in as somebody else reusing the previous user's live socket -
and it is not reachable in practice: `disconnect` defers its close by 150 ms, and
no auth cycle completes inside that window.

Three things below. None of them blocks this, and I am merging it - the current
behaviour cancels a streaming answer every time the console refreshes its own
token, which is far worse than any of them. Recording them so they are not lost.

---

**1. `interrupted` is not scoped to the conversation.**

`lastUsage` already draws this line - `onThisConversation` exists so "a value
from the thread somebody just left is not a value about this one" - and the new
state does not. If the socket drops mid-answer in conversation A and the reader
opens B before it returns, the effect fires on B: the amber notice is drawn over
a thread that was never interrupted, and `handleTurnInterrupted` re-reads B
rather than the thread whose turn is still being written. The window is the
length of a reconnect, so it is narrow, but the fix is the shape already in the
file - capture the conversation id with `interruptedRef`, and compare.

**2. A drop that never comes back still spins.**

The effect only acts on the way back up (`returned && interruptedRef.current`),
so a reader who loses connectivity outright gets the old behaviour: no notice,
no ending, a composer that spins. That is not a regression and it is arguably
the harder half, but the notice's own copy promises the agent "finishes the turn
and saves it" - which is true in the offline case too, and nothing says so.

**3. Asking something new during the grace runs two turns on one conversation.**

`doSend` clears `interrupted`, which is deliberate and the comment says why. The
consequence is not written down anywhere: the lost turn is still running under
the old session for up to ten minutes, so the new turn loads a history without
its answer and that answer lands in the transcript after the later question. It
is not corruption, and a second tab could already do it - but a reader who
follows the notice's own instruction to wait, then gives up and types, is the
common path into it rather than an exotic one.

I have filled in the pull request description from the diff, since it was still
the unedited template and that body becomes the squash commit on `main`.

Issue #1787, opened 2026-09-18
Workflows: visual editor, configuration, testing and publication


DEENUU1:
Build the production visual editor using the SDK decision, with AgenticOS styling and a backend-driven catalog.

## Delivery scope

- Owner: MichalZelasko.
- Milestone: Workflows + Virtual Tables (W8–W12).
- Planning window: 2026-09-21 to 2026-10-23.
- Parent: #56.

## Scope

- Provide workflow list, draft/published/archived states, creation, duplication and templates; canvas pan/zoom, ports, selection, keyboard interactions, undo/redo and copy/paste.
- Use catalog-driven forms plus agent/version, collection, table/column, secret and typed binding pickers. Show only fields available on the selected execution path.
- Provide error ports, nested foreach body editing and clear validation messages. Do not serialize temporary validation state into the executable spec.
- Autosave uses expected_revision and shows conflicts; publish creates an immutable validated version. Test runs use a frozen draft snapshot.
- Distinguish mock tests from real runs that can make network calls or write data. Show per-step inputs/outputs, costs and approval state from persisted execution events.
- Triggers and reply channels are configured beside the graph, all mapping to one core.input. Demos guide behavior only.

## Acceptance criteria

- [ ] A user builds the two-agent RAG example and a table update workflow without code.
- [ ] Publish, draft editing, restore-to-draft and concurrent edit conflicts preserve version integrity.
- [ ] Copied nodes/scopes retain valid remapped bindings; errors and retries can be configured in the UI.
- [ ] No organization data or pending save leaks when switching workflow/organization.

## Dependencies

- #1781 — Evaluate Workflow Builder SDK in the AgenticOS frontend
- #1786 — Workflows: Pydantic node contracts, registry and graph validation

## Design references

Source files in the planning bundle: `workflows-implementation.md §3; workflows-demo.html`.
The planning documents are in Polish; this issue defines the implementation scope in English. HTML demos are illustrative, use simulated data and do not prescribe the production styling.
Related workflow/table design and demo bundle: #56.

Issue #1781, opened 2026-09-18
Evaluate Workflow Builder SDK in the AgenticOS frontend


DEENUU1:
Timebox an integration prototype to decide whether Synergy Codes Workflow Builder SDK accelerates our editor without forcing a different domain model.

## Delivery scope

- Owner: MichalZelasko.
- Milestone: Workflows + Virtual Tables (W8–W12).
- Planning window: 2026-09-21 to 2026-10-23.
- Estimate: 1–2 working days, included in the workflow milestone.
- Parent: #56.

## Scope

- Evaluate @workflowbuilder/sdk 2.3.0 against reviewed upstream commit b926e94298623df479133825172028a30819feeb. Mount client-side in the current Next.js app and adapt styling to AgenticOS.
- Exercise an agent/version picker, a Virtual Table write configuration, nested foreach editing and round-trip serialization to our typed graph. Mock unfinished APIs explicitly.
- Test global singleton state and registries across workflow/organization navigation, pending autosaves and Strict Mode. Only one active editor instance is supported upstream.
- Fix or isolate the save callback bug: the active runtime integration treats the strings error and alreadyStarted as truthy success. Exercise expected_revision and HTTP 409.
- Evaluate undo/redo, copy/paste ID and binding remapping, schema adaptation from Pydantic, keyboard interactions and representative bundle/graph performance. Paid placeholder plugins are not available implementations.

## Acceptance criteria

- [ ] A working integration example demonstrates the listed cases; results identify mocks versus real integration.
- [ ] The decision record lists patches, adapter work, measurements and licensing obligations.
- [ ] Adopt SDK only if adaptations are bounded; otherwise use React Flow directly. Do not automatically extend the experiment or cut the milestone estimate.

## Design references

Source files in the planning bundle: `workflowbuilder-reuse-analysis.md; workflows-implementation.md §3`.
The planning documents are in Polish; this issue defines the implementation scope in English. HTML demos are illustrative, use simulated data and do not prescribe the production styling.
Related workflow/table design and demo bundle: #56.

Issue #56, opened 2026-08-01
Build UI workflows and Virtual Tables (W8–W12)


DEENUU1:
Let users build and run processes in the UI using existing agents, knowledge collections, HTTP calls and shared tables. Virtual Tables provide typed records with table, kanban and list views, available through the UI, agent tools, workflow nodes and the public API.

## Delivery scope

- Owner: MichalZelasko.
- Milestone: Workflows + Virtual Tables (W8–W12).
- Planning window: 2026-09-21 to 2026-10-23.

## Scope

- This replaces the earlier dynamic-workflow capability proposal in this issue. A workflow is its own versioned resource, not an agent capability or arbitrary user-authored Python graph. Users build processes visually; developers implement typed nodes. An optional Python node handles advanced transformations.
- One milestone owned by Michał Żelasko covers implementation, integration, tests, documentation and fixes. Do not split workflow deliverables into artificial weekly commitments.
- Include core agent/RAG/data/HTTP/table nodes, error handling, durable foreach, approvals, isolated Python, files/images, version publication, history, budgets and permissions.
- Inputs in the first delivery: API, WebSocket, webhook, /chat, schedule and table.record.created. Hosted pages and Slack follow through the same adapter contract.
- Start with a timeboxed 1–2 working-day SDK integration evaluation. Reuse the existing Python service layer, AgentRunner, Prefect, storage, approvals and governance.
- Historical estimates of one week for tables and three for workflows are not a commitment for the expanded scope. Re-estimate after the SDK and execution design checks.

## Acceptance criteria

- [ ] A user can build, publish, test and run a workflow without writing its graph in code.
- [ ] A question from /chat passes through retrieval and a pinned agent version and produces one final response.
- [ ] A new table record triggers a workflow once; the workflow updates it without retriggering a created event.
- [ ] Files can pass through foreach, conversion, agent, Python normalization and table upsert with error handling and a final report.
- [ ] All child acceptance criteria pass, including restart recovery, access revocation, conflicts and cost limits.
- [ ] User/developer documentation and illustrative demos are available from this issue. Production UI uses existing AgenticOS components.

## Delivery issues

- [ ] #1781
- [ ] #1782
- [ ] #1783
- [ ] #1784
- [ ] #1785
- [ ] #1786
- [ ] #1787
- [ ] #1788
- [ ] #1789
- [ ] #1790
- [ ] #1791
- [ ] #1792
- [ ] #1793

## Design references

[...]

PR #1813, opened 2026-09-21
docs(frontend): record the Workflow Builder SDK evaluation

MichalZelasko:
### Summary

Record the decision on the Workflow Builder SDK for the workflow editor: do not adopt `@workflowbuilder/sdk`, build the editor on `@xyflow/react` directly. The branch also holds the evaluation lab, kept for review only.

### Business Context

#1781 is the timeboxed SDK evaluation that opens the workflow work in #56. It decides whether #1787 (the visual editor) starts from the SDK or from React Flow.

### Changes

- `docs/plans/workflow-builder-sdk-evaluation.md`: the decision record with the adoption threshold set before building, results per case, real versus mocked, patches, adapter work, measurements, licensing and what was not evaluated.
- **Evaluation lab (for review only, removed before merge).** A development-only route `/dev/workflow-sdk` in `frontend/src/components/dev/workflow-sdk/` with the adapters, mocks, a bare React Flow baseline and 7 vitest specs.
- Lab-only dependencies `@workflowbuilder/sdk@2.3.0` and `@xyflow/react@12.11.6`, their licence entries (`licenses/policy.toml`, `THIRD_PARTY_NOTICES.md`, a Poppins row in `docs/licenses.md` and its translations) and a CHANGELOG entry.

Why not adopt:
- The SDK stylesheet restyles the whole document, survives navigation and cannot be contained with tokens. An iframe is refused by the console's `frame-ancestors 'none'`.
- Every integration point that touches process-wide state needed a guard. Unguarded, a pending autosave overwrote one workflow with another's nodes.
- Undo/redo, copy/paste, nested scopes and a list-of-rows control are not in the package, so they are ours either way.
- Costs: `.next/static` grows from 4.8 MB to 13.4 MB, the frontend image gains 132 npm packages, and a 1,000-node graph settles in 485 ms against 298 ms for React Flow alone.
- Two patches would be needed against a bound of five, so the count alone would pass. The decision rests on the stylesheet, the global state and the missing features.

### Verification

- Passed: `make lint`, `make test-frontend-cov`, `make build-frontend`, `make licenses-check`, `make docs-build` and the docs i18n check.
- Not run: `make check`, backend tests and e2e.
- To reproduce the lab: `cd frontend && bun run dev`, then open `/dev/workflow-sdk` (needs a session). `bunx vitest run src/components/dev` runs the adapter specs.

### Limitations

- No backend was started. The agent picker ran against stubbed responses on the real endpoint paths, and the workflow server, tables list and Pydantic schema are mocks.
- Measured on the published 2.3.0 build. The reviewed upstream commit `b926e94` is four commits ahead of the `v2.3.0` tag.
- Not evaluated: two editors at once, swapping documents by props, the SDK's Polish locale, Ctrl+A select-all and the iframe beyond the refusal. Performance figures are single runs.

### Linked Issues

Refs #1781. Part of #56.


MichalZelasko:
This branch still contains the SDK evaluation lab (`frontend/src/components/dev/workflow-sdk/`, the `/dev/workflow-sdk` route, the two lab dependencies and their licence entries). It is here as a reference for review, so the findings can be checked against the code that produced them.

The lab is going to be removed before this PR merges. What merges is the decision record, `docs/plans/workflow-builder-sdk-evaluation.md`, and its references to lab files will be updated to point at a commit on this branch. The PR is a draft until then.

PR #1814, opened 2026-09-21
fix(chat): copy picked files before resetting the file input

henfircreo:
Attaching a file through the composer's paperclip did nothing: no attachment card, no toast, no spinner, and no request. Nothing reached either service's logs, because the file never left the browser.

`handleFileSelect` captured `e.target.files` - the input's own live `FileList`, not a copy - then set `e.target.value = ""`, which empties that same object in place in Blink and WebKit. The following `Array.from(files)` therefore saw an empty list, and `uploadFiles` returned at its own `accepted.length === 0` guard without queueing a card, toasting, or calling `uploadFile`. Every symptom follows from that one early return.

Drag-and-drop and paste were unaffected: neither goes through the input. Every other picker in the codebase already materializes the `File` before resetting; this was the lone offender.

The array is now built before the reset, and the reset moved ahead of the early return so a cancelled dialog no longer leaves a stale value. `uploadFiles` also gains the `tErrors` dependency it reads in the upload-failure toast.

jsdom hands the input a fresh `FileList` on reset rather than clearing the captured one, so the existing picker tests passed throughout. The unit regression models the browser's semantics explicitly - it fails with "expected upload to be called once, got 0 times" on the old code - and a Playwright spec attaches a real file in the composer, which is the layer that would have caught this.

## Checks

- [ ] Tests cover the new behaviour, and would fail without the change
- [ ] Platform-layer coverage is still 100%
- [ ] If a capability was added or changed, its `README.md` says why
- [ ] If a migration was added, `alembic downgrade base && alembic upgrade head` works


DEENUU1 (approved):
Reviewed the mechanism, the tests and the rest of the codebase for the same shape. **This is right, and the reasoning in the comments is the part worth keeping.**

## The fix

`e.target.files` is the input's own live `FileList`, and `e.target.value = ""` empties *that object*, not a copy of it. The old order — capture, guard, reset, `Array.from` — therefore read an empty list in every browser that implements the clear, so `uploadFiles` returned at its own empty guard: no card, no toast, no request. Copying first is the fix, and the three lines are now in the only order that works.

Two details I checked rather than assumed, both correct:

- **The reset is unconditional and sits before the early return.** That is what makes picking the same file twice in a row fire `change` again — the case somebody hits right after a refusal they have since fixed. Moving it below the guard would have traded one silent failure for another.
- **`tErrors` in the `uploadFiles` dependency array is a real missing dependency**, not noise: it is used at the `toast.error(...)` line inside that callback, so without it a language change left the error translator captured from the previous render.

## The tests

The unit test earns its place because of *why* it exists: jsdom hands the input a fresh list on reset, which is exactly why this shipped with a green suite. Modelling Blink's in-place clear (`blinkFileList` + the `value` setter wired to it) is the honest way to hold that shut, and `Array.from` reading `length` then indices means the double fails the same way the browser does.

The e2e is the one that actually proves it, and the two together are the right pair rather than a duplication: one pins the contract, the other pins the browser.

## One thing outside this PR

The same class of bug — an input that is never reset — is live in the knowledge-base uploader. `app/[locale]/(dashboard)/rag/[id]/page.tsx` wires `onChange={(e) => handleFiles(e.target.files)}` and nothing clears `fileInputRef`'s value, so re-picking the same document after a failed upload fires no `change` at all and the page does nothing. `handleFiles` copies with `Array.from` immediately, so it does not have *this* PR's live-list bug — only the no-second-attempt half of it.

Not this PR's job, and I am not asking for it here. Recording it so it is not found twice. Every other picker I looked at is safe: the `?.[0]` sites take a `File` reference out of the list, which survives the clear, and `components/skills/skill-files.tsx` already copies before resetting.

Approving.

Issue #1811, opened 2026-09-21
platform: concurrent tool-call auxiliary LLM calls double-count ambient usage


MichalZelasko:
Auxiliary LLM calls made inside a tool (self-query, query-expansion) meter spend by snapshotting `ctx.usage` before the call and computing a delta after (`usage_counts`/`usage_delta` → `record_ambient_usage`). When a model emits several tool calls in one response, Pydantic AI runs them concurrently and they share one `ctx.usage`; two concurrent auxiliary calls can capture the same `before` snapshot, and the one finishing second computes a delta that includes the first's tokens — so `record_ambient_usage` books part of the spend twice.

The over-count is in the **budget-conservative** direction (it can only refuse a run earlier, never overspend), but it inflates reported ambient cost. Surfaced in the Codex review of **PR #1806** (self-query, #1650); the same pattern is in **PR #1807** (query-expansion, #1649). `compaction`/`system_reminders` use the same helpers but run in the request wrapper (serialised), so they don't hit the race — the exposure is specific to the concurrent tool-call path.

> **Origin & merge order:** found reviewing PR #1806. Follow-up that changes the shared ambient-metering helpers those PRs use; merge **after** #1805, #1806 and #1807.

## Context

- Helpers: `budget/_capability.py::usage_counts` / `usage_delta` / `record_ambient_usage`, and `reserved_limits`.
- `agent_runner.py` runs the tool calls for one response concurrently.
- The delta approach is correct for serialised callers (compaction, reminders) but races for concurrent tool-call auxiliaries.

## Scope

- Meter each auxiliary call by its own run's usage (e.g. a request-local `RunUsage` merged once into the shared total) rather than a delta over the shared object, without breaking the `reserved_limits` request-slot enforcement.

## Acceptance

- Two concurrent unfiltered searches (self-query on) book ambient usage equal to the sum of the two auxiliary runs, not more.
- The `request_limit` reservation still holds.

## Relates

- Found reviewing **PR #1806** (#1650): https://github.com/vstorm-co/agenticos/pull/1806#discussion_r4060489710
- Same pattern in **PR #1807** (#1649)
- Issue map #168, cluster "What a run costs" (with #15)

Issue #1810, opened 2026-09-21
platform: propagate the run's model settings (timeout, max_tokens) to ambient auxiliary LLM calls


MichalZelasko:
Ambient auxiliary agents are constructed as a bare `Agent(ctx.model, ...)`, which inherits the run's model but not the per-request `ModelSettings` (timeout, max_tokens, temperature, …) attached to the outer `PydanticAgent` in `factory.py`. So an auxiliary call runs with provider defaults: it can outlive the agent's configured `timeout` and generate more tokens than the agent permits.

Surfaced in the Codex reviews of **PR #1806** (self-query, #1650) and **PR #1807** (query-expansion, #1649). Pre-existing and platform-wide — `compaction` and `system_reminders` do the same. The fix belongs across all auxiliary calls.

> **Origin & merge order:** found reviewing PR #1806 and PR #1807. Follow-up; merge **after** #1805, #1806 and #1807.

## Context

- Model settings are merged onto the outer agent in `factory.py` (per profile/spec), not carried on `ctx.model`.
- Auxiliary construction sites: `compaction`, `system_reminders`, `knowledge/_self_query.py` (#1806), `knowledge/_toolset.py::_model_generate` (#1807).
- Bounded fan-out and `reserved_limits` already cap the *amount* of auxiliary work; this is about honouring the per-request settings.

## Scope

- Pass the run's effective `ModelSettings` to the auxiliary agent (or its `run` call), or define explicit bounded settings for these calls — uniformly across the four sites.

## Acceptance

- An agent with a configured `timeout`/`max_tokens` bounds its auxiliary calls by the same settings.
- A test asserts the settings reach the auxiliary run.

## Relates

- Found reviewing **PR #1806** (#1650): https://github.com/vstorm-co/agenticos/pull/1806#discussion_r4060489972 — and **PR #1807** (#1649): https://github.com/vstorm-co/agenticos/pull/1807#discussion_r4060490379
- Issue map #168, cluster "What a run costs"

Issue #1809, opened 2026-09-21
platform: auxiliary agents leak prompt content to tracing under observability.content="none"


MichalZelasko:
When an agent sets `observability.content="none"`, `factory.py` pins the *configured* agent to content-free instrumentation via `suppress_content`, because the deployment instruments Pydantic AI globally with content **on**. Auxiliary agents that a capability constructs ad-hoc (`Agent(ctx.model, ...)`) do not inherit that suppression, so they trace their prompts and outputs — the search query and the text they generate — to the operator's Logfire project despite the agent's explicit no-content setting.

Surfaced in the Codex review of **PR #1806** (self-query, #1650). It is a pre-existing, platform-wide gap, not specific to that PR: `compaction` and `system_reminders` build their nested agents the same way and trace content identically. The new capability followed the established pattern; the fix must cover all of them together.

> **Origin & merge order:** found reviewing PR #1806. Follow-up to the RAG retrieval-quality PRs; merge **after** #1805, #1806 and #1807.

## Context

- `factory.py::_instrument` applies `suppress_content(agent)` to the configured agent only; global instrumentation defaults content on.
- Auxiliary agents with no suppression: `compaction/_capability.py`, `system_reminders/_capability.py`, `knowledge/_self_query.py` (#1806), `knowledge/_toolset.py::_model_generate` (#1807).
- The content policy (`observability.content`) is not currently available where these agents are constructed, so plumbing it there is part of the work.

## Scope

- Make every ad-hoc auxiliary agent honour the outer agent's `observability.content` setting (apply `suppress_content` when it is `"none"`).
- Plumb the content policy to the construction sites, or centralise auxiliary-agent construction so the policy is applied once.

## Acceptance

- With `observability.content="none"`, no auxiliary call (compaction, reminder, self-query, query-expansion) records prompt or output content to any exporter.
- A test asserts content-free instrumentation on the auxiliary spans for a `content="none"` agent.

## Relates

- Found reviewing **PR #1806** (#1650): https://github.com/vstorm-co/agenticos/pull/1806#discussion_r4060489864
- Related observability item #561; issue map #168, cluster "What a run costs"

Issue #1808, opened 2026-09-21
platform: enforce budget/spend limits on ambient auxiliary LLM calls


MichalZelasko:
Ambient auxiliary LLM calls run through an `Agent` a capability constructs itself, so they never pass `BudgetGuard.wrap_model_request` — the guard is a capability on the *outer* agent, not on the one built for the auxiliary call. The call books its spend against `ctx.usage` only afterwards (`record_ambient_usage`) and never re-checks the organization's dollar cap (`SpendLimit`) before issuing the request. This is a deliberate, documented characteristic (the `compaction` module docstring states it: *"What this cannot do is stop the spend"*), bounded today only by holding back one request slot (`reserved_limits`) so an auxiliary call cannot push the run past `request_limit`.

Surfaced in the Codex review of **PR #1807** (query analysis/expansion, #1649) and **PR #1806** (self-query, #1650). It is not specific to them: `compaction` and `system_reminders` have the identical characteristic. The new capabilities correctly followed the established pattern, so the fix belongs platform-wide rather than in one capability.

> **Origin & merge order:** found reviewing PR #1807 (and PR #1806). This is a follow-up that touches the shared budget path those PRs now use, so it must merge **after** #1805, #1806 and #1807 land.

## Context

- `BudgetGuard.wrap_model_request` refuses a run once the ledger is at/over the monthly cap, but only on the *outer* agent's requests.
- Auxiliary agents are built as `Agent(ctx.model, ...)` in `compaction/_capability.py`, `system_reminders/_capability.py`, `knowledge/_self_query.py` (#1806) and `knowledge/_toolset.py::_model_generate` (#1807).
- Each meters after the fact via `record_ambient_usage` and reserves one request slot via `reserved_limits`, so the overshoot is bounded to what the platform already accepts for compaction/reminders — but LLM query-expansion can add more than one auxiliary request per tool call.

## Scope

- Give ambient auxiliary calls a shared way to consult the run's `SpendLimit`/`BudgetGuard` immediately before the call, degrading to the plain (unexpanded/unfiltered) path when the cap is already reached.
- Apply it uniformly to all four call sites; do not let one capability diverge from the others.
- Keep the default-off / bounded-fan-out behaviour — this is about enforcing the cap, not removing the feature.

## Acceptance

- An outer request that takes the ledger to/over the cap while returning a tool call does not let the auxiliary call issue a further paid request; it degrades to the plain path.
[...]

PR #1818, opened 2026-09-21
fix(agents): budget, content, settings and metering for ambient auxiliary LLM calls

MichalZelasko:
### Summary

Centralise how **ambient auxiliary LLM calls** — the sub-agent calls capabilities make inside a run (compaction summaries, system reminders, and, once they merge, self-query and query-expansion) — are constructed and metered, so each one honours the run's budget cap, its `observability.content="none"` setting, and its model settings, and is metered race-free. Introduces shared helpers and applies them to the sites that exist on `main`.

### Business Context

Auxiliary agents were each hand-built as `Agent(ctx.model, …)`, which bypasses the outer agent's `BudgetGuard`, its content-suppression, and its `ModelSettings`, and meters spend with a snapshot/delta over the shared `ctx.usage` that double-counts under concurrent tool calls. These were filed as #1808–#1811 after review of the RAG PRs; the pattern is platform-wide (compaction and system_reminders behave identically), so the fix is one shared mechanism rather than per-capability patches.

### Changes

- **Shared budget helpers** (`budget/_capability.py`, exported): `can_afford_next_request()` / `can_afford_ambient_call()` plus a `guarding()` ContextVar (mirrors the existing `metered_by`); `reserved_limits()` promoted from system_reminders; `metered_nested_run()`, which meters each nested call against a **private `RunUsage` copy merged once** — fixing the concurrent double-count while preserving the reserved-request slot.
- **`observability.py::auxiliary_model(ctx)`** — returns a content-free `InstrumentedModel` when the outer run is not recording content (`ctx.trace_include_content` False), else the model unchanged; reaches even harness-built agents.
- **`capabilities/_ambient.py::run_ambient_agent(...)`** — the single code-built ambient runner: cap gate (raises `AmbientCallSkipped`), content-free model, `model_settings=ctx.model_settings`, `reserved_limits`, and race-free metering. The RAG call sites adopt this once #1806/#1807 merge.
- **`agent_runner.py`** — `PreparedRun.execute`/`.iterate` open `guarding(self.built.budget)` alongside `metered_by`, so the run's cap is reachable from auxiliary calls (including delegates).
- **Per-site (on `main`):** system_reminders routes through `run_ambient_agent`; compaction pins a content-free model + the run's model-settings and skips the summary at the cap (zero-LLM tiers still run); tool_output_limits gates its reduction at the cap.
- Corrected the now-stale "cannot stop the spend" / "inherits the run's model" docstrings.

### Verification

- `make test` — PASS, 9889 passed / 27 skipped, **100% coverage** on the platform layer.
- `make lint-backend` — PASS (ruff, format, ty, vulture, deptry, guard scripts, codespell, docs-i18n).
- `tests/test_security_marker.py` — PASS.
- New tests cover every new branch, including a deterministic #1811 race regression (the old delta pattern double-books; `metered_nested_run` books exactly the sum, requests end at start+2), `auxiliary_model` (content-on/off/realtime), and each per-site at-cap path.
- A local Codex review (`gpt-5.6-sol`) raised one P2 (tool_output_limits skips its free reductions at the cap); rejected after verifying the run is terminal at the cap — so there is no within-run effect — and that an action-aware gate would reintroduce the coupling the uniform wrapper deliberately avoids.
- `make db-check` — no schema change; no migration.

### Linked Issues

Part-of #1808, #1809, #1810, #1811 (kept open). This lands the shared mechanism and the on-`main` sites. Remaining work tracked in those issues: the two RAG call sites (`knowledge/_self_query.py`, `knowledge/_toolset.py::_model_generate`) adopt `run_ambient_agent` once #1806/#1807 merge; and tool_output_limits content-suppression (#1809) / model-settings (#1810), whose harness strategy forwards neither without a hack.


MichalZelasko:
@codex review

